PR #10, opened 2021-03-03
Add migration doc

colin-axner:


docs/migrations/ibc-migration-042.md
```diff
@@ -0,0 +1,54 @@
+# Migrating to ibc-go
+
+This file contains information on how to migrate from the IBC module contained in the SDK 0.41.x line to the IBC module in the ibc-go repository based on the 0.42 SDK version.
```
fedekunze, line 3: I think we also need to add the migration command that's run on `x/genutil/legacy` to the ibc-go repo. So that a migration on gaia uses the `Migration` funcs from both the SDK and ibc-go

fedekunze (commented):
ACK pending follow up issue for migration command

fedekunze (approved):

docs/migrations/ibc-migration-042.md
colin-axner: I see, this migration function is primarily for migrating genesis though? Is there anything we need to do in that migration for this upgrade? Opened: #11

docs/migrations/ibc-migration-042.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Import Changes
+
+The most obvious changes is import name changes. We need to change:
+- applications -> apps
+- cosmos-sdk/x/ibc -> ibc-go
+
+On my GNU/Linux based machine I used the following commands, executed in order:
+
+`grep -RiIl 'cosmos-sdk\/x\/ibc\/applications' | xargs sed -i 's/cosmos-sdk\/x\/ibc\/applications/ibc-go\/apps/g'`
```
AdityaSripal, line 13:
```suggestion
`grep -RiIl 'cosmos-sdk\/x\/ibc\/applications' | xargs sed -i 's/cosmos-sdk\/x\/ibc\/applications\/ibc-go\/apps/g'`
```

docs/migrations/ibc-migration-042.md
```diff
@@ -0,0 +1,54 @@
...
+
+`grep -RiIl 'cosmos-sdk\/x\/ibc' | xargs sed -i 's/cosmos-sdk\/x\/ibc/ibc-go/g'`
```
AdityaSripal, line 15:
```suggestion
`grep -RiIl 'cosmos-sdk\/x\/ibc' | xargs sed -i 's/cosmos-sdk\/x\/ibc\/ibc-go/g'`
```

AdityaSripal (dismissed):
I believe there's a slight mistake in your commands?

docs/migrations/ibc-migration-042.md
colin-axner: No they are correct. I will add a link to the [source](https://www.internalpointers.com/post/linux-find-and-replace-text-multiple-files)

Adding the `\` causes that to be interpreted as one long path `cosmos-sdk/x/ibc/applications/ibc-go/apps` as opposed to the 
search: `cosmos-sdk/x/ibc/applications`
replace: `ibc-go/apps`

docs/migrations/ibc-migration-042.md
colin-axner: ditto

colin-axner:
Updated the migration doc based on experience of updating gaia. The diffs look nice

PR #12, opened 2021-03-04
add codecov.yml

marbar3778:
## Description

Copy over codecov from sdk


.codecov.yml
```diff
@@ -0,0 +1,54 @@
...
+      - "app/"
+      - "baseapp/"
+  modules:
+    paths:
+      - "core/"
+      - "!core/**/client/" # ignore client package
+  client:
+    paths:
+      - "client/"
+      - "core/**/client/"
```
colin-axner, line 46: I think we should probably add `apps/` and `light-clients/`? 

I think the current code layout is a little confusing. Maybe we should add a `src` file which contains
`core/`
`light-clients/`
`apps/`

thoughts?

colin-axner (approved):

.codecov.yml
colin-axner: alternatively we could use `modules` like ibc-rs

PR #14, opened 2021-03-04
Add "modules" top level directory

colin-axner:
Cleaner and more intuitive top level directory. Also aligns with `ibc-rs`


marbar3778 (approved):

marbar3778:
should testing be brought in as well?

colin-axner:
> should testing be brought in as well?

My thought is no, I'd like `testing` to eventually become `relayer`. Relayer wouldn't really be a module (and neither is testing). This also would align with ibc-rs

PR #75, opened 2021-03-05
ADR 002: Localhost Client

AdityaSripal:
Labelled 02, since I think ADR 001 from the SDK should be ported here as its more relevant to IBC


docs/architecture/adr-002-local-host-client.md
```diff
@@ -0,0 +1,38 @@
...
+### Consequences
+
+### Positive
+
+- A working localhost client
+
+### Negative
+
+- Without the readonly change, this would give write access to a potentially faulty client implementation
+- The readonly change will require a change in the SDK
```
fedekunze, line 38: not necessarily, you can create a `ReadOnlyStore` interface:

```
	// Get returns nil iff key doesn't exist. Panics on nil key.
	Get(key []byte) []byte

	// Has checks if a key exists. Panics on nil key.
	Has(key []byte) bool
```

fedekunze (approved):

docs/architecture/adr-002-local-host-client.md
colin-axner: genius

colin-axner, line 38: I'd be nice if we could remove localhost logic from `02-client` handlers. Is there a solution that fixes that issue as well? [Related discussion](https://github.com/cosmos/cosmos-sdk/issues/7706#issuecomment-718591993)

AdityaSripal: Correct, but I believe this interface should exist in `store/types` so other modules can use it.

AdityaSripal: Yes those will all be removed, since the localhost will maintain no separate state

colin-axner: So in begin blocker, we won't call update on the localhost?

colin-axner: Yes I agree, this sounds like a useful feature

AdityaSripal: No

docs/architecture/adr-002-local-host-client.md
```diff
@@ -0,0 +1,47 @@
...
+    }
+	clientPrefix := []byte(fmt.Sprintf("%s/%s/", host.KeyClientStorePrefix, clientID))
+	return prefix.NewStore(ctx.KVStore(k.storeKey), clientPrefix)
+}
+```
+
+The localhost implementation can remain unchanged, but now it will have access to the entire IBC store, and thus getting connections and channels using the paths specified by `24-host` will work as expected.
+
+All of the special case code in the IBC client handlers can be removed as localhost will not have any separate state.
+
```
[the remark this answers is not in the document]
colin-axner: I guess we will still need to keep the begin blocker handler (to update latest height and chain id [in the case of an upgrade])

fedekunze:
should we merge this ADR?

colin-axner:
> should we merge this ADR?

I was waiting on a response to [this](https://github.com/cosmos/ibc-go/pull/75#discussion_r590211829). It seems likely @AdityaSripal might not be the one to implement this ADR, so I think there should be a clarification on what to do in the instance I pointed out. I'd prefer not to leave parts of the implementation up to interpretation

docs/architecture/adr-002-local-host-client.md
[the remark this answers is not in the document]
colin-axner: bump @AdityaSripal

colin-axner:
I'm closing this. It is linked in the localhost issues, so it can be reopened if it becomes relevant again, but for now it seems this won't go anywhere (and isn't a priority at all)

PR #79, opened 2021-03-09
Update main to SDK release

colin-axner:
This is a tracking branch to prepare for the next v0.43.0 release of the SDK which will contain a removed IBC module

Currently using commit on master branch: `e3e89f52607b9d205e1ddea4af2d47e98e0ff2b1`


colin-axner:
marking ready for review, lots of changes already in this pr so I think it'd be good to get this through a review cycle. The main branch is already on an arbitrary SDK commit, so I don't think it hurts to just move that commit much closer to wait will be the next SDK release

modules/core/02-client/keeper/proposal_test.go
```diff
@@ -179,17 +179,6 @@ func (suite *KeeperTestSuite) TestHandleUpgradeProposal() {
 				}
 			}, false,
 		},
-		{
-			"schedule upgrade fails - plan sets time and height", func() {
```
colin-axner, line 183: upgrade time got removed

fedekunze (approved):

PR #82, opened 2021-03-12
client: upgraded consensus state

### LINKED ISSUE #5: Add gRPC for upgraded consensus state
Add a gRPC for upgraded consensus state into 02-client. Should be similar to the upgraded client state gRPC

fedekunze:
closes #5


proto/ibc/core/client/v1/query.proto
```diff
@@ -148,11 +155,13 @@ message QueryClientParamsResponse {
   // params defines the parameters of the module.
   Params params = 1;
 }
+
 // QueryUpgradedClientStateRequest is the request type for the
 // Query/UpgradedClientState RPC method
 message QueryUpgradedClientStateRequest {
-  // client state unique identifier
-  string client_id = 1;
```
fedekunze, line 155: this is not used

proto/ibc/core/client/v1/query.proto
```diff
@@ -148,11 +155,13 @@ message QueryClientParamsResponse {
...
+  // client id field has been deprecated.
+  reserved 1;
+  reserved "option";
   // plan height of the current chain must be sent in request
   // as this is the height under which upgraded client state is stored
   int64 plan_height = 2;
```
colin-axner, line 167: we should deprecate this as well imo. 

There should only ever exist 1 upgrade plan at a time. It might be an IBC upgrade and it might not, we should return this info in the response rather than take it as an arg in the request

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -204,11 +204,13 @@ func (q Keeper) UpgradedClientState(c context.Context, req *types.QueryUpgradedC
...
 	}
 
-	if err := host.ClientIdentifierValidator(req.ClientId); err != nil {
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+	if req.PlanHeight < 1 {
+		return nil, status.Error(codes.InvalidArgument, "non positive plan height")
 	}
 
 	ctx := sdk.UnwrapSDKContext(c)
+	ctx = ctx.WithBlockHeight(req.PlanHeight)
```
colin-axner, line 212: Hmm, we might want to be able to query if the client state is set before the plan height

proto/ibc/core/client/v1/query.proto
```diff
@@ -148,14 +155,13 @@ message QueryClientParamsResponse {
...
 // Query/UpgradedClientState RPC method
 message QueryUpgradedClientStateRequest {
-  // client state unique identifier
-  string client_id = 1;
-  // plan height of the current chain must be sent in request
-  // as this is the height under which upgraded client state is stored
-  int64 plan_height = 2;
+  // client id and plan height field have been deprecated. Use gRPC header for setting the height instead.
+  reserved 1, 2;
+  reserved "client_id", "plan_height";
```
colin-axner, line 164: I think we can just remove these. I don't think we have done a release with this proto definition (since it was added when removing IBC from x/upgrade)

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -240,3 +234,40 @@ func (q Keeper) UpgradedClientState(c context.Context, req *types.QueryUpgradedC
 		UpgradedClientState: any,
 	}, nil
 }
+
+// UpgradedConsensusState implements the Query/UpgradedConsensusState gRPC method
```
colin-axner, line 238: Could you add some comments on this function indicating that the context height should be set for the expected plan height (otherwise this will always return a consensus state not found)

colin-axner (commented):
Looks great! Thanks @fedekunze. Left some minor requests

modules/core/02-client/keeper/grpc_query.go
AdityaSripal: Also this will only return a valid consensus state on the last block of the chain and not for any other height

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -240,3 +234,40 @@ func (q Keeper) UpgradedClientState(c context.Context, req *types.QueryUpgradedC
...
+	ctx := sdk.UnwrapSDKContext(c)
+
+	plan, found := q.GetUpgradePlan(ctx)
+	if !found {
+		return nil, status.Error(
+			codes.NotFound, "upgrade plan not found",
+		)
+	}
+
+	bz, found := q.upgradeKeeper.GetUpgradedConsensusState(ctx, plan.Height)
```
AdityaSripal, line 253: This can simply return early by checking plan.Height rather than doing a store query

proto/ibc/core/client/v1/query.proto
```diff
@@ -51,7 +51,14 @@ service Query {
   rpc UpgradedClientState(QueryUpgradedClientStateRequest)
       returns (QueryUpgradedClientStateResponse) {
     option (google.api.http).get =
-        "/ibc/core/client/v1/upgraded_client_states/{client_id}";
+        "/ibc/core/client/v1/upgraded_client_states";
```
AdityaSripal, line 54:
```suggestion
        "/ibc/core/client/v1/upgraded_client_state";
```

There is only one upgraded client state at a time

proto/ibc/core/client/v1/query.proto
```diff
@@ -51,7 +51,14 @@ service Query {
...
+  }
+
+  // UpgradedConsensusState queries an Upgraded IBC consensus state.
+  rpc UpgradedConsensusState(QueryUpgradedConsensusStateRequest)
+      returns (QueryUpgradedConsensusStateResponse) {
+    option (google.api.http).get =
+        "/ibc/core/client/v1/upgraded_consensus_states";
```
AdityaSripal, line 61: ditto

AdityaSripal (approved):
looks good @fedekunze , have some nits below

proto/ibc/core/client/v1/query.proto
colin-axner: this is an API thing. I think it needs to be plural

modules/core/02-client/keeper/grpc_query.go
fedekunze: > Also this will only return a valid consensus state on the last block of the chain and not for any other height

you can set the height on the grpc req metadata

colin-axner: I think this still needs godoc?

Issue #106, opened 2021-04-07
Create testing package interface


colin-axner:
## Summary

An interface should be created for the `App` used in the `Chain` type. This interface will likely return the necessary keepers for the `Chain` functions to be performed. This will allow other applications which do not exist within this repository to create their own App that implements this interface and therefore allows them to use the testing package without forking or copying its contents

cc @seantking 

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged/assigned

colin-axner:
To Do:
- [ ] attempt to simplify time tracking in coordinator
- [ ] remove redundancy and rename coordinator to relayer
- [ ] create App interface
- [ ] add readmes and docs

colin-axner:
Current Plan:

Bring global time into the coordinator, pass a pointer of the coordinator into each chain. Allow chains to set the block time right before sending a message. This would allow us to call a chain directly to execute some message which would increase with readability of functions and reduce redundant code. 

In conjunction I want to add a convenience struct probably called Endpoint. This will contain the information necessary to understand a channel endpoint completely (it'll contain source and counterparty chains, clients, connection, and channel). This will allow us to reduce the number of arguments passed into every function. 

Now testing code would utilize two endpoints which may be partially or fully filled. I'll probably wrap endpoints in some sort of Path struct. This would contain two endpoints and the associated two chains. I'll probably allow endpoints to have names for convenience

After refactoring all this, I'll then add an App interface and see what functions I need this interface to have

colin-axner:
If time: 
- implement a function to generate proof of an existing key using the latest counterparty consensus state height (for example in the case of packet delays)

PR #114, opened 2021-04-09
IBC Testing Package v2

colin-axner:
## Description

closes: #106

---

write a little note why.


colin-axner:
Note sure why this is failing now:
```go
go test
--- FAIL: TestKeeperTestSuite (2.13s)
    --- FAIL: TestKeeperTestSuite/TestUpdateClientEventEmission (0.05s)
        client_test.go:622: 
            	Error Trace:	client_test.go:622
            	Error:      	Not equal: 
            	            	expected: &types.Header{SignedHeader:(*types.SignedHeader)(0xc0013142f0), ValidatorSet:(*types.ValidatorSet)(0xc00184a600), TrustedHeight:types.Height{RevisionNumber:0x0, RevisionHeight:0x2}, TrustedValidators:(*types.ValidatorSet)(0xc0019864b0)}
            	            	actual  : &types.Header{SignedHeader:(*types.SignedHeader)(0xc001874570), ValidatorSet:(*types.ValidatorSet)(0xc001b23680), TrustedHeight:types.Height{RevisionNumber:0x0, RevisionHeight:0x2}, TrustedValidators:(*types.ValidatorSet)(0xc001b236b0)}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -10,13 +10,5 @@
            	            	    Time: (time.Time) {
            	            	-    wall: (uint64) 13840744187271170818,
            	            	-    ext: (int64) 26756020222,
            	            	-    loc: (*time.Location)({
            	            	-     name: (string) "",
            	            	-     zone: ([]time.zone) <nil>,
            	            	-     tx: ([]time.zoneTrans) <nil>,
            	            	-     extend: (string) "",
            	            	-     cacheStart: (int64) 0,
            	            	-     cacheEnd: (int64) 0,
            	            	-     cacheZone: (*time.zone)(<nil>)
            	            	-    })
            	            	+    wall: (uint64) 89770754,
            	            	+    ext: (int64) 63753571719,
            	            	+    loc: (*time.Location)(<nil>)
            	            	    },
            	            	@@ -96,13 +88,5 @@
            	            	      Timestamp: (time.Time) {
            	            	-      wall: (uint64) 13840744187271170818,
            	            	-      ext: (int64) 26756020222,
            	            	-      loc: (*time.Location)({
            	            	-       name: (string) "",
            	            	-       zone: ([]time.zone) <nil>,
            	            	-       tx: ([]time.zoneTrans) <nil>,
            	            	-       extend: (string) "",
            	            	-       cacheStart: (int64) 0,
            	            	-       cacheEnd: (int64) 0,
            	            	-       cacheZone: (*time.zone)(<nil>)
            	            	-      })
            	            	+      wall: (uint64) 89770754,
            	            	+      ext: (int64) 63753571719,
            	            	+      loc: (*time.Location)(<nil>)
            	            	      },
            	Test:       	TestKeeperTestSuite/TestUpdateClientEventEmission
FAIL
exit status 1
FAIL	github.com/cosmos/ibc-go/modules/core/02-client/keeper	2.176s
```

testing/simapp/test_helpers.go
```diff
@@ -336,19 +336,19 @@ func SignCheckDeliver(
...
-	}
+	//_, res, err := app.Simulate(txBytes)
+
+	//	if expSimPass {
+	//		require.NoError(t, err)
+	//		require.NotNil(t, res)
+	//	} else {
+	//		require.Error(t, err)
+	//		require.Nil(t, res)
+	//	}
```
colin-axner, line 351: `checkState` is set by the block header when the app state is committed. `deliverState` is set by request begin blocker and check state isn't updated. Removing the CheckTx allows us to send a msg without committing an empty block right before while updating the clock over time to keep synchrony. 

Will remove and rename the function later

colin-axner:
The IBC testing refactor will be a multi part refactor. 

Part 1 introduces the new design of the testing package
- global time held by coordinator and chains only need to update time before executing messages
- paths/endpoints are added. Most functionality will happen via endpoints and relaying will happen via paths 

Part 2 refactor tests to use the new design and remove all the old functions

Part 3 will implement a TestingApp interface so outside repos can use the testing package

Part 4 will update docs and do polishing as needed

colin-axner:
I've tested that `SetupNew` works (this will be renamed to `Setup` when the old functions are removed)

testing/app.go
```diff
@@ -0,0 +1,9 @@
+package ibctesting
+
+import (
+	"github.com/cosmos/ibc-go/modules/core/keeper"
+)
+
+type TestingApp interface {
+	IBCKeeper() keeper.Keeper
+}
```
colin-axner, line 9: this will be filled out in part 3

modules/core/03-connection/keeper/verify_test.go
```diff
@@ -53,6 +53,7 @@ func (suite *KeeperTestSuite) TestVerifyClientState() {
 				connection.ClientId = ibctesting.InvalidID
 			}
 
+			suite.coordinator.IncrementTime()
```
colin-axner, line 56: there was a reason these were needed, going to see if I can remove them later

testing/chain.go
```diff
@@ -137,14 +138,15 @@ func NewTestChain(t *testing.T, chainID string) *TestChain {
 	header := tmproto.Header{
 		ChainID: chainID,
 		Height:  1,
-		Time:    globalStartTime,
+		Time:    coord.CurrentTime.UTC(),
```
colin-axner, line 141: NOTE: using non-UTC time can cause issues with proto (like if you use `time.Now()` which uses your local clock)

testing/chain.go
```diff
@@ -281,6 +288,9 @@ func (chain *TestChain) sendMsgs(msgs ...sdk.Msg) error {
 // number and updates the TestChain's headers. It returns the result and error if one
 // occurred.
 func (chain *TestChain) SendMsgs(msgs ...sdk.Msg) (*sdk.Result, error) {
+
+	chain.Coordinator.IncrementTime()
```
colin-axner, line 292: this should be the only placed increment time needs to be done except for testing purposes

testing/chain.go
```diff
@@ -85,6 +85,7 @@ var (
 type TestChain struct {
 	t *testing.T
 
+	Coordinator   *Coordinator
```
colin-axner, line 88: chains now have a reference to the coordinator where they can get the current time from

testing/coordinator.go
```diff
@@ -26,26 +26,136 @@ var (
...
+func (coord *Coordinator) IncrementTimeBy(increment time.Duration) {
+	coord.CurrentTime = coord.CurrentTime.Add(increment).UTC()
+	for _, chain := range coord.Chains {
+		chain.CurrentHeader.Time = coord.CurrentTime.UTC()
+		chain.App.BeginBlock(abci.RequestBeginBlock{Header: chain.CurrentHeader})
 	}
+
+}
+
+// SetupNew constructs a TM client, connection, and channel on both chains provided. It will
```
colin-axner, line 69: these will all be renamed not to have `New` as a suffix, but it is there to avoid name collisions for now

testing/coordinator.go
```diff
@@ -26,26 +26,136 @@ var (
...
+	require.NoError(coord.t, err)
+
+	err = path.EndpointB.ConnOpenTry()
+	require.NoError(coord.t, err)
+
+	err = path.EndpointA.ConnOpenAck()
+	require.NoError(coord.t, err)
+
+	err = path.EndpointB.ConnOpenConfirm()
+	require.NoError(coord.t, err)
```
colin-axner, line 114: :heart:

modules/light-clients/07-tendermint/types/proposal_handle_test.go
```diff
@@ -79,8 +79,7 @@ func (suite *TendermintTestSuite) TestCheckSubstituteUpdateStateBasic() {
 				err = suite.coordinator.UpdateClient(suite.chainA, suite.chainB, substitute, exported.Tendermint)
 				suite.Require().NoError(err)
 
-				suite.chainA.ExpireClient(tmClientState.TrustingPeriod)
-				suite.chainB.ExpireClient(tmClientState.TrustingPeriod)
+				suite.coordinator.IncrementTimeBy(tmClientState.TrustingPeriod)
```
AdityaSripal, line 82: could you add comments that this will expire the client

testing/config.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func NewConnectionConfig() *ConnectionConfig {
+	return &ConnectionConfig{
+		DelayPeriod: DefaultDelayPeriod,
+		Version:     ConnectionVersion,
+	}
+}
+
+type ChannelConfig struct {
```
AdityaSripal, line 52: Port as well? We should allow coordinator to create channels on custom ports

testing/coordinator.go
```diff
@@ -26,26 +26,142 @@ var (
 type Coordinator struct {
 	t *testing.T
 
-	Chains map[string]*TestChain
+	CurrentTime time.Time
```
AdityaSripal, line 29: Looks like all chains will be moving synchronously in time. This is fine as a default, but how would I create tests where chainA is at a different time than chainB, i imagine we'll want to create these tests to make sure the protocol works in asynchronous environments

testing/coordinator.go
```diff
@@ -26,26 +26,142 @@ var (
...
+}
+
+// SetupNew constructs a TM client, connection, and channel on both chains provided. It will
+// fail if any error occurs. The clientID's, TestConnections, and TestChannels are returned
+// for both chains. The channels created are connected to the ibc-transfer application.
+func (coord *Coordinator) SetupNew(path *Path) {
+	coord.SetupConnectionsNew(path)
+
+	// channels can also be referenced through the returned connections
+	coord.CreateMockChannelsNew(path)
```
AdityaSripal, line 82: coordinator should look at desired port specified in path and create the channel rather than hardcoding a mock channel.

This will make the function much nicer to use for 3rd party app developers

testing/endpoint.go
```diff
@@ -0,0 +1,394 @@
...
+// Endpoint
+type Endpoint struct {
+	Chain        *TestChain
+	Counterparty *Endpoint
+	ClientID     string
+	ConnectionID string
+	ChannelID    string
+	PortID       string
+
+	ClientConfig     ClientConfig
```
AdityaSripal, line 28: I like this approach 👍

testing/endpoint.go
```diff
@@ -0,0 +1,394 @@
...
+)
+
+// Endpoint
+type Endpoint struct {
+	Chain        *TestChain
+	Counterparty *Endpoint
+	ClientID     string
+	ConnectionID string
+	ChannelID    string
+	PortID       string
```
AdityaSripal, line 26: PortID should be in ChannelConfig

testing/endpoint.go
```diff
@@ -0,0 +1,394 @@
...
+
+	ClientConfig     ClientConfig
+	ConnectionConfig *ConnectionConfig
+	ChannelConfig    *ChannelConfig
+}
+
+func NewEndpoint(chain *TestChain) *Endpoint {
```
AdityaSripal, line 33:
```suggestion
func NewDefaultEndpoint(chain *TestChain) *Endpoint {
```

Then create a constructor `NewEndpoint` that allows you to pass in each field

testing/config.go
```diff
@@ -0,0 +1,62 @@
...
+	Version string
+	Order   channeltypes.Order
+}
+
+func NewChannelConfig() *ChannelConfig {
```
AdityaSripal, line 57: Channel Config should also contain a successful ACK field and an unsuccessful ACK field, so app developers can specify to the testing package what their acks look like.

Testing package can just write these acks as needed to test app functionality

testing/endpoint.go
```diff
@@ -0,0 +1,394 @@
...
+	if err := endpoint.Chain.sendMsgs(recvMsg); err != nil {
+		return err
+	}
+
+	return endpoint.Counterparty.UpdateClient()
+}
+
+// WriteAcknowledgement writes an acknowledgement on the channel associated with the endpoint.
+// The counterparty client is updated.
+func (endpoint *Endpoint) WriteAcknowledgement(packet exported.PacketI) error {
```
AdityaSripal, line 375: This shouldn't just write MockAcknowledgement. Applications should be able to tell testing package what ack to write.

AdityaSripal (changes requested):
Looks great! I think the channel testing needs to be abstracted away from transfer/mock to be usable by 3rd part app developers

testing/config.go
colin-axner: It is possible acknowledgements might contain specific information. I'd prefer to leave this flexible for now rather than only supporting two types of acknowledgements

testing/endpoint.go
colin-axner: updated, the ack is passed in

testing/coordinator.go
colin-axner: I added a function to update only one chain. The send message execution will now only update its own clock

modules/core/keeper/msg_server_test.go
```diff
@@ -131,28 +138,28 @@ func (suite *KeeperTestSuite) TestHandleRecvPacket() {
...
+			path.SetChannelOrdered()
+			suite.coordinator.SetupNew(path)
+			packet = channeltypes.NewPacket(ibctesting.MockPacketData, 1, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, timeoutHeight, 0)
 
-			err := suite.coordinator.SendPacket(suite.chainA, suite.chainB, packet, clientB)
+			err := path.EndpointA.SendPacket(packet)
 			suite.Require().NoError(err)
 
-			err = suite.coordinator.RecvPacket(suite.chainA, suite.chainB, clientA, packet)
+			err = suite.coordinator.RecvPacket(suite.chainA, suite.chainB, path.EndpointA.ClientID, packet)
```
colin-axner, line 152: these need to be updated, but will happen eventually (technically it is doing the same logic as the endpoint recv func)

modules/core/keeper/msg_server_test.go
```diff
@@ -131,28 +138,28 @@ func (suite *KeeperTestSuite) TestHandleRecvPacket() {
...
-			packet = channeltypes.NewPacket(ibctesting.MockPacketData, 1, channelA.PortID, channelA.ID, channelB.PortID, channelB.ID, timeoutHeight, 0)
+			suite.coordinator.SetupNew(path)
+			packet = channeltypes.NewPacket(ibctesting.MockPacketData, 1, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, timeoutHeight, 0)
 		}, false, false},
 		{"ORDERED: packet already received (replay)", func() {
-			clientA, clientB, _, _, channelA, channelB := suite.coordinator.Setup(suite.chainA, suite.chainB, channeltypes.ORDERED)
-			packet = channeltypes.NewPacket(ibctesting.MockPacketData, 1, channelA.PortID, channelA.ID, channelB.PortID, channelB.ID, timeoutHeight, 0)
+			path.SetChannelOrdered()
+			suite.coordinator.SetupNew(path)
+			packet = channeltypes.NewPacket(ibctesting.MockPacketData, 1, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, timeoutHeight, 0)
```
colin-axner, line 147: I'll probably add a helper func for this

modules/core/keeper/msg_server_test.go
```diff
@@ -483,41 +500,42 @@ func (suite *KeeperTestSuite) TestHandleTimeoutOnClosePacket() {
 		expPass  bool
 	}{
 		{"success: ORDERED", func() {
-			clientA, clientB, _, _, channelA, channelB := suite.coordinator.Setup(suite.chainA, suite.chainB, channeltypes.ORDERED)
-			packet = channeltypes.NewPacket(ibctesting.MockPacketData, 1, channelA.PortID, channelA.ID, channelB.PortID, channelB.ID, timeoutHeight, 0)
+			path.SetChannelOrdered()
+			suite.coordinator.SetupNew(path)
+			packet = channeltypes.NewPacket(ibctesting.MockPacketData, 1, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, timeoutHeight, 0)
 			counterpartyChannel = ibctesting.TestChannel{
```
AdityaSripal, line 506: Why do you need to create this? Why can't I just call `path.EndpointB.CloseChannel()`

AdityaSripal (changes requested):
Small request, everything else looks great!

modules/core/keeper/msg_server_test.go
colin-axner: yea, it'll be switched to that later

AdityaSripal (approved):

PR #115, opened 2021-04-12
Hybrid Packet Delay

AdityaSripal:
## Description

closes: #XXXX

---

write a little note why.


AdityaSripal:
These are pretty major breaking changes to ibc-core. I wonder if there's a less disruptive solution. Let's discuss on core call

colin-axner:
If you want to make this backwards compatible, you should be able to discard the block delay period from the proto definition before marshalling the connection state and storing it in state

proto/ibc/core/connection/v1/tx.proto
```diff
@@ -12,34 +12,31 @@ import "ibc/core/connection/v1/connection.proto";
...
-  Counterparty counterparty = 2 [ (gogoproto.nullable) = false ];
-  Version version = 3;
-  uint64 delay_period = 4 [ (gogoproto.moretags) = "yaml:\"delay_period\"" ];
-  string signer = 5;
+  string       client_id           = 1 [(gogoproto.moretags) = "yaml:\"client_id\""];
+  Counterparty counterparty        = 2 [(gogoproto.nullable) = false];
+  Version      version             = 3;
+  uint64       delay_time_period   = 4 [(gogoproto.moretags) = "yaml:\"delay_time_period\""];
+  uint64       delay_height_period = 5 [(gogoproto.moretags) = "yaml:\"delay_height_period\""];
+  string       signer              = 6;
```
colin-axner, line 39: modifying proto field numbers is protobuf breaking. That means we should either create a v2 or follow the [protobuf update recommendations](https://developers.google.com/protocol-buffers/docs/overview#updating)

You should just be able to append the delay height

colin-axner: this would allow relayers to use the new protobuf while still successfully submitting messages to chains using the old code

proto/ibc/core/connection/v1/tx.proto
```diff
@@ -49,37 +46,30 @@ message MsgConnectionOpenInitResponse {}
...
-      [ (gogoproto.moretags) = "yaml:\"counterparty_versions\"" ];
-  ibc.core.client.v1.Height proof_height = 7 [
-    (gogoproto.moretags) = "yaml:\"proof_height\"",
-    (gogoproto.nullable) = false
-  ];
+  string                    previous_connection_id = 2 [(gogoproto.moretags) = "yaml:\"previous_connection_id\""];
+  google.protobuf.Any       client_state           = 3 [(gogoproto.moretags) = "yaml:\"client_state\""];
+  Counterparty              counterparty           = 4 [(gogoproto.nullable) = false];
+  uint64                    delay_time_period      = 5 [(gogoproto.moretags) = "yaml:\"delay_time_period\""];
+  uint64                    delay_height_period    = 6 [(gogoproto.moretags) = "yaml:\"delay_height_period\""];
```
colin-axner, line 59: ditto

proto/ibc/core/connection/v1/connection.proto
```diff
@@ -17,37 +17,41 @@ import "ibc/core/commitment/v1/commitment.proto";
...
   State state = 4;
   // counterparty chain associated with this connection.
-  Counterparty counterparty = 5 [ (gogoproto.nullable) = false ];
-  // delay period associated with this connection.
-  uint64 delay_period = 6 [ (gogoproto.moretags) = "yaml:\"delay_period\"" ];
+  Counterparty counterparty = 5 [(gogoproto.nullable) = false];
+  // delay time period associated with this connection.
+  uint64 delay_time_period = 6 [(gogoproto.moretags) = "yaml:\"delay_time_period\""];
+  // delay height period associated with this connection.
+  uint64 delay_height_period = 7 [(gogoproto.moretags) = "yaml:\"delay_height_period\""];
```
colin-axner, line 54: This is more tricky. We cannot rely on [protobuf versioning](https://developers.google.com/protocol-buffers/docs/overview#updating) since we construct the connection and then serialize it. We should add the delay height in a backwards compatible way (I think delay height just needs to be changed to optional?)

and 

when we go to serialize our connection in the keeper, we should check its connection version. IF the connection version == 1, then we should use the legacy connection protobuf definition, otherwise we should use this connection definition

The legacy connection should be a copy of the old definition but renamed to `LegacyConnection`. It should only be used to serialize connections. 

We should deserialize using the updated connection. From my understanding, protobuf should still successfully deserialize the old version into the new version 

I'm not 100% sure this will work, but we should find out

cc @cwgoes @fedekunze

colin-axner: We should also determine an end date for support of the legacy connection (this is not something we should indefinitely support imo, chains should update especially since there are significant security increases in the next coordinated upgrade)

colin-axner (changes requested):
I think we can do this in a backwards compatible way, but we need to test that'll work. Lets sync offline check that it will work before making all the changes

If we cannot do this in a backwards compatible way, then I see this as a problem with the current design and we ensure future situations can gracefully handle adding fields

AdityaSripal:
Closing in favor of non-breaking fix

PR #116, opened 2021-04-12
Cross Chain Validation

AdityaSripal:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/apps/ccv/types/expected_keepers.go
+++ b/modules/apps/ccv/types/expected_keepers.go
@@ -1,0 +1,17 @@
+package types
+
+// StakingKeeper defines the contract expected by parent-chain ccv module.
+// StakingKeeper is responsible for keeping track of latest validator set of all baby chains
+type StakingKeeper interface {
+	GetValidatorSetChanges(chainID string) []abci.ValidatorUpdate
+	// This method is not required by CCV module explicitly but necessary for init protocol
+	GetInitialValidatorSet(chainID string) []sdk.Tx
+}
+
+// CNSKeeper defines the contract expected by parent-chain ccv module
+// CNSKeeper is responsible for verifying that chain creator is authorized to create a chain with given chain-id
+type CNSKeeper interface {
+	AuthorizeChainCreator(chainID, creator string)
+}
+
+// TODO: Expected interfaces for distribution on parent and baby chains
--- a/proto/ibc/apps/cross-chain-validation/ccv.proto
+++ b/proto/ibc/apps/cross-chain-validation/ccv.proto
@@ -1,0 +1,19 @@
+syntax = "proto3";
+
+package ibc.apps.ccv.v1;
+
+option go_package = "github.com/cosmos/ibc-go/modules/apps/ccv/types";
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+// This packet is sent from parent chain to baby chain if the validator set for baby chain
+// changes (due to new bonding/unbonding messages or slashing events)
+// The acknowledgement from baby chain will be sent asynchronously once unbonding period is over,
+// and this will function as `UnbondingOver` message for this packet.
+message ValidatorSetChangePacketData {
+  repeated .tendermint.abci.ValidatorUpdate validator_updates = 1 [
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"validator_updates\""];
+}
+
--- a/proto/ibc/apps/cross-chain-validation/tx.proto
+++ b/proto/ibc/apps/cross-chain-validation/tx.proto
@@ -1,0 +1,56 @@
+syntax = "proto3";
+
+package ibc.apps.ccv.v1;
+
+option go_package = "github.com/cosmos/ibc-go/modules/apps/ccv/types";
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+// MsgSubmitNewChain defines a msg to submit a new baby chain that can be validated
+// by the parent chain validators.
+message MsgSubmitNewChain {
+  option (gogoproto.goproto_getters) = false;
+  
+  // the proposed chain-id of new chain
+  string chain_id = 1 [(gogoproto.moretags) = "yaml:\"chain_id\""];
+  // minimum stake that must commit to validate new chain before it can start
+  cosmos.base.v1beta1.Coin minimum_stake = 2 [(gogoproto.nullable) = false];
+  // optional grace period to allow validators to join initial validator set after minimum stake is reached
+  // grace period is a duration in nanoseconds
+  uint64 grace_period = 3 [(gogoproto.moretags) = "yaml:\"grace_period\""];
+  // hash of the genesis file with no staking genesis. Staking genesis will be filled in using gen-txs of initial validator set
+  bytes genesis_hash = 4 [(gogoproto.moretags) = "yaml:\"genesis_hash\""];
+  // the sender address
+  string sender = 5;
+}
+
+// MsgSubmitNewChainResponse defines the Msg/SubmitNewChain response type.
+message MsgSubmitNewChainResponse {}
+
+// MsgBondStake defines a msg for a parent chain validator to stake on a baby chain.
+message MsgBondStake {
+  option (gogoproto.goproto_getters) = false;
+
+  // the chain-id of baby-chain
+  string chain_id = 1 [(gogoproto.moretags) = "yaml:\"chain_id\""];
+  // amount that validator wants to stake on baby chain. This is in addition to current stake on baby chain,
+  // thus the maximum they may stake on baby chain is: `total stake on parent chain - current stake on baby chain`
+  cosmos.base.v1beta1.Coin stake = 2 [(gogoproto.nullable) = false];
+  // the sender address
+  string sender = 3;
+}
+
+message MsgBondStakeResponse {}
+
+// MsgUnbondStake defines a msg for parent chain validator to unbond their stake on baby chain.
+message MsgUnbondStake {
+  option (gogoproto.goproto_getters) = false;
+
+  // the chain-id of baby-chain
+  string chain_id = 1 [(gogoproto.moretags) = "yaml:\"chain_id\""];
+  // amount that validator wants to unbond from baby chain. This may be a maximum of the current stake on baby chain
+   cosmos.base.v1beta1.Coin stake = 2 [(gogoproto.nullable) = false];
+  // the sender address
+  string sender = 3;
+}


modules/apps/ccv/types/expected_keepers.go
```diff
@@ -0,0 +1,17 @@
+package types
+
+// StakingKeeper defines the contract expected by parent-chain ccv module.
+// StakingKeeper is responsible for keeping track of latest validator set of all baby chains
+type StakingKeeper interface {
```
AdityaSripal, line 5: These are the responsibilities of the staking module. This is the only interface CCV will expect. Staking module may make the logic as complex or simple as they like. This will allow upgrades in cross-chain delegation functionality to occur with staking module upgrades without affecting CCV's IBC module

modules/apps/ccv/types/expected_keepers.go
```diff
@@ -0,0 +1,17 @@
...
+	GetValidatorSetChanges(chainID string) []abci.ValidatorUpdate
+	// This method is not required by CCV module explicitly but necessary for init protocol
+	GetInitialValidatorSet(chainID string) []sdk.Tx
+}
+
+// CNSKeeper defines the contract expected by parent-chain ccv module
+// CNSKeeper is responsible for verifying that chain creator is authorized to create a chain with given chain-id
+type CNSKeeper interface {
```
AdityaSripal, line 13: This will allow CNS to enforce that only chain-id owners create new chains with given chain-id. Other parent chains may choose to implement a no-op here, in which case CCV will simply ensure that a given baby chain-id is not already in use before accepting it.

proto/ibc/apps/cross-chain-validation/ccv.proto
```diff
@@ -0,0 +1,19 @@
...
+option go_package = "github.com/cosmos/ibc-go/modules/apps/ccv/types";
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+// This packet is sent from parent chain to baby chain if the validator set for baby chain
+// changes (due to new bonding/unbonding messages or slashing events)
+// The acknowledgement from baby chain will be sent asynchronously once unbonding period is over,
+// and this will function as `UnbondingOver` message for this packet.
+message ValidatorSetChangePacketData {
```
AdityaSripal, line 14: This is the only packet that will get sent across the ORDERED channel so sequence number is implicitly enforced by IBC.

We can implement the UnbondingOver message as an asynchronous acknowledgement of this packet. This fits very well with the semantic purpose of the acknowledgement and will make things simpler.

proto/ibc/apps/cross-chain-validation/tx.proto
```diff
@@ -0,0 +1,56 @@
+syntax = "proto3";
```
AdityaSripal, line 1: All of these messages will likely be implemented and processed in the new staking module, but I've defined them here for now.

proto/ibc/apps/cross-chain-validation/tx.proto
```diff
@@ -0,0 +1,56 @@
...
+package ibc.apps.ccv.v1;
+
+option go_package = "github.com/cosmos/ibc-go/modules/apps/ccv/types";
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+
+// MsgSubmitNewChain defines a msg to submit a new baby chain that can be validated
+// by the parent chain validators.
+message MsgSubmitNewChain {
```
AdityaSripal, line 12: This is my proposal for the init protocol. The owner of the chain-id can create this message along with non-stake genesis file. They can specify the minimum stake they would like, and then allow validators to post their gen-txs on the parent chain.

Once minimum stake and grace period have been reached, the initial validator set is finalized. All validators can retrieve gen-txs from parent chain, collect them into intitial genesis file and start the chain.

This minimizes the need for validator communication/coordination and only requires that 2/3 of validators come online for baby chain to start (as opposed to 100%)

Will write this up as a proposal to the spec once it is in ibc repo. cc: @milosevic @josef-widder

proto/ibc/apps/cross-chain-validation/tx.proto
```diff
@@ -0,0 +1,56 @@
...
+  bytes genesis_hash = 4 [(gogoproto.moretags) = "yaml:\"genesis_hash\""];
+  // the sender address
+  string sender = 5;
+}
+
+// MsgSubmitNewChainResponse defines the Msg/SubmitNewChain response type.
+message MsgSubmitNewChainResponse {}
+
+// MsgBondStake defines a msg for a parent chain validator to stake on a baby chain.
+message MsgBondStake {
```
AdityaSripal, line 32: These are the bonding/unbonding messages currently operating at the validator level. Though future improvements to staking module may push this to delegator level.

AdityaSripal (commented):
I've written up the basic interfaces I expect Interchain staking to require.

1.  Interface to users in `tx.proto` which will be sdk messages that users submit
2. Interface to other chains in `ccv.proto` which will be packets sent between parent blockchain and baby blockchain
3. Interface between modules in `expected_keepers.go` which will be the functions the CCV IBC module expects from the other modules on chain (staking/CNS so far)

There are 2 main outstanding pieces that I haven't included yet.

1. Evidence: We need to decide first if evidence processing will be responsibility of parent chain or baby chain. There is a tradeoff between security model and flexibility here. Parent chain responsibility provides greater guarantee that honest validator won't be maliciously slashed by malicious validator set on baby chain. Baby chain responsibility allows for application-specific evidence to be submitted.

2. Distribution: Fee/Reward distribution is a complicated piece that isn't specified yet. Baby chain will be in charge of inflating its own supply for rewards, and collecting fees for validator set. However, baby chain should not need to know full delegator set of parent chain. We should specify a way such that baby chain can keep track of the reward pool of each validator, and parent chain can distribute that reward pool among individual delegators appropriately. I haven't thought too much about how to do this yet. The simplest approach might require epoched staking

colin-axner:
Sorry, haven't read through any of this great documentation, but could the `ccv` proto types live on `cosmos/ibc` and then add a command to the Makefile to grab the proto from there and put it in `third_party`

I can help with importing the proto files, if needed, but I think it is best to start moving the shared proto files to `cosmos/ibc` as previously agreed

modules/apps/ccv/types/expected_keepers.go
```diff
@@ -0,0 +1,17 @@
+package types
+
+// StakingKeeper defines the contract expected by parent-chain ccv module.
+// StakingKeeper is responsible for keeping track of latest validator set of all baby chains
+type StakingKeeper interface {
+	GetValidatorSetChanges(chainID string) []abci.ValidatorUpdate
+	// This method is not required by CCV module explicitly but necessary for init protocol
+	GetInitialValidatorSet(chainID string) []sdk.Tx
```
fedekunze, line 8: why is this a tx slice instead of a Valset type? How do we represent validator sets for non-Tendermint chains on the leader chain? Should this also return the self height at which the validator set for the follower chain was initialized on the leader chain (baby = follower, parent = leader)

modules/apps/ccv/types/expected_keepers.go
fedekunze, line 5: I think this should be a separate module that imports the staking module. I think it's beneficial to keep the standalone and the cvv implementations in separate ctxs?

modules/apps/ccv/types/expected_keepers.go
```diff
@@ -0,0 +1,17 @@
...
+}
+
+// CNSKeeper defines the contract expected by parent-chain ccv module
+// CNSKeeper is responsible for verifying that chain creator is authorized to create a chain with given chain-id
```
fedekunze, line 12: It took me a couple minutes to realize you were talking about the chain name service / registry
```suggestion
// The Chain Name Service (CNS) is responsible for verifying that the chain creator is authorized to register a chain with the given identifier
```

modules/apps/ccv/types/expected_keepers.go
```diff
@@ -0,0 +1,17 @@
...
+type CNSKeeper interface {
+	AuthorizeChainCreator(chainID, creator string)
```
fedekunze, line 14: `creator` is a string to support different address formats (i.e bech32 HRP)?

proto/ibc/apps/cross-chain-validation/tx.proto
fedekunze: I'd keep the (un)delegation instead of (un)bond for consistency with the staking module

proto/ibc/apps/cross-chain-validation/tx.proto
fedekunze: redelegation is missing

proto/ibc/apps/cross-chain-validation/ccv.proto
fedekunze: I assume there will also be (un/re)delegation requests from follower chain -> leader chain? Should we write the packet data too for them?

proto/ibc/apps/cross-chain-validation/tx.proto
```diff
@@ -0,0 +1,56 @@
...
+  // hash of the genesis file with no staking genesis. Staking genesis will be filled in using gen-txs of initial validator set
+  bytes genesis_hash = 4 [(gogoproto.moretags) = "yaml:\"genesis_hash\""];
+  // the sender address
+  string sender = 5;
+}
+
+// MsgSubmitNewChainResponse defines the Msg/SubmitNewChain response type.
+message MsgSubmitNewChainResponse {}
+
+// MsgBondStake defines a msg for a parent chain validator to stake on a baby chain.
```
fedekunze, line 31: > MsgBondStake defines a msg for a parent chain validator to stake on a baby chain.

How do we handle delegation requests from follower -> leader chain?

fedekunze (commented):
great start!!

proto/ibc/apps/cross-chain-validation/ccv.proto
AdityaSripal: No I believe all (un/re)delegation messages should be sent to parent chain. The resulting valset changes will be communicated to baby chain through IBC packet

proto/ibc/apps/cross-chain-validation/tx.proto
AdityaSripal: redelegation in what sense? If you mean redelegation as in a delegator moving their stake from one validator to another, this is not really in purview of CCV. The staking keeper will keep track of this and propogate any valset changes to CCV.

If you mean a validator wants to move some of their stake away from babyChainA and stake those atoms on babyChainB, I think we could support that without waiting for the unbondingperiod to be over since the atoms will still be staked. Need to verify with spec team.

proto/ibc/apps/cross-chain-validation/tx.proto
AdityaSripal: I've kept it bond/unbond because this was an action taken at the validator level and not at delegator level, so I didn't want to confuse the two. I can change if there's consensus to keep it the same.

modules/apps/ccv/types/expected_keepers.go
```diff
@@ -0,0 +1,17 @@
+package types
+
+// StakingKeeper defines the contract expected by parent-chain ccv module.
+// StakingKeeper is responsible for keeping track of latest validator set of all baby chains
+type StakingKeeper interface {
+	GetValidatorSetChanges(chainID string) []abci.ValidatorUpdate
```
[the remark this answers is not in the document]
AdityaSripal: CCV will ensure that parent chain is not validating two chains named `ethermint`. It is responsibility of CNS that Tharsis is the only entity allowed to start a chain named ethermint, for example

modules/apps/ccv/types/expected_keepers.go
AdityaSripal: This will be a list of genesis transactions that can be queried from the parent chain.

They can then be collected into the genesis file of baby chain to populate all staking genesis info. This is part of the init protocol that I have roughly sketched out above but yet to fully specify

modules/apps/ccv/types/expected_keepers.go
AdityaSripal: Yes, I'm agnostic on whether this is built into x/staking or if there's a bridge module between CCV and staking.

A decision for @shahankhatch

modules/apps/ccv/types/expected_keepers.go
AdityaSripal: Correct

proto/ibc/apps/cross-chain-validation/tx.proto
AdityaSripal: Delegation only happens on leader chain

modules/apps/ccv/types/expected_keepers.go
AdityaSripal: Will change to Chain Name Service. However CCV is asking CNS if the creator can call `SubmitNewChain` with this chainID. For CNS to authorize successfully, the creator must have already registered the chainID

modules/apps/ccv/types/expected_keepers.go
colin-axner: +1 on separate module. The staking keeper is already complex enough

modules/apps/ccv/types/expected_keepers.go
colin-axner: validators can be initialized without a genesis transaction. They could be prespecified in the staking section of the genesis file

proto/ibc/apps/cross-chain-validation/ccv.proto
colin-axner: would be good to specify in the comments that this is to be used only on an ORDERED channel

proto/ibc/apps/cross-chain-validation/ccv.proto
```diff
@@ -0,0 +1,19 @@
...
+  repeated .tendermint.abci.ValidatorUpdate validator_updates = 1 [
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"validator_updates\""];
+}
```
colin-axner, line 18: is CCV only going to support Tendermint chains or is this a divergence from the ICS specs?

proto/ibc/apps/cross-chain-validation/ccv.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package ibc.apps.ccv.v1;
+
+option go_package = "github.com/cosmos/ibc-go/modules/apps/ccv/types";
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
```
colin-axner, line 8: why is coin imported?

modules/apps/ccv/types/expected_keepers.go
shahankhatch: +1 to having a separate module.

modules/apps/ccv/types/expected_keepers.go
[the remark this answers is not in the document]
shahankhatch: I suspect it's simpler to remove dependence on CNS for identifier management for the core ccv module, and to ensure that this chainID is unique only within ccv.
A custom module could then extend/wrap this module with CNS-based identifier management. It would also:
- remove any semantic complexity as CNS evolves how it handles chainIDs
- allow modular use with other identifier services, e.g., DID, addresses, keys

modules/apps/ccv/types/expected_keepers.go
shahankhatch: I suggest separating the authorization layers from the core ccv functionality. There are a few modules which are increasing the number of paths regarding authentication and authorization, and defining an auth/auth API for ccv now could lead to some re-work later.

proto/ibc/apps/cross-chain-validation/tx.proto
```diff
@@ -0,0 +1,56 @@
...
+message MsgSubmitNewChain {
+  option (gogoproto.goproto_getters) = false;
+  
+  // the proposed chain-id of new chain
+  string chain_id = 1 [(gogoproto.moretags) = "yaml:\"chain_id\""];
+  // minimum stake that must commit to validate new chain before it can start
+  cosmos.base.v1beta1.Coin minimum_stake = 2 [(gogoproto.nullable) = false];
+  // optional grace period to allow validators to join initial validator set after minimum stake is reached
+  // grace period is a duration in nanoseconds
+  uint64 grace_period = 3 [(gogoproto.moretags) = "yaml:\"grace_period\""];
```
shahankhatch, line 21: Time-based units in a module can lead to problems during upgrades or chain halts. In general, I suggest sticking to block-based units.

In any case, I suggest removing `grace_period` since it doesn't seem like prerequisite to core ccv functionality.

proto/ibc/apps/cross-chain-validation/tx.proto
shahankhatch: +1 to keeping with bond/unbond terminology

proto/ibc/apps/cross-chain-validation/tx.proto
shahankhatch: Sounds like gen-txs will require a separate tx message and type?

modules/apps/ccv/parent/keeper/relay.go
```diff
@@ -0,0 +1,80 @@
...
+		)
+	}
+
+	// Send ValidatorSet changes in IBC packet
+	packet := channeltypes.NewPacket(
+		packetDataBytes, sequence,
+		types.PortID, channelID,
+		channel.Counterparty.PortId, channel.Counterparty.ChannelId,
+		clienttypes.Height{}, uint64(types.GetTimeoutTimestamp(ctx.BlockTime()).UnixNano()),
+	)
```
cezarad, line 46: This can trigger OnTimeoutPacket. Could we have some ```None```  for the timestamp ? 
The current definition equals with the unbending time and does not consider any delays from the relayer (which comes at best empirical guarantees).  

For ordered channels we should close the channel due to an expired/ timeout valSetChange, which seems a bit harsh, since (at least for now) it's not a user value.

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,383 @@
...
+	if !am.keeper.AuthenticateCapability(ctx, chanCap, host.ChannelCapabilityPath(portID, channelID)) {
+		// Only claim channel capability passed back by IBC module if we do not already own it
+		if err := am.keeper.ClaimCapability(ctx, chanCap, host.ChannelCapabilityPath(portID, channelID)); err != nil {
+			return err
+		}
+	}
+
+	if err := am.keeper.VerifyChildChain(ctx, channelID); err != nil {
+		return err
+	}
```
cezarad, line 255: VerifyChildChain requires the channel to be ```ccv.Initializing```. Where is it set to this value ?

modules/apps/ccv/parent/types/utils.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+import "time"
+
+// TimeDelay is 4 weeks to ensure channel doesn't close on timeout
+const TimeDelay = 4 * 7 * 24 * time.Hour
+
+func GetTimeoutTimestamp(currentTime time.Time) time.Time {
+	return currentTime.Add(TimeDelay)
+}
```
cezarad, line 10: Equal to the un-bonding time, we should add an estimate for the relayer's delay.

modules/apps/ccv/types/errors.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// CCV sentinel errors
+var (
+	ErrInvalidPacketTimeout = sdkerrors.Register(ModuleName, 2, "invalid packet timeout")
```
cezarad, line 9: Not used.

modules/apps/ccv/types/errors.go
```diff
@@ -0,0 +1,16 @@
...
+	ErrInvalidVersion       = sdkerrors.Register(ModuleName, 3, "invalid CCV version")
+	ErrInvalidChannelFlow   = sdkerrors.Register(ModuleName, 4, "invalid message sent to channel end")
+	ErrInvalidChildChain    = sdkerrors.Register(ModuleName, 5, "invalid child chain")
+	ErrInvalidParentChain   = sdkerrors.Register(ModuleName, 6, "invalid parent chain")
```
cezarad, line 13: Not used.

modules/apps/ccv/types/genesis.go
```diff
@@ -0,0 +1,70 @@
...
+	return ChildGenesisState{
+		ParentChainId:        "testparentchainid",
+		NewChain:             true,
+		ParentClientState:    &ibctmtypes.ClientState{},
+		ParentConsensusState: &ibctmtypes.ConsensusState{},
+	}
+}
+
+// Validate performs basic genesis state validation returning an error upon any failure.
+// TODO: Validate UnbondingSequences
```
cezarad, line 40: Unclear who calls this ```Validate()```

modules/apps/ccv/types/genesis.go
```diff
@@ -0,0 +1,70 @@
...
+			return sdkerrors.Wrap(ErrInvalidGenesis, "parent consensus state cannot be nil for new chain")
+		}
+		if err := gs.ParentConsensusState.ValidateBasic(); err != nil {
+			return sdkerrors.Wrapf(ErrInvalidGenesis, "parent client state cannot be nil for new chain %s", err.Error())
+		}
+		if gs.ParentChannelId != "" {
+			return sdkerrors.Wrap(ErrInvalidGenesis, "parent channel id cannot be set for new chain. must be established on handshake")
+		}
+	} else {
+		if gs.ParentChannelId == "" {
```
cezarad, line 62: Maybe to discuss what is the restart scenario. If prior to the restart there were not enough client updates, on the baby chain for the parent chain, the client and consensus state might be too far behind to be trusted.

modules/apps/ccv/parent/keeper/relay.go
```diff
@@ -0,0 +1,80 @@
...
+	k.SetChannelStatus(ctx, packet.DestinationChannel, ccv.Invalid)
+	// TODO: Unbonding everything?
+	return nil
+}
+
+// EndBlockCallback is called for each baby chain in Endblock. It sends latest validator updates to each baby chain
+// in a packet over the CCV channel.
+func (k Keeper) EndBlockCallback(ctx sdk.Context, chainID string) bool {
+	valUpdates := k.registryKeeper.GetValidatorSetChanges(chainID)
+	k.SendPacket(ctx, chainID, valUpdates)
```
jovankomatovic, line 78: Shall we check whether valUpdates is empty before sending the packet?

modules/apps/ccv/child/keeper/genesis.go
```diff
@@ -0,0 +1,43 @@
...
+	k.SetParentChain(ctx, state.ParentChainId)
+	if state.NewChain {
+		// Create the parent client in InitGenesis for new child chain. CCV Handshake must be established with this client id.
+		clientID, err := k.clientKeeper.CreateClient(ctx, state.ParentClientState, state.ParentConsensusState)
+		panic(err)
+		// set parent client id.
+		k.SetParentClient(ctx, clientID)
+	} else {
+		// set parent channel id.
+		k.SetParentChannel(ctx, state.ParentChannelId)
```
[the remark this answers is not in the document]
AdityaSripal: This is checked in `ValidateGenesis`

modules/apps/ccv/parent/keeper/relay.go
```diff
@@ -0,0 +1,80 @@
...
+		return err
+	}
+	k.registryKeeper.UnbondValidators(ctx, chainID, data.ValidatorUpdates)
+	k.DeleteUnbondingChanges(ctx, chainID, packet.Sequence)
+	return nil
+}
+
+func (k Keeper) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet, data ccv.ValidatorSetChangePacketData) error {
+	k.SetChannelStatus(ctx, packet.DestinationChannel, ccv.Invalid)
+	// TODO: Unbonding everything?
```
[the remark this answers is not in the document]
AdityaSripal: correct, perhaps core IBC should enable sending packets with no timeouts to avoid this possibility

modules/apps/ccv/parent/keeper/relay.go
AdityaSripal: Currently, a timeout must be set for now, but I think in a future version of core IBC, we may do this.

> For ordered channels we should close the channel due to an expired/ timeout valSetChange

This does seem very harsh. What is the safety risk in the case where we don't receive valset update by the unbonding period? There doesn't seem to be any at first glance, but I could be wrong

modules/apps/ccv/parent/types/utils.go
AdityaSripal: I would expect that any time delay on normal relayer to be negligible with respect to unbonding period. Even if we give relayer 24 hrs, 28 days and 29 days doesn't matter so much

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,389 @@
...
+
+	if counterpartyVersion != ccv.Version {
+		return sdkerrors.Wrapf(ccv.ErrInvalidVersion, "invalid counterparty version: got: %s, expected %s", counterpartyVersion, ccv.Version)
+	}
+
+	// Module may have already claimed capability in OnChanOpenInit in the case of crossing hellos
+	// (ie chainA and chainB both call ChanOpenInit before one of them calls ChanOpenTry)
+	// If module can already authenticate the capability then module already owns it so we don't need to claim
+	// Otherwise, module does not have channel capability and we must claim it from IBC
+	if !am.keeper.AuthenticateCapability(ctx, chanCap, host.ChannelCapabilityPath(portID, channelID)) {
```
jovankomatovic, line 246: Can you, please, clarify how this can happen?
Thank you!

AdityaSripal:
Only accept a single ACK on baby chain, and guarantee a Confirm on the parent chain

modules/apps/ccv/types/genesis.go
```diff
@@ -0,0 +1,115 @@
...
+			return sdkerrors.Wrap(ErrInvalidGenesis, "parent channel id must be set for a restarting child genesis state")
+		}
+		if gs.ParentClientState != nil || gs.ParentConsensusState != nil {
+			return sdkerrors.Wrap(ErrInvalidGenesis, "parent client state and consensus states must be nil for a restarting genesis state")
+		}
+		for _, us := range gs.UnbondingSequences {
+			if err := us.Validate(); err != nil {
+				return sdkerrors.Wrap(err, "invalid genesis")
```
colin-axner, line 68:
```suggestion
				return sdkerrors.Wrap(err, "invalid unbonding sequences")
```

modules/apps/ccv/types/events.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+// IBC transfer events
```
colin-axner, line 3:
```suggestion
// CCV events
```

modules/apps/ccv/types/events.go
```diff
@@ -0,0 +1,12 @@
...
+const (
+	EventTypeTimeout      = "timeout"
+	EventTypePacket       = "ccv_packet"
+	EventTypeChannelClose = "channel_closed"
+
+	AttributeKeyAckSuccess = "success"
+	AttributeKeyAck        = "acknowledgement"
+	AttributeKeyAckError   = "error"
```
colin-axner, line 11: everything but the `EventTypePacket` look like they could be imported from 04-channel

modules/apps/ccv/types/errors.go
```diff
@@ -0,0 +1,20 @@
+package types
+
+import (
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// CCV sentinel errors
+var (
+	ErrInvalidPacketData        = sdkerrors.Register(ModuleName, 2, "invalid packet data")
```
colin-axner, line 9:
```suggestion
	ErrInvalidPacketData        = sdkerrors.Register(ModuleName, 2, "invalid CCV packet data")
```

modules/apps/ccv/types/codec.go
```diff
@@ -0,0 +1,35 @@
+package types
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+)
+
```
colin-axner, line 7: godoc in this file is outdated, references ibc transfer

modules/apps/ccv/types/ccv.go
```diff
@@ -0,0 +1,25 @@
...
+// ValidateBasic is used for validating the CCV packet data.
+func (vsc ValidatorSetChangePacketData) ValidateBasic() error {
+	if len(vsc.ValidatorUpdates) == 0 {
+		return sdkerrors.Wrap(ErrInvalidPacketData, "validator updates cannot be empty")
+	}
+	return nil
+}
+
+func (vsc ValidatorSetChangePacketData) GetBytes() []byte {
+	valUpdateBytes, _ := vsc.Marshal()
```
colin-axner, line 23: this should panic on error (use `MustMarshal`) and it should have godoc indicating the encoding type

modules/apps/ccv/types/ccv.go
```diff
@@ -0,0 +1,25 @@
...
+
+func NewValidatorSetChangePacketData(valUpdates []abci.ValidatorUpdate) ValidatorSetChangePacketData {
+	return ValidatorSetChangePacketData{
+		ValidatorUpdates: valUpdates,
+	}
+}
+
+// ValidateBasic is used for validating the CCV packet data.
+func (vsc ValidatorSetChangePacketData) ValidateBasic() error {
+	if len(vsc.ValidatorUpdates) == 0 {
```
colin-axner, line 16: do `ValidatorUpdates` have a validate function?

modules/apps/ccv/child/types/keys.go
```diff
@@ -0,0 +1,80 @@
+package types
+
+import (
+	"encoding/binary"
+	"time"
+)
+
+const (
+	// ModuleName defines the CCV child module name
+	ModuleName = "child"
```
colin-axner, line 10:
```suggestion
	ModuleName = "ccvchild"
```
?

colin-axner (commented):
still got lots more to review. Looking clean!

proto/ibc/applications/ccv/v1/genesis.proto
```diff
@@ -0,0 +1,49 @@
...
+
+import "ibc/lightclients/tendermint/v1/tendermint.proto";
+import "ibc/core/channel/v1/channel.proto";
+import "gogoproto/gogo.proto";
+import "tendermint/abci/types.proto";
+import "ibc/applications/ccv/v1/ccv.proto";
+
+// ChildGenesisState defines the CCV child chain genesis state
+message ChildGenesisState {
+  bool   disabled          = 1;
```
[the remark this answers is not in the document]
AdityaSripal: If childchain genesis is disabled then this chain is not a child chain of any other chain. It allows a binary to have the child chain module in it, while allowing individual chains to choose whether they want to enable that feature or not

modules/apps/ccv/types/genesis.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (us UnbondingSequence) Validate() error {
+	if us.UnbondingTime == 0 {
+		return sdkerrors.Wrap(ErrInvalidUnbondingSequence, "cannot have 0 unbonding time")
+	}
+	if err := us.UnbondingPacket.ValidateBasic(); err != nil {
+		return sdkerrors.Wrap(err, "invalid unbonding packet")
+	}
+	if us.UnbondingPacket.Sequence != us.Sequence {
```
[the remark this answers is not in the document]
AdityaSripal: Simplicity and easy to read/understand. I could remove it if desired

modules/apps/ccv/types/genesis.go
```diff
@@ -0,0 +1,115 @@
...
+func (gs ParentGenesisState) Validate() error {
+	for _, cs := range gs.ChildStates {
+		if err := cs.Validate(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (cs ChildState) Validate() error {
```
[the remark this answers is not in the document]
AdityaSripal: hmm well it could have some very basic validation. But it's valid to be nil

proto/ibc/applications/ccv/v1/genesis.proto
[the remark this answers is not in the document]
colin-axner: Oooohh. `ChildGenesisState` is the genesis state for the _module_ not a child chain (referencing one as a parent chain)

modules/apps/ccv/types/genesis.go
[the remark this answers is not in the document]
colin-axner: Up to you. Adding a `GetSequence` function seems just as easy and requires less correctness consideration, but I don't have strong opinions here

modules/apps/ccv/types/genesis.go
```diff
@@ -0,0 +1,115 @@
...
+// unless explicitly specified in genesis.
+func DefaultChildGenesisState() ChildGenesisState {
+	return ChildGenesisState{
+		Disabled: true,
+	}
+}
+
+// Validate performs basic genesis state validation returning an error upon any failure.
+func (gs ChildGenesisState) Validate() error {
+	if gs.Disabled {
```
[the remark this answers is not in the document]
colin-axner: It makes more sense now that I understand what a `ChildGenesisState` is

modules/apps/ccv/types/genesis.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		if gs.ParentChannelId != "" {
+			return sdkerrors.Wrap(ErrInvalidGenesis, "parent channel id cannot be set for new chain. must be established on handshake")
+		}
+		if gs.UnbondingSequences != nil {
+			return sdkerrors.Wrap(ErrInvalidGenesis, "unbonding sequences must be nil for new chain")
+		}
+	} else {
+		if gs.ParentChannelId == "" {
+			return sdkerrors.Wrap(ErrInvalidGenesis, "parent channel id must be set for a restarting child genesis state")
```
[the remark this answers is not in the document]
colin-axner: I completely misunderstood what a `ChildGenesisState` is. So this error message makes more sense now, but I think the terminology could be worked on since it took me this long to realize my misunderstanding

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+		// Only claim channel capability passed back by IBC module if we do not already own it
+		if err := am.keeper.ClaimCapability(ctx, chanCap, host.ChannelCapabilityPath(portID, channelID)); err != nil {
+			return err
+		}
+	}
+
+	am.keeper.SetChannelStatus(ctx, channelID, ccv.INITIALIZING)
+
+	if err := am.keeper.VerifyChildChain(ctx, channelID); err != nil {
```
jovankomatovic, line 256: Do we need this line if a channel was open before?

modules/apps/ccv/types/codec.go
```diff
@@ -0,0 +1,35 @@
...
+// RegisterLegacyAminoCodec registers the necessary x/ibc transfer interfaces and concrete types
+// on the provided LegacyAmino codec. These types are used for Amino JSON serialization.
+func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
```
colin-axner, line 10: This function can be removed. Amino is only needed to enable ledger signed txs but I don't see any messages in this module

modules/apps/ccv/types/codec.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// ModuleCdc references the global x/ibc-transfer module codec. Note, the codec
+	// should ONLY be used in certain instances of tests and for JSON encoding.
+	//
+	// The actual codec used for serialization should be provided to x/ibc transfer and
+	// defined at the application level.
+	ModuleCdc = codec.NewProtoCodec(codectypes.NewInterfaceRegistry())
+
+	// AminoCdc is a amino codec created to support amino json compatible msgs.
+	AminoCdc = codec.NewAminoCodec(amino)
```
colin-axner, line 29: I'd remove the amino references unless they are needed for some reason?

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+	version string,
+) error {
+	if order != channeltypes.ORDERED {
+		return sdkerrors.Wrapf(channeltypes.ErrInvalidChannelOrdering, "expected %s channel, got %s ", channeltypes.ORDERED, order)
+	}
+
+	// Require portID is the portID CCV module is bound to
+	boundPort := keeper.GetPort(ctx)
+	if boundPort != portID {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid port: %s, expected %s", portID, boundPort)
```
colin-axner, line 200:
```suggestion
		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "expected %s, got %s", boundPort, portID)
```
to be consistent with error message above

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	if version != ccv.Version {
+		return sdkerrors.Wrapf(ccv.ErrInvalidVersion, "got %s, expected %s", version, ccv.Version)
```
colin-axner, line 204:
```suggestion
		return sdkerrors.Wrapf(ccv.ErrInvalidVersion, "expected %s, got %s", ccv.Version, version)
```
ditto

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+	counterparty channeltypes.Counterparty,
+	version,
+	counterpartyVersion string,
+) error {
+	if err := ValidateParentChannelParams(ctx, am.keeper, order, portID, channelID, version); err != nil {
+		return err
+	}
+
+	if counterpartyVersion != ccv.Version {
+		return sdkerrors.Wrapf(ccv.ErrInvalidVersion, "invalid counterparty version: got: %s, expected %s", counterpartyVersion, ccv.Version)
```
colin-axner, line 240:
```suggestion
		return sdkerrors.Wrapf(ccv.ErrInvalidVersion, "invalid counterparty version: expected %s, got %s", ccv.Version, counterpartyVersion)
```

modules/apps/ccv/parent/module.go
colin-axner: I don't think it can happen. This was copied over from ics20 where it could happen. The first if statement should be removed. An error should be returned if the capability cannot be claimed

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+// OnChanOpenAck implements the IBCModule interface
+func (am AppModule) OnChanOpenAck(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+	counterpartyVersion string,
+) error {
+	return sdkerrors.Wrap(ccv.ErrInvalidChannelFlow, "channel handshake must be initiated by child chain")
```
colin-axner, line 269:
```suggestion
	return sdkerrors.Wrap(ccv.ErrInvalidChannelFlow, "channel handshake must be acknowledged by child chain")
```

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+// OnChanOpenConfirm implements the IBCModule interface
+func (am AppModule) OnChanOpenConfirm(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+) error {
+	err := am.keeper.SetChildChain(ctx, channelID)
```
colin-axner, line 278: `SetChildChain` returning an error breaks from common usage of a `Set` function. Maybe it should be renamed?

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+	return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "user cannot close channel")
+}
+
+// OnChanCloseConfirm implements the IBCModule interface
+func (am AppModule) OnChanCloseConfirm(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+) error {
+	return nil
```
colin-axner, line 301: Should something occur if the channel is closed?

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+// OnRecvPacket implements the IBCModule interface. A successful acknowledgement
```
colin-axner, line 304: godoc can be updated

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+// OnAcknowledgementPacket implements the IBCModule interface
+func (am AppModule) OnAcknowledgementPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	acknowledgement []byte,
+	_ sdk.AccAddress,
+) (*sdk.Result, error) {
+	var ack channeltypes.Acknowledgement
+	if err := ccv.ModuleCdc.UnmarshalJSON(acknowledgement, &ack); err != nil {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal parent packet acknowledgement: %v", err)
```
colin-axner, line 324:
```suggestion
		return nil, sdkerrors.Wrap(err, "cannot unmarshal parent packet acknowledgement")
```

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+	var data ccv.ValidatorSetChangePacketData
+	if err := ccv.ModuleCdc.UnmarshalJSON(packet.GetData(), &data); err != nil {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal parent packet data: %s", err.Error())
```
colin-axner, line 328:
```suggestion
		return nil, sdkerrors.Wrap(err, "cannot unmarshal parent packet data")
```

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+		return nil, err
+	}
+
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			ccv.EventTypePacket,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+			sdk.NewAttribute(ccv.AttributeKeyAck, ack.String()),
+		),
+	)
```
colin-axner, line 341: Is this needed? Shouldn't the ack be emitted in 04-keeper? If so, the ack should probably be hex encoded?

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+
+// OnTimeoutPacket implements the IBCModule interface
+func (am AppModule) OnTimeoutPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	_ sdk.AccAddress,
+) (*sdk.Result, error) {
+	var data ccv.ValidatorSetChangePacketData
+	if err := ccv.ModuleCdc.UnmarshalJSON(packet.GetData(), &data); err != nil {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal parent packet data: %s", err.Error())
```
colin-axner, line 373:
```suggestion
		return nil, sdkerrors.Wrapf(err, "cannot unmarshal parent packet data")
```

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+	if err := am.keeper.OnTimeoutPacket(ctx, packet, data); err != nil {
+		return nil, err
+	}
+
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			ccv.EventTypeTimeout,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+		),
+	)
```
colin-axner, line 385: I think this can be removed?

modules/apps/ccv/parent/module.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+// RegisterServices registers module services.
+// TODO
+func (am AppModule) RegisterServices(cfg module.Configurator) {
+}
+
+// InitGenesis performs genesis initialization for the parent module. It returns
+// no validator updates.
+// TODO
```
colin-axner, line 128: Looks like the todo can be removed?

modules/apps/ccv/parent/keeper/genesis.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/cosmos/ibc-go/modules/apps/ccv/parent/types"
+)
+
+func (k Keeper) InitGenesis(ctx sdk.Context) {
+	k.SetPort(ctx, types.PortID)
+
+	// Only try to bind to port if it is not already bound, since we may already own
+	// port capability from capability InitGenesis
+	if !k.IsBound(ctx, types.PortID) {
+		// transfer module binds to the transfer port on InitChain
```
colin-axner, line 16: comment needs updating

modules/apps/ccv/parent/keeper/keeper.go
```diff
@@ -0,0 +1,256 @@
...
+		registryKeeper:   registryKeeper,
+	}
+}
+
+// Logger returns a module-specific logger.
+func (k Keeper) Logger(ctx sdk.Context) log.Logger {
+	return ctx.Logger().With("module", "x/"+host.ModuleName+"-"+types.ModuleName)
+}
+
+// IsBound checks if the transfer module is already bound to the desired port
```
colin-axner, line 57: godoc for this file needs updating

modules/apps/ccv/parent/keeper/keeper.go
```diff
@@ -0,0 +1,256 @@
...
+func (k Keeper) IsBound(ctx sdk.Context, portID string) bool {
+	_, ok := k.scopedKeeper.GetCapability(ctx, host.PortPath(portID))
+	return ok
+}
+
+// BindPort defines a wrapper function for the ort Keeper's function in
```
colin-axner, line 63:
```suggestion
// BindPort defines a wrapper function for the Port Keeper's function in
```

modules/apps/ccv/parent/keeper/keeper.go
```diff
@@ -0,0 +1,256 @@
...
+	return ccv.Status(bz[0])
+}
+
+// VerifyChildChain verifies that the chain trying to connect on the channel handshake
+// is the expected baby chain.
+func (k Keeper) VerifyChildChain(ctx sdk.Context, channelID string) error {
+	// Verify CCV channel is in Initialized state
+	status := k.GetChannelStatus(ctx, channelID)
+	if status != ccv.INITIALIZING {
+		return sdkerrors.Wrap(ccv.ErrInvalidStatus, "CCV channel status must be in Initializing state")
```
colin-axner, line 185:
```suggestion
		return sdkerrors.Wrap(ccv.ErrInvalidStatus, "expected %s, got %s", ccv.INITIALIZING, status)
```

modules/apps/ccv/parent/keeper/keeper.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	_, tmClient, err := k.getUnderlyingClient(ctx, channelID)
+	if err != nil {
+		return err
+	}
+	// TODO: Verify consensus state against initial validator set of the child chain
+
+	// Verify that there isn't already a CCV channel for the child chain
+	if prevChannel, ok := k.GetChainToChannel(ctx, tmClient.ChainId); ok {
```
colin-axner, line 194:
```suggestion
	if prevChannel, found := k.GetChainToChannel(ctx, tmClient.ChainId); found {
```
?

modules/apps/ccv/parent/keeper/keeper.go
```diff
@@ -0,0 +1,256 @@
...
+		return sdkerrors.Wrapf(ccv.ErrDuplicateChannel, "CCV channel with ID: %s already created for child chain %s", prevChannel, tmClient.ChainId)
+	}
+	return nil
+}
+
+// SetChildChain ensures that the child chain has not already been set by a different channel, and then sets the child chain mappings in keeper,
+// and set the channel status to validating.
+// If there is already a ccv channel between the parent and child chain then close the channel, so that another channel can be made.
+func (k Keeper) SetChildChain(ctx sdk.Context, channelID string) error {
```
colin-axner, line 203: I'd recommend renaming this function since it does more than setting the child chain. I see why it is useful to wrap these checks before setting the child chain

modules/apps/ccv/parent/keeper/keeper.go
```diff
@@ -0,0 +1,256 @@
...
+	chainID, tmClient, err := k.getUnderlyingClient(ctx, channelID)
+	if err != nil {
+		return err
+	}
+	// Verify that there isn't already a CCV channel for the child chain
+	// If there is, then close the channel.
+	if prevChannel, ok := k.GetChannelToChain(ctx, chainID); ok {
```
colin-axner, line 210:
```suggestion
	if prevChannel, found := k.GetChannelToChain(ctx, chainID); found {
```

modules/apps/ccv/parent/keeper/keeper.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	connectionID := channel.ConnectionHops[0]
+	conn, ok := k.connectionKeeper.GetConnection(ctx, connectionID)
+	if !ok {
+		return "", nil, sdkerrors.Wrapf(conntypes.ErrConnectionNotFound, "connection not found for connection ID: %s", connectionID)
+	}
+	client, ok := k.clientKeeper.GetClientState(ctx, conn.ClientId)
+	if !ok {
+		return "", nil, sdkerrors.Wrapf(clienttypes.ErrClientNotFound, "client not found for client ID: %s", conn.ClientId)
+	}
```
colin-axner, line 240: use [`GetChannelClientState`](https://github.com/cosmos/ibc-go/blob/a647546486d162f0b58a0a505bfbc1dbbb68004f/modules/core/04-channel/keeper/keeper.go#L385)

modules/apps/ccv/parent/keeper/keeper.go
```diff
@@ -0,0 +1,256 @@
...
+	tmClient, ok := client.(*ibctmtypes.ClientState)
+	if !ok {
+		return "", nil, sdkerrors.Wrapf(clienttypes.ErrInvalidClientType, "invalid client type. expected %s, got %s", ibcexported.Tendermint, client.ClientType())
```
colin-axner, line 243:
```suggestion
		return "", nil, sdkerrors.Wrapf(clienttypes.ErrInvalidClientType, "expected %s, got %s", ibcexported.Tendermint, client.ClientType())
```

modules/apps/ccv/child/types/keys.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// UnbondingTimeKey returns the key for storing unbonding time for a given received packet sequence
+func UnbondingTimeKey(sequence uint64) []byte {
+	seqBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(seqBytes, sequence)
+	return append([]byte(UnbondingTimePrefix), seqBytes...)
+}
+
+func GetSequenceFromUnbondingTimeKey(key []byte) uint64 {
```
colin-axner, line 78: godoc

colin-axner (commented):
left only code cleanup suggestions. If this code is being moved to its own repository, I'd recommend doing that sooner than later. It isn't sustainable to review this pr or suggest changes as it stands. Reviewing small prs and opening issues will speed up development, especially as another contributor begins to onboard

Issue #119, opened 2021-04-13
IBC `ConnectionOpenAck` fails for solo machine client


devashishdxt:
## Summary of Bug

After creating an IBC solo machine client, the connection handshake fails at `ConnectionOpenAck`.

For a solo machine, in `MsgConnectionOpenAck`, we send `proof_height`, `proof_try`, `proof_client` and `proof_consensus`. On cosmos-sdk side, [this](https://github.com/cosmos/cosmos-sdk/blob/c4864e9f85011b3e971885ea995a0021c01a885d/x/ibc/core/03-connection/keeper/handshake.go#L195)  function is called to verify all the details of the message. For solo machine, `proof_try` contains signed connection data, `proof_client` contains signed client state of tendermint client on solo machine, `proof_consensus` contains signed consensus state of tendermint client on solo machine and `proof_height` is the height of the solo machine at which these proofs were created.

Each of these proofs are verified by cosmos chain in these functions:

- [`VerifyConnectionState`](https://github.com/cosmos/cosmos-sdk/blob/c4864e9f85011b3e971885ea995a0021c01a885d/x/ibc/core/03-connection/keeper/verify.go#L62)
- [`VerifyClientState`](https://github.com/cosmos/cosmos-sdk/blob/c4864e9f85011b3e971885ea995a0021c01a885d/x/ibc/light-clients/06-solomachine/types/client_state.go#L103)
- [`VerifyClientConsensusState`](https://github.com/cosmos/cosmos-sdk/blob/c4864e9f85011b3e971885ea995a0021c01a885d/x/ibc/core/03-connection/keeper/verify.go#L36)

All these functions are also supplied with `proof_height` which are verified against the `sequence` stored in solo machine client state on cosmos chain. But, all the above three functions increment the `sequence` stored in solo machine client state. [here](https://github.com/cosmos/cosmos-sdk/blob/c4864e9f85011b3e971885ea995a0021c01a885d/x/ibc/light-clients/06-solomachine/types/client_state.go#L207), [here](https://github.com/cosmos/cosmos-sdk/blob/c4864e9f85011b3e971885ea995a0021c01a885d/x/ibc/light-clients/06-solomachine/types/client_state.go#L132) and [here](https://github.com/cosmos/cosmos-sdk/blob/c4864e9f85011b3e971885ea995a0021c01a885d/x/ibc/light-clients/06-solomachine/types/client_state.go#L170).

[...]

PR #120, opened 2021-04-14
Fix solo machine handshake verification bug

colin-axner:
## Description

closes: #119 

---

write a little note why.

### DIFF
--- a/modules/light-clients/06-solomachine/types/client_state.go
+++ b/modules/light-clients/06-solomachine/types/client_state.go
@@ -100,7 +100,7 @@ func (cs ClientState) VerifyUpgradeAndUpdateState(
 
 // VerifyClientState verifies a proof of the client state of the running chain
 // stored on the solo machine.
-func (cs ClientState) VerifyClientState(
+func (cs *ClientState) VerifyClientState(
 	store sdk.KVStore,
 	cdc codec.BinaryMarshaler,
 	height exported.Height,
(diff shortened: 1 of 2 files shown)


colin-axner:
I don't know of an efficient way to test this in an end to end manner so I opened #121. We definitely need end to end tests for this, but I don't see a reason to block this fix unless someone is unconvinced this will fix the issue

fedekunze (approved):

AdityaSripal (approved):
utACK, this looks correct to me though it means that the solomachine must sign each message with a new sequence in the correct order during the handshake.

So, solomachine must sign connection with sequence 0, clientstate with sequence 1, and consensus state with sequence 2 in order to successfully pass `ConnOpenAck`.

This should be documented somewhere

devashishdxt:
@colin-axner Thanks for a quick fix. But, I'm still a bit unsure of how this fix solves the issue. I understand that solo machine has to increment the sequence after creating a proof. That means that a sequence/height should be passed along with each proof, but in `ConnectionOpenAck` (or any other call), all the three proofs are checked against `proofHeight` in [`produceVerificationArgs`](https://github.com/cosmos/ibc-go/blob/5d3a80060ed4ed78752596103d8200c3c07337cb/modules/light-clients/06-solomachine/types/client_state.go#L468).

Maybe I'm missing something, but, shouldn't we send a sequence number associated with each proof. 

NOTE: `SignBytes` does contain that sequence number but implementation verifies sequence against `proofHeight` and not the sequence number in `SignBytes`.

colin-axner:
@devashishdxt You are 100% correct that this didn't fix the issue, the proof height being passed in by the connection keeper should be discarded and we should only use the client state lastest sequence as done in the [ICS spec](https://github.com/cosmos/ibc/tree/master/spec/client/ics-006-solo-machine-client)

The proof height presumes that all the proofs are produced at the same height, but in the case of solo machines, this would be misbehaviour

I'll open a fix shortly. Thanks a ton for paying careful attention!! Hopefully end to end tests can mitigate these issues in the future

I'm also very interested in developer feedback for improving interaction with the solo machine since it was largely designed in a black box. Feel free to open any issues with suggestions!

colin-axner:
Actually upon further inspection, we need to verify that the proof height is correct since it used in timeouts. The ICS spec `verifyPacketReceiptAbsence` function contains a bug. Since it doesn't check the proof height against the latest sequence, a user could specify a proof height different than the sequence verified potentially resulting in timing out a packet that shouldn't have been timed out

PR #125, opened 2021-04-15
Efficient Consensus State Iteration

AdityaSripal:
## Description

This approach implements a backward compatible fix for efficient consensus state iteration without changing the host path, which is what the counterparty chain expects as the consensus state key.

- [x] Prune Consensus States
- [ ] ~~Enforce monotonic time~~ (will address in subsequent PR)

closes: #XXXX

---

write a little note why.


modules/light-clients/07-tendermint/types/update.go
```diff
@@ -60,6 +60,34 @@ func (cs ClientState) CheckHeaderAndUpdateState(
...
 
+	// Check the earliest consensus state to see if it is expired, if so then set the prune height
+	// so that we can delete consensus state and all associated metadata.
+	var (
+		pruneHeight exported.Height
+		pruneError  error
+	)
+	pruneCb := func(height exported.Height) bool {
+		consState, err := GetConsensusState(clientStore, cdc, height)
+		if err != nil {
```
AdityaSripal, line 71: This should never happen.

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -60,6 +60,34 @@ func (cs ClientState) CheckHeaderAndUpdateState(
...
+			pruneError = err
+			return true
+		}
+		if consState.IsExpired(cs.TrustingPeriod, ctx.BlockTime()) {
+			pruneHeight = height
+		}
+		return true
+	}
+	IterateConsensusStateAscending(clientStore, pruneCb)
```
AdityaSripal, line 80: This will fetch earliest consensus state, we can implement pruning now without breaking changes.

Can similarly test for monotonic time as well

modules/light-clients/07-tendermint/types/consensus_state.go
```diff
@@ -38,6 +38,12 @@ func (cs ConsensusState) GetTimestamp() uint64 {
 	return uint64(cs.Timestamp.UnixNano())
 }
 
+// IsExpired returns whether this consensus state is expired given the trusting period and current timestamp
+func (cs ConsensusState) IsExpired(trustingPeriod time.Duration, now time.Time) bool {
+	expirationTime := cs.Timestamp.Add(trustingPeriod)
+	return !expirationTime.After(now)
+}
```
colin-axner, line 45: Is this needed? Should we remove the `IsExpired` in the client state?

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -60,6 +60,34 @@ func (cs ClientState) CheckHeaderAndUpdateState(
...
+	)
+	pruneCb := func(height exported.Height) bool {
+		consState, err := GetConsensusState(clientStore, cdc, height)
+		if err != nil {
+			pruneError = err
+			return true
+		}
+		if consState.IsExpired(cs.TrustingPeriod, ctx.BlockTime()) {
+			pruneHeight = height
+		}
```
colin-axner, line 77: I'd just use the existing client state `IsExpired` function, doesn't make sense to me to maintain two copies

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -1,18 +1,25 @@
...
 )
 
-// KeyProcessedTime is appended to consensus state key to store the processed time
-var KeyProcessedTime = []byte("/processedTime")
+const KeyIterateConsensusStatePrefix = "iterateConsensusStates"
+
+var (
+	// KeyProcessedTime is appended to consensus state key to store the processed time
+	KeyProcessedTime = []byte("/processedTime")
+	KeyIteration     = []byte("/iterationKey")
```
colin-axner, line 21: Could you add a very long detailed comments at the top of this file explaining why we have `KeyIteration`, what purpose it serves and that (IMO) it should eventually be replaced by changing how the keys are stored (when we update to a breaking version of IBC)

I think we should also open an issue to track removing this workaround as I don't see this fix as a sustainable, but very useful in the short term

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -94,3 +107,151 @@ func GetProcessedTime(clientStore sdk.KVStore, height exported.Height) (uint64,
...
+				"invalid consensus type %T, expected %T", consensusState, &ConsensusState{},
+			)
+		}
+
+		if cb(*consensusState) {
+			break
+		}
+	}
+	return nil
+}
```
colin-axner, line 192: How is this descending? This is the same as Ascending except it returns the consensus state in the callback instead of the height

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -94,3 +107,151 @@ func GetProcessedTime(clientStore sdk.KVStore, height exported.Height) (uint64,
...
+
+func GetPreviousConsensusState(clientStore sdk.KVStore, cdc codec.BinaryMarshaler, height exported.Height) (*ConsensusState, bool) {
+	iterateStore := prefix.NewStore(clientStore, []byte(KeyIterateConsensusStatePrefix))
+	iterator := iterateStore.ReverseIterator(nil, bigEndianHeightBytes(height))
+	defer iterator.Close()
+
+	if !iterator.Valid() {
+		return nil, false
+	}
```
colin-axner, line 230: Why isn't `iterator.Next()` called? Can you add a comment?

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -94,3 +107,151 @@ func GetProcessedTime(clientStore sdk.KVStore, height exported.Height) (uint64,
...
+
+	consensusStateI, err := clienttypes.UnmarshalConsensusState(cdc, bz)
+	if err != nil {
+		return nil, false
+	}
+
+	consensusState, ok := consensusStateI.(*ConsensusState)
+	if !ok {
+		return nil, false
+	}
```
colin-axner, line 218: could we de-duplicate this?

modules/light-clients/07-tendermint/types/store_test.go
```diff
@@ -1,8 +1,12 @@
 package types_test
 
 import (
+	math "math"
```
colin-axner, line 4:
```suggestion
	"math"
```

modules/light-clients/07-tendermint/types/update.go
colin-axner: want to add an inline comment?

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -60,6 +60,34 @@ func (cs ClientState) CheckHeaderAndUpdateState(
...
+	IterateConsensusStateAscending(clientStore, pruneCb)
+	if pruneError != nil {
+		return nil, nil, pruneError
+	}
+	// if pruneHeight is set, delete consensus state and metadata
+	if pruneHeight != nil {
+		deleteConsensusState(clientStore, pruneHeight)
+		deleteProcessedTime(clientStore, pruneHeight)
+		deleteIterationKey(clientStore, pruneHeight)
+	}
```
colin-axner, line 89: Interesting, only pruning one height at a time? Shouldn't you just delete these in the callback so one iteration deletes all expired states?

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -60,6 +60,34 @@ func (cs ClientState) CheckHeaderAndUpdateState(
 		return nil, nil, err
 	}
 
+	// Check the earliest consensus state to see if it is expired, if so then set the prune height
+	// so that we can delete consensus state and all associated metadata.
+	var (
+		pruneHeight exported.Height
+		pruneError  error
+	)
+	pruneCb := func(height exported.Height) bool {
```
colin-axner, line 69: It is a little odd to me that we cannot prune consensus states for expired clients. I guess this can be fixed in a future improvment

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -60,6 +60,34 @@ func (cs ClientState) CheckHeaderAndUpdateState(
 		return nil, nil, err
 	}
 
+	// Check the earliest consensus state to see if it is expired, if so then set the prune height
+	// so that we can delete consensus state and all associated metadata.
```
colin-axner, line 64: can you add a test for this logic? I guess it can wait until the testing refactor part 2 is merged

colin-axner (commented):
I agree with the solution :+1: Clever fix!! I do also think this isn't a sustainable fix and is definitely a hack to support backwards compatibility. I think we should open a followup issue (or better yet, an ADR!) which explains in detail the sustainable fix so if IBC ever becomes non-backwards compatible we can be ready to swap out this solution for a more robust one that supports all light clients

I left various comments on areas for improvement in the code. I'd mostly like more code documentation explaining that this is a backwards compatible hack. I can see it being very confusing trying to understand why the code is constructed how it is and I think it is good to keep this code isolated so we can easily replace it when we are ready

modules/light-clients/07-tendermint/types/update.go
AdityaSripal: We can't just do variable size deletions because it will mess with gas costs. Here because we're doing store operations, each iteration consumes gas.

modules/light-clients/07-tendermint/types/store.go
AdityaSripal: Yes it's a confusing quirk in ReverseIterator that took me a while to figure out. Explaining in function go-doc

modules/light-clients/07-tendermint/types/update.go
AdityaSripal: Yes this can be added in future PR

modules/light-clients/07-tendermint/types/store.go
AdityaSripal: Sure, but i imagine this might just be something that sticks around. It's not very costly to do this (from a state space or developer maintanence point of view), and its a pain to change from a operational point of view.

It may get packaged with other breaking changes in IBC 2.0

AdityaSripal:
This PR implements (minimal) consensus state pruning. Future PRs will address monotonic time as well as more complete pruning solution

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -94,3 +124,133 @@ func GetProcessedTime(clientStore sdk.KVStore, height exported.Height) (uint64,
...
+	return consensusState, true
+}
+
+func bigEndianHeightBytes(height exported.Height) []byte {
+	revisionBytes := make([]byte, 8)
+	heightBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(revisionBytes, height.GetRevisionNumber())
+	binary.BigEndian.PutUint64(heightBytes, height.GetRevisionHeight())
+	return append(revisionBytes, heightBytes...)
+}
```
colin-axner, line 256: I believe it is more efficient to create one 16 byte slice. Appending the two slices causes a new slice to be allocated in order to double the capacity. 

I ran some benchmarks to verify.
Before:
```
Benchmark1-32    	 9852082	       110.4 ns/op	      40 B/op	       3 allocs/op
```
After:
```
Benchmark1-32    	18250534	        71.37 ns/op	      32 B/op	       2 allocs/op
```

Note, the first number is the amount of times the benchmark ran the for loop. As we can see allocating a 16 byte array is quite a bit faster. Almost twice as fast.

Here is the code I updated it to:
```go
func bigEndianHeightBytes(height exported.Height) []byte {
    heightBytes := make([]byte, 16) 
    binary.BigEndian.PutUint64(heightBytes[0:8], height.GetRevisionNumber())
    binary.BigEndian.PutUint64(heightBytes[8:], height.GetRevisionHeight())
    return heightBytes
}
```

Here is the benchmark I ran:

```go
package types_test

import (
    "testing"

    clienttypes "github.com/cosmos/ibc-go/modules/core/02-client/types"
    "github.com/cosmos/ibc-go/modules/light-clients/07-tendermint/types"
)

func Benchmark1(b *testing.B) {
    benchmarkBigEndian(b)
}

func benchmarkBigEndian(b *testing.B) {
    b.ReportAllocs()

    for i := 0; i < b.N; i++ {

        height := clienttypes.NewHeight(uint64(i), uint64(i))

        types.BigEndianHeightBytes(height)
    }   
}
```

modules/light-clients/07-tendermint/types/store.go
colin-axner: these errors are also missing codecov, but I believe de duplicating logic from the existing `GetConsensusState` function could allow reusal of that test to ensure codecov

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -60,6 +65,35 @@ func (cs ClientState) CheckHeaderAndUpdateState(
...
+	var (
+		pruneHeight exported.Height
+		pruneError  error
+	)
+	pruneCb := func(height exported.Height) bool {
+		consState, err := GetConsensusState(clientStore, cdc, height)
+		// this error should never occur
+		if err != nil {
+			pruneError = err
```
colin-axner, line 78: I think maybe we should wrap this error with some message that says "please open an issue". I wouldn't want a relayer to get this issue trying to debug it since really there'd need to be something really wrong

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -279,3 +280,66 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
+
+	// create 2 consensus states with the clientTime along with the metadata that would be stored by UpdateClient
+	consensusState := types.NewConsensusState(suite.clientTime, commitmenttypes.NewMerkleRoot(suite.header.Header.GetAppHash()), suite.valsHash)
+	suite.chainA.App.IBCKeeper.ClientKeeper.SetClientConsensusState(ctx, clientID, height, consensusState)
+	types.SetProcessedTime(suite.chainA.App.IBCKeeper.ClientKeeper.ClientStore(ctx, clientID), height, uint64(ctx.BlockTime().UnixNano()))
+	types.SetIterationKey(suite.chainA.App.IBCKeeper.ClientKeeper.ClientStore(ctx, clientID), height)
+	nextHeight := height.Increment()
+	suite.chainA.App.IBCKeeper.ClientKeeper.SetClientConsensusState(ctx, clientID, nextHeight, consensusState)
+	types.SetProcessedTime(suite.chainA.App.IBCKeeper.ClientKeeper.ClientStore(ctx, clientID), nextHeight, uint64(ctx.BlockTime().UnixNano()))
+	types.SetIterationKey(suite.chainA.App.IBCKeeper.ClientKeeper.ClientStore(ctx, clientID), nextHeight)
```
colin-axner, line 295: when you update to the new testing package, I think you can just create and update the client

colin-axner: I also don't think you should be manually setting the keys. We should test that regular client creation/updates set these properly

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -279,3 +280,66 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
+
+	// Set consensus state to a timestamp a week from now so that it is valid for the next consensus state which will be two weeks from now.
```
colin-axner, line 297: Why not just create a small trusting period?

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -279,3 +280,66 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
 		}
 	}
 }
+
+func (suite *TendermintTestSuite) TestPruneConsensusState() {
```
colin-axner, line 284: I think this test can be written in a cleaner fashion with the new testing package. I believe time increments are 5 seconds. So you should be able to create a client with 10 second trusting period. Update it once and then every subsequent update should prune the oldest consensus state. 

At the very least we should not be testing this function by manually setting the keys

colin-axner (commented):
LGTM, left some suggestions and nits

cwgoes:
This seems safe, but the extra layer of indirection is a bit wonky, I think we should try to implement a proper (backwards-incompatible) fix sometime in the future. Maybe it's worth keeping a list of "pending fixes" like this so we don't accumulate too technical debt and then lose track of what was necessary vs. what was merely done to avoid breaking backwards compatibility.

colin-axner:
> This seems safe, but the extra layer of indirection is a bit wonky, I think we should try to implement a proper (backwards-incompatible) fix sometime in the future. Maybe it's worth keeping a list of "pending fixes" like this so we don't accumulate too technical debt and then lose track of what was necessary vs. what was merely done to avoid breaking backwards compatibility.

I very much agree. I've added a `ibcv2` label as a start to keep track of issues that should be tackled when we are ready to do a backwards-incompatible fix

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -282,64 +281,65 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
-	_, _, err := clientState.CheckHeaderAndUpdateState(
-		currentCtx,
-		suite.cdc,
-		suite.chainA.App.IBCKeeper.ClientKeeper.ClientStore(currentCtx, clientID),
-		newHeader,
-	)
-	suite.Require().NoError(err)
+	path := ibctesting.NewPath(suite.chainA, suite.chainB)
+
+	clientState := path.EndpointA.GetClientState().(*types.ClientState)
```
colin-axner, line 286: you need to run `suite.coordinator.Setup(path)` first

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -282,64 +281,65 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
 
 	// check that second expired consensus state doesn't get deleted
 	// this ensures that there is a cap on gas cost of UpdateClient
-	consState, err = types.GetConsensusState(suite.chainA.App.IBCKeeper.ClientKeeper.ClientStore(currentCtx, clientID), suite.cdc, nextHeight)
-	suite.Require().Equal(consensusState, consState, "consensus state is unexpectedly pruned")
-	suite.Require().NoError(err)
-	// check processed time metadata is not pruned
-	processTime, ok = types.GetProcessedTime(suite.chainA.App.IBCKeeper.ClientKeeper.ClientStore(currentCtx, clientID), nextHeight)
-	suite.Require().Equal(uint64(ctx.BlockTime().UnixNano()), processTime, "processed time metadata is incorrect")
+	consState, ok = path.EndpointA.Chain.GetConsensusState(path.EndpointA.ClientID, pruneHeight)
```
colin-axner, line 335: looks like this should be expired height and same for below

modules/light-clients/07-tendermint/types/update_test.go
AdityaSripal: Good call, realized I forgot to SetIterationKey in UpdateClient because I was manually setting in tests. 😬

modules/light-clients/07-tendermint/types/update_test.go
AdityaSripal: Decided not to do this because I want to test that an Update only prunes a single consensus state. This requires non uniform time intervals between updates. So i had to use `TimeIncrementBy` so no point changing the Trusting period anyway

modules/light-clients/07-tendermint/types/update_test.go
AdityaSripal: Done! Thanks for suggestion!

modules/light-clients/07-tendermint/types/update.go
AdityaSripal: Hmm should I just log and ignore the issue. If this is caught in wild, it would suck that no updates can happen because clientstore can't prune properly. It should still just revert to old updating logic, while we implement a fix since this is a nice-to-have and not critical for IBC functioning

modules/light-clients/07-tendermint/types/update_test.go
AdityaSripal: No there are two expired consensus states. I'm calling the one that is pruned `pruneHeight` and the one that is expired but not pruned `expiredHeight`

modules/light-clients/07-tendermint/types/update.go
colin-axner: I'm ok with this. How do you plan to log the error?

AdityaSripal: I could have used logger in context.

I decided against this because logging instead of returning error would make it easy for this to be missed in unit tests. I'd rather just ensure our test coverage makes us confident about this, rather than implementing with mistakes and not catching them until deployment

colin-axner: Sounds like to correct decision to me. There is something severely wrong if this error occurs

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -229,7 +215,13 @@ func GetPreviousConsensusState(clientStore sdk.KVStore, cdc codec.BinaryMarshale
 	}
 
 	csKey := iterator.Value()
-	bz := clientStore.Get(csKey)
+
+	return getTmConsensusState(clientStore, cdc, csKey)
+}
+
+// Helper function for GetNextConsensusState and GetPreviousConsensusState
+func getTmConsensusState(clientStore sdk.KVStore, cdc codec.BinaryMarshaler, key []byte) (*ConsensusState, bool) {
```
colin-axner, line 223: GetConsensusState can make use of this as well

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -35,28 +38,19 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
 	testCases := []struct {
 		name    string
-		setup   func()
+		setup   func(suite *TendermintTestSuite)
 		expPass bool
 	}{
 		{
 			name: "successful update with next height and same validator set",
-			setup: func() {
+			setup: func(suite *TendermintTestSuite) {
```
colin-axner, line 53: what was the purpose of this? I'm pretty sure all the setup funcs have access to suite regardless

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -233,49 +228,123 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
+
+	// call update client twice. When pruning occurs, only first consensus state should be pruned.
+	// this height will be pruned
+	path.EndpointA.UpdateClient()
+	pruneHeight := path.EndpointA.GetClientState().GetLatestHeight()
+
+	// this height will be expired but not pruned
+	path.EndpointA.UpdateClient()
+	expiredHeight := path.EndpointA.GetClientState().GetLatestHeight()
+	// store expected values that must still remain in store after pruning
```
colin-axner, line 301:
```suggestion
```

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -233,49 +228,123 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
+	expectedConsState, ok := path.EndpointA.Chain.GetConsensusState(path.EndpointA.ClientID, expiredHeight)
+	suite.Require().True(ok)
+	ctx := path.EndpointA.Chain.GetContext()
+	clientStore := path.EndpointA.Chain.App.GetIBCKeeper().ClientKeeper.ClientStore(ctx, path.EndpointA.ClientID)
+	expectedProcessTime, ok := types.GetProcessedTime(clientStore, expiredHeight)
+	suite.Require().True(ok)
+	// check iteration key metadata is pruned
```
colin-axner, line 308:
```suggestion
```

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -233,49 +228,123 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
+	consKey := types.GetIterationKey(clientStore, pruneHeight)
+	suite.Require().Nil(consKey, "iteration key not pruned")
+
+	// check that second expired consensus state doesn't get deleted
+	// this ensures that there is a cap on gas cost of UpdateClient
+	consState, ok = path.EndpointA.Chain.GetConsensusState(path.EndpointA.ClientID, expiredHeight)
+	suite.Require().Equal(expectedConsState, consState, "consensus state incorrectly pruned")
+	suite.Require().True(ok)
+	// check processed time metadata is pruned
```
colin-axner, line 343:
```suggestion
	// check processed time metadata is not pruned
```

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -233,49 +228,123 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
+	processTime, ok = types.GetProcessedTime(clientStore, expiredHeight)
+	suite.Require().Equal(expectedProcessTime, processTime, "processed time metadata incorrectly pruned")
+	suite.Require().True(ok)
+	// check iteration key metadata is pruned
```
colin-axner, line 347:
```suggestion
	// check iteration key metadata is not pruned
```

colin-axner (approved):
Great work!! ACK pending CI fix and changelog entry

modules/light-clients/07-tendermint/types/update_test.go
AdityaSripal: Keeping this and updating comment

modules/light-clients/07-tendermint/types/update_test.go
AdityaSripal: Need to reset the suite on each case to avoid metadata writes messing things up. Have to pass in suite so that the function uses the current suite's signer values rather than original suite (which will be different)

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -180,7 +214,9 @@ func update(ctx sdk.Context, clientStore sdk.KVStore, clientState *ClientState,
 
 	// set context time as processed time as this is state internal to tendermint client logic.
 	// client state and consensus state will be set by client keeper
+	// set iteration key to provide ability for efficient ordered iteration of consensus states.
 	SetProcessedTime(clientStore, header.GetHeight(), uint64(ctx.BlockTime().UnixNano()))
+	SetIterationKey(clientStore, header.GetHeight())
```
colin-axner, line 219: I think we need this to be set during the upgrade function right? We should add a test as well, but it seems to me and upgrade would cause pruning to break as the code currently stands

Issue #98, opened 2021-03-30
Modify 'IsFrozen' to be a 'Status' function


colin-axner:
## Summary

Modify `IsFrozen` to become a `Status` function and block any packet processing if the status of the IBC client is not active. 

## Problem Definition

Core IBC does not account for specific light client notions. It blocks packet processing if the client is frozen but does not check if a tendermint is expired. This can lead to users accidentally sending packets via expired clients if their user interface doesn't make an explicit check before hand. If the counterparty client is expired as well then any tokens sent are effectively lost. 

## Proposal

Change `IsFrozen` to return a `Status` of the client. If the `Status` != exported.Active, then no packet processing should occur. Alternatively, we could modify `IsFrozen` to be `IsActive`

In order for 07-tendermint to determine if a client is expired, it must have access to the latest consensus state. In most situations this can be generalized to all light clients. It is possible a light client might need access to its own client store in order to use metadata set in that store to determine the status of the client. Either signature should be fine, but in order to get the latest consensus state, we need access to the client store so maybe it is best to supply the client store in the function signature. 

The benefit of using `Status()` as opposed to `IsActive()` is trivial support for a Status gRPC route. If the client state interface only supported `IsActive` then we would likely need to add a `Status` gRPC route for every light client as opposed to defining it once in 02-client. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #140, opened 2021-04-26
IsFrozen() changed to Status()

colin-axner:
## Description

closes: #98

---

write a little note why.


colin-axner:
Marking r4r, just need to add gRPC + gRPC test

modules/light-clients/07-tendermint/types/client_state.go
```diff
@@ -19,6 +19,11 @@ import (
 	"github.com/cosmos/ibc-go/modules/core/exported"
 )
 
+// Expired is a potential status of a client. A client is expired
+// if its latest consensus state timestamp added to its trusting period
+// is before or equal to the current time.
+const Expired = "Expired"
```
fedekunze, line 25: nit, this should be defined in exported too?

modules/core/exported/client.go
```diff
@@ -21,6 +21,15 @@ const (
...
 	Localhost string = "09-localhost"
+
+	// Active is a status type of a client. An active client is allowed to be used.
+	Active string = "Active"
+
+	// Frozen is a status type of a client. A frozen client is not allowed to be used.
+	Frozen string = "Frozen"
+
+	// Unknown indicates there was an error in determining the status of a client
+	Unknown string = "Unknown"
```
fedekunze, line 32: should these be proto enums?

modules/light-clients/06-solomachine/types/client_state.go
```diff
@@ -40,6 +40,18 @@ func (cs ClientState) GetLatestHeight() exported.Height {
 	return clienttypes.NewHeight(0, cs.Sequence)
 }
 
+// Status returns the status of the solo machine client.
+// The client may be:
+// - Active
+// - Frozen
```
fedekunze, line 46: nit, I'd briefly define the cases for each status
```suggestion
// - Active: if frozen sequence is 0
// - Frozen: otherwise
```

modules/light-clients/07-tendermint/types/client_state.go
```diff
@@ -58,9 +63,34 @@ func (cs ClientState) GetLatestHeight() exported.Height {
...
 }
 
-// IsFrozen returns true if the frozen height has been set.
-func (cs ClientState) IsFrozen() bool {
-	return !cs.FrozenHeight.IsZero()
+// Status returns the status of the tendermint client.
+// The client may be:
+// - Active
+// - Frozen
+// - Expired
```
fedekunze, line 70: ditto

fedekunze (approved):
changes LGTM, and thanks for the cleanup! I left some minor comments. Also, can you add a changelog entry?

modules/light-clients/07-tendermint/types/client_state.go
colin-axner: My reasoning is that this is a tendermint specific concept. Is expiry common amongst light clients? cc @AdityaSripal @cwgoes

modules/core/exported/client.go
colin-axner: No? They aren't used in communication. So I think it is probably simplest to leave them as strings. (only use proto when necessary)

modules/core/02-client/keeper/grpc_query_test.go
```diff
@@ -28,6 +28,12 @@ func (suite *KeeperTestSuite) TestQueryClientState() {
 		malleate func()
 		expPass  bool
 	}{
+		{"req is nil",
+			func() {
+				req = nil
+			},
+			false,
+		},
```
colin-axner, line 36: this test case panics using the old old testing setup. Another reason to use the new one :slightly_smiling_face:

proto/ibc/apps/transfer/v1/genesis.proto
```diff
@@ -9,11 +9,11 @@ import "gogoproto/gogo.proto";
 
 // GenesisState defines the ibc-transfer genesis state
 message GenesisState {
```
colin-axner, line 11: this is all from running `make proto-all`

modules/light-clients/06-solomachine/types/client_state_test.go
```diff
@@ -25,6 +25,18 @@ var (
 	consensusHeight = clienttypes.ZeroHeight()
 )
 
+func (suite *SoloMachineTestSuite) TestStatus() {
+	clientState := suite.solomachine.ClientState()
+	// solo machine discards arguements
+	status := clientState.Status(suite.chainA.GetContext(), nil, nil)
+	suite.Require().Equal(exported.Active, status)
+
+	// free solo machine
```
colin-axner, line 34:
```suggestion
	// freeze solo machine
```

modules/core/02-client/keeper/keeper_test.go
```diff
@@ -66,6 +66,7 @@ type KeeperTestSuite struct {
 	now            time.Time
 	past           time.Time
 
+	// TODO: deprecate
```
colin-axner, line 69: there's a couple tests left, will open an issue to track

testing/endpoint.go
```diff
@@ -433,6 +433,20 @@ func (endpoint *Endpoint) GetClientState() exported.ClientState {
...
+
+// GetConsensusState retrieves the Consensus State for this endpoint at the provided height.
+// The consensus state is expected to exist otherwise testing will fail.
+func (endpoint *Endpoint) GetConsensusState(height exported.Height) exported.ConsensusState {
+	consensusState, found := endpoint.Chain.GetConsensusState(endpoint.ClientID, height)
+	require.True(endpoint.Chain.t, found)
+
+	return consensusState
+}
+
```
AdityaSripal, line 449: Shouldn't we have `SetConsensusState`? I know it gets used a lot in tests

AdityaSripal (changes requested):
Good work!! I think Status should be a go enum, and it would be nice to have a `RefreshStatus` function with the arguments clientstore, cdc, etc.

And then just a getter function `Status` that returns the latest status set by `RefreshStatus`

colin-axner:
> it would be nice to have a `RefreshStatus` function with the arguments clientstore, cdc, etc.
> 
> And then just a getter function `Status` that returns the latest status set by `RefreshStatus`

I don't understand the usefulness of this. If you don't call `RefreshStatus` you run the risk of querying a stale status. Clients would then need to keep internal logic as to what height (and block time) they last refreshed their status. It seems simplest t to just always refresh the status

modules/core/exported/client.go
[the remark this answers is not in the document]
colin-axner: Can you elaborate on your concern of using strings? My primary concern of using enums is modularity and value registration

Enums are useful when you know the full range of constants (connection/channel state is a good example) since it provides uniqueness. We cannot know all types of status a client may have. The problem is that every light client developer would need to register their own enums in order to avoid registration conflict. Registering status types seems unnecessarily complicated without a concrete benefit. I think the performance impact of using constants should be pretty low (if any impact).

It would also make maintaining backwards compatibility difficult since we would either need to pre-reserve a selected amount of enums (otherwise we could never make a backwards compatible addition)

I tried using a middle ground solution of using a typed constant 
```
type Status string
```

This changes the function signature to use `exported.Status` instead of `string`, but it makes using the gRPC clunky since we cannot use `exported.Status` as a return type in the gRPC route. Thus we need to cast to/from the exported.Status when returning. Relayers and external third parties would rely on `string` (unless they decided to cast) whereas the code would rely on its typed string. 

I think my preference is continuing to use string. Client types are strings and we use those in identifiers. I don't see the benefit in increasing complexity and potentially creating a divide between what type is used in core IBC vs outside core IBC. 

Thoughts?

testing/endpoint.go
colin-axner: Sure, I've just been adding helper functions as needed

modules/core/02-client/keeper/client_test.go
```diff
@@ -42,141 +42,101 @@ func (suite *KeeperTestSuite) TestCreateClient() {
...
+	// Must create header creation functions since suite.header gets recreated on each test case
+	createFutureUpdateFn := func(trustedHeight clienttypes.Height) *ibctmtypes.Header {
+		header, err := suite.chainA.ConstructUpdateTMClientHeaderWithTrustedHeight(path.EndpointB.Chain, path.EndpointA.ClientID, trustedHeight)
+		suite.Require().NoError(err)
+		return header
+	}
+
+	createPastUpdateFn := func(fillHeight, trustedHeight clienttypes.Height) *ibctmtypes.Header {
+		consState, found := suite.chainA.App.GetIBCKeeper().ClientKeeper.GetClientConsensusState(suite.chainA.GetContext(), path.EndpointA.ClientID, trustedHeight)
+		suite.Require().True(found)
```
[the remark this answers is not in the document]
colin-axner: I added this because the consensus state is required to know the correct timestamp. The future function doesn't need to pass in a timestamp and thus the consensus state isn't needed. Checking for the existence of the consensus state would break a test case below, I'm fairly sure - "consensus state not found"

modules/core/exported/client.go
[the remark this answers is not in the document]
AdityaSripal: Thinking more on this, I guess I'm ok with strings.

My concern is there isn't a very clear global view on what a Status means, beyond just `Active` or some other value.

At the code level this is really just acting as a boolean.

I would be ok with even just an enum: `Active` or `Inactive`. And either log the specific issue for inactiveness, or allow it as a specific query on the client state.

But in the absence of that, strings are fine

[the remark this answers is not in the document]
colin-axner: I agree that `"Active"` is arbitrary and is likely something developers are capable of missing. Same with `"Frozen"`

The primary benefit for `Active` to be used not to be used as a boolean is trivial support for a gRPC route, as stated in the linked [issue](https://github.com/cosmos/ibc-go/issues/98):
```
The benefit of using Status() as opposed to IsActive() is trivial support for a Status gRPC route. If the client state interface only supported IsActive then we would likely need to add a Status gRPC route for every light client as opposed to defining it once in 02-client.
```

What about this as the function signature:
```
Status(ctx, clientStore, cdc) exported.Status, string
```
Where exported.Status is an enum of either `Active, Frozen, InActive` and string is any additional information such as "Expired"?

Core IBC only uses the string in error messages and for gRPC support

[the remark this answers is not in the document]
AdityaSripal: I like this idea! Though I think we just need `Active, Inactive` no? Frozen is a particular type of inactive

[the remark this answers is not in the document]
colin-axner: > I like this idea! Though I think we just need `Active, Inactive` no? Frozen is a particular type of inactive

I believe #141 requires knowing if the client is Frozen? That's the only use case that comes to mind. No strong preference on this decision. Happy to do whichever you think is best

AdityaSripal:
My concern is if I want to make a Status check, I need to have `ctx, clientStore, cdc`. If i don't have them (let's say im in a stateless helper function), then I need to add those arguments to my helper function. Or i need to make the status check earlier even if it doesn't make the most sense from a code-organization POV.

My proposal was that you call `RefreshStatus` once at the start of each client handler. And then all subsequent logic can just use `Status` and get the latest status. I think that's a bit cleaner, but I don't have a strong preference so I leave it to your decision

colin-axner:
I agree `Status` is a hefty function. Here's my current reasoning. 

The primary motivation for this change was based on 2 issues
- developer demand for a standard way of determining the client status
- preventing verification/packet sends on clients which are not active

There are a few constraints:
- requiring clientstore, cdc, and ctx to determine status
- cannot add fields to `ClientState` without breaking IBC
- gRPC callers need to set context block time

There's no way around the first constraint, thus the question is whether we should cache the status

Trying to cache the status in the client state causes us to deal with backwards compatibility. I believe it isn't possible to do a backwards compatible fix without introducing client versions. Client versions are useful regardless, but that fix is out of scope of this issue. Furthermore, a modification to the client state proto file should really require an ICS spec change

gRPC callers need to set the current time in the context. This might be difficult for non-go implementations. I'm not sure if it is possible for relayers like `ibc-rs` to set sdk context specific fields. There's no way around this, unless we set the context time to `time.Now` in the gRPC call. I'm not sure of the ramifications of this. 

Lets say we wanted to cache the status. We know that because the context is passed in, we can probably assume each status is tied to a block height (of the current chain, note this block height is not the same as the consensus state block height as they represent different chains). Thus we could create a mapping from block height -> status. This doesn't solve the gRPC constraint since queries at a height which didn't have a status set would need to depend on the context block time which may or may not be set. If a status isn't set for a block height, it is not safe to assume it has the same status as the next/prev block height which has a status set. 

Using the client store mapping, a helper function would still need access to the client store and ctx (and potentially cdc)

Based on the above reasoning, without any specific use case in mind for caching the status, I'd prefer to do the simplest solution possible. Otherwise we increase the amount of things all light clients would need to do or break IBC. I think the heftiness of `Status` may be revisited so I think it is best to build a first step solution that is compatible with a future extension. 

I like the naming of `Status` as it is given the function signature since it returns the current status based on the context. 

I'm not really sure how you could cache the status within the client state since the status is dependent upon the block height it was updated against. Therefore a client status would need to keep its current status and the block height it was updated against, otherwise it is impossible to verify if the status is stale

> My proposal was that you call RefreshStatus once at the start of each client handler. And then all subsequent logic can just use Status and get the latest status

My main concern is there is no clear contract here. The `Status` function will likely be accessible without requiring `RefreshStatus`. You'd need to add a check to ensure the status isn't stale or increase the risk for bugs within core IBC or by external developers

modules/core/exported/client.go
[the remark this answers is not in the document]
colin-axner: Going with `Active, Frozen, Inactive` since misbehaviour/freezing is already a concept in core IBC, it makes sense to have this status type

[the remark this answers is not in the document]
colin-axner: Also going to use protobuf enums. Turns out go doesn't have an enum type. The only way to replicate enums is via a typed constant, but then you have to add a to/from conversion function to get the string representation. Using proto enums seems like the most sensible option

modules/light-clients/07-tendermint/types/client_state.go
cwgoes: Expiry will be common to light clients other than Tendermint, yes. Anything for proof-of-stake should have this.

modules/core/exported/client.go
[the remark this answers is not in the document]
colin-axner: I decided to use a typed string. Initially I implemented a typed int32 and the function signature above (hence the force push), but I realized separating the status and info felt unnecessary. The main critique in this thread is separating a string from a status and a typed string accomplishes that. 

It also is self descriptive. The value `1` indicates `Frozen` a lot less than the string `"Frozen"`. Furthermore, backwards compatibility will be easier since the next value added wouldn't have to be `3`. 

Since we are using a typed string, I kept `Active, Frozen, Expired, Unknown` as the default supported status types. Core IBC should only ever look for `Active` and potentially `Frozen` in exceptional cases. 

I tried using proto enums but proto enums will just create a typed int32 and requires the active status to be suffixed with `_UNSPECIFIED` per proto linting. In the gRPC response, since I couldn't use a typed Status, I went with string as I figured relayers would prefer to compare against a self descriptive string rather than an arbitrary number. 

I think there are also less risks for errors, since core IBC defensively uses the status (must be Active to do anything). If I used a typed int32, a typed status return value of 0 would always be interpreted as Active by core IBC. For example
```
var inactive exported.Status = 0
return inactive
```

With a typed string, returning a typed status of `"Active"` is self descriptive. You wouldn't use that return value to indicate the client is inactive. 

This last part is a minor nit, but a typed int32 just felt like unnecessary indirection.

AdityaSripal (approved):
Looks good, clean fix!

PR #141, opened 2021-04-26
Time Monotonicity Enforcement

AdityaSripal:
## Description

- [x] enforce time monotoncity
- [x] create time misbehaviour
- [x] early return on identical updates
- [x] misbehaviour detection within updateclient

closes: #XXXX

---

write a little note why.


modules/core/02-client/keeper/client.go
```diff
@@ -61,63 +61,80 @@ func (k Keeper) UpdateClient(ctx sdk.Context, clientID string, header exported.H
...
+		// set default consensus height with header height
+		consensusHeight = header.GetHeight()
+
+	}
+
+	k.SetClientState(ctx, clientID, newClientState)
+	// If client state is not frozen after clientState CheckHeaderAndUpdateState,
+	// then write the update state changes, and set new consensus state.
+	// Else the update was proof of misbehaviour and we must emit appropriate misbehaviour events.
+	if !newClientState.IsFrozen() {
```
AdityaSripal, line 95: There was too much confusion regarding separation of responsibilities for detecting misbehaviour here. Because conflicting header can be detected here, but time monotonicity can't. Thus, it makes more sense to just make it the responsibility of client developers to do this correctly so we have clear separation of responsibility.

Here i just check if new clientstate is frozen and if so emit appropriate events/write state

I think resulting code is cleaner

AdityaSripal (commented):
I've decided to keep the `FrozenHeight` to being the height of the offending header that caused misbehaviour. With time misbehaviour i use the larger height. Since we no longer use its specific value, it doesn't really matter so long as its non-zero. @colin-axner argues for just making the height {0-1} since it isn't used for anything other than a boolean value.

I figure if we're keeping it, we may as well make it meaningful. Care to weigh in @cwgoes ?

modules/core/02-client/keeper/client_test.go
```diff
@@ -68,6 +68,7 @@ func (suite *KeeperTestSuite) TestUpdateClientTendermint() {
 		name     string
 		malleate func() error
 		expPass  bool
+		freeze   bool // only true if update freezes the client to a new frozen height. false if client is already frozen and header is valid past update
```
colin-axner, line 71: I refactored this test in #140 to use the new testing package since consensus states were not being stored for all client states resulting status to be `Unknown` on valid cases. 

Could you copy over those changes and adjust?

modules/core/02-client/keeper/client_test.go
```diff
@@ -114,29 +115,39 @@ func (suite *KeeperTestSuite) TestUpdateClientTendermint() {
...
+			// updateHeader will fill in consensus state between prevConsState and suite.consState
+			// clientState should not be updated
+			updateHeader = createPastUpdateFn(suite)
+			// set updateHeader's consensus state in store to create duplicate UpdateClient scenario
+			suite.keeper.SetClientConsensusState(suite.ctx, clientID, updateHeader.GetHeight(), updateHeader.ConsensusState())
 			return nil
-		}, false},
-		{"valid past update before client was frozen", func() error {
+		}, true, false},
+		{"misbehaviour detection: conflicting header", func() error {
```
colin-axner, line 149: This test case be reduced once you copy over the changes:
```go
{"misbehaviour", func() {
            // perform update and then replicate update
            clientState := path.EndpointA.GetClientState()
            trustedHeight := clientState.GetLatestHeight().(types.Height)
            path.EndpointA.UpdateClient()
            clientState = path.EndpointA.GetClientState()
            fillHeight := clientState.GetLatestHeight().(types.Height)
            updateHeader = createPastUpdateFn(fillHeight, trustedHeight)
        }, false},
        ```

File: modules/light-clients/07-tendermint/types/misbehaviour_handle.go [hunk 4]
```diff
@@ -30,12 +31,37 @@ func (cs ClientState) CheckMisbehaviourAndUpdateState(
 		return nil, sdkerrors.Wrapf(clienttypes.ErrInvalidClientType, "expected type %T, got %T", misbehaviour, &Misbehaviour{})
 	}
 
-	// If client is already frozen at earlier height than misbehaviour, return with error
-	if cs.IsFrozen() && cs.FrozenHeight.LTE(misbehaviour.GetHeight()) {
-		return nil, sdkerrors.Wrapf(clienttypes.ErrInvalidMisbehaviour,
-			"client is already frozen at earlier height %s than misbehaviour height %s", cs.FrozenHeight, misbehaviour.GetHeight())
+	// If client is already frozen, return with error
+	if cs.IsFrozen() {
+		return nil, sdkerrors.Wrapf(clienttypes.ErrInvalidMisbehaviour, "client is already frozen at height %s", cs.FrozenHeight)
```
colin-axner, line 36: we can remove this. This is checked in 02-client

modules/light-clients/07-tendermint/types/misbehaviour_handle_test.go
```diff
@@ -50,15 +50,30 @@ func (suite *TendermintTestSuite) TestCheckMisbehaviourAndUpdateState() {
...
+		{
+			"valid time misbehaviour",
+			types.NewClientState(chainID, types.DefaultTrustLevel, trustingPeriod, ubdPeriod, maxClockDrift, height, commitmenttypes.GetSDKSpecs(), upgradePath, false, false),
+			types.NewConsensusState(suite.now, commitmenttypes.NewMerkleRoot(tmhash.Sum([]byte("app_hash"))), bothValsHash),
+			height,
+			types.NewConsensusState(suite.now, commitmenttypes.NewMerkleRoot(tmhash.Sum([]byte("app_hash"))), bothValsHash),
+			height,
+			&types.Misbehaviour{
+				Header1:  suite.chainA.CreateTMClientHeader(chainID, int64(height.RevisionHeight+3), height, suite.now, bothValSet, bothValSet, bothSigners),
+				Header2:  suite.chainA.CreateTMClientHeader(chainID, int64(height.RevisionHeight+1), height, suite.now, bothValSet, bothValSet, bothSigners),
```
colin-axner, line 76: can you add a test where header 1 time is < header 2

modules/light-clients/07-tendermint/types/misbehaviour_handle_test.go
```diff
@@ -323,8 +368,8 @@ func (suite *TendermintTestSuite) TestCheckMisbehaviourAndUpdateState() {
 			types.NewConsensusState(suite.now, commitmenttypes.NewMerkleRoot(tmhash.Sum([]byte("app_hash"))), bothValsHash),
 			height,
 			&types.Misbehaviour{
-				Header1:  suite.chainA.CreateTMClientHeader(chainID, int64(height.RevisionHeight), height, suite.now, altValSet, bothValSet, altSigners),
-				Header2:  suite.chainA.CreateTMClientHeader(chainID, int64(height.RevisionHeight), height, suite.now.Add(time.Minute), altValSet, bothValSet, altSigners),
+				Header1:  suite.chainA.CreateTMClientHeader(chainID, int64(height.RevisionHeight+1), height, suite.now, altValSet, bothValSet, altSigners),
+				Header2:  suite.chainA.CreateTMClientHeader(chainID, int64(height.RevisionHeight+1), height, suite.now.Add(time.Minute), altValSet, bothValSet, altSigners),
```
colin-axner, line 372: can you adjust line 400 to use `suite.chainA.App.AppCodec()`. We should be consistent in either using the old or new testing functions (preferably the new so it is easier to maintain)

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -191,12 +192,19 @@ func GetNextConsensusState(clientStore sdk.KVStore, cdc codec.BinaryMarshaler, h
 	iterateStore := prefix.NewStore(clientStore, []byte(KeyIterateConsensusStatePrefix))
 	iterator := iterateStore.Iterator(bigEndianHeightBytes(height), nil)
 	defer iterator.Close()
-	// ignore the consensus state at current height and get next height
-	iterator.Next()
 	if !iterator.Valid() {
 		return nil, false
 	}
 
+	// if iterator is at current height, ignore the consensus state at current height and get next height
```
colin-axner, line 199: can you update the godoc?

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -100,6 +105,70 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
 			},
 			expPass: true,
 		},
+		{
+			name: "successful update with identical header to a previous update",
```
colin-axner, line 109: what is this testing?

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -100,6 +105,70 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
+			name: "misbehaviour detection: previous consensus state time is not before header time. time monotonicity violation",
+			setup: func(suite *TendermintTestSuite) {
+				clientState = types.NewClientState(chainID, types.DefaultTrustLevel, trustingPeriod, ubdPeriod, maxClockDrift, height, commitmenttypes.GetSDKSpecs(), upgradePath, false, false)
+				// swap header and client time to create violation
+				consensusState = types.NewConsensusState(suite.clientTime, commitmenttypes.NewMerkleRoot(suite.header.Header.GetAppHash()), suite.valsHash)
+				newHeader = suite.chainA.CreateTMClientHeader(chainID, int64(heightPlus5.RevisionHeight), height, suite.headerTime, suite.valSet, suite.valSet, signers)
+				currentTime = suite.now
+				prevConsensusState := types.NewConsensusState(suite.headerTime, commitmenttypes.NewMerkleRoot(suite.header.Header.GetAppHash()), suite.valsHash)
+				ctx := suite.chainA.GetContext().WithBlockTime(currentTime)
+				suite.chainA.App.GetIBCKeeper().ClientKeeper.SetClientConsensusState(ctx, clientID, heightPlus1, prevConsensusState)
```
colin-axner, line 148: it's hard to verify this test. There appears to be 3 times:
- client time
- header time
- suite.now

I'm not sure what is before or after without looking into the helper file. Can you add comments as to why this test functions correctly?

I'd assume header time to be after client time. The header is made with the header time, the consensus state is made with the client time, and the previous consensus state is made with header time and the context is set to suite.now. What is the consensus state used for?

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -100,6 +105,70 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
...
+				clientStore := suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(ctx, clientID)
+				types.SetIterationKey(clientStore, heightPlus1)
+			},
+			expFrozen: true,
+			expPass:   true,
+		},
+		{
+			name: "misbehaviour detection: next consensus state time is not after header time. time monotonicity violation",
```
colin-axner, line 156: ditto

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -100,6 +105,70 @@ func (suite *TendermintTestSuite) TestCheckHeaderAndUpdateState() {
 			},
 			expPass: true,
 		},
+		{
+			name: "successful update with identical header to a previous update",
+			setup: func(suite *TendermintTestSuite) {
+				clientState = types.NewClientState(chainID, types.DefaultTrustLevel, trustingPeriod, ubdPeriod, maxClockDrift, heightPlus1, commitmenttypes.GetSDKSpecs(), upgradePath, false, false)
```
colin-axner, line 111: It worries me a little to use so much manual setup as opposed to just running `SetupClients`, but I guess updating these tests can be done another day

colin-axner (commented):
Overall LGTM, nice work!

colin-axner:
changelog entry is needed (I keep forgetting as well)

modules/core/02-client/keeper/client.go
AdityaSripal: > I don't understand why we cache the context if it is now the full responsibility of the client developers to handle all instances of misbehaviour correctly

They are responsible for telling core IBC if an update was misbehaviour. They are not responsible for rolling back all state changes.

> What if a client wanted to write metadata everytime it handled misbehaviour in an update client message?

This is definitely possible, i guess its up to us what we want to enable. The downside is accidentally leaving in metadata writes after misbehaviour that we intend to write only for valid updates. Think @cwgoes can weigh in on tradeoff between flexibility and opinionated code. I believe being opinionated here and having the ClientKeeper write metadata on valid update makes more sense. Light client implementations are fully responsible for doing update logic (UpdateClient will do none of that).
But ClientKeeper will take the returned output, and do all of the necessary store writes. I think that's a clean separation of responsibility.

> I'd actually prefer to be as defensive as possible. In which case, we should keep the cached context and return an error if a duplicate update occurs without the client detecting misbehavior

I tried doing this and the code got super ugly because there was freezing logic in both the ClientKeeper and in tendermint's update function. That could have been much cleaner if it was just in one place.

Furthermore, I think it's possible to take all client developer checks that must be done by every light client and put them in the ClientKeeper to minimize the possibility of light-client developer error.
But I think in practice, this would make things less secure if it trades off too much on separation of concerns.
Critically, I think it just needs to be clear to a reviewer/developer where a particular check is supposed to happen.
My proposal is that we create a very clear separation of concern that acts as a contract between core IBC and light client developer.

Light client implementation must give core IBC the updated clientstate/consensus state. And it **must** return a frozen client state if the update was evidence of misbehaviour.
Core IBC will in turn store the clientstate (and consensus if valid update), write all callback state changes on successful updates, and emit appropriate events.

This means that there may be redundant checks happening in light clients, that may be missed by some of them. But it gives a very clear rule for what a light client implementation is responsible for. Even though i place responsibility of all misbehaviour checks on light client. As a reader and reviewer I can analyze the light-client implementation **in isolation** and check that it is catching all misbehaviour and holding up its side of bargain.

Without this, I need to be checking whether ClientKeeper+misbehaviour together are catching all misbehaviour. And I need to make sure together they don't miss a gap between them. And that they aren't doing redundant checks. It's also harder as time goes on to determine where a check should go. We would need to make a subjective decision on whether we think some check is universal or not.

For these reasons I think clear separation of concerns is more important than putting all universal checks (even subtle ones) in the ClientKeeper. But yes, this should absolutely be documented in `light_client.md`. Will do so once there's consensus on this point

modules/core/02-client/keeper/client.go
```diff
@@ -178,8 +196,12 @@ func (k Keeper) CheckMisbehaviourAndUpdateState(ctx sdk.Context, misbehaviour ex
 		return sdkerrors.Wrapf(types.ErrClientNotFound, "cannot check misbehaviour for client with ID %s", misbehaviour.GetClientID())
 	}
 
-	if clientState.IsFrozen() && clientState.GetFrozenHeight().LTE(misbehaviour.GetHeight()) {
-		return sdkerrors.Wrapf(types.ErrInvalidMisbehaviour, "client is already frozen at height ≤ misbehaviour height (%s ≤ %s)", clientState.GetFrozenHeight(), misbehaviour.GetHeight())
+	if clientState.IsFrozen() {
+		return sdkerrors.Wrap(types.ErrInvalidMisbehaviour, "client is already frozen")
+	}
+
+	if err := misbehaviour.ValidateBasic(); err != nil {
```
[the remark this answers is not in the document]
AdityaSripal: My concern is if misbehaviour is passed in from another module rather than through the SDK msg that we designed

modules/light-clients/07-tendermint/types/misbehaviour_handle.go
```diff
@@ -68,6 +94,7 @@ func (cs ClientState) CheckMisbehaviourAndUpdateState(
 	}
 
 	cs.FrozenHeight = tmMisbehaviour.GetHeight().(clienttypes.Height)
```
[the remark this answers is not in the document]
AdityaSripal: The frozen height is the height of the offending header. In the case of time misbehaviour, it is the existence of the larger height that breaks monotonicity.

This feeds into the intuition we had with submitting misbehaviour that lowers frozen height. That logic rested on the understanding that frozen height was height of offending header. You could lower frozen height by submitting proof of an earlier offending header.

No longer matters now, but if we choose to re-enable this would follow that intuition.

modules/core/02-client/keeper/client.go
AdityaSripal: Regardless of if we allow metadata writes on misbehaviour, we still want to cache so we can discard on error.

Developers shouldn't be forced to revert state themselves on error

modules/core/02-client/keeper/client.go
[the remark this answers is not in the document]
AdityaSripal: Out of scope for this PR, let's open separate issue to discuss if there's disagreement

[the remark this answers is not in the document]
colin-axner: > My concern is if misbehaviour is passed in from another module rather than through the SDK msg that we designed

This can be applied to all our handlers. Is there a use case in mind? We could turn these functions into message servers such that we aren't exposing functions that don't take messages

modules/light-clients/07-tendermint/types/update_test.go
AdityaSripal: It's testing that submitting a header for which the consensus state already exists passes (no-op)

modules/light-clients/07-tendermint/types/misbehaviour_handle.go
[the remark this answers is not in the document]
colin-axner: Couldn't an attacker construct a fill in header causing the larger height to become the offending header despite it being signed by > 66% of validators and thus steal funds since the frozen height would have been at the larger height

modules/light-clients/07-tendermint/types/update_test.go
AdityaSripal: The consensus state with header time is just used to create the time violation. I create a monotonic time violation by setting a consensus state at a different height to be the same time as header.

I still need a valid consensus state in order to even pass the update. Note: the trusted height of header points to the valid consensus state at client time.

Documented to make things clearer.

Agreed that client time and header time can be confusing, but it's not just in this test case but throughout tendermint package. future testing pr should resolve this

modules/core/02-client/keeper/client.go
colin-axner: I think you make great points. 

> it must return a frozen client state if the update was evidence of misbehaviour.

I agree with this.

> As a reader and reviewer I can analyze the light-client implementation in isolation and check that it is catching all misbehaviour and holding up its side of bargain.

I like this, and I think we can still achieve this with a duplicate check. My concern is that allowing a duplicate update at an existing height is a critical security vulnerability and I'm hesitant to let it go by when we have the capacity to do the check. This is the code I have in mind:

```go
consState, exists := keeper.GetConsensusState()

newClientState, newConsensusState, err := CheckHeaderAndUpdateState()
if err != nil {
    return err
}

// write client state, errors returned later revert state changes
switch {
case: newCilentState.IsFrozen()
    // use logic you have
case: exists && !reflect.DeepEqual(consState, newConsensusState)
    // return err, light client implementation missed misbehaviour handling
default:
    // regular update code
}
```

I don't see why this code gets ugly? It allows light clients to fully implement misbehaviour logic without relying on 02-client and it allows 02-client to prevent duplicate updates which are misbehaviour

modules/core/02-client/keeper/client.go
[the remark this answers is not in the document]
AdityaSripal: Suppose with Interchain Staking I get a packet of Evidence which is chain-wide misbehaviour. I could freeze the underlying client during processing

modules/core/02-client/keeper/client.go
colin-axner: > Regardless of if we allow metadata writes on misbehaviour, we still want to cache so we can discard on error.
> 
> Developers shouldn't be forced to revert state themselves on error

Do you have the use case in mind that update is being called by an external module? Messages that result in errors always have state changes reverted

AdityaSripal: > Do you have the use case in mind that update is being called by an external module? Messages that result in errors always have state changes reverted by baseapp. I think this is a safe assumption to make

Oh yes you're correct about this. We should only cache if we discard metadata on misbehaviour

modules/core/02-client/keeper/client_test.go
AdityaSripal: I don't have this function: `suite.chainA.ConstructUpdateTMClientHeaderWithTrustedHeight`

modules/light-clients/07-tendermint/types/misbehaviour.go
```diff
@@ -35,8 +34,6 @@ func (misbehaviour Misbehaviour) GetClientID() string {
 }
 
 // GetHeight returns the height at which misbehaviour occurred
-//
-// NOTE: assumes that misbehaviour headers have the same height
 func (misbehaviour Misbehaviour) GetHeight() exported.Height {
 	return misbehaviour.Header1.GetHeight()
 }
```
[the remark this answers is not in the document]
AdityaSripal: Out of scope, let's create an issue and discuss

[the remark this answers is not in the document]
colin-axner: see #148

modules/core/02-client/keeper/client.go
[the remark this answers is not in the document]
colin-axner: see #150

modules/core/02-client/keeper/client_test.go
```diff
@@ -42,216 +42,186 @@ func (suite *KeeperTestSuite) TestCreateClient() {
...
 
+			height3 := types.NewHeight(0, 3)
 			// updateHeader will fill in consensus state between prevConsState and suite.consState
 			// clientState should not be updated
-			updateHeader = createPastUpdateFn(suite)
+			updateHeader = createPastUpdateFn(height3, height1)
 			// set updateHeader's consensus state in store to create duplicate UpdateClient scenario
-			suite.keeper.SetClientConsensusState(suite.ctx, clientID, updateHeader.GetHeight(), updateHeader.ConsensusState())
-			return nil
+			suite.chainA.App.GetIBCKeeper().ClientKeeper.SetClientConsensusState(suite.chainA.GetContext(), clientID, updateHeader.GetHeight(), updateHeader.ConsensusState())
```
colin-axner: I think in the future it is best to avoid setting all the consensus states manually, but this is fine. I think setting the update headers consensus state is fine, but the previous/intermediate consensus state (setup code) should just be done by calling `path.EndpointA.UpdateClient`. I think it helps in maintainability/readability in testing code

In my opinion, hard coding the heights is generally harder to maintain as opposed to referencing the latest height before/after an update.

modules/core/02-client/keeper/client_test.go
```diff
@@ -42,216 +42,186 @@ func (suite *KeeperTestSuite) TestCreateClient() {
...
+			clientState := path.EndpointA.GetClientState().(*ibctmtypes.ClientState)
+			clientID := path.EndpointA.ClientID
+			trustedHeight := clientState.GetLatestHeight().(types.Height)
 
 			// store intermediate consensus state at a time greater than updateHeader time
 			// this will break time monotonicity
-			incrementedClientHeight := testClientHeight.Increment().(types.Height)
+			incrementedClientHeight := clientState.GetLatestHeight().Increment().(types.Height)
 			intermediateConsState := &ibctmtypes.ConsensusState{
 				Timestamp:          suite.header.Header.Time.Add(2 * time.Hour),
```
colin-axner, line 164: `suite.header.Header.Time.Add(2 * time.Hour)` should be `suite.coordinator.CurrentTime.Add(2 * time.Hour)`

colin-axner (commented):
LGTM, waiting to approve once we resolve decisions on
- using cache ctx (no preference)
- returning error if client did not handle misbehaviour (duplicate update produces different consensus state)

I have a preference for doing the consensus state check, but willing to concede if others consider it unnecessary

cwgoes:
> I figure if we're keeping it, we may as well make it meaningful. Care to weigh in @cwgoes ?

Just to refresh my memory, the goal here is backwards compatibility, right? Otherwise it seems odd to me to keep a field with a range (integer) far greater than the set of semantically meaningful values (0, non-zero). For backwards compatibility I guess there isn't too much difference between 0, 1 and 0, nonzero height - although the latter does run the risk of appearing to be more semantically relevant than it actually is - and we don't really want IBC users to start using this value for other kinds of data processing, if they really care about misbehaviour heights they should read them from the event logs - those are both some (weak) reasons to prefer 0, 1 I think.

modules/core/02-client/keeper/client_test.go
AdityaSripal: These are all past updates so it should be ok

modules/light-clients/07-tendermint/types/misbehaviour_handle.go
```diff
@@ -88,7 +89,7 @@ func (cs ClientState) CheckMisbehaviourAndUpdateState(
 		return nil, sdkerrors.Wrap(err, "verifying Header2 in Misbehaviour failed")
 	}
 
-	cs.FrozenHeight = tmMisbehaviour.GetHeight().(clienttypes.Height)
+	cs.FrozenHeight = clienttypes.NewHeight(0, 1)
```
colin-axner, line 92: can we have a const `FrozenHeight`? This would be useful in tests

AdityaSripal: Struct can't be a constant but I can make it a var inside tendermint package

modules/core/02-client/keeper/client.go
cwgoes: > My concern is that allowing a duplicate update at an existing height is a critical security vulnerability and I'm hesitant to let it go by when we have the capacity to do the check.

Here's a question - is this always true? Certainly it is a problem if unequal consensus states at the same height would allow for violation of exactly-once packet delivery guarantees or timeouts, but there could conceivably be client types which _allow_ duplicate consensus states, just not verification at them (so they are only intermediate update points) - for example, a (non-Tendermint) consensus algorithm could have a block history which looks like this:

blocks

Is this a case we want to consider? There is something to be said for not overly constraining what it means for clients to be "correct", since clients implement all of the packet data / timeout / etc. verification functions anyways.

colin-axner: Great question! I didn't realize intermediate update points were a possibility. 

In light of our [discussion yesterday](https://github.com/cosmos/ibc/issues/569#issuecomment-832051686), I don't see the usefulness of adding this check if in the near future, light client implementations will be responsible for getting/setting client/consensus states. In this design, light clients should definitely be aware to guard against duplicate updates which constitute misbehaviour

colin-axner (approved):
Excellent work! In light of our [discussion yesterday](https://github.com/cosmos/ibc/issues/569#issuecomment-832051686), I think the current solution aligns perfectly with future changes

colin-axner:
@AdityaSripal I fixed the merge conflicts. Everything was straightforward except for the check in update client for if the client is frozen. I update the code to be semantically equivalent:

Before
```go
if !client.IsFrozen {

} else {

}
```

after
```go
if status := client.Status; status != exported.Frozen {

} else {

}
```

It is just slightly odd because we aren't checking if the client is active, just that it is not frozen

modules/light-clients/07-tendermint/types/misbehaviour_handle.go
```diff
@@ -32,6 +34,32 @@ func (cs ClientState) CheckMisbehaviourAndUpdateState(
...
+		if err != nil {
+			return nil, sdkerrors.Wrap(err, "invalid block ID from header 2 in misbehaviour")
+		}
+
+		// Ensure that Commit Hashes are different
+		if bytes.Equal(blockID1.Hash, blockID2.Hash) {
+			return nil, sdkerrors.Wrap(clienttypes.ErrInvalidMisbehaviour, "headers block hashes are equal")
+		}
+	} else {
+		// Header1 is at greater height than Header2, therefore Header1 time must be less than or equal to
```
cwgoes, line 54: just to confirm, do we check this somewhere?

cwgoes (approved):
ACK modulo quick confirmation question

modules/light-clients/07-tendermint/types/misbehaviour_handle.go
AdityaSripal: We check it on `ValidateBasic` of Misbehaviour

PR #144, opened 2021-04-28
Hex Packet Data

AdityaSripal:
## Description

Encoding packet data in hex allows for more robust encoding/decoding for relayers. Proto-encoded packets were breaking with old bare string conversion. Raw binary can be passed through events using hex.

Tested live with go-relayer and two-chainz script

closes: #XXXX

---

write a little note why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### Improvements
 
 * (modules/core/04-channel) [\#7949](https://github.com/cosmos/cosmos-sdk/issues/7949) Standardized channel `Acknowledgement` moved to its own file. Codec registration redundancy removed.
+* (modules/core/04-channel) [\#144](https://github.com/cosmos/ibc-go/pull/144) Introduced a `packet_data_hex` attribute to emit the hex-encoded packet data in events. This allows for raw binary (proto-encoded message) to be sent over events and decoded correctly on relayer. Original `packet_data` is DEPRECATED. All relayers and IBC event consumers are encouraged to switch to `packet_data_hex` as soon as possible.
 * (modules/light-clients/07-tendermint) [\#125](https://github.com/cosmos/ibc-go/pull/125) Implement efficient iteration of consensus states and pruning of earliest expired consensus state on UpdateClient.
 
 ## IBC in the Cosmos SDK Repository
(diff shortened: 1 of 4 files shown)


colin-axner (changes requested):
I disagree with the proposed solution. Applications should hex encode before passing it to core IBC. We should check that the packet data is utf-8 in packet ValidateBasic, if anything

AdityaSripal:
cc relayer teams: @ethanfrey @ancazamfir

`packet_data` attribute is getting deprecated in favor of `packet_data_hex`. This will allow app developers to marshal their packet data with protobuf and relay without any issues

AdityaSripal:
@colin-axner Why? That doesn't make any sense to me. If I want to protobuf encode my packet data i should just be able to do that.

With your proposal, I either shouldn't use protobuf. Or I need to do a secondary encoding to make it utf-8, and then decode twice in my application RecvPacket callback. This doesn't make any sense to me.

Packet Data is raw data (marshaled in whatever preference of the app developer), it's not a string. It's up to us as core IBC to make sure that raw data gets to the other end correctly so that it can be unmarshaled as expected by app developer. It's not up to app developer to make their raw data look like a "string" for our convenience.

colin-axner (approved):
I gave it some more thought and this solution sounds good since it only changes how events are emitted.

If applications are using proto to marshal their packet data, then I don't see why other applications won't run into the same issue we are facing here. Is there usefulness in not hex encoding packet data before passing it to core IBC?

colin-axner:
can you add a section to the migrations docs?

AdityaSripal:
> If applications are using proto to marshal their packet data, then I don't see why other applications won't run into the same issue we are facing here. Is there usefulness in not hex encoding packet data before passing it to core IBC?

This fix makes it such that applications that use proto to marshal won't face this issue at all. Assuming relayers upgrade to use the new attribute. So regardless of the encoding format that applications choose, IBC will work. As I tested, proto-marshalling transfer packet data works seamlessly if you encode/decode with hex. It breaks when you try to just do a raw conversion.

Without this fix, we are limiting application choice on encoding without a good reason. Note applications are returning a byte slice, not a string in `packet.GetData`. It's still incumbent on core IBC to turn that into a string to emit in events. And the obvious choice is to use a standardized encoding like hex rather than relying on `string()` and have applications run into this issue again.

ethanfrey:
Thanks for the quick fix.

Issue #100, opened 2021-03-30
Implement hybrid packet delays


colin-axner:
## Summary

Implement spec [changes](https://github.com/cosmos/ibc/pull/552)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #171, opened 2021-05-11
Block Delay Parameter

AdityaSripal:
## Description

- Implements Block Delay
- Sets consensus metadata on updateproposal and upgradeclient

closes: #100 💯 

---

write a little note why.


modules/light-clients/07-tendermint/types/client_state.go
```diff
@@ -499,18 +507,29 @@ func (cs ClientState) VerifyNextSequenceRecv(
...
+	// check that executing chain's timestamp has passed consensusState's processed time + delay time period
 	processedTime, ok := GetProcessedTime(store, proofHeight)
 	if !ok {
 		return sdkerrors.Wrapf(ErrProcessedTimeNotFound, "processed time not found for height: %s", proofHeight)
 	}
-	validTime := processedTime + delayPeriod
-	// NOTE: delay period is inclusive, so if currentTimestamp is validTime, then we return no error
+	validTime := processedTime + delayTimePeriod
+	// NOTE: delay time period is inclusive, so if currentTimestamp is validTime, then we return no error
 	if validTime > currentTimestamp {
```
colin-axner, line 518: nit: `currentTimestamp < validTime` to match if statement below. I think it might be easier to read

modules/light-clients/07-tendermint/types/errors.go
```diff
@@ -10,16 +10,17 @@ const (
...
-	ErrInvalidValidatorSet    = sdkerrors.Register(SubModuleName, 13, "invalid validator set")
+	ErrInvalidChainID          = sdkerrors.Register(SubModuleName, 2, "invalid chain-id")
+	ErrInvalidTrustingPeriod   = sdkerrors.Register(SubModuleName, 3, "invalid trusting period")
+	ErrInvalidUnbondingPeriod  = sdkerrors.Register(SubModuleName, 4, "invalid unbonding period")
+	ErrInvalidHeaderHeight     = sdkerrors.Register(SubModuleName, 5, "invalid header height")
+	ErrInvalidHeader           = sdkerrors.Register(SubModuleName, 6, "invalid header")
+	ErrInvalidMaxClockDrift    = sdkerrors.Register(SubModuleName, 7, "invalid max clock drift")
+	ErrProcessedTimeNotFound   = sdkerrors.Register(SubModuleName, 8, "processed time not found")
+	ErrProcessedHeightNotFound = sdkerrors.Register(SubModuleName, 9, "processed height not found")
+	ErrDelayPeriodNotPassed    = sdkerrors.Register(SubModuleName, 10, "packet-specified delay period has not been reached")
```
colin-axner, line 21: Going forward, I would consider changing error numbers a breaking change since others might be relying on the number value. I think understanding errors as a relayer is useful functionality. I think it is best to just append even if the ordering isn't logical

modules/core/03-connection/keeper/verify.go
```diff
@@ -159,9 +161,17 @@ func (k Keeper) VerifyPacketCommitment(
 		return sdkerrors.Wrapf(clienttypes.ErrClientNotActive, "client (%s) status is %s", clientID, status)
 	}
 
+	// calculate minimum block delay by dividing time delay period
+	// by the maximum expected time per block. Round up the block delay.
+	timeDelay := connection.GetDelayPeriod()
+	maxTimePerBlock := k.GetMaxTimePerBlock(ctx)
+	blockDelay := uint64(math.Ceil(float64(timeDelay) / float64(maxTimePerBlock)))
```
colin-axner, line 168: I think this deserves a private helper function

modules/core/03-connection/keeper/verify.go
```diff
@@ -159,9 +161,17 @@ func (k Keeper) VerifyPacketCommitment(
...
+
+	// verify that block delay has passed
```
colin-axner, line 170:
```suggestion
```
? seems misleading since the `VerifyPacketCommitment` is doing much more than just the block delay

modules/core/03-connection/keeper/verify.go
```diff
@@ -195,9 +205,15 @@ func (k Keeper) VerifyPacketAcknowledgement(
 		return sdkerrors.Wrapf(clienttypes.ErrClientNotActive, "client (%s) status is %s", clientID, status)
 	}
 
+	// calculate minimum block delay by dividing time delay period
+	// by the maximum expected time per block. Round up the block delay.
+	timeDelay := connection.GetDelayPeriod()
+	maxTimePerBlock := k.GetMaxTimePerBlock(ctx)
+	blockDelay := uint64(math.Ceil(float64(timeDelay) / float64(maxTimePerBlock)))
```
colin-axner, line 212: this enforces block delays even for time delays of 0 (since it rounds up). Is there documentation for why you chose to round up?

For the case where time delay > 0 and < maxTimePerBlock, the packet cannot be processed until the next block anyways, regardless of block delay since the time delay is > 0. So the cases we only care about are time delay == 0 and when timeDelay > MaxTimePerBlock and timeDelay % MaxTimePerBlock != 0

Lets say MaxTimePerBlock is 10s and my time delay is 35s. Rounding up puts us at 4 blocks and rounding down puts us at 3 blocks. In the usual worst case scenario, 3 blocks would be produced in 30 seconds, thus even if the block delay was 3, we would still need to wait for the 4th block. So I guess it doesn't really matter whether we round up or down for this case

For timeDelay == 0, we need to add an if statement if we round up, but no if statement is needed if we round down. This seems like the only reason to influence which way to round

Thoughts?

modules/core/03-connection/keeper/verify.go
colin-axner: what happens if maxTimePerBlock is 0? I know that isn't a valid param, but in the test `params_test.go` you set it to 0 as a test. I guess the code would panic?

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -131,6 +134,43 @@ func deleteProcessedTime(clientStore sdk.KVStore, height exported.Height) {
 	clientStore.Delete(key)
 }
 
+// ProcessedHeight Store code
```
colin-axner, line 137: My preference would be to remove this comment and same comment for processed time, the function names in `store.go` seem self descriptive to me

colin-axner (commented):
LGTM, what does the upgrade process for this look like via the x/upgrade module? Will chains need to add the setting of this parameter into their scripts?

colin-axner:
The processed height and iteration keys are missing from the proposal handle (processed time is already set)

I believe these need to be set for upgade as well?

modules/core/exported/client.go
```diff
@@ -156,8 +160,10 @@ type ClientState interface {
 		store sdk.KVStore,
 		cdc codec.BinaryCodec,
 		height Height,
+		currentHeight Height,
 		currentTimestamp uint64,
```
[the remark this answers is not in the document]
AdityaSripal: Yea we could do that.

modules/core/03-connection/types/params.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"time"
+
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+)
+
+// DefaultTimePerBlock is the default value for maximum time
+// expected per block.
+const DefaultTimePerBlock = time.Hour
```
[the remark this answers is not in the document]
AdityaSripal: This is the maximum time expected per block so we want it to be much larger than normal but not **too** much larger than normal. IDK what a sensible default here is.

cc: @cwgoes @shahankhatch

modules/light-clients/07-tendermint/types/errors.go
AdityaSripal: Should I do this for this PR or for future ones?

modules/core/03-connection/keeper/verify.go
AdityaSripal: > this enforces block delays even for time delays of 0 (since it rounds up). Is there documentation for why you chose to round up?

Not true: https://play.golang.org/p/5B4Sdhvs4PU

> Lets say MaxTimePerBlock is 10s and my time delay is 35s. Rounding up puts us at 4 blocks and rounding down puts us at 3 blocks. In the usual worst case scenario, 3 blocks would be produced in 30 seconds, thus even if the block delay was 3, we would still need to wait for the 4th block. So I guess it doesn't really matter whether we round up or down for this case

Rounding up is the safer approach. If the expected blocks are 3.3 blocks, its safer to wait 4 blocks because that is still fitting with our invariant that we wait **at least** "3.3" blocks. If we round down, it's not a huge deal but it's also breaking the security assumption. Since block delay itself is a conservative decision may as well stick to that philosophy. Also in cases where the calculated delay is 0.3 blocks, we want to at least wait one block. If we round down, we need to create a special case for this.

modules/light-clients/07-tendermint/types/errors.go
colin-axner: I think just future ones should be fine. I don't think relayers are using these values yet

modules/core/03-connection/keeper/verify.go
AdityaSripal: That was a dumb test, removed. I'm assuming params are validated elsewhere

modules/light-clients/07-tendermint/types/client_state.go
```diff
@@ -497,20 +501,33 @@ func (cs ClientState) VerifyNextSequenceRecv(
 	return nil
 }
 
-// verifyDelayPeriodPassed will ensure that at least delayPeriod amount of time has passed since consensus state was submitted
-// before allowing verification to continue.
-func verifyDelayPeriodPassed(store sdk.KVStore, proofHeight exported.Height, currentTimestamp, delayPeriod uint64) error {
-	// check that executing chain's timestamp has passed consensusState's processed time + delay period
+// verifyDelayPeriodPassed will ensure that at least delayTimePeriod amount of time and delayBlockPerion number of blocks have passed
```
colin-axner, line 504:
```suggestion
// verifyDelayPeriodPassed will ensure that at least delayTimePeriod amount of time and delayBlockPeriod number of blocks have passed
```

modules/core/03-connection/types/params.go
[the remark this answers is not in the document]
cwgoes: This is the default, but it can be set by governance, correct?

I think a minute is a sane default, an hour is certainly safer but probably unnecessary.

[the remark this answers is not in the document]
colin-axner: > This is the default, but it can be set by governance, correct?

Correct, although I don't expect all chains to deviate from the default.

The higher the max expected block time, the smaller the block delay will be correct? One minute sounds good to me. For comparison, a 30 minute packet delay would result in a 30 block delay for one minute and 1 block delay for one hour. Is a longer time safer? I think a longer time per block just guards against bad UX in the case the average block time is greater than the expected block delay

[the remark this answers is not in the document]
cwgoes: Hmm, now I'm a little confused. The goal is to ensure that we don't allow packets to be confirmed before enough blocks have passed for evidence to be included. If we want the block delay and the time delay to "match", we should calculate the block delay from the time delay in a way which ensures that at least `X` blocks pass with an expected time of some `t` per block, for a time delay `X * t` - so maybe what we want is _minimum_ time per block, not maximum?

Maybe I misunderstand though, @AdityaSripal can you clarify the calculation intended here?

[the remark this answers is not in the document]
AdityaSripal: `Minimum Number of Blocks = Time Delay / Maximum Expected Time Per Block`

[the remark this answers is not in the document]
AdityaSripal: bump @cwgoes

modules/core/keeper/keeper.go
```diff
@@ -36,8 +37,16 @@ func NewKeeper(
 	stakingKeeper clienttypes.StakingKeeper, upgradeKeeper clienttypes.UpgradeKeeper,
 	scopedKeeper capabilitykeeper.ScopedKeeper,
 ) *Keeper {
+	// register paramSpace at top level keeper
+	// set KeyTable if it has not already been set
+	if !paramSpace.HasKeyTable() {
+		keyTable := clienttypes.ParamKeyTable()
+		keyTable.RegisterParamSet(&connectiontypes.Params{})
```
AdityaSripal, line 44: Registers client and connection params

modules/core/03-connection/types/params.go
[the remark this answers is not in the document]
AdityaSripal: I see the block delay as handling the edge case where the receiving chain slows down unexpectedly. I'd rather not cause a packet to fail on block delay if there was sufficient number of blocks to submit evidence so I went with a large default so that this error would be hit **very rarely**. Though I can change the default if this is not the desired behaviour.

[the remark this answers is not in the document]
colin-axner: The above reasoning makes sense to me, but it also assumes non-malicious chain halts. In the presence of a > 1/3 malicious validator set on the receiving chain which intentionally halts the chain for a set amount of time, it appears they could always engineer a solution which gets the packet processed before misbehaviour submission unless there exists mempool prioritization. Is this correct reasoning? The block delay above is only protecting against non-malicious chain halts?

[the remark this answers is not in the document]
cwgoes: See https://github.com/cosmos/ibc/issues/573#issuecomment-843101054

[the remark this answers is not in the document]
AdityaSripal: > The above reasoning makes sense to me, but it also assumes non-malicious chain halts. In the presence of a > 1/3 malicious validator set on the receiving chain which intentionally halts the chain for a set amount of time, it appears they could always engineer a solution which gets the packet processed before misbehaviour submission unless there exists mempool prioritization. Is this correct reasoning? The block delay above is only protecting against non-malicious chain halts?

This is if the executing chain becomes malicious. If the chain you are on becomes malicious all bets are off. They can do whatever they want, there's no point trying to mitigate that.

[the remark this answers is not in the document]
colin-axner: > This is if the executing chain becomes malicious. If the chain you are on becomes malicious all bets are off. They can do whatever they want, there's no point trying to mitigate that.

There's an important distinction here though. 1/3 malicious on the executing chain doesn't result in "all bets are off". It certainly can cause problems via halting, but I believe 2/3 malicious is the "all bets are off" magic number. If 1/3 could perform a profitable attack by halting the chain and colluding with a counterparty, I still see this as an attack that should be documented. Halting the chain is not misbehaviour right? So we could have 1/3 malicious halt the chain without consequences and the rest of the validators be trustworthy. Maybe after the attack folks undelegate from the 1/3 and they are no longer in the validator set, but they still performed a successful attack without *any* slashing

From my understanding, ordering the transactions could still occur with a 1/3 attack. If the 1/3 have a good setup, they could halt the chain when one of their validators is proposing the block. They could wait the trusting period out and then order the transactions such that no misbehavior submission is allowed in the block that is being processed after the halt. There doesn't seem to be anything atm that we can do about this, but it does seem like a possible 1/3 attack that could be performed by a validator set colluding with a counterparty validator set (or potentially itself in the case of cross chain validation). A block delay > 1 seems like the best defense against this attack atm (and perhaps this should be recommended in our user facing documents when deciding upon a connection delay time)

Am I missing something? Obviously this sort of attack requires a very advanced setup and solid understanding of IBC/TM, but I wouldn't say its impossible. It's even potentially incentivized (if the channel has more funds then the amount the counterparty is slashed) since there is no penalty for halting

modules/core/keeper/keeper.go
colin-axner: should this if be removed from 02-client?

modules/core/03-connection/keeper/verify.go
```diff
@@ -277,3 +295,18 @@ func (k Keeper) VerifyNextSequenceRecv(
...
+
+// getBlockDelay calculates the block delay period from the time delay of the connection
+// and the expected time per block.
+func (k Keeper) getBlockDelay(ctx sdk.Context, connection exported.ConnectionI) uint64 {
+	// expectedTimePerBlock should never be zero, however if it is then return a 0 blcok delay for safety
+	// as the expectedTimePerBlock parameter was not set.
+	expectedTimePerBlock := k.GetExpectedTimePerBlock(ctx)
+	if expectedTimePerBlock == 0 {
+		return 0
+	}
```
colin-axner, line 307: :+1:

proto/ibc/core/connection/v1/connection.proto
```diff
@@ -104,3 +104,9 @@ message Version {
   // list of features compatible with the specified identifier
   repeated string features = 2;
 }
+
+// Params defines the set of Connection parameters.
+message Params {
+  // expected time per block, used to enforce block delay.
+  uint64 expected_time_per_block = 1 [(gogoproto.moretags) = "yaml:\"expected_time_per_block\""];
```
colin-axner, line 111:
```suggestion
  uint64 max_expected_time_per_block = 1 [(gogoproto.moretags) = "yaml:\"max_expected_time_per_block\""];
```
thoughts? I like having max since it implies the person choosing the parameter should maybe read more documentation about this and use a value larger than the average block time. 

If I was setting this for my chain and saw `expected_time_per_block`, I'd go to mintscan, see that the hub produces blocks every 7 seconds and then set this to `7s`

I think we could add more comments to this proto definition explaining that the value should be a larger than the average block time. Maybe 3-5 times larger

colin-axner (approved):
ACK

This pr needs a changelog entry and updated migration docs (chains should be aware they need to set a value here and how to set the value). A nice-to-have is connection delay docs explaining the time delay and block delay in more detail, but this can be done in a followup

We also should test this feature manually. We should test setting it and not setting it during upgrade testing

modules/core/03-connection/types/params.go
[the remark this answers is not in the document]
AdityaSripal: 1/3 malicious results in arbitrary censorship of messages.
2/3 malicious results in arbitrary inclusion of messages.

1/3+ malicious can stop any misbehaviour from being submitted to a block by simply refusing to commit to any block that includes the misbehaviour. They don't have to halt the chain. The chain can keep going indefinitely with a commit only happening for a block that doesn't include the messages they don't want. Nothing we write into the state machine (time/block delay) saves us from this. The message can be censored indefinitely.

This is true of any messages, you can't even slash a malicious validator since slashing msgs can be censored. 

Once you have >1/3 malicious colluding validators the chain is broken. There is no tiered security model that we can safely reason about.

As an application on top of Tendermint, we shouldn't worry about what happens once we break the Tendermint security model. It is the executing chain's responsibility to make sure that never happens.

modules/core/keeper/keeper.go
AdityaSripal: No if client keeper is registered seperately (e.g. for testing) this will make sure params are still registered.

Issue #177, opened 2021-05-18
TM chain-id validation


colin-axner:
## Summary

The current code relies on the tendermint constant for the MaxChainID. If this value ever changes it will become problematic between chains using different versions and should likely be dealt with. 

We should add a note in code, and 2 tests which will allow us to detect if the max chain id increases or decreases. The test should indicate that if this test breaks, this issue (which will be closed at the time) should be reopened and dealt with. 

The problem scenario is lets say the hub is running a version of ibc-go with the MaxChainID as 50 characters. A counterparty chain could run a tendermint version with a max chain id of 70 characters. This mismatch would prevent a light client from being created. If the max chain id decreased, it could render previously valid IBC clients invalid (althought this case is less likely to occur). 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
If the tests written in #180 break, reopen this issue

PR #180, opened 2021-05-19
Chain-ID validation notes/tests

colin-axner:
## Description

I've added some notes and tests so if the value of the chain-id changes, at least the current maintainers of this codebase can be aware that this may be problematic

There isn't much that can be done in the case that a counterparty upgrade to the tendermint version before the other chains are even aware this value may be changed, but it is good for folks to be aware so they don't launch a chain that cannot interact with IBC until everyone upgrades

closes: #177

---

write a little note why.


fedekunze (commented):
I think a chain ID validation function should be incorporated daemon setup process. This way you could avoid this issue entirely and enforce chains that use IBC to validate their chainID before initializing the chain.

See [here](https://github.com/tharsis/ethermint/blob/18d1f01e828bbfbd08fd1dc4dcf87d437b728f35/cmd/ethermintd/root.go#L95) for reference

colin-axner:
> I think a chain ID validation function should be incorporated daemon setup process. This way you could avoid this issue entirely and enforce chains that use IBC to validate their chainID before initializing the chain.

The IBC client cannot assume the counterparty validates its own chain-id. The counterparty could be malicious. Without this check, an user could create an IBC client with no max chain-id length

AdityaSripal (approved):

Issue #11, opened 2021-03-03
Add v1.0.0 in-place store migration and genesis migrations


colin-axner:
We need to prune:
- [ ] solo machines (previously broken, new proto definition)
- [ ] expired consensus states for tendermint clients

We need to add:
- [ ] genutil genesis migration command
- [ ] in-place store migration via incremented consensus version

PR #205, opened 2021-05-31
Add in-place and genesis migrations

colin-axner:
## Description

Adds an in-place migration for SDK v0.40 -> ibc-go v1.0.0 and genesis migrations

Solo machine proto definition changed (FrozenSequence because a boolean). This may lead to indeterminate decoding errors. During in-place and genesis migrations, we will decode the solomachine into the v1 definition, then we will migrate it to v2 and set this in the store or genesis. All consensus states for solo machines are removed. The legacy solo machine is generated into `legacy/v100`

Also all expired tendermint consensus states are pruned

ref: #11 

---

write a little note why.


modules/core/02-client/legacy/v100/store.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/cosmos/ibc-go/modules/core/02-client/types"
+	host "github.com/cosmos/ibc-go/modules/core/24-host"
+	"github.com/cosmos/ibc-go/modules/core/exported"
+	ibctmtypes "github.com/cosmos/ibc-go/modules/light-clients/07-tendermint/types"
+)
+
+// MigrateStore performs in-place store migrations from SDK v0.40 of the IBC module to v1.0.0 of ibc-go.
+// The migration includes:
+//
+// - Pruning solo machine clients
```
colin-axner, line 20: Since the proto definition changed, the first thought of @AdityaSripal and I is to delete the solo machine, but I'm realizing there could be connection in INIT attached to that solo machine. I could delete those connections as well

Alternatively, I could add back solomachine proto definition v1. Unmarshal into that, copy over the fields to the new proto definition and set the updated proto in state. Requires adding back v1 of the solo machine proto definition

Thoughts? @AdityaSripal @cwgoes @fedekunze

cwgoes: Are there any opened channels using solo machines? If so, it seems like deleting clients would be a problem.

I don't think channels in INIT state are an issue, those can be deleted as well, but if any are open that's a problem.

colin-axner: No channels. The connection handshake failed in OpenTry/OpenAck previously

cwgoes: OK, subject to that holding true this seems safe to me, although isn't it always possible that channels could be created in between when this code is deployed and when the switchover actually happens (e.g. if the upgrade module is used)?

colin-axner: No I don't think so. Unless a chain upgraded to an unreleased commit of this code and/or backported the fixes we added. The previous version being upgraded from could never create an open channel with solo machines. When this upgrade occurs, the migrations will be run first (same idea as init genesis but for x/upgrade)

cwgoes: Ah yes - sorry, I had forgotten that this wasn't possible at the moment anyways - makes sense & seems safe to me.

modules/core/02-client/legacy/v100/store.go
```diff
@@ -0,0 +1,69 @@
...
+		clientState := types.MustUnmarshalClientState(cdc, iterator.Value())
+
+		clientType, _, err := types.ParseClientIdentifier(clientID)
+		if err != nil {
+			return err
+		}
+
+		switch clientType {
+		case exported.Solomachine:
+			store.Delete([]byte(fmt.Sprintf("%s/%s", host.KeyClientStorePrefix, clientID)))
```
AdityaSripal, line 46: Is there consensus states we need to delete here as well?

AdityaSripal (changes requested):
Looks great! Just have a question on solo machine cons states

modules/core/02-client/legacy/v100/store.go
colin-axner: I suppose `store.Delete` does not delete the paths underneath?

ie

ibc/clients/06-solomachine-1 

vs

ibc/clients/06-solomachine-1/clientState
and
ibc/clients/06-solomachine-1/consensusStates/0-1

Ideally I want something that wipes the entire prefixed client store

modules/core/04-channel/legacy/v100/store.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/cosmos/ibc-go/modules/core/04-channel/types"
+	host "github.com/cosmos/ibc-go/modules/core/24-host"
+)
+
+// MigrateStore performs in-place store migrations from SDK v0.40 of the IBC module to v1.0.0 of ibc-go.
+// The migration includes:
+//
+// - Pruning all channels whose connection has been removed (solo machines)
+func MigrateStore(ctx sdk.Context, storeKey sdk.StoreKey, cdc codec.BinaryCodec) (err error) {
```
colin-axner, line 15: hmm, @AdityaSripal do I need to do something about capabilities?

modules/core/02-client/legacy/v100/store.go
colin-axner: updated, my only question now is about channel capabilities. The logic for removing connections/channels turned out good

- remove connection if client does not exist
- remove channel if connection does not exist

modules/core/02-client/legacy/v100/store.go
colin-axner: updated, I also fixed a bug with iteration. There's a contract not to do writes in the domain during iteration

colin-axner:
I thought about this some more. Instead of deleting the solo machines I'm going to migrate the proto definitions. I will generate the v1 solo machine proto files in the legacy dir, iterate over the client store and migrate the old proto definition to the new one (changing frozen sequence to be a boolean)

The existing solo machines are still considered active so they don't need to be deleted. This avoids risk of deleting the wrong clients/connections/channels and it'll make genesis migration a lot easier. I'm not sure how I would be capable of efficiently iterating all channels to delete those associated with solo machines

Genesis migration will do the same, iterate over the clients, unmarshal if solo machine, migrate to new definition, marshal and set in genesis

modules/core/02-client/legacy/v100/store.go
colin-axner: see updated changes. Decided migrating from v1 to v2 was the better option

colin-axner:
Our genesis migrations will not fit into the interface defined by the SDK since it needs access to the genesis block time (this block time must also be updated before the migration is ran). Even if we didn't need the block time, the interface for adding genesis migrations outside of the SDK is clumsy. We would need to add another version migration. Normally the SDK might go from v0.40 -> v0.43. We would need to add a migration which goes from v0.43 -> v0.43.0-ibc or something like that. Basically doing the migrations for v0.43.0 for IBC

The better way to do this is to just explicitly call the IBC migrations without abiding to the migration interface. 

It'll look something like 
```go
newIBCGenState, err := v100.Migrate(IBCGenState, genesisTime)
if err != nil {
    return err
}
genesis.AppState[ibchost.ModuleName] = newIBCGenState
```

I'll add the code explicitly into our migration doc and open a pr on gaia

cc @shahankhatch

colin-axner:
Actually, gaia won't need this as it should use the in-place migrations (not a genesis restart)

docs/migrations/ibc-migration-043.md
```diff
@@ -27,6 +27,44 @@ Feel free to use your own method for modifying import names.
 NOTE: Updating to the `v0.43.0` SDK release and then running `go mod tidy` will cause a downgrade to `v0.42.0` in order to support the old IBC import paths.
 Update the import paths before running `go mod tidy`.  
 
+## Chain Upgrades
+
+Chains may choose to upgrade via an upgrade proposal or genesis upgrades. Both in-place store migrations and genesis migrations are supported. 
+
+**WARNING**: Please read at least the quick guide for [IBC client upgrades](../ibc/upgrades/README.md) before upgrading your chain. It is highly recommended to not change the chain id during an upgrade, otherwise you must follow the IBC client upgrade instructions.
```
colin-axner, line 34:
```suggestion
**WARNING**: Please read at least the quick guide for [IBC client upgrades](../ibc/upgrades/README.md) before upgrading your chain. It is highly recommended you do not change the chain-ID during an upgrade, otherwise you must follow the IBC client upgrade instructions.
```

docs/migrations/ibc-migration-043.md
```diff
@@ -27,6 +27,44 @@ Feel free to use your own method for modifying import names.
...
+
+Both in-place store migrations and genesis migrations will:
+- update the solo machine client state to using the v2 solo machine protobuf defintion
```
colin-axner, line 37:
```suggestion
- migrate the solo machine client state from v1 to v2 protobuf definitions
```

docs/migrations/ibc-migration-043.md
```diff
@@ -27,6 +27,44 @@ Feel free to use your own method for modifying import names.
...
+In place store migrations will automatically be run after the binary is swapped and the application is started again during an upgrade proposal. 
+
+### Genesis Migrations
+
+To perform genesis migrations, the following code must be added to your existing migration code.
+
+```go
+// add imports as necessary
+import (
+    "github.com/cosmos/cosmos-sdk/codec"
```
colin-axner, line 52:
```suggestion
```

docs/migrations/ibc-migration-043.md
```diff
@@ -27,6 +27,44 @@ Feel free to use your own method for modifying import names.
...
+
+...
+
+// add in migrate cmd function
+newGenState, err = ibcv100.MigrateGenesis(clientCtx.InterfaceRegistry, newGenState, genDoc.GenesisTime)
+if err != nil {
+    return err
+}
+```
+
```
colin-axner, line 65: untested, will test when I do upgrade testing and update accordingly

modules/core/02-client/keeper/migrations.go
```diff
@@ -0,0 +1,28 @@
...
+	return Migrator{keeper: keeper}
+}
+
+// Migrate1to2 migrates from version 1 to 2.
+// This migration prunes:
+// - solo machine consensus states
+// - expired tendermint consensus states
+//
+// This migration migrates:
+// - solo machine client states from v1 to v2 protobuf definition
```
colin-axner, line 25: update godoc

modules/core/02-client/legacy/v100/genesis_test.go
```diff
@@ -0,0 +1,272 @@
...
+	// store migration and genesis migration should produce identical results
+	err := v100.MigrateStore(path1.EndpointA.Chain.GetContext(), path1.EndpointA.Chain.GetSimApp().GetKey(host.StoreKey), path1.EndpointA.Chain.App.AppCodec())
+	suite.Require().NoError(err)
+	expectedClientGenState := ibcclient.ExportGenesis(path1.EndpointA.Chain.GetContext(), path1.EndpointA.Chain.App.GetIBCKeeper().ClientKeeper)
+
+	migrated, err := v100.MigrateGenesis(codec.NewProtoCodec(clientCtx.InterfaceRegistry), &clientGenState, suite.coordinator.CurrentTime)
+	suite.Require().NoError(err)
+
+	// check path 1 client pruning
+	//	fmt.Println(migrated.ClientsConsensus)
```
colin-axner, line 206:
```suggestion
```

modules/core/02-client/legacy/v100/genesis_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// Indent the JSON bz correctly.
+	err = json.Unmarshal(bz, &jsonObj)
+	suite.Require().NoError(err)
+	indentedBz, err := json.MarshalIndent(jsonObj, "", "\t")
+	suite.Require().NoError(err)
+
+	fmt.Println(string(indentedBz))
+
+	fmt.Println(string(expectedIndentedBz))
```
colin-axner, line 269:
```suggestion
```

modules/core/02-client/legacy/v100/genesis_test.go
```diff
@@ -0,0 +1,272 @@
...
+	bz, err = clientCtx.JSONCodec.MarshalJSON(migrated)
+	suite.Require().NoError(err)
+
+	// Indent the JSON bz correctly.
+	err = json.Unmarshal(bz, &jsonObj)
+	suite.Require().NoError(err)
+	indentedBz, err := json.MarshalIndent(jsonObj, "", "\t")
+	suite.Require().NoError(err)
+
+	fmt.Println(string(indentedBz))
```
colin-axner, line 267: the tests I followed printed the actual. I can remove, but it is useful in quickly verifying things

modules/core/keeper/migrations.go
```diff
@@ -0,0 +1,33 @@
...
+	return Migrator{keeper: keeper}
+}
+
+// Migrate1to2 migrates from version 1 to 2.
+// This migration prunes:
+// - solo machine consensus states
+// - expired tendermint consensus states
+//
+// This migration migrates:
+// - solo machine client state from protobuf definition v1 to v2
```
colin-axner, line 25: updated go doc

modules/core/module.go
```diff
@@ -157,7 +161,7 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
 }
 
 // ConsensusVersion implements AppModule/ConsensusVersion.
-func (AppModule) ConsensusVersion() uint64 { return 1 }
+func (AppModule) ConsensusVersion() uint64 { return 2 }
```
colin-axner, line 164: that was easy

modules/core/02-client/legacy/v100/store.go
```diff
@@ -0,0 +1,141 @@
...
+		case exported.Solomachine:
+			any := &codectypes.Any{}
+			if err := cdc.Unmarshal(bz, any); err != nil {
+				return sdkerrors.Wrap(err, "failed to unmarshal client state bytes into solo machine client state")
+			}
+
+			clientState := &ClientState{}
+			if err := cdc.Unmarshal(any.Value, clientState); err != nil {
+				return sdkerrors.Wrap(err, "failed to unmarshal client state bytes into solo machine client state")
+			}
```
colin-axner, line 70: This took me a while to figure out. Since I regenerate v1 solomachine into legacy, it doesn't have any implementation code and isn't a exported.ClientState, so you cannot use the usual methods for decoding. 

The work around is to decode to Any and then decode the Any.Value (by passing the any.Type lookup)

modules/core/legacy/v100/genesis.go
```diff
@@ -0,0 +1,40 @@
...
+	host "github.com/cosmos/ibc-go/modules/core/24-host"
+	"github.com/cosmos/ibc-go/modules/core/types"
+)
+
+// MigrateGenesis accepts exported v1.0.0 IBC client genesis file and migrates it to:
+//
+// - Update solo machine client state protobuf definition (v1 to v2)
+// - Remove all solo machine consensus states
+// - Remove all expired tendermint consensus states
+func MigrateGenesis(appState genutiltypes.AppMap, clientCtx client.Context, genesisBlockTime time.Time) (genutiltypes.AppMap, error) {
```
colin-axner, line 20: this needs to be tested. I think I need to add the unpack interfaces functions for solo machine. Will work on this tomorrow

modules/core/02-client/legacy/v100/store.go
```diff
@@ -0,0 +1,141 @@
...
+
+		case exported.Tendermint:
+			var clientState exported.ClientState
+			if err := cdc.UnmarshalInterface(bz, &clientState); err != nil {
+				return sdkerrors.Wrap(err, "failed to unmarshal client state bytes into tendermint client state")
+			}
+
+			if err = ibctmtypes.PruneAllExpiredConsensusStates(ctx, clientStore, cdc, clientState.(*ibctmtypes.ClientState)); err != nil {
+				return err
+			}
```
AdityaSripal, line 92: We also need to set iteration and height keys for unexpired consensus states

AdityaSripal (changes requested):
I believe you need to set processedHeight and Iteration keys for the unexpired tendermint consensus states that you migrate over.

modules/core/02-client/legacy/v100/genesis.go
```diff
@@ -0,0 +1,133 @@
...
+								height := consState.Height
+
+								// iterate through metadata and find metadata which should be pruned
+								// only unexpired consensus state heights should be added
+								var clientMetadata []types.GenesisMetadata
+								for _, metadata := range identifiedGenMetadata.ClientMetadata {
+									if bytes.Equal(metadata.Key, ibctmtypes.IterationKey(height)) ||
+										bytes.Equal(metadata.Key, ibctmtypes.ProcessedTimeKey(height)) ||
+										bytes.Equal(metadata.Key, ibctmtypes.ProcessedHeightKey(height)) {
+										clientMetadata = append(clientMetadata, metadata)
```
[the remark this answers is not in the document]
colin-axner: it's a good thing we decided to prune expired consensus states and migrate the solo machine, don't think we would have realized this otherwise

modules/core/02-client/legacy/v100/store.go
```diff
@@ -139,3 +148,36 @@ func pruneSolomachineConsensusStates(clientStore sdk.KVStore) {
 		clientStore.Delete(host.ConsensusStateKey(height))
 	}
 }
+
+// addConsensusMetadata adds the iteration key and processed height for all unexpired tendermint consensus states
+// These keys were not included in the previous release of the IBC module.
+func addConsensusMetadata(ctx sdk.Context, clientStore sdk.KVStore, cdc codec.BinaryCodec, clientState *ibctmtypes.ClientState) error {
```
colin-axner, line 155: I decided not to modify `PruneAllExpiredConsensusStates` as that seems like a useful ibctmtypes function by itself. Since the insertion of the iteration key and processed height key is logic specific to this migration, I decided to iterate the consensus state again and do it in a separate iteration. It is less efficient, but I think better code wise. Open to changing if efficiency is a big concern

modules/core/02-client/legacy/v100/solomachine.go
```diff
@@ -0,0 +1,208 @@
...
+
+// NOTE: this is a mock implmentation for exported.ClientState. This implementation
+// should only be registered on the InterfaceRegistry during cli command genesis migration.
+// This implementation is only used to successfully unmarshal the previous solo machine
+// client state and consensus state and migrate them to the new implementations. When the proto
+// codec unmarshals, it calls UnpackInterfaces() to create a cached value of the any. The
+// UnpackInterfaces function for IdenitifiedClientState will attempt to unpack the any to
+// exported.ClientState. If the solomachine v1 type is not registered against the exported.ClientState
+// the unmarshal will fail. This implementation will panic on every interface function.
+// The same is done for the ConsensusState.
```
colin-axner, line 21: there's really no way around this. go semantic versioning will help a lot in the future

modules/core/legacy/v100/genesis_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	// solo machine clients must come before tendermint in expected
+	clientGenState.Clients = append(clients, clientGenState.Clients...)
+
+	// migrate store get expected genesis
+	// store migration and genesis migration should produce identical results
+	err := clientv100.MigrateStore(path.EndpointA.Chain.GetContext(), path.EndpointA.Chain.GetSimApp().GetKey(host.StoreKey), path.EndpointA.Chain.App.AppCodec())
+	suite.Require().NoError(err)
+	expectedClientGenState := ibcclient.ExportGenesis(path.EndpointA.Chain.GetContext(), path.EndpointA.Chain.App.GetIBCKeeper().ClientKeeper)
+
```
colin-axner, line 150: this is the main test code added (this was copied from previous tests)

docs/migrations/ibc-migration-043.md
```diff
@@ -27,6 +27,66 @@ Feel free to use your own method for modifying import names.
...
+            // https://github.com/cosmos/ibc-go/blob/release/v1.0.x/docs/ibc/proto-docs.md#params-2
+            app.IBCKeeper.ConnectionKeeper.SetParams(ctx, ibcconnectiontypes.DefaultParams())
+
+            fromVM := map[string]uint64{
+                ... // other modules
+                "ibc":          1,
+                ... 
+            }   
+            return app.mm.RunMigrations(ctx, app.configurator, fromVM)
+        })
```
colin-axner, line 61: tested

docs/migrations/ibc-migration-043.md
```diff
@@ -27,6 +27,66 @@ Feel free to use your own method for modifying import names.
...
+
+...
+
+// add in migrate cmd function
+// expectedTimePerBlock is a new connection parameter
+// https://github.com/cosmos/ibc-go/blob/release/v1.0.x/docs/ibc/proto-docs.md#params-2
+newGenState, err = ibcv100.MigrateGenesis(newGenState, clientCtx, *genDoc, expectedTimePerBlock)
+if err != nil {
+    return err 
+}
```
colin-axner, line 84: tested, got as far as I could. Final confirmation is blocked on upstream changes (SDK is moving protobuf definition for upgrade/gov from v1beta1 to v1 and the current pr has minor bugs)

I confirmed the migration updated the pr, as far as I can tell, everything works on the IBC end, although I didn't get a chance to test that IBC clients were successfully upgraded. At the very least the chain started fine

AdityaSripal (approved):
Looks good, just a couple nits

modules/core/02-client/legacy/v100/genesis.go
```diff
@@ -102,10 +104,20 @@ func MigrateGenesis(cdc codec.BinaryCodec, clientGenState *types.GenesisState, g
...
 								var clientMetadata []types.GenesisMetadata
 								for _, metadata := range identifiedGenMetadata.ClientMetadata {
-									if bytes.Equal(metadata.Key, ibctmtypes.IterationKey(height)) ||
-										bytes.Equal(metadata.Key, ibctmtypes.ProcessedTimeKey(height)) ||
-										bytes.Equal(metadata.Key, ibctmtypes.ProcessedHeightKey(height)) {
+									// the previous version of IBC only contained the processed time metadata
+									// if we find the processed time metadata for an unexpired height, add the
+									// iteration key and processed height keys.
+									if bytes.Equal(metadata.Key, ibctmtypes.ProcessedTimeKey(height)) {
+										clientMetadata = append(clientMetadata, types.GenesisMetadata{
```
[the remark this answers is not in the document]
colin-axner: wow, I've always worked under the impression `append()` took only 2 arguments

modules/core/02-client/legacy/v100/store.go
```diff
@@ -139,3 +149,36 @@ func pruneSolomachineConsensusStates(clientStore sdk.KVStore) {
...
+	var heights []exported.Height
+
+	metadataCb := func(height exported.Height) bool {
+		consState, err := ibctmtypes.GetConsensusState(clientStore, cdc, height)
+		// this error should never occur
+		if err != nil {
+			return true
+		}
```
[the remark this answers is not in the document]
colin-axner: will return an error, I think a panic could corrupt the database

modules/core/02-client/legacy/v100/store_test.go
```diff
@@ -185,4 +185,21 @@ func (suite *LegacyTestSuite) TestMigrateStoreTendermint() {
...
+
+	processedTime, ok := ibctmtypes.GetProcessedTime(clientStore, height)
+	suite.Require().True(ok)
+	suite.Require().NotEqual(uint64(0), processedTime)
+
+	processedHeight, ok := ibctmtypes.GetProcessedHeight(clientStore, height)
+	suite.Require().True(ok)
+	suite.Require().Equal(types.GetSelfHeight(path.EndpointA.Chain.GetContext()), processedHeight)
+
+	consKey := ibctmtypes.GetIterationKey(clientStore, height)
```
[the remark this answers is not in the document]
colin-axner: Yea, the purpose of this to ensure the unexpired consensus state metadata did not get pruned. I wrote this before adding the `addConsensusMetadata` function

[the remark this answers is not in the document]
colin-axner: found a bug, since I was using the iteration keys for iterating, it wasn't updating the metadata since the iteration keys hadn't been set

changed the structure, set new metadata for all consensus states (using traditional iteration), then prune expired consensus states

modules/core/02-client/legacy/v100/store_test.go
```diff
@@ -118,88 +118,114 @@ func (suite *LegacyTestSuite) TestMigrateStoreSolomachine() {
...
+		unexpiredHeights = append(unexpiredHeights, path.EndpointA.GetClientState().GetLatestHeight())
+		path.EndpointA.UpdateClient()
+		unexpiredHeights = append(unexpiredHeights, path.EndpointA.GetClientState().GetLatestHeight())
+
+		// remove processed height and iteration keys since these were missing from previous version of ibc module
+		clientStore := path.EndpointA.Chain.App.GetIBCKeeper().ClientKeeper.ClientStore(path.EndpointA.Chain.GetContext(), path.EndpointA.ClientID)
+		for _, height := range unexpiredHeights {
+			clientStore.Delete(ibctmtypes.ProcessedHeightKey(height))
+			clientStore.Delete(ibctmtypes.IterationKey(height))
+		}
```
colin-axner, line 179: this now enforces a check on the metadata being added

modules/core/02-client/legacy/v100/store.go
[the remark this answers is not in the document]
colin-axner: removed entirely

modules/core/02-client/legacy/v100/store_test.go
```diff
@@ -118,88 +118,114 @@ func (suite *LegacyTestSuite) TestMigrateStoreSolomachine() {
...
-	pruneHeights = append(pruneHeights, path.EndpointA.GetClientState().GetLatestHeight())
+	path2 := ibctesting.NewPath(suite.chainA, suite.chainB)
+	suite.coordinator.SetupClients(path2)
+	pruneHeightMap := make(map[*ibctesting.Path][]exported.Height)
+	unexpiredHeightMap := make(map[*ibctesting.Path][]exported.Height)
 
-	// these heights will be expired and also pruned
-	for i := 0; i < 3; i++ {
-		path.EndpointA.UpdateClient()
+	for _, path := range []*ibctesting.Path{path1, path2} {
```
colin-axner, line 129: I just added an extra for loop to test multiple paths

PR #233, opened 2021-07-06
Fix missing events in OnRecvPacket

bluele:
## Description

The events set in the cache context need to be set in the original context.

---

write a little note why.

### DIFF
--- a/modules/core/keeper/msg_server.go
+++ b/modules/core/keeper/msg_server.go
@@ -502,6 +502,7 @@ func (k Keeper) RecvPacket(goCtx context.Context, msg *channeltypes.MsgRecvPacke
 	ack := cbs.OnRecvPacket(cacheCtx, msg.Packet, relayer)
 	// This doesn't cause duplicate events to be emitted.
 	// NOTE: The context returned by CacheContext() refers to a new EventManager, so it needs to explicitly set events to the original context.
+	// Events from callback are emitted regardless of acknowledgement success
 	ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
 	if ack == nil || ack.Success() {
 		// write application state changes for asynchronous and successful acknowledgements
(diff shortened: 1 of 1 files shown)


modules/core/keeper/msg_server.go
```diff
@@ -503,6 +503,7 @@ func (k Keeper) RecvPacket(goCtx context.Context, msg *channeltypes.MsgRecvPacke
 	if ack == nil || ack.Success() {
 		// write application state changes for asynchronous and successful acknowledgements
 		writeFn()
+		ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
```
colin-axner, line 506: could you add a comment that indicates that `ctx.CacheContext()` sets a new event manager and thus won't cause duplicate events to be returned by `cacheCtx.EventManager().Events()`

At first I thought this line might cause events emitted by `k.ChannelKeeper.RecvPacket` to be represented twice but I realized the cache context reset the events by setting a new event manager

It'd be great to have this written in the code

colin-axner (approved):
Very nice catch!! Just left a suggestion for adding a in code comment for easier verifiability of the code. Tested ACK

colin-axner (changes requested):
On second thought, I think we should emit events in all cases. So we should move the line added above the if statement. This allows failed acks to be emitted as well

AdityaSripal:
Great catch! Agreed with Colin that we should emit events regardless of callback success. We must document to developers that state gets reverted but any emitted events still get emitted even in error case

bluele:
@colin-axner
Thanks for your comments! I've fixed them and squashed commits.

modules/core/keeper/msg_server.go
```diff
@@ -500,6 +500,9 @@ func (k Keeper) RecvPacket(goCtx context.Context, msg *channeltypes.MsgRecvPacke
 	// Cache context so that we may discard state changes from callback if the acknowledgement is unsuccessful.
 	cacheCtx, writeFn := ctx.CacheContext()
 	ack := cbs.OnRecvPacket(cacheCtx, msg.Packet, relayer)
+	// This doesn't cause duplicate events to be emitted.
+	// NOTE: The context returned by CacheContext() refers to a new EventManager, so it needs to explicitly set events to the original context.
```
AdityaSripal, line 504:
```suggestion
	// NOTE: The context returned by CacheContext() refers to a new EventManager, so it needs to explicitly set events to the original context.
	// Events from callback are emitted regardless of acknowledgement success
```

AdityaSripal (approved):

colin-axner (approved):
LGTM, thanks for the contribution @bluele and the fast response time!

Issue #245, opened 2021-07-07
No-op on redundant relay messages


colin-axner:
## Summary

Instead of returning an error, we should do a no-op on relay messages which have already been relayed. This will prevent relay transactions with multi messages from failing due to one redundant message

We already have error codes for indicating a redundant message, we should simply return early here and ensure no state changes have occurred

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #268, opened 2021-07-15
perform a no-op on redundant relay messages

colin-axner:
## Description

closes: #245 

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/timeout.go
+++ b/modules/core/04-channel/keeper/timeout.go
@@ -74,6 +74,7 @@ func (k Keeper) TimeoutPacket(
 	commitment := k.GetPacketCommitment(ctx, packet.GetSourcePort(), packet.GetSourceChannel(), packet.GetSequence())
 
 	if len(commitment) == 0 {
+		EmitTimeoutPacketEvent(ctx, packet, channel)
 		// This error indicates that the timeout has already been relayed
 		// or there is a misconfigured relayer attempting to prove a timeout
 		// for a packet never sent. Core IBC will treat this error as a no-op in order to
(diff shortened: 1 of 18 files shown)


modules/core/keeper/msg_server.go
```diff
@@ -492,13 +492,21 @@ func (k Keeper) RecvPacket(goCtx context.Context, msg *channeltypes.MsgRecvPacke
 	}
 
 	// Perform TAO verification
-	if err := k.ChannelKeeper.RecvPacket(ctx, cap, msg.Packet, msg.ProofCommitment, msg.ProofHeight); err != nil {
+	// If the packet was already received we want to perform a no-op. We use a cached context to prevent
+	// accidental state changes
+	cacheCtx, writeFn := ctx.CacheContext()
```
colin-axner, line 497: this seems like the best solution to me. I need to return before performing the application callbacks, but I also don't want to skip some of the earlier IBC checks in case the sender misconfigured their msg (used wrong channel id or something like that). I also don't want to allow state changes to slip through

modules/core/keeper/msg_server.go
```diff
@@ -556,9 +564,18 @@ func (k Keeper) Timeout(goCtx context.Context, msg *channeltypes.MsgTimeout) (*c
 	}
 
 	// Perform TAO verification
-	if err := k.ChannelKeeper.TimeoutPacket(ctx, msg.Packet, msg.ProofUnreceived, msg.ProofHeight, msg.NextSequenceRecv); err != nil {
+	// If the timeout was already received we want to perform a no-op. We use a cached context to prevent
+	// accidental state changes
+	cacheCtx, writeFn := ctx.CacheContext()
+	if err := k.ChannelKeeper.TimeoutPacket(cacheCtx, msg.Packet, msg.ProofUnreceived, msg.ProofHeight, msg.NextSequenceRecv); err != nil {
+		// timeout already received
+		// NOTE: ordered channels will error on a closed channel upon replay
```
colin-axner, line 572: I need to give some thought about considering closed channels as redundant relaying

colin-axner (commented):
still need to fix up some stuff, but here is the general outline I'm thinking

modules/core/04-channel/keeper/timeout.go
```diff
@@ -34,13 +34,6 @@ func (k Keeper) TimeoutPacket(
 		)
 	}
 
-	if channel.State != types.OPEN {
-		return sdkerrors.Wrapf(
-			types.ErrInvalidChannelState,
-			"channel state is not OPEN (got %s)", channel.State.String(),
-		)
-	}
-
```
colin-axner, line 43: If a packet is timed out on an ordered channel, all future timeout packets need to be timed out via timeout on close (since the channel is closed). I want this "channel closed" error message to be sent to relayers only when the relay isn't redundant (packet commitment doesn't exist). Thus I moved it below the redundant check

testing/mock/mock.go
```diff
@@ -194,7 +197,12 @@ func (am AppModule) OnChanCloseConfirm(sdk.Context, string, string) error {
 // OnRecvPacket implements the IBCModule interface.
 func (am AppModule) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, relayer sdk.AccAddress) exported.Acknowledgement {
 	// set state by claiming capability to check if revert happens return
-	am.scopedKeeper.NewCapability(ctx, MockCanaryCapabilityName)
+	_, err := am.scopedKeeper.NewCapability(ctx, MockRecvCanaryCapabilityName+strconv.Itoa(int(packet.GetSequence())))
+	if err != nil {
+		// application callback called twice on same packet sequence
+		// must never occur
+		panic(err)
+	}
```
AdityaSripal, line 205: Is this purely a way to check it isn't called twice?

AdityaSripal (approved):
Looks good! Nice fix!

modules/core/04-channel/keeper/timeout.go
AdityaSripal: This is so if there's a redundant timeout for the first timed out packet on an ORDERED channel, the relayer gets a no-op rather than this error?

colin-axner: That's correct. When I did the change though I was thinking more packets would return a no op here. I'm fine to revert this in order to keep consistency with the spec. Thoughts?

testing/mock/mock.go
colin-axner: I'm fairly certain the capability name change needed to occur in order to receive 2 packets successfully on the same channel?

The panic was added to ensure the application callback isn't called twice as you say. Based on the structure of the code, it isn't even close to occurring, but if it did occurred it'd be a bad bug, so I figured it didn't hurt to add these panics

AdityaSripal (approved):
looks clean

Issue #266, opened 2021-07-15
Fee Middleware proto files


colin-axner:
- [x] Msgs
- [x] grpc
- [x] genesis

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #149, opened 2021-04-29
Add heights to every gRPC request


colin-axner:
## Summary

Heights are not included in gRPC requests as the height can be set in the SDK context. This isn't very useful for non-go implementations which might have a difficult time setting the SDK context correctly. 

We should add a height request into the gRPC requests. `ibc-rs` is currently having difficulty doing gRPC requests at specific heights. 

The height in the request seems sufficiently generalized that other implementations could implement the same gRPC routes with the same proto files 

cc @fedekunze 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
[ref](https://github.com/cosmos/ibc/issues/568)

noot:
@colin-axner any progress/updates on this? I'm working on integrating IBC in Rust into Astria and have run into this as a major issue, since I would rather use only gRPC requests rather than the ABCI + gRPC combo. I'm happy to take this issue on!

DimitrisJim:
doesn't appear that this was closed due to some blocker so don't see why you couldn't take it on and push it through, rationale still appears to hold :smile:. cc @crodriguezvega for a second opinion.

crodriguezvega:
No, this hasn't been implemented and there hasn't been any progress, so thanks @noot if you want to work on it. For reference, [here's a PR](https://github.com/cosmos/ibc-go/pull/248) where Colin started to work on this but eventually closed.

noot:
great, thanks @DimitrisJim and @crodriguezvega ! I'll take this on then :)

crodriguezvega:
I am going to tentatively put this in v11...

gjermundgaraba:
I'll close this for now. While this might be useful, it is not going to be prioritized unless there is a clear need for it right now. We can reopen and reconsider in the future if needed.

PR #272, opened 2021-07-20
feat: adding proto files for fee payment middleware

seantking:
## Description

Adding proto files for the fee middleware module. See: https://github.com/cosmos/ibc/tree/master/spec/app/ics-029-fee-payment#fee-middleware-contract

closes: https://github.com/cosmos/ibc-go/issues/266

## How to run 
```
make proto-all
```
---

write a little note why.


proto/ibc/applications/middleware/fee/v1/genesis.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+package ibc.applications.middleware.fee.v1;
+
+import "gogoproto/gogo.proto";
+
+// GenesisState defines the ibc-transfer genesis state
+message GenesisState {}
```
seantking, line 8: Not too sure what state we want here @AdityaSripal

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,69 @@
...
+message MsgRegisterCounterPartyAddress {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string address              = 1 [(gogoproto.moretags) = "yaml:\"address\""];
+  string counterparty_address = 2 [(gogoproto.moretags) = "yaml:\"address\""];
+}
+
+// MsgRegisterCounterPartyAddressResponse defines the Msg/RegisterCounteryPartyAddress response type
+message MsgRegisterCounterPartyAddressResponse {}
```
seantking, line 38: Not clear on what we want to be returning here. wdyt @AdityaSripal @colin-axner

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,69 @@
...
+message MsgEscrowPacketFee {
+  string                                 packet_id   = 1 [(gogoproto.moretags) = "yaml:\"packet_id\""];
+  ibc.applications.middleware.fee.v1.Fee receive_fee = 2 [(gogoproto.moretags) = "yaml:\"receive_fee\""];
+  ibc.applications.middleware.fee.v1.Fee ack_fee     = 3 [(gogoproto.moretags) = "yaml:\"ack_fee\""];
+  ibc.applications.middleware.fee.v1.Fee timeout_fee = 4 [(gogoproto.moretags) = "yaml:\"timeout_fee\""];
+  repeated string                        relayers    = 5 [(gogoproto.moretags) = "yaml:\"relayers\""];
+}
+
+// MsgEscrowPacketFeeResponse defines the response type for Msg/EscrowPacketFee
+message MsgEscrowPacketFeeResponse {}
```
seantking, line 50: ditto

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,69 @@
...
+
+// MsgPayFee defines the request type for PayFee the RPC
+message MsgPayFee {
+  string packet_id       = 1 [(gogoproto.moretags) = "yaml:\"packet_id\""];
+  string forward_relayer = 2 [(gogoproto.moretags) = "yaml:\"forward_relayer\""];
+  string reverse_relayer = 3 [(gogoproto.moretags) = "yaml:\"reverse_relayer\""];
+}
+
+// MsgPayFeeResponse defines the response type for Msg/PayFee
+message MsgPayFeeResponse {}
```
seantking, line 60: ditto

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,69 @@
...
+
+// MsgPayTimmeout defines the request type for the PayTimeoutFee RPC
+message MsgPayTimeoutFee {
+  string packet_id       = 1 [(gogoproto.moretags) = "yaml:\"packet_id\""];
+  string timeout_relayer = 2 [(gogoproto.moretags) = "yaml:\"timeout_relayer\""];
+}
+
+// MsgPayTimeoutFeeResponse defines the response type for Msg/PayTimeoutFee
+message MsgPayTimeoutFeeResponse {}
```
seantking, line 69: ditto

proto/ibc/applications/middleware/fee/v1/tx.proto
AdityaSripal: Just leave it blank. This is what the other msg servers do

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,70 @@
...
+  // EscrowPacketFee defines a rpc handler method for MsgEscrowPacketFee
+  // EscrowPacketFee is an open callback that may be called by any module/user that wishes to escrow funds in order to
+  // incentivize the relaying of the given packet.
+  rpc EscrowPacketFee(MsgEscrowPacketFee) returns (MsgEscrowPacketFeeResponse);
+  // PayFee defines a rpc handler method for MsgPayFee
+  // PayFee is a callback implemented by fee module called by the ICS-4 AcknowledgePacket handler.
+  rpc PayFee(MsgPayFee) returns (MsgPayFeeResponse);
+  // PayTimeoutFee defines a rpc handler method for MsgPayTimeoutFee
+  // PayFee is a callback implemented by fee module called by the ICS-4 TimeoutPacket handler.
+  rpc PayTimeoutFee(MsgPayTimeoutFee) returns (MsgPayTimeoutFeeResponse);
```
AdityaSripal, line 26: These should not be rpcs because they are not called by users. But are callbacks within the module

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,70 @@
...
+  ibc.applications.middleware.fee.v1.Fee ack_fee     = 3 [(gogoproto.moretags) = "yaml:\"ack_fee\""];
+  ibc.applications.middleware.fee.v1.Fee timeout_fee = 4 [(gogoproto.moretags) = "yaml:\"timeout_fee\""];
+  repeated string                        relayers    = 5 [(gogoproto.moretags) = "yaml:\"relayers\""];
+}
+
+// MsgEscrowPacketFeeResponse defines the response type for Msg/EscrowPacketFee
+message MsgEscrowPacketFeeResponse {}
+
+// MsgPayFee defines the request type for PayFee the RPC
+message MsgPayFee {
```
AdityaSripal, line 54: This should be removed, as should its response

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,70 @@
...
+  ibc.core.channel.v1.PacketId packet_id       = 1 [(gogoproto.moretags) = "yaml:\"packet_id\""];
+  string                       forward_relayer = 2 [(gogoproto.moretags) = "yaml:\"forward_relayer\""];
+  string                       reverse_relayer = 3 [(gogoproto.moretags) = "yaml:\"reverse_relayer\""];
+}
+
+// MsgPayFeeResponse defines the response type for Msg/PayFee
+message MsgPayFeeResponse {}
+
+// MsgPayTimmeout defines the request type for the PayTimeoutFee RPC
+message MsgPayTimeoutFee {
```
AdityaSripal, line 64: This should be removed along with response

proto/ibc/applications/middleware/fee/v1/fee.proto
```diff
@@ -0,0 +1,13 @@
...
+package ibc.applications.middleware.fee.v1;
+
+// Fee interface
+// See Fee Payment Middleware spec:
+// https://github.com/cosmos/ibc/tree/master/spec/app/ics-029-fee-payment#fee-middleware-contract
+message Fee {
+  // the token denomination the fee will be paid in
+  string denom = 1;
+  // the token amount to be transferred
+  uint64 amount = 2;
```
AdityaSripal, line 12: Since this is only written on source side, we can use SDK structs here. Should change this to just use sdk.Coins. And I believe we can make this multidenom. cc: @colin-axner @ethanfrey

proto/ibc/applications/middleware/fee/v1/genesis.proto
AdityaSripal: mapping from packetID to escrowed fees

proto/ibc/core/channel/v1/channel.proto
```diff
@@ -132,6 +132,18 @@ message PacketState {
...
 }
 
+// PacketId is an identifer for a unique Packet
+message PacketId {
+  option (gogoproto.goproto_getters) = false;
+
+  // channel port identifier.
+  string port_id = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
+  // channel unique identifier.
+  string channel_id = 2 [(gogoproto.moretags) = "yaml:\"channel_id\""];
```
AdityaSripal, line 142: I think for consistency we should always specify this is **source** port and channel. cc: @colin-axner @ethanfrey

AdityaSripal (changes requested):
Looks great! Some of the pay functions got refactored from being exposed for IBC core to call, to being completely internal with the middleware implementation. Forgot to move these when making the spec. Sorry for the confusion!

proto/ibc/applications/middleware/fee/v1/tx.proto
seantking: Gotcha. Makes sense.

proto/ibc/applications/middleware/fee/v1/genesis.proto
seantking: what type are you thinking for escrowed fees? sdk.Coins?

proto/ibc/core/channel/v1/channel.proto
colin-axner: That would mean a packet is only uniquely identifiable on the source chain, which is the case for 29-fee, but then this seems less compelling to bring into 04-channel since then this is really packet id as identified on the source chain

Why can't the destination chain refer to a packet using the destination port and channel?

AdityaSripal: I mean both sides get access to source and dest port and channelID. So I think both sides using the same semantics will avoid confusion.

Even a destination chain can refer to packet using its source (counterparty) port/channel. It's mainly about keeping very consistent semantics

AdityaSripal: Oh I see what you're saying, it's possible for collisions on dest chains. So we should make the semantics that the portID/channelID of the PacketID always refers to the port/channel on the chain in question

So source chains refer to packets by source port/channel
dest chains refer to packets by dest port/channel.

Yea let's make this the consistent rule. Good catch @colin-axner

proto/ibc/applications/middleware/fee/v1/fee.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
```
colin-axner, line 2: all files need this:
```proto
option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
```
then rerun `make proto-all`

seantking: I left this out just because I figured it depends on the scaffolding. I'll do you as you suggest tho

seantking: Ah, I see that it was merged in. makes sense :+1:

seantking: Updated :+1:

proto/ibc/applications/middleware/fee/v1/fee.proto
seantking: Updated this. Can you double-check? What do you mean by making this multdenom?

proto/ibc/core/channel/v1/channel.proto
seantking: Updated the comment

proto/ibc/applications/middleware/fee/v1/genesis.proto
```diff
@@ -0,0 +1,14 @@
...
+
+import "gogoproto/gogo.proto";
+import "ibc/applications/middleware/fee/v1/fee.proto";
+import "ibc/core/channel/v1/channel.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+
+// GenesisState defines the fee middleware genesis state
+message GenesisState {
+  map<string, ibc.applications.middleware.fee.v1.Fee> packet_fees = 1;
```
seantking, line 13: @AdityaSripal As you can see here its not possible to have a map with a struct as the key (or bytes) https://developers.google.com/protocol-buffers/docs/proto3#maps

I think we will need some sort of .String() fn on the PacketId regardless, in order to store this in a map. Wdyt?

proto/ibc/applications/middleware/fee/v1/fee.proto
```diff
@@ -0,0 +1,12 @@
...
+package ibc.applications.middleware.fee.v1;
+import "cosmos/base/v1beta1/coin.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// Fee interface
+// See Fee Payment Middleware spec:
+// https://github.com/cosmos/ibc/tree/master/spec/app/ics-029-fee-payment#fee-middleware-contract
+message Fee {
+  cosmos.base.v1beta1.Coin amount = 1;
```
colin-axner, line 11: We can have this be `Coins`, there's an example in the sdk proto files of how to do this. I think you need to use gogo proto to cast it to `Coins`

proto/ibc/applications/middleware/fee/v1/genesis.proto
colin-axner, line 13: I'm not sure it is safe to use maps since they aren't determinate and genesis files need to be ordered the same (for verification), there's usually a lot of concern about ensuring `ExportGenesis` is determinate 

I'm also fairly certain this is why the genesis file is filled with a ton of arrays and structs which do indirect mappings

seantking: Also, what do you want to call the variable here? `packet_fees` seems wrong

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+package ibc.applications.middleware.fee.v1;
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/query/v1beta1/pagination.proto";
+import "ibc/applications/middleware/fee/v1/fee.proto";
+import "google/api/annotations.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// Query provides defines the gRPC querier service.
+service Query {
```
colin-axner, line 12: relayers will want a service which queries all incentivized packets

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  // Gets the fee expected for submitting TimeoutPacket msg for the given packet
+  rpc TimeoutFee(QueryTimeoutFeeRequest) returns (QueryTimeoutFeeResponse) {
+    option (google.api.http).get = "/ibc/apps/middleware/fee/v1/timeout_fee";
+  }
+}
+
+// QueryReceiveRequestFee is the request type for quering the receive fee
+message QueryReceiveFeeRequest {
```
colin-axner, line 30: lets add the `QueryHeight` arg while we are at it #149

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+  // Source Port ID
+  string port_id = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
+  // Source Channel ID
+  string channel_id = 2 [(gogoproto.moretags) = "yaml:\"channel_id\""];
+  // Packet sequence
+  string sequence = 3 [(gogoproto.moretags) = "yaml:\"sequence\""];
+  // Caller should provide the intended relayer address in case the fee is dependent on specific relayer(s).
+  string relayer_address = 4 [(gogoproto.moretags) = "yaml:\"relayer_address\""];
```
colin-axner, line 38: Maybe we should leave this off and add it later? It suggests we implemented this functionality

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// QueryAckFeeResponse is the response type for the AckFee RPC
+message QueryAckFeeResponse {
+  ibc.applications.middleware.fee.v1.Fee fee = 1 [(gogoproto.moretags) = "yaml:\"fee\""];
+}
+
+// QueryTimeoutFee is the request type for quering the timeout fee
```
colin-axner, line 63:
```suggestion
// QueryTimeoutFeeRequest is the request type for querying the timeout fee
```

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+  // Caller should provide the intended relayer address in case the fee is dependent on specific relayer(s).
+  string relayer_address = 4 [(gogoproto.moretags) = "yaml:\"relayer_address\""];
+}
+
+// QueryRequestFeeResponse is the response type for the ReceiveFee RPC
+message QueryReceiveFeeResponse {
+  ibc.applications.middleware.fee.v1.Fee fee = 1 [(gogoproto.moretags) = "yaml:\"fee\""];
+}
+
+// QueryAckRequestFee is the request type for quering the acknowledgement fee
```
colin-axner, line 46:
```suggestion
// QueryAckFeeRequest is the request type for querying the acknowledgement fee
```

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+  string port_id = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
+  // Source Channel ID
+  string channel_id = 2 [(gogoproto.moretags) = "yaml:\"channel_id\""];
+  // Packet sequence
+  string sequence = 3 [(gogoproto.moretags) = "yaml:\"sequence\""];
+  // Caller should provide the intended relayer address in case the fee is dependent on specific relayer(s).
+  string relayer_address = 4 [(gogoproto.moretags) = "yaml:\"relayer_address\""];
+}
+
+// QueryRequestFeeResponse is the response type for the ReceiveFee RPC
```
colin-axner, line 41:
```suggestion
// QueryReceiveFeeResponse is the response type for the ReceiveFee RPC
```

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+    option (google.api.http).get = "/ibc/apps/middleware/fee/v1/ack_fee";
+  }
+
+  // Gets the fee expected for submitting TimeoutPacket msg for the given packet
+  rpc TimeoutFee(QueryTimeoutFeeRequest) returns (QueryTimeoutFeeResponse) {
+    option (google.api.http).get = "/ibc/apps/middleware/fee/v1/timeout_fee";
+  }
+}
+
+// QueryReceiveRequestFee is the request type for quering the receive fee
```
colin-axner, line 29:
```suggestion
// QueryReceiveFeeRequest is the request type for querying the receive fee
```

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+message QueryReceiveFeeRequest {
+  // Source Port ID
+  string port_id = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
+  // Source Channel ID
+  string channel_id = 2 [(gogoproto.moretags) = "yaml:\"channel_id\""];
+  // Packet sequence
+  string sequence = 3 [(gogoproto.moretags) = "yaml:\"sequence\""];
```
colin-axner, line 36: yaml is only needed when we have an underscore. The go proto compiler defaults to camelcase, so we need to add the underscore when we have two words

proto/ibc/core/channel/v1/channel.proto
```diff
@@ -132,6 +132,20 @@ message PacketState {
...
+// Source chains refer to packets by source port/channel
+// Destination chains refer to packets by destination port/channel
+message PacketId {
+  option (gogoproto.goproto_getters) = false;
+
+  // channel port identifier
+  string port_id = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
+  // channel unique identifier
+  string channel_id = 2 [(gogoproto.moretags) = "yaml:\"channel_id\""];
+  // packet sequence.
```
colin-axner, line 145:
```suggestion
  // packet sequence
```

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,46 @@
...
+package ibc.applications.middleware.fee.v1;
+
+import "gogoproto/gogo.proto";
+import "ibc/applications/middleware/fee/v1/fee.proto";
+import "ibc/core/channel/v1/channel.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// Msg defines the ibc/fee Msg service.
+service Msg {
+  // RegisterCounterPartyAddress defines a rpc handler method for MsgRegisterCounterPartyAddress
```
colin-axner, line 12:
```suggestion
  // RegisterCounterpartyAddress defines a rpc handler method for MsgRegisterCounterpartyAddress
```

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,46 @@
...
+  // may be called more than once by a relayer, in which case, latest counterparty address is always used.
+  rpc RegisterCounterPartyAddress(MsgRegisterCounterPartyAddress) returns (MsgRegisterCounterPartyAddressResponse);
+  // EscrowPacketFee defines a rpc handler method for MsgEscrowPacketFee
+  // EscrowPacketFee is an open callback that may be called by any module/user that wishes to escrow funds in order to
+  // incentivize the relaying of the given packet.
+  rpc EscrowPacketFee(MsgEscrowPacketFee) returns (MsgEscrowPacketFeeResponse);
+}
+
+// MsgRegisterCounterPartyAddress is the request type for registering the counter party address
+message MsgRegisterCounterPartyAddress {
```
colin-axner, line 25:
```suggestion
message MsgRegisterCounterpartyAddress {
```

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,46 @@
...
+  // destination chain must send back relayer's source address (counterparty address) in acknowledgement This function
+  // may be called more than once by a relayer, in which case, latest counterparty address is always used.
+  rpc RegisterCounterPartyAddress(MsgRegisterCounterPartyAddress) returns (MsgRegisterCounterPartyAddressResponse);
+  // EscrowPacketFee defines a rpc handler method for MsgEscrowPacketFee
+  // EscrowPacketFee is an open callback that may be called by any module/user that wishes to escrow funds in order to
+  // incentivize the relaying of the given packet.
+  rpc EscrowPacketFee(MsgEscrowPacketFee) returns (MsgEscrowPacketFeeResponse);
+}
+
+// MsgRegisterCounterPartyAddress is the request type for registering the counter party address
```
colin-axner, line 24:
```suggestion
// MsgRegisterCounterpartyAddress is the request type for registering the counter party address
```

proto/ibc/applications/middleware/fee/v1/tx.proto
```diff
@@ -0,0 +1,46 @@
...
+message MsgRegisterCounterPartyAddress {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string address              = 1 [(gogoproto.moretags) = "yaml:\"address\""];
+  string counterparty_address = 2 [(gogoproto.moretags) = "yaml:\"address\""];
```
colin-axner, line 30: fix yaml

proto/ibc/applications/middleware/fee/v1/tx.proto
colin-axner: `CounterParty` -> `Counterparty`

colin-axner (commented):
left nits/formatting, looking good!

proto/ibc/applications/middleware/fee/v1/genesis.proto
seantking: Ahhhh, of course, yeah makes sense! Same reason we don't use maps in smart contracts I suppose.

AdityaSripal: I don't think it needs to be explicitly a mapping. I don't think we do that in other genesis files either.

I would make a separate struct

```
GenesisFee{
  packetID    PacketID
  fee Fee
}

GenesisState{
  repeated GenesisFee
}
```

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+
+  // Gets the fee expected for submitting TimeoutPacket msg for the given packet
+  rpc TimeoutFee(QueryTimeoutFeeRequest) returns (QueryTimeoutFeeResponse) {
+    option (google.api.http).get = "/ibc/apps/middleware/fee/v1/timeout_fee";
+  }
+}
+
+// QueryReceiveRequestFee is the request type for quering the receive fee
+message QueryReceiveFeeRequest {
+  // Source Port ID
```
AdityaSripal, line 31: Should we just replace these with a `PacketID`?

proto/ibc/applications/middleware/fee/v1/query.proto
AdityaSripal: No I'd prefer we do it now, so we don't have to break API later. Especially don't want to be breaking proto and rpc apis. We can just  ignore the field for now

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+
+// QueryRequestFeeResponse is the response type for the ReceiveFee RPC
+message QueryReceiveFeeResponse {
+  ibc.applications.middleware.fee.v1.Fee fee = 1 [(gogoproto.moretags) = "yaml:\"fee\""];
+}
+
+// QueryAckRequestFee is the request type for quering the acknowledgement fee
+message QueryAckFeeRequest {
+  // Source Port ID
+  string port_id = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
```
AdityaSripal, line 49: ditto

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,78 @@
...
+
+// QueryAckFeeResponse is the response type for the AckFee RPC
+message QueryAckFeeResponse {
+  ibc.applications.middleware.fee.v1.Fee fee = 1 [(gogoproto.moretags) = "yaml:\"fee\""];
+}
+
+// QueryTimeoutFee is the request type for quering the timeout fee
+message QueryTimeoutFeeRequest {
+  // Source Port ID
+  string port_id = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
```
AdityaSripal, line 66: ditto

proto/ibc/applications/middleware/fee/v1/tx.proto
AdityaSripal: interesting it wasn't caught by compiler

AdityaSripal (changes requested):
Nice work, just need to fix GenesisState and think we can replace (portid, channelid, sequence) with PacketID where applicable

proto/ibc/applications/middleware/fee/v1/genesis.proto
colin-axner: I think a helper struct might be useful here, a struct which contains the packet id, and each of the fees. Then the genesis can have a repeated set of those

proto/ibc/applications/middleware/fee/v1/query.proto
colin-axner: Adding proto fields doesn't break API, it is a backwards compatible definition

proto/ibc/applications/middleware/fee/v1/fee.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package ibc.applications.middleware.fee.v1;
```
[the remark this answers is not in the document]
seantking: Good point. I like this also. 

I considered `ibc.applications.middleware.v1.fee.v1` also as potentially the middleware spec could change but :shrug:

proto/ibc/applications/middleware/fee/v1/genesis.proto
```diff
@@ -0,0 +1,19 @@
...
+
+import "gogoproto/gogo.proto";
+import "ibc/applications/middleware/fee/v1/fee.proto";
+import "ibc/core/channel/v1/channel.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// GenesisState defines the fee middleware genesis state
+message GenesisState {
+  // An entry for the PacketsFees map
+  message PacketsFeesMapEntry {
```
colin-axner, line 13: Let's make this `IdentifiedPacketFee` and defined outside of genesis. This follows our existing pattern

proto/ibc/applications/middleware/fee/v1/genesis.proto
seantking: I pushed up a change before I saw this. Let me know what you think

seantking:
```
  // An entry for the PacketsFees map
  message PacketsFeesMapEntry {
    ibc.core.channel.v1.PacketId           packet_id = 1;
    ibc.applications.middleware.fee.v1.Fee fee       = 2;
  }
  // A mapping of packets -> escrowed fees
  repeated PacketsFeesMapEntry packets_fees_map = 1;
```

I pushed this up before I saw your comment. Want me to change the naming?

proto/ibc/applications/middleware/fee/v1/fee.proto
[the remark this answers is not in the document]
colin-axner: Proto versioning should be independent of the specs. A v2 middleware spec could use the exact same proto definitions which would still be v1

proto/ibc/applications/middleware/fee/v1/tx.proto
seantking: Maybe I'm missing something obvious but what is the change here? 

``` 
// RegisterCounterPartyAddress defines a rpc handler method for MsgRegisterCounterPartyAddress 
```

This is already the comment

proto/ibc/applications/middleware/fee/v1/genesis.proto
AdityaSripal: Still inside genesis.proto just outside of GenesisState correct? Or do you mean outside genesis.proto?

proto/ibc/applications/middleware/fee/v1/tx.proto
seantking: Ahh its the capital P

proto/ibc/applications/middleware/fee/v1/genesis.proto
colin-axner: I would put it in `fee.proto` to be consistent what we [already do](https://github.com/cosmos/ibc-go/blob/0afa15af41997cfb4c5dd5f98de8043961e26195/proto/ibc/core/channel/v1/channel.proto#L31). This will be used by query.proto as well

proto/ibc/applications/middleware/fee/v1/query.proto
AdityaSripal: Yes, but I'd like relayers today to still submit their address. Because their maybe a case in future where some chains are using the address and some chains are not.

I'd like them to just be able to use a single rpc call and not have to change anything as the ecosystem evolves

proto/ibc/applications/middleware/fee/v1/genesis.proto
seantking: Updated

proto/ibc/applications/middleware/fee/v1/query.proto
seantking: Done.

proto/ibc/applications/middleware/fee/v1/query.proto
seantking: gotcha. updated.

proto/ibc/applications/middleware/fee/v1/tx.proto
seantking: Updated.

proto/ibc/applications/middleware/fee/v1/query.proto
seantking: Updated.

proto/ibc/applications/middleware/fee/v1/query.proto
seantking: I added a service for this. Not sure if it's as you wanted. Also, not sure what we want in the request body -> https://github.com/cosmos/ibc-go/pull/272/files#diff-62029a85ec933b1fcec98cf25e2de0aa23db41b56a1cb8758b44dc6fcf806045R82

proto/ibc/applications/middleware/fee/v1/fee.proto
seantking: I'm doing it the same way with the interchain acc helper module. However, I get a bunch of linter errors when I try it here. Any ideas?

```
repeated cosmos.base.v1beta1.Coin amount = 3
      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
```

` unknown extension gogoproto.nullable`
` unknown extension gogoproto.castrepeated`
`unknown extension gogoproto.moretags`

proto/ibc/applications/middleware/fee/v1/query.proto
colin-axner: I think just [optional pagination is good](https://github.com/cosmos/ibc-go/blob/0afa15af41997cfb4c5dd5f98de8043961e26195/proto/ibc/applications/transfer/v1/query.proto#L48)

docs/ibc/proto-docs.md
```diff
@@ -785,6 +786,25 @@ Packet defines a type that carries data across different chains through IBC
```
colin-axner, line 786: I think we need to modify something to get the ics29 proto files to be added

proto/ibc/applications/middleware/fee/v1/fee.proto
colin-axner: Add `import "gogoproto/gogo.proto";` to the top of the file

proto/ibc/applications/middleware/fee/v1/fee.proto
```diff
@@ -0,0 +1,19 @@
...
+// Fee interface
+// See Fee Payment Middleware spec:
+// https://github.com/cosmos/ibc/tree/master/spec/app/ics-029-fee-payment#fee-middleware-contract
+message Fee {
+  cosmos.base.v1beta1.Coin amount = 1;
+}
+
+// Fee associated with a packet_id
+message IdentifiedPacketFee {
+  ibc.core.channel.v1.PacketId           packet_id = 1;
```
colin-axner, line 17: yaml

proto/ibc/applications/middleware/fee/v1/genesis.proto
```diff
@@ -0,0 +1,15 @@
...
+
+import "gogoproto/gogo.proto";
+import "ibc/applications/middleware/fee/v1/fee.proto";
+import "ibc/core/channel/v1/channel.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// GenesisState defines the fee middleware genesis state
+message GenesisState {
+  // A mapping of packets -> escrowed fees
+  repeated ibc.applications.middleware.fee.v1.IdentifiedPacketFee packets_fees_map = 1
```
colin-axner, line 13: I'd remove the `map` part since it is just an array of identified packets. `IdentifiedPacketFees` would be most consistent with what we already have

proto/ibc/applications/middleware/fee/v1/genesis.proto~
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
```
colin-axner, line 2: remove file

proto/ibc/applications/middleware/fee/v1/query.proto
seantking: Added :+1:

docs/ibc/proto-docs.md
seantking: ye just noticed this also\

proto/ibc/applications/middleware/fee/v1/fee.proto
seantking: good catch. updated :+1:

proto/ibc/applications/middleware/fee/v1/genesis.proto~
seantking: good catch. done

docs/ibc/proto-docs.md
AdityaSripal: Need to increase max depth in proto script

proto/ibc/applications/middleware/fee/v1/fee.proto
AdityaSripal: i don't see the fix?

proto/ibc/applications/middleware/fee/v1/genesis.proto~
AdityaSripal: ditto

proto/ibc/applications/middleware/fee/v1/query.proto
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  // Gets the fee expected for submitting TimeoutPacket msg for the given packet
+  rpc TimeoutFee(QueryTimeoutFeeRequest) returns (QueryTimeoutFeeResponse) {
+    option (google.api.http).get = "/ibc/apps/middleware/fee/v1/timeout_fee";
+  }
+
+  // Gets all incentivized packets
+  rpc IncentivizedPackets(QueryIncentivizedPacketsRequest) returns (QueryIncentivizedPacketsResponse) {
+    option (google.api.http).get = "/ibc/apps/middleware/fee/v1/incentivized_packets";
```
AdityaSripal, line 33: These RPC URLs need to take in the arguments sent in the request. See the examples here: https://github.com/cosmos/ibc-go/blob/main/proto/ibc/core/channel/v1/query.proto#L18

As part of this, I'm realizing you may have to unravel the PacketId in the query request and place the strings at the top level of the request field so they can be properly parsed/unparsed from the URL. Sorry!

AdityaSripal (changes requested):
RPC for queries need to change. Everything else looks good!

proto/ibc/applications/middleware/fee/v1/query.proto
AdityaSripal: perhaps there's a way to reference the channel id in the URL as `{packet_id.channel_id}`? But I've never tried so not sure if it will work

proto/ibc/applications/middleware/fee/v1/fee.proto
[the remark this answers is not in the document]
colin-axner: I'm starting to think we should drop the `middleware`. Unless we changed our implementation layout and the ICS layout, I don't see why we should differ. It seems possible a module is partially middleware and then we have to have decide whether it is or isn't a middleware module. Also, at some point in the future these proto files should live in cosmos/ibc

proto/ibc/applications/middleware/fee/v1/fee.proto
```diff
@@ -0,0 +1,21 @@
...
+// https://github.com/cosmos/ibc/tree/master/spec/app/ics-029-fee-payment#fee-middleware-contract
+message Fee {
+  repeated cosmos.base.v1beta1.Coin amount = 1
+      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+}
+
+// Fee associated with a packet_id
+message IdentifiedPacketFee {
+  ibc.core.channel.v1.PacketId           packet_id = 1;
+  ibc.applications.middleware.fee.v1.Fee fee       = 2;
```
colin-axner, line 20: I think this can just be `Fee` when it is defined in the same file
```suggestion
  Fee fee       = 2;
```

proto/ibc/applications/middleware/fee/v1/fee.proto
```diff
@@ -0,0 +1,21 @@
...
+// Fee interface
+// See Fee Payment Middleware spec:
+// https://github.com/cosmos/ibc/tree/master/spec/app/ics-029-fee-payment#fee-middleware-contract
+message Fee {
+  repeated cosmos.base.v1beta1.Coin amount = 1
+      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+}
+
+// Fee associated with a packet_id
+message IdentifiedPacketFee {
```
colin-axner, line 18: We need to be able to determine which relay type the fee corresponds to. There's two approaches, define a type with:
```
  ibc.core.channel.v1.PacketId           packet_id   = 1 [(gogoproto.moretags) = "yaml:\"packet_id\""];
  ibc.applications.middleware.fee.v1.Fee receive_fee = 2 [(gogoproto.moretags) = "yaml:\"receive_fee\""];
  ibc.applications.middleware.fee.v1.Fee ack_fee     = 3 [(gogoproto.moretags) = "yaml:\"ack_fee\""];
  ibc.applications.middleware.fee.v1.Fee timeout_fee = 4 [(gogoproto.moretags) = "yaml:\"timeout_fee\""];
  repeated string                        relayers    = 5;
  ```
 as MsgEscrowFee does now, or create individual mappings as the rpc's do now (different endpoint for recv, ack, timeout)
 
 I don't see a strong design reason not to just bundle all 3 fee's into one type. Thoughts @AdityaSripal @seantking

File: proto/ibc/applications/middleware/fee/v1/query.proto (hunk 36 above)
Comment (colin-axner) [reply proto/ibc/applications/middleware/fee/v1/query.proto:33]: I would look at how SDK handles this

File: proto/ibc/applications/middleware/fee/v1/fee.proto (hunk 38 above)
Comment (AdityaSripal) [reply proto/ibc/applications/middleware/fee/v1/fee.proto:18]: Yea I think its fine to store them in state together as 1 struct containing all 3 fees. I'd still like to have the different rpcs, and have the query funcs split out each fee and return the relevant one to relayer. This would at least conform to spec as is.

Making it so we only send back bundle is a difference from the spec. We could do it, but we would need to make the API changes on spec level as well.

Comment (colin-axner) [reply proto/ibc/applications/middleware/fee/v1/fee.proto:18]: Lets do both for now and ask relayers which they'd prefer?

File: proto/ibc/applications/middleware/fee/v1/query.proto (hunk 36 above)
Comment (seantking) [reply proto/ibc/applications/middleware/fee/v1/query.proto]: `{packet_id.channel_id}` etc worked.

Comment (seantking) [reply proto/ibc/applications/middleware/fee/v1/query.proto]: Updated :+1:

File: proto/ibc/applications/middleware/fee/v1/fee.proto (hunk 33 above)
Comment (seantking) [reply proto/ibc/applications/middleware/fee/v1/fee.proto]: Try now?

File: proto/ibc/applications/middleware/fee/v1/fee.proto (hunk 38 above)
Comment (seantking) [reply proto/ibc/applications/middleware/fee/v1/fee.proto:18]: I updated the `IdentifiedPacketFee` structure. 

Just so we're clear, we are also saying that we want three new queries for getting all of a specific type of fee (timeout etc)?

File: proto/ibc/applications/middleware/fee/v1/genesis.proto~ (hunk 35 above)
Comment (seantking) [reply proto/ibc/applications/middleware/fee/v1/genesis.proto~]: somehow got readded. deleted again :shrug:

File: proto/ibc/applications/middleware/fee/v1/fee.proto (hunk 30 above)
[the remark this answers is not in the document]
Comment (seantking) [reply proto/ibc/applications/middleware/fee/v1/fee.proto:3]: Yeah, fair points. I think this makes sense. @AdityaSripal thoughts?

**Review by AdityaSripal — APPROVED** [review]
Great work!!

File: proto/ibc/applications/middleware/fee/v1/fee.proto (hunk 30 above)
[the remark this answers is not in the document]
Comment (seantking) [reply proto/ibc/applications/middleware/fee/v1/fee.proto]: I updated this

File: proto/ibc/applications/middleware/fee/v1/fee.proto (hunk 38 above)
Comment (seantking) [reply proto/ibc/applications/middleware/fee/v1/fee.proto]: Added another query as requested

**Comment by seantking** [comment]
@colin-axner I believe this is ready to merge

**Review by AdityaSripal — APPROVED** [review]

File: proto/ibc/applications/fee/v1/query.proto [hunk 39]
```diff
@@ -0,0 +1,123 @@
...
+
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// Query provides defines the gRPC querier service.
+service Query {
+  // Gets the fee expected for submitting ReceivePacket msg for the given packet
+  rpc ReceiveFee(QueryReceiveFeeRequest) returns (QueryReceiveFeeResponse) {
+    option (google.api.http).get =
+        "/ibc/apps/fee/v1/receive_fee/port/{packet_id.port_id}/channel/{packet_id.channel_id}/sequence/"
+        "{packet_id.sequence}/relayer/{relayer_address}/height/{query_height}";
```
colin-axner, line 20: I think we should leave off the relayer address and query height so that they are [query parameters](https://github.com/aspnet/AspLabs/blob/c1e59cacf7b9606650d6ec38e54fa3a82377f360/src/GrpcHttpApi/sample/Proto/google/api/http.proto#L90)

I think query height is pretty naturally a query parameter and I think it is useful to have the relayer address as a query parameter. This can allow default behaviour to query incentivized packets which are unassigned (can be relayed by any relayer). I imagine for legal reasons, some relayers won't want to query for packets which have assigned relayers even if they are an assigned relayer

proto/ibc/applications/fee/v1/query.proto
```diff
@@ -0,0 +1,123 @@
...
+        "/ibc/apps/fee/v1/timeout_fee/port/{packet_id.port_id}/channel/{packet_id.channel_id}/sequence/"
+        "{packet_id.sequence}/{relayer_address}/height/{query_height}";
+  }
+
+  // Gets all incentivized packets
+  rpc IncentivizedPackets(QueryIncentivizedPacketsRequest) returns (QueryIncentivizedPacketsResponse) {
+    option (google.api.http).get = "/ibc/apps/fee/v1/incentivized_packets/height/{query_height}";
+  }
+
+  // Gets all incentivized packets
```
colin-axner, line 42:
```suggestion
  // Gets the specified incentivized packet
```

proto/ibc/applications/fee/v1/query.proto
```diff
@@ -0,0 +1,123 @@
...
+  rpc IncentivizedPacket(QueryIncentivizedPacketRequest) returns (QueryIncentivizedPacketResponse) {
+    option (google.api.http).get =
+        "/ibc/apps/fee/v1/incentivized_packet/port/{packet_id.port_id}/channel/{packet_id.channel_id}/sequence/"
+        "{packet_id.sequence}/height/{query_height}";
+  }
+}
+
+// QueryReceiveRequestFee is the request type for quering the receive fee
```
colin-axner, line 50:
```suggestion
// QueryReceiveFeeRequest is the request type for querying the receive fee
```

proto/ibc/applications/fee/v1/query.proto
```diff
@@ -0,0 +1,123 @@
...
+  // Height to query at
+  uint64 query_height = 3 [(gogoproto.moretags) = "yaml:\"query_height\""];
+}
+
+// QueryTimeoutFeeResponse is the response type for the timeout RPC
+message QueryTimeoutFeeResponse {
+  ibc.applications.fee.v1.Fee fee = 1;
+}
+
+// QueryIncentivizedPacketRequest is the request type for querying for all incentivized packets
```
colin-axner, line 95:
```suggestion
// QueryIncentivizedPacketsRequest is the request type for querying for all incentivized packets
```

proto/ibc/applications/fee/v1/query.proto
```diff
@@ -0,0 +1,123 @@
...
+message QueryIncentivizedPacketsRequest {
+  // pagination defines an optional pagination for the request.
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+  // Height to query at
+  uint64 query_height = 2 [(gogoproto.moretags) = "yaml:\"query_height\""];
+}
+
+// QueryTimeoutFeeResponse is the response type for the timeout RPC
```
colin-axner, line 103:
```suggestion
// QueryIncentivizedPacketsResponse is the response type for the incentivized packets RPC
```

proto/ibc/applications/fee/v1/query.proto
```diff
@@ -0,0 +1,123 @@
...
+
+// QueryIncentivizedPacketRequest is the request type for querying for all incentivized packets
+message QueryIncentivizedPacketRequest {
+  // PacketID
+  ibc.core.channel.v1.PacketId packet_id = 1 [(gogoproto.moretags) = "yaml:\"packet_id\""];
+  // Height to query at
+  uint64 query_height = 2 [(gogoproto.moretags) = "yaml:\"query_height\""];
+}
+
+// QueryTimeoutFeeResponse is the response type for the timeout RPC
```
colin-axner, line 118:
```suggestion
// QueryIncentivizedPacketsResponse is the response type for the incentivized packet RPC
```

proto/ibc/applications/fee/v1/query.proto
```diff
@@ -0,0 +1,123 @@
...
+    option (google.api.http).get =
+        "/ibc/apps/fee/v1/incentivized_packet/port/{packet_id.port_id}/channel/{packet_id.channel_id}/sequence/"
+        "{packet_id.sequence}/height/{query_height}";
+  }
+}
+
+// QueryReceiveRequestFee is the request type for quering the receive fee
+message QueryReceiveFeeRequest {
+  // PacketID
+  ibc.core.channel.v1.PacketId packet_id = 1 [(gogoproto.moretags) = "yaml:\"packet_id\""];
```
colin-axner, line 53: sorry for mentioning this so late, but `query.proto` doesn't need yaml for some reason. At least [SDK](https://github.com/cosmos/cosmos-sdk/blob/bf049d36d1f955302f0f4aacadb773a90496813c/proto/cosmos/group/v1beta1/query.proto#L130) doesn't do it (neither do [we](https://github.com/cosmos/ibc-go/blob/0afa15af41997cfb4c5dd5f98de8043961e26195/proto/ibc/core/client/v1/query.proto#L64)). I will remove them

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -0,0 +1,49 @@
...
+import "gogoproto/gogo.proto";
+import "ibc/applications/fee/v1/fee.proto";
+import "ibc/core/channel/v1/channel.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// Msg defines the ibc/fee Msg service.
+service Msg {
+  // RegisterCounterpartyAddress defines a rpc handler method for MsgRegisterCounterpartyAddress
+  // RegisterCounterpartyAddress is called by the relayer on each channelEnd and allows them to specify their
+  // counterparty address before relaying This ensures they will be properly compensated for forward relaying since
```
colin-axner, line 14:
```suggestion
  // counterparty address before relaying. This ensures they will be properly compensated for forward relaying since
```

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -0,0 +1,49 @@
...
+  // destination chain must send back relayer's source address (counterparty address) in acknowledgement This function
```
colin-axner, line 15:
```suggestion
  // destination chain must send back relayer's source address (counterparty address) in acknowledgement. This function
```

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -0,0 +1,49 @@
...
+// MsgEscrowPacketFee defines the request type EscrowPacketFee RPC
+message MsgEscrowPacketFee {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  ibc.core.channel.v1.PacketId packet_id   = 1 [(gogoproto.moretags) = "yaml:\"packet_id\""];
+  ibc.applications.fee.v1.Fee  receive_fee = 2 [(gogoproto.moretags) = "yaml:\"receive_fee\""];
+  ibc.applications.fee.v1.Fee  ack_fee     = 3 [(gogoproto.moretags) = "yaml:\"ack_fee\""];
+  ibc.applications.fee.v1.Fee  timeout_fee = 4 [(gogoproto.moretags) = "yaml:\"timeout_fee\""];
+  repeated string              relayers    = 5;
```
colin-axner, line 45: This should use `IdentifiedPacketFee`. Any reason not to?

colin-axner (approved):
Fantastic work!

colin-axner:
enabling auto merge, I would like the two suggestions I left to be addressed in a follow up or discussed:
- [ ] [using query params for query height/relayer](https://github.com/cosmos/ibc-go/pull/272#discussion_r676486321)
- [ ] [using incentivized packet fee](https://github.com/cosmos/ibc-go/pull/272#discussion_r676495836)

proto/ibc/applications/fee/v1/tx.proto
AdityaSripal: None think it was just outdated

PR #276, opened 2021-07-21
ICS 29: Fee Middleware

AdityaSripal:
## Description

Implements fee middleware as specified [here](https://github.com/cosmos/ibc/blob/master/spec/app/ics-029-fee-payment/README.md)

closes: #XXXX

---

write a little note why.


modules/apps/transfer/module.go
```diff
@@ -213,6 +214,10 @@ func ValidateTransferChannelParams(
 		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid port: %s, expected %s", portID, boundPort)
 	}
 
+	if portID != counterparty.PortId {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "portID: %s does not match counterparty portID: %s", portID, counterparty.PortId)
+	}
```
colin-axner, line 219: why is this code not included in main? We should open a separate pr with any changes made outside of `29-fee`

modules/apps/transfer/module.go
```diff
@@ -190,6 +190,7 @@ func ValidateTransferChannelParams(
 	ctx sdk.Context,
 	keeper keeper.Keeper,
 	order channeltypes.Order,
+	counterparty channeltypes.Counterparty,
```
colin-axner, line 193: this is API breaking

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/cosmos/ibc-go/v3/modules/apps/29-fee/types"
+	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+)
+
+var _ types.MsgServer = Keeper{}
+
+// RegisterCounterpartyAddress is called by the relayer on each channelEnd and allows them to specify their counterparty address before relaying
+// This ensures they will be properly compensated for forward relaying on the source chain since the destination chain must send back relayer's source address (counterparty address) in acknowledgement
+// This function may be called more than once by relayers, in which case, the previous counterparty address will be overwritten by the new counterparty address
+func (k Keeper) RegisterCounterpartyAddress(goCtx context.Context, msg *types.MsgRegisterCounterpartyAddress) (*types.MsgRegisterCounterpartyAddressResponse, error) {
```
soareschen, line 17: This looks to me like a global API, as I do not see the `channelEnd` being specified anywhere. But in a multi-chain setting, a relayer usually use the same wallet to relay from multiple chains. We cannot use the same counterparty address address for all channels that are relayed by the same relayer wallet, as the relayer's counterparty address for each chain is different.

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, channeltypes.ErrSequenceSendNotFound
+	}
+
+	packetId := channeltypes.NewPacketId(
+		msg.SourceChannelId,
+		msg.SourcePortId,
+		sequence,
+	)
+
+	identifiedPacket := types.NewIdentifiedPacketFee(packetId, msg.Fee, msg.Signer, msg.Relayers)
```
soareschen, line 46: Are the `Coin` values in `msg.Fee` properly sanitized and validated using `Coin.NewCoin` during deserialization from protobuf? If not, the `Coin` values may contain negative values, and `coins.Add` may panic if the coins are not sorted.

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  ibc.applications.fee.v1.Fee fee = 1 [(gogoproto.nullable) = false];
+  // source channel port identifier
+  string source_port_id = 2 [(gogoproto.moretags) = "yaml:\"source_port_id\""];
+  // source channel unique identifier
+  string source_channel_id = 3 [(gogoproto.moretags) = "yaml:\"source_channel_id\""];
+  // account address to refund fee if necessary
+  string          signer   = 4;
```
soareschen, line 53: Is this field being validated anywhere that it is the same signer as the one that signed the transaction? If not, anyone will be able to escrow fees of any amount on behalf of anyone else.

proto/ibc/applications/fee/v1/fee.proto
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// IdentifiedPacketFee contains the relayer fee along with the associated metadata needed to process it.
+// This includes the PacketId identifying the packet the fee is paying for,
+// the refund address to which any unused funds are refunded,
+// and an optional list of relayers that are permitted to receive the fee.
+message IdentifiedPacketFee {
+  ibc.core.channel.v1.PacketId packet_id      = 1 [(gogoproto.nullable)     = false, (gogoproto.moretags) = "yaml:\"packet_id\""];
+  Fee                          fee            = 2 [(gogoproto.nullable) = false];
+  string                       refund_address = 3 [(gogoproto.moretags) = "yaml:\"refund_address\""];
```
soareschen, line 37: This has the same issue as `MsgPayPacketFee`. `IdentifiedPacketFee` is used in `MsgPayPacketFeeAsync`. Not validating this will allow anyone to escrow fees from any account.

soareschen (changes requested):

modules/apps/29-fee/keeper/msg_server.go
soareschen: Nvm, found the validation in `ValidateBasic`.

proto/ibc/applications/fee/v1/tx.proto
soareschen: Found them being handled in `sdk.Msg.GetSigners`. This can be disregarded.

proto/ibc/applications/fee/v1/fee.proto
soareschen: Found them being handled in `sdk.Msg.GetSigners`. This can be disregarded.

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,251 @@
...
+	err := im.keeper.RefundFeesOnChannel(ctx, portID, channelID)
+	// error should only be non-nil if there is a bug in the code
+	// that causes module account to have insufficient funds to refund
+	// all escrowed fees on the channel.
+	// Disable all channels to allow for coordinated fix to the issue
+	// and mitigate/reverse damage.
+	// NOTE: Underlying application's packets will still go through, but
+	// fee module will be disabled for all channels
+	if err != nil {
+		im.keeper.DisableAllChannels(ctx)
```
soareschen, line 146: Is it safe to just disable the fee for all channels on one chain, without notifying the counterparty chains?

According to the logic of `OnRecvPacket` and `OnAcknowledgementPacket`, whether the fee module is enabled or not depends on the local state of the chain. So what would happen if one chain has fee enabled and the other chain has fee disabled, either via `DisableAllChannels` or some other corner cases? Would it cause the raw `Acknowledgement` to be serialized and deserialized in different ways, thereby causing the deserialization to fail?

soareschen (changes requested):

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,139 @@
...
+	coins = coins.Add(identifiedFee.Fee.TimeoutFee...)
+
+	if err := k.bankKeeper.SendCoinsFromAccountToModule(
+		ctx, refundAcc, types.ModuleName, coins,
+	); err != nil {
+		return err
+	}
+
+	// Store fee in state for reference later
+	k.SetFeeInEscrow(ctx, identifiedFee)
```
soareschen, line 40: It looks like this will override any fee that are already registered in the store, if `EscrowPacketFee` is called multiple times with the same `PacketId`.

modules/apps/29-fee/keeper/relay.go
```diff
@@ -0,0 +1,31 @@
...
+func (k Keeper) SendPacket(ctx sdk.Context, chanCap *capabilitytypes.Capability, packet ibcexported.PacketI) error {
+	return k.ics4Wrapper.SendPacket(ctx, chanCap, packet)
+}
+
+// WriteAcknowledgement wraps IBC ChannelKeeper's WriteAcknowledgement function
+// ICS29 WriteAcknowledgement is used for asynchronous acknowledgements
+func (k Keeper) WriteAcknowledgement(ctx sdk.Context, chanCap *capabilitytypes.Capability, packet ibcexported.PacketI, acknowledgement []byte) error {
+	// retrieve the forward relayer that was stored in `onRecvPacket`
+	packetId := channeltypes.NewPacketId(packet.GetSourceChannel(), packet.GetSourcePort(), packet.GetSequence())
+	relayer, _ := k.GetForwardRelayerAddress(ctx, packetId)
```
soareschen, line 22: There is no check here on whether fee is enabled for the given channel and port. It also does not handle the case when no forward relayer address is found.

soareschen (changes requested):

modules/apps/29-fee/keeper/relay.go
```diff
@@ -0,0 +1,31 @@
...
+
+	k.DeleteForwardRelayerAddress(ctx, packetId)
+
+	ack := types.NewIncentivizedAcknowledgement(relayer, acknowledgement)
```
soareschen, line 26: If my understanding on the [core logic](https://github.com/cosmos/ibc-go/blob/ics29-fee-middleware/modules/core/keeper/msg_server.go#L516-L531) is correct, the `acknowledgement` we get here is already serialized from the `IncentivizedAcknowledgement` returned from `OnRecvPacket` by the fee middleware. So this would result in double wrapping of the `acknowledgement` bytes, and the inner module at the counterparty chain would not able to parse the inner acknowledgement correctly.

soareschen (changes requested):

soareschen (changes requested):

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,139 @@
...
+	return nil
+}
+
+// DistributePacketFees pays the acknowledgement fee & receive fee for a given packetId while refunding the timeout fee to the refund account associated with the Fee.
+func (k Keeper) DistributePacketFees(ctx sdk.Context, refundAcc, forwardRelayer string, reverseRelayer sdk.AccAddress, feeInEscrow types.IdentifiedPacketFee) {
+	// distribute fee for forward relaying
+	forward, err := sdk.AccAddressFromBech32(forwardRelayer)
+	if err == nil {
+		k.distributeFee(ctx, forward, feeInEscrow.Fee.ReceiveFee)
```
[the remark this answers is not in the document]
soareschen: The `forwardRelayer` may also be an empty string if the relayer did not set a counterparty address on the counterparty chain. So in that case it is also necessary to refund the receive fee.

soareschen (changes requested):

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -0,0 +1,307 @@
...
+// e.g. ModuleAccount has insufficient balance to refund users.
+// In this case, chain developers should investigate the issue, fix it,
+// and then re-enable the fee module in a coordinated upgrade.
+func (k Keeper) DisableAllChannels(ctx sdk.Context) {
+	store := ctx.KVStore(k.storeKey)
+	iterator := sdk.KVStorePrefixIterator(store, []byte(types.FeeEnabledKeyPrefix))
+
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		store.Delete(iterator.Key())
```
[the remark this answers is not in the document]
colin-axner: Great catch! This is a bug. It is not safe to delete key while iterating over it. #780 

As a side note, there should be a linter for this. cc @crodriguezvega this could one of the patterns for the process trail of bits proposed (I forget what it was called)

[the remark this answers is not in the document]
crodriguezvega: Yes, [semgrep](https://semgrep.dev). We could maybe use it for this indeed.

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// distributeFee will attempt to distribute the escrowed fee to the receiver address.
+// If the distribution fails for any reason (such as the receiving address being blocked),
+// the state changes will be discarded.
+func (k Keeper) distributeFee(ctx sdk.Context, receiver sdk.AccAddress, fee sdk.Coins) {
+	// cache context before trying to send to reverse relayer
+	cacheCtx, writeFn := ctx.CacheContext()
+
+	err := k.bankKeeper.SendCoinsFromModuleToAccount(cacheCtx, types.ModuleName, receiver, fee)
```
soareschen, line 96: The call to `SendCoinsFromModuleToAccount` may fail if the [receiver address is blocked](https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/keeper/keeper.go#L302). And since the forward relayer address can be anything, there is a chance that this can be failed deliberately.

We may want to have some ways to handle the error, or at least have a log if the bank transfer fails. That way operators can at least found out that something is wrong is the fee module and begin investigation.

soareschen (changes requested):

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,136 @@
...
+		refundAccAddr, err := sdk.AccAddressFromBech32(identifiedFee.RefundAddress)
+		if err != nil {
+			refundErr = err
+			return true
+		}
+
+		// refund all fees to refund address
+		// Use SendCoins rather than the module account send functions since refund address may be a user account or module address.
+		// if any `SendCoins` call returns an error, we return error and stop iteration
+		if err = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, refundAccAddr, identifiedFee.Fee.RecvFee); err != nil {
```
soareschen, line 120: The call to `SendCoinsFromModuleToAccount` may fail if the [receiver address is blocked](https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/keeper/keeper.go#L302). If someone managed to use an address that is blocked either currently or in the future, there is a chance for exploit that cause all subsequent refunds in this loop to fail.

soareschen (changes requested):

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,253 @@
...
+		return im.app.OnRecvPacket(ctx, packet, relayer)
+	}
+
+	ack := im.app.OnRecvPacket(ctx, packet, relayer)
+
+	forwardRelayer, found := im.keeper.GetCounterpartyAddress(ctx, relayer.String())
+
+	// incase of async aknowledgement (ack == nil) store the ForwardRelayer address for use later
+	if ack == nil && found {
+		im.keeper.SetForwardRelayerAddress(ctx, channeltypes.NewPacketId(packet.GetSourceChannel(), packet.GetSourcePort(), packet.GetSequence()), forwardRelayer)
```
soareschen, line 193: The source channel and source port are not unique on the destination chain. This has to be set with the destination channel and port, otherwise it will get overridden when two source chains have the same source channel and port.

To make things clear, consider two separate functions `NewPacketIdForSourceChain(packet)` and `NewPacketIdForDestinationChain(packet)` to ensure that the packet ID is always derived uniquely on the respective chains.

soareschen (changes requested):

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -0,0 +1,67 @@
...
+
+var _ types.MsgServer = Keeper{}
+
+// RegisterCounterpartyAddress is called by the relayer on each channelEnd and allows them to specify their counterparty address before relaying
+// This ensures they will be properly compensated for forward relaying on the source chain since the destination chain must send back relayer's source address (counterparty address) in acknowledgement
+// This function may be called more than once by relayers, in which case, the previous counterparty address will be overwritten by the new counterparty address
+func (k Keeper) RegisterCounterpartyAddress(goCtx context.Context, msg *types.MsgRegisterCounterpartyAddress) (*types.MsgRegisterCounterpartyAddressResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	k.SetCounterpartyAddress(
```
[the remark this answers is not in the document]
AdityaSripal: > We may want to validate that msg.Address is a valid address for the local chain.

This will already be checked during antehandler, since address must sign message.

Do you have suggestions on Counterparty address validation?

soareschen (changes requested):

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,263 @@
...
+	acknowledgement []byte,
+	relayer sdk.AccAddress,
+) error {
+	if !im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
+		return im.app.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
+
+	ack := new(types.IncentivizedAcknowledgement)
+	if err := types.ModuleCdc.UnmarshalJSON(acknowledgement, ack); err != nil {
+		return sdkerrors.Wrapf(err, "cannot unmarshal ICS-29 incentivized packet acknowledgement: %v", ack)
```
soareschen, line 223: I'm not sure of the exact semantics of `UnmarshalJSON`, but given the prevalent practices in Go, would this unmarshalling succeed for _any_ JSON result regardless of whether it contains the expected fields? If so, the error condition would not be triggered unless it contains non-JSON content.

I have been thinking that it may be more practical if the fee middleware can be enabled/disabled for each chain individually without requiring negotiation. Looking at all the protocol interaction, the only requirement for both sides to enable fee simultaneously is so the ack bytes containing the forward relayer addresss can be decoded correctly.

We should instead re-design this such that in case the decoding of the acknowledgement bytes fail, it would simply refund the escrowed receive fee and forward the raw ack to the inner module. Granted, this would also require `UnmarshalJSON` to indeed fail if the JSON raw content do not contain the forward relayer address and result fields.

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,253 @@
...
+	ack := im.app.OnRecvPacket(ctx, packet, relayer)
+
+	forwardRelayer, found := im.keeper.GetCounterpartyAddress(ctx, relayer.String())
+
+	// incase of async aknowledgement (ack == nil) store the ForwardRelayer address for use later
+	if ack == nil && found {
+		im.keeper.SetForwardRelayerAddress(ctx, channeltypes.NewPacketId(packet.GetSourceChannel(), packet.GetSourcePort(), packet.GetSequence()), forwardRelayer)
+	}
+
+	return types.NewIncentivizedAcknowledgement(forwardRelayer, ack.Acknowledgement())
```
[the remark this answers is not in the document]
soareschen: I have been thinking that it may be more practical if the fee middleware can be enabled/disabled for each chain individually without requiring negotiation. Would it be possible to insert that information into the acknowledgement bytes while maintaining backward compatibility?

For example, can we add an extra field like `extra_data` to the `Acknowledgement` protobuf, so that all middlewares can make use of that for backward-compatible ack processing?

modules/apps/29-fee/keeper/msg_server.go
soareschen: Would it be possible to register the forward relayer address on the source chain after this is registered on the destination chain? In theory, the source chain can also do the mapping with a merkle proof that the mapping exist on the destination chain.

If we can do the mapping, then there is no need for complicated version negotiation to enable fees. Each chain would be able to enable fee on any channel regardless of whether the counterparty chain has been upgraded, because the packet payloads do not need to be modified.

In the case that the source chain has fee enabled but the destination chain have not upgraded to support fee and the register counterparty address API, the only downside would just be that the escrowed receive fee is always refunded instead of being paid to the forward relayer.

On the other hand, if both chains are upgraded to support fee, the fee can still be asynchronously enabled on the source chain without it having to be enabled on the destination chain. In such case the relayer would still be able to register the counterparty and forward address on both chains, and then the source chain can still pay for the receive fee without having to modify the packet format.

soareschen (changes requested):

modules/apps/29-fee/types/tx.pb.go
```diff
@@ -0,0 +1,1535 @@
...
+// is compatible with the proto package it is being compiled against.
+// A compilation error at this line likely means your copy of the
+// proto package needs to be updated.
+const _ = proto.GoGoProtoPackageIsVersion3 // please upgrade the proto package
+
+// MsgRegisterCounterpartyAddress is the request type for registering the counterparty address
+type MsgRegisterCounterpartyAddress struct {
+	Address             string `protobuf:"bytes,1,opt,name=address,proto3" json:"address,omitempty"`
+	CounterpartyAddress string `protobuf:"bytes,2,opt,name=counterparty_address,json=counterpartyAddress,proto3" json:"counterparty_address,omitempty" yaml:"counterparty_address"`
+	ChannelId           string `protobuf:"bytes,3,opt,name=channel_id,json=channelId,proto3" json:"channel_id,omitempty" yaml:"channel_id"`
```
soareschen, line 36: We just realized that with interchain accounts, this may be cumbersome for relayers to have to register the counterparty address for each new channel added. Would it make more sense if we register counterparty address by client ID or connection ID?

crodriguezvega: Issue [#1133](https://github.com/cosmos/ibc-go/issues/1133) opened to discuss.

damiannolan (approved):
🚀

seantking (approved):

charleenfei (approved):

modules/apps/29-fee/types/codec.go
```diff
@@ -0,0 +1,42 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/msgservice"
+)
+
+// RegisterLegacyAminoCodec registers the necessary x/ibc 29-fee interfaces and concrete types
+// on the provided LegacyAmino codec. These types are used for Amino JSON serialization.
+func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
+	cdc.RegisterConcrete(&MsgRegisterCounterpartyAddress{}, "cosmos-sdk/MsgRegisterCounterpartyAddress", nil)
+	cdc.RegisterConcrete(&MsgPayPacketFee{}, "cosmos-sdk/MsgPayPacketFee", nil)
+	cdc.RegisterConcrete(&MsgPayPacketFeeAsync{}, "cosmos-sdk/MsgPayPacketFeeAsync", nil)
```
fedekunze, line 15: why `cosmos-sdk`? Is amino support required?

modules/apps/29-fee/types/codec.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// RegisterInterfaces register the 29-fee module interfaces to protobuf
+// Any.
+func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
+	registry.RegisterImplementations((*sdk.Msg)(nil), &MsgRegisterCounterpartyAddress{}, &MsgPayPacketFee{}, &MsgPayPacketFeeAsync{})
```
fedekunze, line 21: nit, split lines

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+
+	fees := []types.PacketFee{packetFee}
+	if feesInEscrow, found := k.GetFeesInEscrow(ctx, packetID); found {
+		fees = append(fees, feesInEscrow.PacketFees...)
+	}
+
+	packetFees := types.NewPacketFees(fees)
+	k.SetFeesInEscrow(ctx, packetID, packetFees)
```
fedekunze, line 41: nit, add comment to explain this in more detail

fedekunze (commented):
first pass

fedekunze (approved):
utACK, will review in detail soon

modules/apps/29-fee/types/fee.go
```diff
@@ -0,0 +1,92 @@
...
+	if !fee.RecvFee.IsValid() {
+		errFees = append(errFees, "recv fee invalid")
+	}
+	if !fee.TimeoutFee.IsValid() {
+		errFees = append(errFees, "timeout fee invalid")
+	}
+
+	if len(errFees) > 0 {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidCoins, "contains invalid fees: %s", strings.Join(errFees, " , "))
+	}
```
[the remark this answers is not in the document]
seantking: Multiple fees may be invalid I think it's good to let the user know.

Issue #261, opened 2021-07-15
Fee Middleware: Implement Handshake callbacks


AdityaSripal:

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #307, opened 2021-08-03
feat: Add handshake logic to ics29

AdityaSripal:
## Description

closes: #261

Opening handshake is done. We should open separate issue for closing handshake to complete once escrow logic is in.

---

write a little note why.


modules/apps/29-fee/module.go
```diff
@@ -191,7 +201,23 @@ func (am AppModule) OnChanOpenInit(
 	counterparty channeltypes.Counterparty,
 	version string,
 ) error {
-	return nil
+	feeVersion, appVersion := channeltypes.SplitChannelVersion(version)
+	if feeVersion != types.Version {
+		return sdkerrors.Wrapf(types.ErrInvalidVersion, "expected: %s, got: %s", types.Version, feeVersion)
+	}
+	// Claim channel capability passed back by IBC module
+	if err := am.scopedKeeper.ClaimCapability(ctx, chanCap, host.ChannelCapabilityPath(portID, channelID)); err != nil {
```
seantking, line 209: Question: Will every middleware application want to do this claim capability step? Is the idea that the first middleware will always create the capability and just pass it along to the next?

modules/apps/29-fee/module.go
```diff
@@ -206,7 +232,43 @@ func (am AppModule) OnChanOpenTry(
 	version,
 	counterpartyVersion string,
 ) error {
-	return nil
+	feeVersion, appVersion := channeltypes.SplitChannelVersion(version)
+	cpFeeVersion, cpAppVersion := channeltypes.SplitChannelVersion(counterpartyVersion)
+
+	if feeVersion != types.Version {
+		return sdkerrors.Wrapf(types.ErrInvalidVersion, "expected: %s, got: %s", types.Version, feeVersion)
```
seantking, line 239:
```suggestion
		return sdkerrors.Wrapf(types.ErrInvalidVersion, "expected version: %s, got: %s", types.Version, feeVersion)
```

modules/apps/transfer/keeper/genesis.go
```diff
@@ -9,20 +9,22 @@ import (
 
 // InitGenesis initializes the ibc-transfer state and binds to PortID.
 func (k Keeper) InitGenesis(ctx sdk.Context, state types.GenesisState) {
-	k.SetPort(ctx, state.PortId)
+	k.SetPorts(ctx, state.PortIds)
```
seantking, line 12: Question: Just wondering why we changed this to handle multiple port-ids?

seantking (approved):
LGTM approved.

modules/apps/29-fee/module.go
AdityaSripal: If the middleware is acting on information coming from app to IBC, then this is a useful pattern to enforce that communication from app to IBC must go through middleware first.

This is a good design pattern to enforce anyway since it enforces the middleware setup in code.

modules/apps/transfer/module.go
```diff
@@ -208,9 +209,19 @@ func ValidateTransferChannelParams(
...
 
 	// Require portID is the portID transfer module is bound to
-	boundPort := keeper.GetPort(ctx)
-	if boundPort != portID {
-		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid port: %s, expected %s", portID, boundPort)
+	boundPorts := keeper.GetPorts(ctx)
+	var portOk bool
+	for _, boundPort := range boundPorts {
+		if boundPort == portID {
+			portOk = true
```
colin-axner, line 216: This could break from the for loop

modules/apps/transfer/module.go
```diff
@@ -208,9 +209,19 @@ func ValidateTransferChannelParams(
 	}
 
 	// Require portID is the portID transfer module is bound to
-	boundPort := keeper.GetPort(ctx)
-	if boundPort != portID {
-		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid port: %s, expected %s", portID, boundPort)
+	boundPorts := keeper.GetPorts(ctx)
```
colin-axner, line 212: The benefit of using `GetPorts` as opposed to storing each port under `ports/` key is less migration code?

modules/apps/transfer/module.go
```diff
@@ -208,9 +209,19 @@ func ValidateTransferChannelParams(
...
+		if boundPort == portID {
+			portOk = true
+		}
+	}
+	if !portOk {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid port: %s not bound by transfer module", portID)
+	}
+
+	if portID != counterparty.PortId {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "portID: %s do not match counterparty portID: %s", portID, counterparty.PortId)
```
colin-axner, line 224:
```suggestion
		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "portID: %s does not match counterparty portID: %s", portID, counterparty.PortId)
```

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,245 @@
...
+		{
+			"hanging delimiter",
+			"fee29-1:ics20-1:",
+			false,
+		},
+	}
+
+	for _, tc := range testCases {
+		ctx := suite.chainA.GetContext()
+		cap, err := suite.chainA.GetSimApp().ScopedIBCKeeper.NewCapability(ctx, tc.name)
```
colin-axner, line 55: Why is the new capability using `tc.name`?

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,245 @@
...
+		suite.Require().NoError(err)
+		err = suite.module.OnChanOpenInit(
+			ctx,
+			channeltypes.UNORDERED,
+			[]string{"connection-1"},
+			transfertypes.FeePortID,
+			"channel-1",
+			cap,
+			channeltypes.NewCounterparty(transfertypes.FeePortID, "channel-2"),
+			tc.version,
```
colin-axner, line 65: Does this need to be hardcoded or can we setup a connection and use those values? Also I believe the oounterparty channel id must be empty in this step

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,245 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		ctx := suite.chainA.GetContext()
+		cap, err := suite.chainA.GetSimApp().ScopedIBCKeeper.NewCapability(ctx, tc.name)
+		suite.Require().NoError(err)
+		err = suite.module.OnChanOpenInit(
+			ctx,
+			channeltypes.UNORDERED,
```
colin-axner, line 59: Can we have tests for both unordered and ordered since both should be supported (happy case ensuring that check isn't added later)

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,245 @@
...
+			// check that capabilities are properly claimed and issued
+			ctx := suite.chainA.GetContext()
+			ibcCap, ok := suite.chainA.GetSimApp().ScopedIBCFeeKeeper.GetCapability(ctx, host.ChannelCapabilityPath(transfertypes.FeePortID, "channel-1"))
+			suite.Require().NotNil(ibcCap, "IBC capability is nil on fee keeper")
+			suite.Require().True(ok)
+
+			appFeeCap, ok := suite.chainA.GetSimApp().ScopedIBCFeeKeeper.GetCapability(ctx, types.AppCapabilityName(transfertypes.FeePortID, "channel-1"))
+			suite.Require().NotNil(appFeeCap, "App capability not created or owned by ibc fee keeper")
+			suite.Require().True(ok)
+			transferCap, ok := suite.chainA.GetSimApp().ScopedTransferKeeper.GetCapability(ctx, host.ChannelCapabilityPath(transfertypes.FeePortID, "channel-1"))
```
colin-axner, line 80: Why does the transfer keeper have access to the capability owned by fee and core IBC? This should be the app capability, but seems problematic that the test is passing

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,245 @@
...
+func (suite *FeeTestSuite) TestOnChanOpenTry() {
+	testCases := []struct {
+		name      string
+		version   string
+		cpVersion string
+		capExists bool
+		expPass   bool
+	}{
+		{
+			"valid fee middleware and transfer version",
```
colin-axner, line 99: can you add a test where the fee and transfer versions are valid but don't match (maybe have an extra nested version)

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,245 @@
...
+		err = suite.module.OnChanOpenTry(
+			ctx,
+			channeltypes.UNORDERED,
+			[]string{"connection-1"},
+			transfertypes.FeePortID,
+			"channel-1",
+			cap,
+			channeltypes.NewCounterparty(transfertypes.FeePortID, "channel-0"),
+			tc.version,
+			tc.cpVersion,
```
colin-axner, line 188: ditto on hard coding

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,245 @@
...
+			// check that capabilities are properly claimed and issued
+			ctx := suite.chainA.GetContext()
+			ibcCap, ok := suite.chainA.GetSimApp().ScopedIBCFeeKeeper.GetCapability(ctx, host.ChannelCapabilityPath(transfertypes.FeePortID, "channel-1"))
+			suite.Require().NotNil(ibcCap, "IBC capability is nil on fee keeper")
+			suite.Require().True(ok)
+
+			appFeeCap, ok := suite.chainA.GetSimApp().ScopedIBCFeeKeeper.GetCapability(ctx, types.AppCapabilityName(transfertypes.FeePortID, "channel-1"))
+			suite.Require().NotNil(appFeeCap, "App capability not created or owned by ibc fee keeper")
+			suite.Require().True(ok)
+			transferCap, ok := suite.chainA.GetSimApp().ScopedTransferKeeper.GetCapability(ctx, host.ChannelCapabilityPath(transfertypes.FeePortID, "channel-1"))
```
colin-axner, line 203: same concern here

colin-axner (commented):
Looking good! Left a lot of comments. My main concern is that I think any code not under `apps/29-fee` should be made in a separate pr against the main branch since most middleware applications will require those changes. We wouldn't want duplicated code or applications to be blocked on the development of 29-fee. I also think it'd be easier to verify these changes in smaller chunks

modules/apps/29-fee/module.go
seantking: Gotcha, thanks!

modules/apps/29-fee/types/keys.go
```diff
@@ -12,4 +14,12 @@ const (
 
 	// QuerierRoute is the querier route for IBC transfer
 	QuerierRoute = ModuleName
+
+	Version = "fee29-1"
+
+	KeyAppCapability = "app_capabilities"
 )
+
+func AppCapabilityName(portID, channelID string) string {
```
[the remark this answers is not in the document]
AdityaSripal: No it's fine because there are all local names. everything is connected in a series. so multiple middleware in a stack is fine.

A middleware will claim the capability it gets using `channel capability path` and issue a capability with a local name of `app capability path`.

the name of module is implicitly included in the scoped keeper

proto/ibc/applications/transfer/v1/genesis.proto
```diff
@@ -9,7 +9,7 @@ import "gogoproto/gogo.proto";
 
 // GenesisState defines the ibc-transfer genesis state
 message GenesisState {
-  string              port_id      = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
+  repeated string     port_ids     = 1 [(gogoproto.moretags) = "yaml:\"port_ids\""];
```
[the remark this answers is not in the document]
AdityaSripal: why would this change go into main first?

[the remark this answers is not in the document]
colin-axner: Any module which wants to build middleware for transfer needs this right?

modules/apps/29-fee/types/keys.go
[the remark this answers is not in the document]
colin-axner: Will the error messages include this implicit name? If I have multiple middlewares using `app capability path`, will it be hard to debug?

testing/simapp/app.go
```diff
@@ -323,6 +330,12 @@ func NewSimApp(
 	)
 	transferModule := transfer.NewAppModule(app.TransferKeeper)
 
+	app.IBCFeeKeeper = ibcfeekeeper.NewKeeper(appCodec, keys[ibcfeetypes.StoreKey], app.GetSubspace(ibcfeetypes.ModuleName),
```
jackzampolin, line 333: do we need to initialize the subspace here too?

AdityaSripal:
Still need to fix proof issue

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,167 @@
...
+	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
+	"github.com/cosmos/ibc-go/modules/apps/29-fee/keeper"
+	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
+	porttypes "github.com/cosmos/ibc-go/modules/core/05-port/types"
+	ibcexported "github.com/cosmos/ibc-go/modules/core/exported"
+)
+
+// IBCModule implements the ICS26 callbacks for the fee middleware given the fee keeper and the underlying application.
+type IBCModule struct {
```
AdityaSripal, line 15: Separated IBCModule struct from AppModule

AppModule should only be instantiated once and will hook up to msg server, genesis, etc

IBCModule may be instantiated any number of times with different underlying applications

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,167 @@
...
+	order channeltypes.Order,
+	connectionHops []string,
+	portID string,
+	channelID string,
+	chanCap *capabilitytypes.Capability,
+	counterparty channeltypes.Counterparty,
+	version string,
+) error {
+	feeVersion, appVersion := channeltypes.SplitChannelVersion(version)
+	if feeVersion != "" {
```
AdityaSripal, line 40: If FeeVersion is blank, we revert to default transfer behaviour.

i.e. fee middleware is a no-op middleware in this case

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,167 @@
...
+	feeVersion, appVersion := channeltypes.SplitChannelVersion(version)
+	cpFeeVersion, cpAppVersion := channeltypes.SplitChannelVersion(counterpartyVersion)
+
+	if feeVersion != "" || cpFeeVersion != "" {
+		if feeVersion != types.Version {
+			return sdkerrors.Wrapf(types.ErrInvalidVersion, "expected: %s, got: %s", types.Version, feeVersion)
+		}
+		if cpFeeVersion != feeVersion {
+			return sdkerrors.Wrapf(types.ErrInvalidVersion, "expected counterparty version: %s, got: %s", types.Version, cpFeeVersion)
+		}
```
AdityaSripal, line 74: Ensures that if relayer on INIT specified a fee version, then fee version must be specified here.

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,167 @@
...
+
+		im.keeper.SetFeeEnabled(ctx, portID, channelID)
+	} else if im.keeper.IsFeeEnabled(ctx, portID, channelID) {
```
AdityaSripal, line 77: I'm pretty sure crossing hello's are not possible on INIT with auto-generated IDs. So I think I can remove this case. @colin-axner

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -1,16 +1,23 @@
...
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
 	host "github.com/cosmos/ibc-go/modules/core/24-host"
+	ibcexported "github.com/cosmos/ibc-go/modules/core/exported"
+)
+
+// Middleware must implement types.ChannelKeeper and types.PortKeeper expected interfaces
+// so that it can wrap IBC channel and port logic for underlying application.
+var (
+	_ types.ChannelKeeper = Keeper{}
+	_ types.PortKeeper    = Keeper{}
```
AdityaSripal, line 20: middleware must also wrap ChannelKeeper and PortKeeper since middleware is a "2-way" wrapper between IBC and app

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,167 @@
+package fee
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
+	"github.com/cosmos/ibc-go/modules/apps/29-fee/keeper"
```
colin-axner, line 7:
```suggestion
	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
	
	"github.com/cosmos/ibc-go/modules/apps/29-fee/keeper"
```

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -3,12 +3,21 @@ package keeper
 import (
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	capabilitykeeper "github.com/cosmos/cosmos-sdk/x/capability/keeper"
+	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
```
colin-axner, line 9: nit: fix import grouping

modules/core/04-channel/types/version.go
```diff
@@ -0,0 +1,27 @@
+package types
```
colin-axner, line 1: should you merge main to remove this diff?

modules/apps/transfer/module.go
```diff
@@ -213,6 +214,10 @@ func ValidateTransferChannelParams(
 		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid port: %s, expected %s", portID, boundPort)
 	}
 
+	if portID != counterparty.PortId {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "portID: %s does not match counterparty portID: %s", portID, counterparty.PortId)
+	}
```
colin-axner, line 219: Hmm, why is this check needed? Applications like interchain accounts have portIDs which do not match counterparty portIDs

modules/apps/29-fee/types/keys.go
```diff
@@ -18,10 +18,16 @@ const (
 	// QuerierRoute is the querier route for IBC transfer
 	QuerierRoute = ModuleName
 
+	Version = "fee29-1"
+
 	// RelayerAddressKeyPrefix is the key prefix for relayer address mapping
 	RelayerAddressKeyPrefix = "relayerAddress"
 )
 
+func FeeEnabledKey(portID, channelID string) []byte {
```
colin-axner, line 27: godoc

modules/apps/29-fee/types/keys.go
```diff
@@ -18,10 +18,16 @@ const (
...
+	return []byte(fmt.Sprintf("fee_enabled/%s/%s", portID, channelID))
```
colin-axner, line 28: fee_enabled should be in a const so we can do range queries with the prefix

modules/apps/29-fee/module.go
```diff
@@ -48,30 +45,32 @@ func (AppModuleBasic) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {}
...
+	// return gs.Validate()
+	return nil
 }
 
 // RegisterRESTRoutes implements AppModuleBasic interface
 func (AppModuleBasic) RegisterRESTRoutes(clientCtx client.Context, rtr *mux.Router) {
 }
 
-// RegisterGRPCGatewayRoutes registers the gRPC Gateway routes for the ibc-transfer module.
+// RegisterGRPCGatewayRoutes registers the gRPC Gateway routes for the ibc-29-fee module.
```
colin-axner, line 73:
```suggestion
// RegisterGRPCGatewayRoutes registers the gRPC Gateway routes for ics29 fee module.
```
?

modules/apps/29-fee/fee_test.go
```diff
@@ -0,0 +1,40 @@
+package fee_test
+
+import (
+	"testing"
+
+	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
+	transfertypes "github.com/cosmos/ibc-go/modules/apps/transfer/types"
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
+	ibctesting "github.com/cosmos/ibc-go/testing"
+	"github.com/stretchr/testify/suite"
```
colin-axner, line 10: nit: suite import should be in middle group isolated from ibc-go imports

colin-axner (commented):
Looking good! I haven't gone through the bulk of this pr (ibc_module.go, app.go), will do that soon. Here are my code nits for now

modules/apps/transfer/module.go
AdityaSripal: Sure but it's true for transfer. Note this is the transfer module not middleware

modules/core/04-channel/types/version.go
AdityaSripal: I'd rather finalize on what we're doing regarding `v2` before merging in `main`

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,281 @@
...
+				Version:        tc.version,
+			}
+
+			module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), ibctesting.TransferPort)
+			suite.Require().NoError(err)
+
+			cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
+			suite.Require().True(ok)
+
+			err = cbs.OnChanOpenTry(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
```
colin-axner, line 222: I think it is possible to call this function directly via `suite.chainA.GetSimApp().FeeKeeper`? (remove looking up module by port)

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,281 @@
...
+			if tc.expPass {
+				suite.Require().NoError(err, "unexpected error from version: %s", tc.version)
+			} else {
+				suite.Require().Error(err, "error not returned for version: %s", tc.version)
+			}
+		})
+	}
+}
+
+func (suite *FeeTestSuite) TestOnChanOpenTry() {
```
colin-axner, line 94: could you add a little note that this testing is calling open try on chainA

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func (suite *FeeTestSuite) TestOnChanOpenAck() {
+	testCases := []struct {
+		name      string
+		cpVersion string
+		expPass   bool
+	}{
+		{
+			"success",
```
colin-axner, line 241: there should be a test where init is called without a fee version and ack is called with a cp fee version

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1,281 @@
...
+			"fee29-1:ics20-1",
+			true,
+		},
+		{
+			"invalid fee version",
+			"fee29-3:ics20-1",
```
colin-axner, line 247: my preference would be to use something like `fee29-300` or `fee29-A`, so this test doesn't need to be changed if we get to `fee29-3`

colin-axner (changes requested):
Excellent! Very clean design! 

I think there's an edge case on `OnChanOpenAck` which needs to be addressed

modules/core/04-channel/types/version_test.go
```diff
@@ -0,0 +1,43 @@
...
+	testCases := []struct {
+		name       string
+		version    string
+		mwVersion  string
+		appVersion string
+	}{
+		{
+			"single wrapped middleware",
+			"fee29-1:ics20-1",
+			"fee29-1",
```
seantking, line 21: nit: should we use the actual types.Version here?

testing/simapp/app.go
```diff
@@ -315,22 +321,35 @@ func NewSimApp(
...
+		app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
+	)
+
+	// Create Transfer Keeper and pass IBCFeeKeeper as expected Channel and PortKeeper
+	// since fee middleware will wrap the IBCKeeper for underlying application.
 	app.TransferKeeper = ibctransferkeeper.NewKeeper(
 		appCodec, keys[ibctransfertypes.StoreKey], app.GetSubspace(ibctransfertypes.ModuleName),
-		app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
+		app.IBCFeeKeeper, &app.IBCFeeKeeper,
```
seantking, line 332: Question: why do we need to pass the `IBCFeeKeeper` here? In the case that we want to use multiple middleware, what would this look like?

seantking (approved):

modules/apps/29-fee/module_test.go
AdityaSripal: I followed how it was done in the transfer tests.

I think its good to keep anyway because i found routing issues with this flow. This makes sure we can actually grab the right module from the router

modules/apps/29-fee/ibc_module.go
AdityaSripal: Turns out this is irrelevant. We don't have to check here because core IBC checks that in case of crossing hellos,

the channel params in current TRY are exact same as the params in previous INIT: https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/handshake.go#L131

which is exactly the behaviour we want out of core IBC, so we don't have to handle that case in the application.

I fixed the incorrect godoc in this PR.

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,168 @@
...
+
+// OnChanOpenAck implements the IBCModule interface
+func (im IBCModule) OnChanOpenAck(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+	counterpartyVersion string,
+) error {
+	cpAppVersion := counterpartyVersion
+	if im.keeper.IsFeeEnabled(ctx, portID, channelID) {
```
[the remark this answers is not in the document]
AdityaSripal: No because in that case the entire counterparty version `fee29-1:ics20-1` would be passed down to transfer which would error.

Though it's a good thing to test. I added a test case to prove it.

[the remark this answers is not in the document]
AdityaSripal: We don't want future relayers to override the initial decisions of past relayers.

Future relayers should only complete the handshake as initial relayers intended it

modules/apps/29-fee/module_test.go
AdityaSripal: done

testing/simapp/app.go
AdityaSripal: Good question.

We need to ensure that all communication from transfer to IBC goes through the fee keeper first. This ensures that the fee keeper can do its logic first before passing on the information to IBC

So that's why IBCFeeKeeper wraps the channel/port keeper methods in `keeper.go`.

In the case of multiple middleware, we again want to ensure the line of communication from the base application through the lowest middleware up the chain to the highest middleware and then to IBC.

Suppose we had the following stack: `fee -> whitelist -> transfer`.

transfer would be given the `whitelist` keeper as its "wrapped" channel and port keeper.
Meanwhile `whitelist` would be given the `fee` keeper as its "wrapped" channel and port keeper
Fee will get the true channel and port keeper.

This ensures that any flow from transfer to IBC. e.g. a `SendPacket` will first go through `whitelist`, then it will go through `fee`, and only then will it reach IBC. At each step, the middleware can perform its custom logic at the right step of the chain, and there is no way for a middleware/app at one part of the stack to "bypass" the higher levels of middleware and communicate with IBC directly

AdityaSripal: This should be written up in the middleware developer guide

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,176 @@
...
+	counterparty channeltypes.Counterparty,
+	version string,
+) error {
+	mwVersion, appVersion := channeltypes.SplitChannelVersion(version)
+	// Since it is valid for fee version to not be specified, the above middleware version may be for a middleware
+	// lower down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying
+	// application.
+	// If an invalid fee version was passed, we expect the underlying application to fail on its version negotiation.
+	if mwVersion == types.Version {
+		im.keeper.SetFeeEnabled(ctx, portID, channelID)
```
AdityaSripal, line 46: Note since we no longer require that the fee version is specified, it's possible that the middleware version we get is actually for some middleware lower in the stack. So we must pass the full version on in this case.

Note: For a stack `fee -> whitelist -> transfer`

the fee middleware cannot differentiate between cases:

1. `wrongfee29-A:whitelist:ics20-1`
2. `whitelist:ics20-1`

So in both cases, it will assume fee is not enabled and pass the full version to the underlying app.

However in case 1, the underlying app will fail its version negotiation since the wrongfee string is unexpected.

modules/apps/29-fee/ibc_module.go
```diff
@@ -0,0 +1,176 @@
...
+			return sdkerrors.Wrapf(types.ErrInvalidVersion, "fee versions do not match. self version: %s, counterparty version: %s", mwVersion, cpMwVersion)
+		}
+
+		im.keeper.SetFeeEnabled(ctx, portID, channelID)
+	} else {
+		// middleware versions are not the expected version for this midddleware. Pass the full version strings along,
+		// if it not valid version for any other lower middleware, an error will be returned by base application.
+		appVersion = version
+		cpAppVersion = counterpartyVersion
+	}
```
AdityaSripal, line 88: ditto the above logic, but for the TRY case

testing/simapp/app.go
colin-axner: Nice explanation @AdityaSripal 

As a side note, the type being passed in is `types.ChannelKeeper` (from expected_interfaces.go) not `feekeeper.Keeper`

modules/apps/29-fee/types/keys.go
```diff
@@ -18,10 +18,21 @@ const (
 	// QuerierRoute is the querier route for IBC transfer
 	QuerierRoute = ModuleName
 
+	Version = "fee29-1"
+
+	// FeeEnabledPrefix is the key prefix for storing fee enabled flag
+	FeeEnabledKeyPrefix = "fee_enabled"
```
colin-axner, line 24: I think this should be `feeEnabled` to keep with existing key prefix naming

modules/core/04-channel/types/version_test.go
colin-axner: The merge function won't lead to circular testing since no code imports `types_test`

Using the raw strings sounds fine though

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -3,12 +3,23 @@ package keeper
...
+	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
+	host "github.com/cosmos/ibc-go/modules/core/24-host"
+	ibcexported "github.com/cosmos/ibc-go/modules/core/exported"
+
 	"github.com/tendermint/tendermint/libs/log"
 
 	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
```
colin-axner, line 15: could you fix the import grouping?

modules/apps/29-fee/ibc_module.go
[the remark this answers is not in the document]
colin-axner: That makes sense. Can you add a inline comment to the code?

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	for _, tc := range testCases {
+		tc := tc
+
+		suite.Run(tc.name, func() {
+			// reset suite
+			suite.SetupTest()
+			suite.coordinator.SetupClients(suite.path)
+			suite.coordinator.SetupConnections(suite.path)
+
```
colin-axner, line 67:
```suggestion
			suite.coordinator.SetupConnections(suite.path)
```

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -0,0 +1,305 @@
...
+			false,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest()
+			suite.coordinator.SetupClients(suite.path)
+			suite.coordinator.SetupConnections(suite.path)
```
colin-axner, line 283:
```suggestion
			suite.coordinator.SetupConnections(suite.path)
```

colin-axner (approved):
Outstanding work! Big step forward for relayer incentivization :rocket:

testing/simapp/app.go
seantking: Thanks for the explanation. :+1:

Issue #316, opened 2021-08-06
Add counterparty connection id to controller port-id


seantking:
## Summary
[As defined](https://github.com/seantking/ibc/tree/sean/ics-27-updates/spec/app/ics-027-interchain-accounts#registering--controlling-flows) in the specification the controller side port-id is made up of `ics27-1-{connection-number}-{counterparty-connection-number}-{owner-address}`

At present, the `generatePortId` function only builds a portID with the source connection number. 

## Proposal

Update the `generatePortId` [function](https://github.com/cosmos/ibc-go/blob/38ebbff9543c89fd42a330ff5ca4956fdc98656f/modules/apps/27-interchain-accounts/keeper/keeper.go#L132) to build a portID based on both the source + counterparty connection ids.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #297, opened 2021-07-26
GeneratePortId should be in types package


colin-axner:
## Summary

`GeneratePortId` should be stateless and thus in the types package. It should have tests as well

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #319, opened 2021-08-06
Add counterparty port ID to controller portID

colin-axner:
## Description

closes: #316 
closes: #297 

---

write a little note why.


modules/apps/27-interchain-accounts/keeper/account.go
```diff
@@ -17,21 +17,24 @@ import (
 // call 04-channel 'ChanOpenInit'. An error is returned if the port identifier is
 // already in use. Gaining access to interchain accounts whose channels have closed
 // cannot be done with this function. A regular MsgChanOpenInit must be used.
-func (k Keeper) InitInterchainAccount(ctx sdk.Context, connectionId, owner string) error {
-	portId := k.GeneratePortId(owner, connectionId)
+func (k Keeper) InitInterchainAccount(ctx sdk.Context, connectionID, counterpartyConnectionID, owner string) error {
```
colin-axner, line 20: Id -> ID

Historically we always used `ID`. We only use Id on proto defintions because of an issue with gRPC gateway not allowing `ID`. All other variables names in core IBC use `ID`

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -3,18 +3,47 @@ package types
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+
+	connectiontypes "github.com/cosmos/ibc-go/modules/core/03-connection/types"
 )
 
 const (
-	IcaPrefix string = "ics27-1-"
+	ICAPrefix string = "ics-27"
```
colin-axner, line 19: the spec specifies two different formats:
`ics27-1` and `ics-27`

my preference is not to reference any version number in the ports, we should never have a reason to change the port ID generation as backwards compatibility/migration would be a pain, it'd be a different module at that point

Either `ics27` or `ics-27` sounds fine to me

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -3,18 +3,47 @@ package types
...
 
+// GeneratePortID generates the portID for a specific owner
+// on the controller chain in the format:
+//
+// 'ics-27-<connectionSequence>-<counterpartyConnectionSequence>-<owner-address>'
+// https://github.com/seantking/ibc/tree/sean/ics-27-updates/spec/app/ics-027-interchain-accounts#registering--controlling-flows
+// TODO: update link to spec
+// TODO:
+func GeneratePortID(owner, connectionID, counterpartyConnectionID string) (string, error) {
+	ownerID := strings.TrimSpace(owner)
```
colin-axner, line 30: we should consider enforcing limits on ownerID

proto/ibc/applications/interchain_accounts/v1/query.proto
```diff
@@ -18,9 +18,8 @@ message QueryInterchainAccountAddressRequest {
   option (gogoproto.equal)           = false;
   option (gogoproto.goproto_getters) = false;
 
-  // Owner address is the owner of the interchain account on the controller chain
-  string owner_address = 1;
-  string connection_id = 2;
+  // Counterparty PortID is the portID on the controller chain
+  string counterparty_port_id = 1;
```
colin-axner, line 22: This seemed like a better request argument, rather than specifying each individual piece and rebuilding the portID

modules/apps/27-interchain-accounts/types/account.go
seantking: Lets go with `ics-27` :)

proto/ibc/applications/interchain_accounts/v1/query.proto
seantking: My reasoning fore having the previous arguements was that the connection-id represents the chain essentially, so that will be easy to query for. Likewise, it will be easy for whoever is querying to find the owner's address. 

AFAIK we are not able to query for portIDs right now, so it's maybe less ideal? 

I guess, if I'm reasoning this from first principles, who/what is going to be the client that might be using this query? If it is a human using the CLI, maybe it's less likely they'll have the portID on hand. If it's a client-side application using cosmJS or something where would the client get the portID from?

seantking: Also, just thinking about it, we should probably be able to query this on both the controller and host chain. 

Sidenote: Do we even want/need this query at all?

modules/apps/27-interchain-accounts/types/account.go
seantking: I actually just remembered why I kept the `-1` version. I figured in the situation whereby we might have a version `-2` we may still have interchain accounts from version `1` still active and we will want to differentiate. A version two could potentially have 1 owner for N accounts rather than 1 to 1 mapping.

colin-axner: Can't this be represented in the actual interchain accounts version?

proto/ibc/applications/interchain_accounts/v1/query.proto
colin-axner: PortID's will always be known after ChanOpenInit is called. This query is stateless so it isn't necessary. We could just have it be a cli command

modules/apps/27-interchain-accounts/types/account.go
seantking: Let's say we have a bunch of interchain accounts registered with version 1. If we upgrade the architecture to version 2 we have to either 1. keep supporting the version 1 account or 2 provide a migration path for all of the currently existing interchain accounts. I imagined the first choice to be easier and considered using the version in the port-id as a future-proofing step if it's ever required. 

It's a minimal addition that may save time in the future. It probably will never come up, but if it does it could be useful.

colin-axner: But version 1 portID's still need to be supported? You cannot change port ID's (as this is negotiated in the handshake)

seantking: I meant that the functionality associated with an account registered with version 1 of interchain accounts would still need to be supported and by keeping the version number in the prefix here would be a quick way to distinguish between what accounts were registered with what version of the module.

seantking: If a DAO registers an account on the hub with version 1 of interchain accounts, but version 2 (hypothetically) ended up having a different architecture, we would need to either migrate this account (ownership + any tokens) to version a version 2 account or keep supporting version 1.

colin-axner: The channel would still be accessible on both the host and controller side even if versions changed? My main hesitation is it seems odd to overload the port ID with information that could be contained in places it is designated for (app version in the channel version)

seantking: Ah, you're saying if both sides update to version 2 of interchain accounts, it will be possible to get the version string that was originally used to create the channel, and that we can use that to distinguish between accounts registered under different versions?

Getting very edge case here, but what would happen if the channel closed? Would we still be able to access the version string?

colin-axner: It depends on the structure of tracking the active channels. The channel version is always accessible even if the channel is closed. The active channel mapping would just need to know the previous channel which was active

seantking: Alright, let's go with this for now. Before we go live with ICA I think it's a good idea to have some plan in place for how we would handle this kind of issue if it arises.

proto/ibc/applications/interchain_accounts/v1/query.proto
seantking: Let's go with this, no point in blocking work on this for the time being.

seantking (approved):
ACK

Issue #318, opened 2021-08-06
Create GetInterchainAccountAddressFromAck function 


seantking:
## Summary
As defined in the [specification](https://github.com/seantking/ibc/tree/sean/ics-27-updates/spec/app/ics-027-interchain-accounts#contract) we need a way to set the address of the interchain-account on the controller chain upon successful registration on the host chain. 

A function that sends a generic IBC packet to the host chain, and returns the interchain account address on packet sequence 1 needs to be defined. Also, the logic in the OnRecvPacket/OnAckPacket needs to be updated as per definition [here](https://github.com/seantking/ibc/tree/sean/ics-27-updates/spec/app/ics-027-interchain-accounts#packet-relay).

## Proposal

```
// This helper function is required for the controller chain to get the address of a newly registered interchain account on a host chain.
// Because the registration of an interchain account happens during the channel creation handshake, there is no way for the controller chain to know what the address of the interchain account is on the host chain in advance. 
// This function sends an IBC packet to the host chain, on the owner port + active channel with the sole intention of eventually parsing the interchain account address from the Acknowledgement packet on the controller chain side.
// The OnAcknowledgePacket function on the controller chain will handle the parsing + setting the interchain account address in state.
// The controller chain builds the messages (before sending via IBC in the TrySendTx fn) that the host side will eventually execute. Therefore, the interchain account address must be known by the controller chain.
function GetInterchainAccountAddressFromAck(connectionId: string, counterPartyConnectionId: string, ownerAddress: string) returns (nil){
    // Sends a generic IBC packet to the host chain with the intention of parsing the interchain account address associated with this port/connection/channel from the Acknowledgement packet.
}
```
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

seantking:
closing as discussed :https://github.com/cosmos/ibc-go/pull/333#issuecomment-911830049

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #333, opened 2021-08-13
Add interchain account address to packet 1 acknowledgement

colin-axner:
## Description

closes: #318 

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/module.go
+++ b/modules/apps/27-interchain-accounts/module.go
@@ -1,4 +1,4 @@
-package ibc_account
+package interchain_account
 
 import (
 	"encoding/json"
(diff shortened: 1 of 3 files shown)


modules/apps/27-interchain-accounts/module.go
```diff
@@ -243,6 +250,11 @@ func (am AppModule) OnAcknowledgementPacket(
 		return err
 	}
 
+	if packet.Sequence == 1 && ack.Success() {
+		// pass interchain account address into base applications
+		// connected to the middleware
+	}
```
colin-axner, line 256: this information shouldn't be set in state since we could have domain conflicts (this module acts as a host and controller in some cases). It's usefulness is passing this information to helper modules, but even so we will likely end up using the version to convey this information

seantking:
> Sweet, I finally got the codeowner file to work right

Nice!!

seantking (commented):
I'm wondering if we should even go down this path at the moment. 

I think it's not going to be a large amount of work to get the `interchain-account address` set on the controller side via the version string. It's basically just the query endpoint on the `ibc-go` side and then a minor change on the hermes side (you had some ideas for a more complete approach on the hermes side, but a short term solution is to just replace a hardcoded value with a call to the query endpoint on our side). 

I think we could potentially get this done pretty quickly, and it'll give us the nicest UX and feature completeness within `interchain-accounts`.

colin-axner:
> I'm wondering if we should even go down this path at the moment.

Yea I agree. I'd prefer to just move forward with the version solution. Shall we close this pr?

seantking:
> > I'm wondering if we should even go down this path at the moment.
> 
> Yea I agree. I'd prefer to just move forward with the version solution. Shall we close this pr?

Yep, yep. I'll update the issue as well. Thank you for starting this :handshake:

Issue #329, opened 2021-08-11
Support Big Int IBC Transfer Amounts


timlind:
## Summary

Support a larger amount field on FungibleTokenPacketData instead of uint64. This would make it's support in line with what MsgTransfer.Amount supports.

The SDK currently panics when trying to convert a large MsgTransfer.Amount to the underlying FungibleTokenPacketData.Amount.

## Problem Definition

Transferring tokens with 18 decimal places, like many Ethereum tokens for instance, are limited to insignificant amounts (< $1). 

Transferring these tokens by splitting the transfer into multiple smaller transfers would require several hundred or thousands messages.

The fees increase significantly to an uneconomical state when needing to split a transfer to this degree.

Keprl also does not handle this well in the UI, as this many messages shouldn't really be a use case. 

Some relayers also don't handle these large transactions well.

## Proposal

Introduce a larger amount field (i.e string) on FungibleTokenPacketData proto message, to replace the current amount field.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
I've opened an issue on [cosmos/ibc](https://github.com/cosmos/ibc/issues/595). We won't change the packet structure until a change is approved on the specifications

The packet structure will not adopt any SDK structures. This is intentional as packets need to be sent/received on non-SDK chains. They may decide to introduce their own amount type which may or may not be the same as sdk.Int. 

In this repo, we can at least add a check that the MsgTransfer amount can be converted to uint64 and return an error instead of panici'ing

albertchon:
Why not represent the amount as a string?

fedekunze:
This is relevant for Ethermint as well. Happy to provide resources or reviews to get this fixed ASAP @colin-axner

jtremback:
> The packet structure will not adopt any SDK structures. This is intentional as packets need to be sent/received on non-SDK chains. They may decide to introduce their own amount type which may or may not be the same as sdk.Int.

This is kind of funny since the only blockchain I've ever seen that uses u64 for numbers is the Cosmos hub. So u64 amounts are already an SDK structure

mccallofthewild:
@colin-axner The IBC spec defines that `FungibleTokenPacketData.amount` should be `uint256`. https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer#data-structures

colin-axner:
[...]

PR #350, opened 2021-08-26
Change ICS 20 packet data amount to be string

colin-axner:
## Description

closes: #329 

---

write a little note why.


modules/apps/transfer/keeper/relay.go
```diff
@@ -244,7 +250,7 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 		defer func() {
 			telemetry.SetGaugeWithLabels(
 				[]string{"ibc", types.ModuleName, "packet", "receive"},
-				float32(data.Amount),
+				float32(transferAmount.Int64()),
```
AdityaSripal, line 253: why is this a float32 of an int64?

colin-axner: well I think float32 is the argument type, not sure why it is int64 (I used the existing code format)

It will be updated to be big int

colin-axner: Did some more digging. float32 is the argument type for metrics, so not getting around that. The SDK handles this by checking if the amount is a Int64 before setting the value. I'm not sure why Int64 is used and Uint64 isn't, since I believe float32 goes up to somewhere around ~2^127

For consistency, I'm sticking with Int64 unless we think we should increase it to uint64?

seantking (approved):

damiannolan (approved):

colin-axner:
Mostly notes to self found while testing this feature:
Stargate versions:
- sending > int64 results in panic
- recv > int64 results in failed tx (requires timeout)

ibc-go v1.1
- sending > int64 results in panic
- recv > int64 results in failed ack being written

modules/apps/transfer/keeper/relay.go
```diff
@@ -163,11 +163,13 @@ func (k Keeper) SendTransfer(
...
-			float32(token.Amount.Int64()),
-			[]metrics.Label{telemetry.NewLabel(coretypes.LabelDenom, fullDenomPath)},
-		)
+		if token.Amount.IsInt64() {
+			telemetry.SetGaugeWithLabels(
+				[]string{"tx", "msg", "ibc", "transfer"},
+				float32(token.Amount.Int64()),
+				[]metrics.Label{telemetry.NewLabel(coretypes.LabelDenom, fullDenomPath)},
+			)
+		}
```
AdityaSripal, line 172: what happens if we're above int64? is it possible to just emit the string or a big Int?

modules/apps/transfer/types/packet_test.go
```diff
@@ -7,8 +7,10 @@ import (
 )
 
 const (
-	denom  = "transfer/gaiachannel/atom"
-	amount = uint64(100)
+	denom              = "transfer/gaiachannel/atom"
+	amount             = "100"
+	largeAmount        = "18446744073709551616"                                                           // one greater than largest uint64 (^uint64(0))
+	invalidLargeAmount = "115792089237316195423570985008687907853269984665640564039457584007913129639936" // 2^256
```
AdityaSripal, line 13: is there a reason we've stopped the limit at `2^256`?

AdityaSripal (approved):
Looks great! I just think the telemetry logic is weird and i don't understand it

modules/apps/transfer/types/packet_test.go
colin-axner: That's what `sdk.Int` supports

modules/apps/transfer/keeper/relay.go
colin-axner: The amount isn't emitted in telemetry. The telemetry API requires a float32 which means nothing > int64 is supported and we cannot emit string or big.Int. This logic was taken from the SDK. The only other alternative is to remove the telemetry entirely

adizere:
> This should be built and tested against the hermes and ts-relayer before merging. We need to ensure it works with existing binaries and that it is capable of sending uint256 amounts to other binaries with this change included

We're now upgrading Hermes & testing with some of the corner-cases released in ibc-go v2, including the one in this PR.

We actually _do not_ use [`FungibleTokenPacketData`](https://github.com/cosmos/ibc-go/blob/f7051429e1cf833a6f65d51e6c3df1609290a549/proto/ibc/applications/transfer/v2/packet.proto#L10) in Hermes at all. To create a ICS20 transfer, Hermes encodes all of the transfer parameters into a [`MsgTransfer`](https://github.com/cosmos/ibc-go/blob/f7051429e1cf833a6f65d51e6c3df1609290a549/proto/ibc/applications/transfer/v1/tx.proto#L20). It's possible that the ICS20 transfer implementation in Hermes is incorrect if it is supposed to encode transfer parameters into a `FungibleTokenPacketData`.

In any case, the [spec](https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer#data-structures) is not very clear in the difference between `FungibleTokenPacketData` vs.  `MsgTransfer`. The proto files do not help, as there is a `MsgTransfer` in [v1](https://github.com/cosmos/ibc-go/blob/f7051429e1cf833a6f65d51e6c3df1609290a549/proto/ibc/applications/transfer/v1/tx.proto#L19) which points to ICS20 spec; this data structure no longer appears in v2 but now [`FungibleTokenPacketData`](https://github.com/cosmos/ibc-go/blob/dc416084694b6ed1d45baf50cc408b747daa90aa/proto/ibc/applications/transfer/v2/packet.proto#L9) points to ICS20 spec. Any help would be appreciated!

colin-axner:
`Msg`s are SDK types. Typically submission of a SDK Msg causes a packet to be constructed and sent via IBC

Relayer submits `MsgTransfer`. Processing of `MsgTransfer` causes the creation of `FungibleTokenPacketData`. Internally this was a uint64 but is now a string (up to uint256 as we use SDK big.Int for MsgTransfer which submits the amount). 

The proto definitions are split into v1 and v2 because of migration issues with proto Any's. The proto definition for `FungibleTokenPacketData` needed to change causing v1 -> v2, but if we tried to move `MsgTransfer` from v1 -> v2, the types would be considered unequivalent despite containing the same definition. Since the sdk.Msg is an interface, all SDK Msgs are packed/unpacked to/from Any's. This relies on the proto URL which in turn uses the proto definition versioning. Thus, if we moved `MsgTransfer` to v2, this would force all external clients (including relayers) to know whether to submit a v1 or v2 `MsgTransfer` based on the chains version. This is obviously something we wanted to avoid, so we opted to leave all types which didn't change into `v1`. 

I don't suspect there to be any issues with Hermes supporting this change

adizere:
Appreciate the thorough answer Colin! I'd like to follow-up with one last detail that is confusing to me --

> Processing of `MsgTransfer` causes the creation of `FungibleTokenPacketData`.

Since `FungibleTokenPacketData` doesn't seem to part of the interface (i.e., external facing), I'm curious why is it captured in the .proto files?

colin-axner:
We proto JSON marshal `FungibleTokenPacketData` so we need it to be a proto definition. It also needs to be unmarshaled by counterparty chains. It is possible external clients could want to interact with `FungibleTokenPacketData`, such as to filter based on the sender or receiver of an ICS20 packet, but in general I don't expect external clients to use this type

PR #357, opened 2021-08-27
feat: ics 29 packet callbacks

charleenfei:
## Description

closes: #262

---

write a little note why.

### DIFF
--- a/modules/apps/29-fee/fee_test.go
+++ b/modules/apps/29-fee/fee_test.go
@@ -6,6 +6,7 @@ import (
 	"github.com/stretchr/testify/suite"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
 	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
 	transfertypes "github.com/cosmos/ibc-go/modules/apps/transfer/types"
 	clienttypes "github.com/cosmos/ibc-go/modules/core/02-client/types"
--- a/modules/apps/29-fee/ibc_module.go
+++ b/modules/apps/29-fee/ibc_module.go
@@ -230,7 +230,7 @@ func (im IBCModule) OnAcknowledgementPacket(
 	if err == nil {
 		// write the cache and then call underlying callback
 		writeFn()
-		// emit the event
+		// NOTE: The context returned by CacheContext() refers to a new EventManager, so it needs to explicitly set events to the original context.
 		ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
 	}
 	// otherwise discard cache and call underlying callback
@@ -266,7 +266,7 @@ func (im IBCModule) OnTimeoutPacket(
 	if err == nil {
 		// write the cache and then call underlying callback
 		writeFn()
-		// emit the event
+		// NOTE: The context returned by CacheContext() refers to a new EventManager, so it needs to explicitly set events to the original context.
 		ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
 	}
 


modules/apps/29-fee/fee_test.go
```diff
@@ -36,6 +36,14 @@ func (suite *FeeTestSuite) SetupTest() {
 	suite.path = path
 }
 
+func NewTransferPath(chainA, chainB *ibctesting.TestChain) *ibctesting.Path {
```
seantking, line 39: Where are we using this?

modules/apps/29-fee/module_test.go
```diff
@@ -0,0 +1 @@
+package fee_test
```
seantking, line 1: Is this necessary?

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -17,6 +17,16 @@ var _ types.MsgServer = Keeper{}
 func (k Keeper) RegisterCounterpartyAddress(goCtx context.Context, msg *types.MsgRegisterCounterpartyAddress) (*types.MsgRegisterCounterpartyAddressResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	_, err := sdk.AccAddressFromBech32(msg.Address)
```
seantking, line 20: These checks are unnecessary as we should do this in the validation

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -54,16 +56,16 @@ func (k Keeper) DistributeFee(ctx sdk.Context, refundAcc, forwardRelayer, revers
 	// get module accAddr
 	feeModuleAccAddr := k.authKeeper.GetModuleAddress(types.ModuleName)
 
-	// send receive fee to forward relayer
-	err := k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, forwardRelayer, feeInEscrow.Fee.ReceiveFee)
+	// send receive fee to dest relayer
```
seantking, line 59: I think we discussed this before but I've forgotten. Is this the terminology we're aligning on? Just wondering :) 

@AdityaSripal

modules/apps/29-fee/ibc_module.go
```diff
@@ -9,10 +9,10 @@ import (
 	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
 	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
 	porttypes "github.com/cosmos/ibc-go/modules/core/05-port/types"
-	ibcexported "github.com/cosmos/ibc-go/modules/core/exported"
+	"github.com/cosmos/ibc-go/modules/core/exported"
 )
 
-// IBCModule implements the ICS26 callbacks for the fee middleware given the fee keeper and the underlying application.
+// IncentivizedIBCModule implements the ICS26 callbacks for the fee middleware given the fee keeper and the underlying application.
```
seantking, line 15: Why change the name in the comment?

modules/apps/29-fee/module_test.go
charleenfei: whoops, no it was a left over file

modules/apps/29-fee/ibc_module.go
```diff
@@ -187,8 +195,23 @@ func (im IBCModule) OnAcknowledgementPacket(
...
-	// TODO: Implement fee specific logic if fee is enabled for the given channel
-	return im.app.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	ack := &types.IncentivizedAcknowledgement{}
+	if im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
+		if err := types.ModuleCdc.UnmarshalJSON(acknowledgement, ack); err != nil {
+			return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal ICS-29 incentivized packet acknowledgement: %v", err)
+		}
+
+		packetId := channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)
+		identifiedPacketFee, exists := im.keeper.GetFeeInEscrow(ctx, packetId)
```
seantking, line 205:
```suggestion
		identifiedPacketFee, found := im.keeper.GetFeeInEscrow(ctx, packetId)
```

modules/apps/29-fee/ibc_module.go
```diff
@@ -187,8 +195,23 @@ func (im IBCModule) OnAcknowledgementPacket(
...
+		if !exists {
```
seantking, line 206:
```suggestion
		if !found {
```

modules/apps/29-fee/fee_test.go
charleenfei: leftover from the integration test attempt

modules/apps/29-fee/ibc_module.go
```diff
@@ -175,9 +175,17 @@ func (im IBCModule) OnRecvPacket(
 	ctx sdk.Context,
 	packet channeltypes.Packet,
 	relayer sdk.AccAddress,
-) ibcexported.Acknowledgement {
-	// TODO: Implement fee specific logic if fee is enabled for the given channel
-	return im.app.OnRecvPacket(ctx, packet, relayer)
+) exported.Acknowledgement {
+	ack := im.app.OnRecvPacket(ctx, packet, relayer)
+	sourceRelayer, found := im.keeper.GetCounterpartyAddress(ctx, relayer.String())
```
colin-axner, line 180: This needs IsFeeEnabled check. If the fee is not enabled for this channel, the acknowledgement should not be wrapped

modules/apps/29-fee/ibc_module.go
```diff
@@ -197,6 +220,15 @@ func (im IBCModule) OnTimeoutPacket(
 	packet channeltypes.Packet,
 	relayer sdk.AccAddress,
 ) error {
-	// TODO: Implement fee specific logic if fee is enabled for the given channel
+	packetId := channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)
+	identifiedPacketFee, exists := im.keeper.GetFeeInEscrow(ctx, packetId)
+	if !exists {
+		return sdkerrors.Wrapf(types.ErrFeeNotFound, "fee not found for packet id %s", packetId)
+	}
```
colin-axner, line 227: first check must be `IsFeeEnabled`, we don't want non fee enabled channels erroring on logic that doesn't apply to it

modules/apps/29-fee/keeper/escrow.go
colin-axner: forward/reverse relayer is my preference. I find source/destination can become confusing

modules/apps/29-fee/types/ack.go
```diff
@@ -0,0 +1,21 @@
...
+// considered successful if the forward relayer address is empty. Otherwise it is
+// considered a failed acknowledgement.
+func (ack IncentivizedAcknowledgement) Success() bool {
+	return ack.ForwardRelayerAddress != ""
+}
+
+// Acknowledgement implements the Acknowledgement interface. It returns the
+// acknowledgement serialised using JSON.
+func (ack IncentivizedAcknowledgement) Acknowledgement() []byte {
+	var SubModuleCdc = codec.NewProtoCodec(codectypes.NewInterfaceRegistry())
```
colin-axner, line 19: nit: use the `ModuleCdc` in codec.go (can rename to `SubModuleCdc` if desired)

modules/apps/transfer/types/expected_keepers.go
```diff
@@ -6,6 +6,7 @@ import (
 	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
 	connectiontypes "github.com/cosmos/ibc-go/modules/core/03-connection/types"
 	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
+	"github.com/cosmos/ibc-go/modules/core/exported"
 	ibcexported "github.com/cosmos/ibc-go/modules/core/exported"
```
colin-axner, line 10: remove the import with the `ibcexported` alias

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +470,301 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest()
+			suite.coordinator.SetupConnections(suite.path)
+
+			// open incentivized channel
+			suite.path.EndpointA.ChanOpenInit()
+			suite.path.EndpointB.ChanOpenTry()
+			suite.path.EndpointA.ChanOpenAck()
+			suite.path.EndpointB.ChanOpenConfirm()
```
colin-axner, line 504: I think this can all be replaced with `suite.coordinator.Setup(suite.path)`

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +470,301 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+			if tc.sourceRelayer {
+				suite.Require().NoError(err, "unexpected error for case: %s", tc.name)
+				ack := types.IncentivizedAcknowledgement{
+					Result:                channeltypes.NewResultAcknowledgement([]byte{1}).Acknowledgement(),
+					ForwardRelayerAddress: suite.chainB.SenderAccount.GetAddress().String(),
+				}
+				suite.Require().Equal(result, ack)
+			} else {
+				suite.Require().NoError(err, "unexpected error for case: %s", tc.name)
+				ack := types.IncentivizedAcknowledgement{
```
colin-axner, line 562: if the fee is not enabled, an IncentivizedAcknowledgement should not be returned (this would break backwards compatibility). In the above test case the returned ack should be from ics20

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +470,301 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+					Result:                channeltypes.NewResultAcknowledgement([]byte{1}).Acknowledgement(),
+					ForwardRelayerAddress: "",
+				}
+				suite.Require().Equal(result, ack)
+			}
+		})
+	}
+}
+
+func (suite *FeeTestSuite) TestOnAcknowledgementPacket() {
```
colin-axner, line 572: this should test non fee enabled channels still work as expected

colin-axner (changes requested):
Great work!! My only request is that the `IsFeeEnabled` check is always done first. Non fee enabled channels must not execute any ics29 logic. They should no-op in this IBC module

modules/apps/29-fee/keeper/escrow.go
seantking: yeah agreed

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +470,301 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+func (suite *FeeTestSuite) TestOnRecvPacket() {
+	testCases := []struct {
+		name          string
+		malleate      func(suite *FeeTestSuite)
+		sourceRelayer bool
+	}{
+		{
+			"success",
+			func(suite *FeeTestSuite) {
+				suite.chainA.GetSimApp().IBCFeeKeeper.SetCounterpartyAddress(suite.chainA.GetContext(), suite.chainA.SenderAccount.GetAddress().String(), suite.chainB.SenderAccount.GetAddress().String())
```
seantking, line 483: I'm not actually sure what the idiomatic way is (maybe @colin-axner @damiannolan disagrees with me here) but I'd prefer to have the success case do nothing and put this `SetCounterpartyAddress` below instead of calling here. Instead, having the malleate change the relayer address to an empty string explicitly for the fail case would be easier to read IMO.

Just a nit, maybe this is totally fine :)

seantking: Actually, I think it makes more sense the way you have it after looking at it closer.

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +470,301 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+				)
+				err := suite.chainA.GetSimApp().IBCFeeKeeper.EscrowPacketFee(suite.chainA.GetContext(), identifiedFee)
+				suite.Require().NoError(err)
+			},
+			true,
+		},
+		{
+			"identified packet fee doesn't exist",
+			func(suite *FeeTestSuite) {
+				suite.Require().Error(types.ErrFeeNotFound)
```
seantking, line 600: This isn't actually testing the `OnAcknowledgementPacket` call, it's just testing your setup functions.

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +470,301 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+
+			// set up module and callbacks
+			module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), ibctesting.TransferPort)
+			suite.Require().NoError(err)
+
+			cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
+			suite.Require().True(ok)
+
+			// malleate test case
+			tc.malleate(suite)
```
seantking, line 664: Should we be passing suite into malleate here? This seems unnecessary @colin-axner

colin-axner: yea it is unnecessary

seantking:
I think the transfer tests are failing in CI. Are they failing locally for you @charleenfei

seantking:
I left some initial comments. It's coming together! I think some of the tests need to be reworked with regards to the use of the malleate & for some reason the transfer tests and some others are failing but it's looking good otherwise. 

Gonna review again in the morning :+1:

modules/apps/29-fee/ibc_module_test.go
seantking: Ahhh, I think see what you're doing here. The setup in `suite.run` fails when you don't explicitly escrow the packet etc in each test case. I think it would be easier to read (and also reduce code)  if you called `EscrowPacketFee` after calling malleate and used each case to malleate/change the data to force the test to fail :)

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +470,301 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+					types.Fee{
+						ReceiveFee: validCoins,
+						AckFee:     validCoins2,
+						TimeoutFee: validCoins3,
+					},
+					suite.chainB.SenderAccount.GetAddress().String(),
+					[]string{},
+				)
+				suite.chainA.GetSimApp().IBCFeeKeeper.EscrowPacketFee(suite.chainA.GetContext(), identifiedFee)
+				suite.Require().Error(types.ErrRefundAccNotFound)
```
seantking, line 619: I think what you have here is the equivalent to not calling `EscrowPacketFee` at all. By passing in a refund account that doesn't exist `EscrowPacketFee` will error and do nothing. Maybe I'm missing something? :)

modules/apps/29-fee/ibc_module_test.go
charleenfei: this test is just for if the sourceRelayer is an empty string, but I will add one for if the fee is not enabled

modules/apps/29-fee/ibc_module_test.go
charleenfei: true, i'll update the test to include a wrong ack format

modules/apps/29-fee/ibc_module_test.go
charleenfei: it's not really that the setup fails but that the OnAck callback checks if an identified fee exists, so if it hasn't been escrowed then that error should surface

seantking: Yeah, I see. I feel like it's a bit hard to read using the malleate functions like this (I think the general approach is you change (malleate) data here to force an error rather than have the default case fail) but I think it's ok for now :)

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -57,13 +59,13 @@ func (k Keeper) DistributeFee(ctx sdk.Context, refundAcc, forwardRelayer, revers
...
 	if err != nil {
-		return sdkerrors.Wrap(err, "failed to send fee to forward relayer")
+		return sdkerrors.Wrap(err, "failed to send fee to destination relayer")
 	}
 
 	// send ack fee to reverse relayer
 	err = k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, reverseRelayer, feeInEscrow.Fee.AckFee)
 	if err != nil {
-		return sdkerrors.Wrap(err, "error sending fee to reverse relayer")
+		return sdkerrors.Wrap(err, "error sending fee to source relayer")
```
colin-axner, line 68: these should be changed back to forward and reverse?

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -83,7 +85,7 @@ func (k Keeper) DistributeFeeTimeout(ctx sdk.Context, refundAcc, timeoutRelayer
 	// check if there is a Fee in escrow for the given packetId
 	feeInEscrow, found := k.GetFeeInEscrow(ctx, packetID)
 	if !found {
-		return sdkerrors.Wrap(types.ErrFeeNotFound, refundAcc.String())
+		return sdkerrors.Wrapf(types.ErrFeeNotFound, "for packetID %s", packetID.String())
```
colin-axner, line 88:
```suggestion
		return sdkerrors.Wrapf(types.ErrFeeNotFound, "for packetID %s", packetID)
```
.String should automatically be called when formatting to string

modules/apps/29-fee/ibc_module.go
```diff
@@ -175,8 +175,19 @@ func (im IBCModule) OnRecvPacket(
 	ctx sdk.Context,
 	packet channeltypes.Packet,
 	relayer sdk.AccAddress,
-) ibcexported.Acknowledgement {
-	// TODO: Implement fee specific logic if fee is enabled for the given channel
+) exported.Acknowledgement {
+	if im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
```
colin-axner, line 179: I think this should be destination port and channel since this is on the receiving side. Could we ensure the tests using different channel id's so that such an bug would be caught?

modules/apps/29-fee/ibc_module.go
```diff
@@ -175,8 +175,19 @@ func (im IBCModule) OnRecvPacket(
...
+		if !found {
+			forwardRelayer = ""
+		}
+
+		return types.IncentivizedAcknowledgement{
+			Result:                ack.Acknowledgement(),
+			ForwardRelayerAddress: forwardRelayer,
+		}
+	}
 	return im.app.OnRecvPacket(ctx, packet, relayer)
```
colin-axner, line 191:
```suggestion
	}
	
	return im.app.OnRecvPacket(ctx, packet, relayer)
```

super small nit/suggestion. Another idea is to move ics29 logic into `keeper/`. Or adjust the code such that:
```go
    if !im.keeper.IsFeeEnabled() {
        return im.app.OnRecvPacket
    }
    
    // do ics29 logic
``` 

Mostly just looking to make sure the no-op is very clear when the fee is not enabled, how this is done is up to you. I don't think there's a "best" way

modules/apps/29-fee/ibc_module.go
```diff
@@ -197,6 +224,15 @@ func (im IBCModule) OnTimeoutPacket(
...
-	// TODO: Implement fee specific logic if fee is enabled for the given channel
+	if im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
+		packetId := channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)
+		identifiedPacketFee, exists := im.keeper.GetFeeInEscrow(ctx, packetId)
+		if !exists {
+			return sdkerrors.Wrapf(types.ErrFeeNotFound, "fee not found for packet id %s", packetId)
+		}
+		if err := im.keeper.DistributeFeeTimeout(ctx, sdk.AccAddress(identifiedPacketFee.RefundAddress), relayer, channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)); err != nil {
+			return err
+		}
```
colin-axner, line 235:
```suggestion
		packetId := channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)
		
		identifiedPacketFee, exists := im.keeper.GetFeeInEscrow(ctx, packetId)
		if !exists {
			return sdkerrors.Wrapf(types.ErrFeeNotFound, "fee not found for packet id %s", packetId)
		}
		
		if err := im.keeper.DistributeFeeTimeout(ctx, sdk.AccAddress(identifiedPacketFee.RefundAddress), relayer, channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)); err != nil {
			return err
		}
```

logical spacing helps with readability

modules/apps/29-fee/ibc_module_test.go
colin-axner: I agree that success case should do nothing. This makes it easier to verify the failing tests cases (otherwise it can be easy to forget to do the required change)

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,331 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+			cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
+			suite.Require().True(ok)
+
+			// malleate test case
+			tc.malleate()
+
+			result := cbs.OnRecvPacket(suite.chainA.GetContext(), packet, suite.chainA.SenderAccount.GetAddress())
+
+			if tc.sourceRelayer && tc.feeEnabled {
+				suite.Require().NoError(err, "unexpected error for case: %s", tc.name)
```
colin-axner, line 561: This error is coming from `LookupModuleByPort`. I think this line can be removed

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,331 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
 		})
 	}
 }
+
+func (suite *FeeTestSuite) TestOnRecvPacket() {
+	testCases := []struct {
+		name          string
+		malleate      func()
+		sourceRelayer bool
```
colin-axner, line 477: rename to forwardRelayer and add a comment what the boolean indicates

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,331 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+
+			cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
+			suite.Require().True(ok)
+
+			// malleate test case
+			tc.malleate()
+
+			result := cbs.OnRecvPacket(suite.chainA.GetContext(), packet, suite.chainA.SenderAccount.GetAddress())
+
+			if tc.sourceRelayer && tc.feeEnabled {
```
colin-axner, line 560: nit: turns ifs into switch statement

modules/apps/29-fee/ibc_module.go
```diff
@@ -175,8 +175,19 @@ func (im IBCModule) OnRecvPacket(
 	ctx sdk.Context,
 	packet channeltypes.Packet,
 	relayer sdk.AccAddress,
-) ibcexported.Acknowledgement {
-	// TODO: Implement fee specific logic if fee is enabled for the given channel
+) exported.Acknowledgement {
+	if im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
+		ack := im.app.OnRecvPacket(ctx, packet, relayer)
+		forwardRelayer, found := im.keeper.GetCounterpartyAddress(ctx, relayer.String())
```
colin-axner, line 181:
```suggestion
		ack := im.app.OnRecvPacket(ctx, packet, relayer)
		
		forwardRelayer, found := im.keeper.GetCounterpartyAddress(ctx, relayer.String())
```

modules/apps/29-fee/ibc_module_test.go
colin-axner: I don't understand the purpose of `suite.Require().Error(types.ErrFeeNotFound)`. It is just asserting that `types.ErrFeeNotFound` is of type error, it isn't checking any logic related to ibc_module.go

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,331 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+				suite.Require().Error(types.ErrFeeNotFound)
+			},
+			types.IncentivizedAcknowledgement{
+				Result:                channeltypes.NewResultAcknowledgement([]byte{1}).Acknowledgement(),
+				ForwardRelayerAddress: suite.chainB.SenderAccount.GetAddress().String(),
+			}.Acknowledgement(),
+			false,
+		},
+		{
+			"ack wrong format",
```
colin-axner, line 646: this appears not to be testing the lines as expected. I see missing code coverage for unmarshling error of unmarshalling into incentivized ack. Maybe try making the ack be some random bytes like []byte("not a valid acknowledgement")

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,331 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+				suite.Require().NoError(err, "unexpected error for case: %s", tc.name)
+				ack := channeltypes.NewResultAcknowledgement([]byte{1})
+				suite.Require().Equal(result, ack)
+			}
+		})
+	}
+}
+
+func (suite *FeeTestSuite) TestOnAcknowledgementPacket() {
+	testCases := []struct {
```
colin-axner, line 586: add test for `DistributeFee` failing. One idea is to have the relayer address the fees are being distributed to be a blocked address

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,331 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+			if tc.expPass {
+				suite.Require().NoError(err, "unexpected error for case: %s", tc.name)
+			} else {
+				suite.Require().Error(err, "%s expected error but returned none", tc.name)
+			}
+		})
+	}
+}
+
+func (suite *FeeTestSuite) TestOnTimeoutPacket() {
```
colin-axner, line 715: add test for error on distribute fee

colin-axner (commented):
Nice work! Thanks for following up on my previous requests. I did a closer review this time. Left a lot of suggestions/nits for code improvement

modules/apps/29-fee/ibc_module.go
```diff
@@ -187,7 +198,23 @@ func (im IBCModule) OnAcknowledgementPacket(
...
+	if im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
+		ack := &types.IncentivizedAcknowledgement{}
+		if err := types.ModuleCdc.UnmarshalJSON(acknowledgement, ack); err != nil {
+			return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal ICS-29 incentivized packet acknowledgement: %v", err)
+		}
+
+		packetId := channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)
+		identifiedPacketFee, found := im.keeper.GetFeeInEscrow(ctx, packetId)
+		if !found {
+			return sdkerrors.Wrapf(types.ErrFeeNotFound, "fee not found for packet id %s", packetId)
```
AdityaSripal, line 210: Don't return error here, since it's possible to send a packet over a fee-enabled channel without escrowing fees.

Just move on to the underlying app

modules/apps/29-fee/ibc_module.go
```diff
@@ -187,7 +198,23 @@ func (im IBCModule) OnAcknowledgementPacket(
...
+		}
+
+		if err := im.keeper.DistributeFee(ctx, sdk.AccAddress(identifiedPacketFee.RefundAddress), sdk.AccAddress(ack.ForwardRelayerAddress), relayer, packetId); err != nil {
+			return err
```
AdityaSripal, line 214: ditto here. We don't want to fail on distribute fee if forward relayer address is invalid

modules/apps/29-fee/ibc_module.go
```diff
@@ -187,7 +198,23 @@ func (im IBCModule) OnAcknowledgementPacket(
...
+			return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal ICS-29 incentivized packet acknowledgement: %v", err)
+		}
+
+		packetId := channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)
+		identifiedPacketFee, found := im.keeper.GetFeeInEscrow(ctx, packetId)
+		if !found {
+			return sdkerrors.Wrapf(types.ErrFeeNotFound, "fee not found for packet id %s", packetId)
+		}
+
+		if err := im.keeper.DistributeFee(ctx, sdk.AccAddress(identifiedPacketFee.RefundAddress), sdk.AccAddress(ack.ForwardRelayerAddress), relayer, packetId); err != nil {
```
AdityaSripal, line 213: we should cache context and then try this code block. If an error occurs, discard the cache  and call the underlying app.

Otherwise, write the cache and call underlying app

modules/apps/29-fee/ibc_module.go
AdityaSripal, line 235: ditto logic

AdityaSripal (changes requested):
Must ensure that ics29 does not affect underlying app

modules/apps/29-fee/ibc_module.go
charleenfei: good catch

modules/apps/29-fee/ibc_module.go
```diff
@@ -171,32 +171,100 @@ func (im IBCModule) OnChanCloseConfirm(
...
+	cacheCtx, writeFn := ctx.CacheContext()
+
+	err := im.keeper.DistributeFeeTimeout(cacheCtx, sdk.AccAddress(identifiedPacketFee.RefundAddress), relayer, channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence))
+	// emit the error in event if any
+	ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
+
+	// if there is no error, write the cache and then call underlying callback
+	if err == nil {
+		writeFn()
+		return im.app.OnTimeoutPacket(ctx, packet, relayer)
```
AdityaSripal, line 266: ditto

AdityaSripal (approved):
Great work!! Just have small nits to make code cleaner.

Didn't read through test files in detail

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -39,6 +39,8 @@ func (k Keeper) EscrowPacketFee(ctx sdk.Context, identifiedFee *types.Identified
 
 	// Store fee in state for reference later
 	// feeInEscrow/<port-id>/<channel-id>/packet/<sequence-id>/ -> Fee (timeout, ack, onrecv)
+
+	// TODO: set refund acc to refer to later
```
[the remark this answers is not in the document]
charleenfei: i can take it out

modules/apps/29-fee/ibc_module.go
```diff
@@ -171,32 +171,100 @@ func (im IBCModule) OnChanCloseConfirm(
...
+	// return underlying callback if no fee found for given packetID
+	if !found {
+		return im.app.OnAcknowledgementPacket(ctx, packet, ack.Result, relayer)
+	}
+	// cache context before trying to distribute the fee
+	cacheCtx, writeFn := ctx.CacheContext()
+
+	err := im.keeper.DistributeFee(cacheCtx, sdk.AccAddress(identifiedPacketFee.RefundAddress), sdk.AccAddress(ack.ForwardRelayerAddress), relayer, packetId)
+	// emit the error in event if any
+	ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
```
damiannolan, line 226: Should these cached events only be emitted if the err is nil and `writeFn()` is being called?
Wouldn't it be misleading to have cached events from bank `SendCoins` emitted if there's an error and `writeFn()` is skipped?

charleenfei: I see your point. But I also think it's important to bubble the errors out somewhere, otherwise it'll be v difficult to debug.

colin-axner: The events from `DistributeFee` should only be emitted if the err is nil. Otherwise a new event should be emitted with the error message from `DistributeFee`

Events are not typically used for error debugging. The error string should provide most of the information used for debugging

By default, I wouldn't expect the error to be emitted in the event (this is an exceptional situation)

modules/apps/29-fee/ibc_module.go
```diff
@@ -171,32 +171,100 @@ func (im IBCModule) OnChanCloseConfirm(
...
+	}
+
+	packetId := channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)
+	identifiedPacketFee, found := im.keeper.GetFeeInEscrow(ctx, packetId)
+
+	// return underlying callback if no fee found for given packetID
+	if !found {
+		return im.app.OnAcknowledgementPacket(ctx, packet, ack.Result, relayer)
+	}
+	// cache context before trying to distribute the fee
```
colin-axner, line 221:
```suggestion
	identifiedPacketFee, found := im.keeper.GetFeeInEscrow(ctx, packetId)
	if !found {
	        // return underlying callback if no fee found for given packetID
		return im.app.OnAcknowledgementPacket(ctx, packet, ack.Result, relayer)
	}
	
	// cache context before trying to distribute the fee
```

just a suggestion, might not want to commit via github UI since I don't think this is gofmt'd

modules/apps/29-fee/ibc_module.go
```diff
@@ -171,32 +171,100 @@ func (im IBCModule) OnChanCloseConfirm(
...
+	}
+
+	packetId := channeltypes.NewPacketId(packet.SourceChannel, packet.SourcePort, packet.Sequence)
+
+	// return underlying callback if fee not found for given packetID
+	identifiedPacketFee, found := im.keeper.GetFeeInEscrow(ctx, packetId)
+	if !found {
+		return im.app.OnTimeoutPacket(ctx, packet, relayer)
+	}
+
```
colin-axner, line 255: could structure should be consistent with OnAck (either way is fine)

modules/apps/29-fee/ibc_module.go
```diff
@@ -171,32 +171,100 @@ func (im IBCModule) OnChanCloseConfirm(
...
+	err := im.keeper.DistributeFee(cacheCtx, sdk.AccAddress(identifiedPacketFee.RefundAddress), sdk.AccAddress(ack.ForwardRelayerAddress), relayer, packetId)
+	// emit the error in event if any
+	ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
+
+	// if there is no error, write the cache and then call underlying callback
+	// otherwise discard cache and call underlying callback
+	if err == nil {
+		writeFn()
+	}
+	// call underlying callback
```
colin-axner, line 233:
```suggestion
	}
	
	// call underlying callback
```
just a suggestion

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,330 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+			types.IncentivizedAcknowledgement{
+				Result:                channeltypes.NewResultAcknowledgement([]byte{1}).Acknowledgement(),
+				ForwardRelayerAddress: suite.chainB.SenderAccount.GetAddress().String(),
+			}.Acknowledgement(),
+			true,
+		},
+		{
+			"ack wrong format",
+			func() {},
+			channeltypes.NewResultAcknowledgement([]byte{1}).Acknowledgement(),
```
colin-axner, line 615:
```suggestion
			[]byte("unsupported acknowledgement format"),
```
this will give you the code coverage

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,330 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+			if !tc.forwardRelayer && tc.feeEnabled {
+				ack := types.IncentivizedAcknowledgement{
+					Result:                channeltypes.NewResultAcknowledgement([]byte{1}).Acknowledgement(),
+					ForwardRelayerAddress: "",
+				}
+				suite.Require().Equal(ack, result)
+			}
+			if tc.forwardRelayer && !tc.feeEnabled {
+				ack := channeltypes.NewResultAcknowledgement([]byte{1})
+				suite.Require().Equal(ack, result)
```
colin-axner, line 577:
```suggestion
                       switch {
                       case !tc.feeEnabled :
				ack := channeltypes.NewResultAcknowledgement([]byte{1})
				suite.Require().Equal(ack, result)
				
			case tc.forwardRelayer:
				ack := types.IncentivizedAcknowledgement{
					Result:                channeltypes.NewResultAcknowledgement([]byte{1}).Acknowledgement(),
					ForwardRelayerAddress: suite.chainB.SenderAccount.GetAddress().String(),
				}
				suite.Require().Equal(ack, result)
				
			case !tc.forwardRelayer:
				ack := types.IncentivizedAcknowledgement{
					Result:                channeltypes.NewResultAcknowledgement([]byte{1}).Acknowledgement(),
					ForwardRelayerAddress: "",
				}
				suite.Require().Equal(ack, result)
			
```
just a suggestion, not sure if it compiles or is gofmt'd

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,330 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+
+			err = cbs.OnAcknowledgementPacket(suite.chainA.GetContext(), packet, tc.ackBytes, suite.chainB.SenderAccount.GetAddress())
+
+			if tc.expPass {
+				suite.Require().NoError(err, "unexpected error for case: %s", tc.name)
+			}
+
+			if !tc.expPass {
+				suite.Require().Error(err, "%s expected error but returned none", tc.name)
+			}
```
colin-axner, line 698:
```suggestion
			if tc.expPass {
				suite.Require().NoError(err, "unexpected error for case: %s", tc.name)
			} else {
			        suite.Require().Error(err, "%s expected error but returned none", tc.name)
			}
```
consistency improves readability (this is how all our tests are structured). Optionally the error message can be removed:
```
suite.Require().Error(err)
```

The `suite.Run` will print the test case name

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,330 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+			tc.malleate()
+			err = cbs.OnTimeoutPacket(suite.chainA.GetContext(), packet, suite.chainA.SenderAccount.GetAddress())
+
+			if tc.expPass {
+				suite.Require().NoError(err, "unexpected error for case: %s", tc.name)
+			}
+
+			if !tc.expPass {
+				suite.Require().Error(err, "%s expected error but returned none", tc.name)
+			}
```
colin-axner, line 795: nit: use common format of if else

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,330 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+				suite.path.EndpointB.ChannelID,
+				clienttypes.NewHeight(0, 100),
+				0,
+			)
+
+			// fund chain A's escrow path so that tokens can be unescrowed upon timeout
+			escrow := transfertypes.GetEscrowAddress(suite.path.EndpointA.ChannelConfig.PortID, suite.path.EndpointA.ChannelID)
+
+			suite.Require().NoError(simapp.FundAccount(suite.chainA.GetSimApp(), suite.chainA.GetContext(), escrow, sdk.NewCoins(coin)))
+
```
colin-axner, line 763: this seems unnecessary since `EscrowPacketFee` is called later?

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,330 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+			packet := channeltypes.NewPacket(
+				fungibleTokenPacket.GetBytes(),
+				suite.chainA.SenderAccount.GetSequence(),
+				suite.path.EndpointA.ChannelConfig.PortID,
+				suite.path.EndpointA.ChannelID,
+				suite.path.EndpointB.ChannelConfig.PortID,
+				suite.path.EndpointB.ChannelID,
+				clienttypes.NewHeight(0, 100),
+				0,
+			)
```
colin-axner, line 757: we could optionally make this into a helper function `constructICS20Packet(suite)` since it is used in all the test functions, just a suggestion, not required by any means

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,330 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+					ReceiveFee: validCoins,
+					AckFee:     validCoins2,
+					TimeoutFee: validCoins3,
+				},
+				refundAcc.String(),
+				[]string{},
+			)
+			// escrow packet fees
+			err = suite.chainA.GetSimApp().IBCFeeKeeper.EscrowPacketFee(suite.chainA.GetContext(), identifiedFee)
+			suite.Require().NoError(err)
```
colin-axner, line 554: this seems unnecessary since we are receiving the packet on chainA (escrow happens on sending chain, but isn't necessary for this test since we aren't testing the sending chain)

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -468,3 +469,330 @@ func (suite *FeeTestSuite) TestOnChanCloseConfirm() {
...
+			identifiedFee := types.NewIdentifiedPacketFee(
+				packetId,
+				types.Fee{
+					ReceiveFee: validCoins,
+					AckFee:     validCoins2,
+					TimeoutFee: validCoins3,
+				},
+				refundAcc.String(),
+				[]string{},
+			)
```
colin-axner, line 551: I think this can be removed after the `EscrowPacketFee` is removed

modules/apps/29-fee/ibc_module.go
```diff
@@ -171,32 +171,100 @@ func (im IBCModule) OnChanCloseConfirm(
...
+// If fees are not enabled, this callback will default to the ibc-core packet callback
 func (im IBCModule) OnRecvPacket(
 	ctx sdk.Context,
 	packet channeltypes.Packet,
 	relayer sdk.AccAddress,
-) ibcexported.Acknowledgement {
-	// TODO: Implement fee specific logic if fee is enabled for the given channel
-	return im.app.OnRecvPacket(ctx, packet, relayer)
+) exported.Acknowledgement {
+	if !im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
```
colin-axner, line 180: This should be destination port and channel. There should be a test which catches this bug (make the receiving chain use a non `channel-0` id). Let me know if you need help with this

colin-axner (commented):
Great work!! Thanks for adding the context caching to the Distrbute fee function. Left several suggestions on code layout, only 1 required change (source -> destination port/channel bug)

modules/apps/29-fee/ibc_module.go
colin-axner: This seems to not have been fixed

modules/apps/29-fee/ibc_module_test.go
colin-axner: ah! Could you add a note that this is setup for the ics20 callback?

charleenfei: will do!

modules/apps/29-fee/ibc_module.go
```diff
@@ -224,9 +222,10 @@ func (im IBCModule) OnAcknowledgementPacket(
 	// cache context before trying to distribute the fee
 	cacheCtx, writeFn := ctx.CacheContext()
 
-	fmt.Println("forward", ack.ForwardRelayerAddress)
+	forwardRelayer, _ := sdk.AccAddressFromBech32(ack.ForwardRelayerAddress)
```
colin-axner, line 225: we should probably check the error

colin-axner (approved):
Outstanding work!! :raised_hands:

modules/apps/29-fee/ibc_module.go
charleenfei: yes...but then should we forego distribution entirely if this error exists?

seantking: I think we handle this here already implicitly -> https://github.com/cosmos/ibc-go/blob/feat/ics-29-packetcallbacks/modules/apps/29-fee/keeper/escrow.go#L56 

So no need to return an error here

Issue #279, opened 2021-05-28
ICA Code Coverage 


seantking:
There code coverage for tests is low & needs to improve. 

Unit tests should be written for :

- [x] https://github.com/cosmos/interchain-accounts/tree/master/x/ibc-account/types
- [x] https://github.com/cosmos/interchain-accounts/tree/master/x/ibc-account/keeper
- [x] https://github.com/cosmos/interchain-accounts/tree/master/x/ibc-account/client/cli

Integration tests should be written for: 

- [x] https://github.com/cosmos/interchain-accounts/tree/master/x/ibc-account/keeper

colin-axner:
Within types files which need tests:
- [ ] account.go
- [ ] genesis.go (needs basic validation logic as well)
- [ ] hook.go
- [ ] packet.go

Within keeper files which need tests:
- [ ] account.go
- [ ] grpc_query.go 
- [ ] handshake.go
- [ ] keeper.go
- [ ] relay.go

Top level files which need tests:
- [ ] module.go (ibc callbacks)
- [ ] genesis.go

CLI tests are difficult to do for IBC due to the structure of in-process app testing. CLI tests are very low priority

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #143, opened 2021-04-28
All gRPC tests should use new testing package


colin-axner:
## Summary

In #140 I noticed some gRPC tests were not using the new testing package. The test case that `req = nil` can only be tested with the new testing package. We should update all tests and remove any old testing code for this setup. It is much easier to maintain the new testing code 
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
I think the idea would be to get rid of `queryClient` field in the test suites for:

- [ ] [transfer](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/apps/transfer/keeper/keeper_test.go#L34)
- [ ] [29-fee](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/apps/29-fee/keeper/keeper_test.go#L63)
- [ ] [02-client](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/core/02-client/keeper/keeper_test.go#L130)

And use call instead the gRPC endpoint using the method in the keeper. For example: for transfer instead of doing `suite.queryClient.DenomTrace`, call `suite.chainA.GetSimApp().TransferKeeper.DenomTrace`.

expertdicer:
Can you please review my changes? Thank you

PR #368, opened 2021-09-02
test: adding interchain accounts grpc query tests

damiannolan:
## Description

- Additional tests for #279 

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/keeper/grpc_query_test.go
+++ b/modules/apps/27-interchain-accounts/keeper/grpc_query_test.go
@@ -1,0 +1,80 @@
+package keeper_test
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+
+	"github.com/cosmos/ibc-go/modules/apps/27-interchain-accounts/types"
+)
+
+func (suite *KeeperTestSuite) TestQueryInterchainAccountAddress() {
+	var (
+		req     *types.QueryInterchainAccountAddressRequest
+		expAddr string
+	)
+
+	testCases := []struct {
+		msg      string
+		malleate func()
+		expPass  bool
+	}{
+		{
+			"empty request",
+			func() {
+				req = &types.QueryInterchainAccountAddressRequest{}
+			},
+			false,
+		},
+		{
+			"invalid counterparty portID",
+			func() {
+				req = &types.QueryInterchainAccountAddressRequest{
+					CounterpartyPortId: "",
+				}
+			},
+			false,
+		},
+		{
+			"interchain account address not found",
+			func() {
+				req = &types.QueryInterchainAccountAddressRequest{
+					CounterpartyPortId: "ics-27",
+				}
+			},
+			false,
+		},
+		{
+			"success",
+			func() {
+				expAddr = authtypes.NewBaseAccountWithAddress(types.GenerateAddress("ics-27")).GetAddress().String()
+				req = &types.QueryInterchainAccountAddressRequest{
+					CounterpartyPortId: "ics-27",
+				}
+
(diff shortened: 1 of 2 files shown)


modules/apps/27-interchain-accounts/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	testCases := []struct {
+		msg      string
+		malleate func()
+		expPass  bool
+	}{
+		{
+			"empty request",
+			func() {
+				req = &types.QueryInterchainAccountAddressRequest{}
```
damiannolan, line 26: This testcase is essentially redundant but also causes no harm as `req` is the same as below in "invalid counterparty portID" due to protobuf defaulting to empty string. 

Using `req = nil` here causes a `panic: invalid memory address or nil pointer dereference` as protobuf attempts to marshal the nil request.

Can be removed if anyone feels it should. Thoughts?

seantking: I think it's fine to leave in.

seantking (approved):
LGTM. Nice!

modules/apps/27-interchain-accounts/keeper/keeper_test.go
```diff
@@ -20,13 +21,19 @@ type KeeperTestSuite struct {
...
 
 func (suite *KeeperTestSuite) SetupTest() {
 	suite.coordinator = ibctesting.NewCoordinator(suite.T(), 3)
 	suite.chainA = suite.coordinator.GetChain(ibctesting.GetChainID(0))
 	suite.chainB = suite.coordinator.GetChain(ibctesting.GetChainID(1))
 	suite.chainC = suite.coordinator.GetChain(ibctesting.GetChainID(2))
+
+	queryHelper := baseapp.NewQueryServerTestHelper(suite.chainA.GetContext(), suite.chainA.GetSimApp().InterfaceRegistry())
+	types.RegisterQueryServer(queryHelper, suite.chainA.GetSimApp().ICAKeeper)
+	suite.queryClient = types.NewQueryClient(queryHelper)
```
colin-axner, line 36: This looks like the legacy grpc testing, you should be able to use the query server without adding a field, see #143

modules/apps/27-interchain-accounts/keeper/grpc_query_test.go
colin-axner: Might be good to update invalid counterparty portID to be empty with spaces and add `strings.TrimSpace` to the grpc.go code?

damiannolan: Agree with this, I think it's a good call. Can handle this in a PR tomorrow

modules/apps/27-interchain-accounts/keeper/keeper_test.go
damiannolan: Ah, nice! I had just referenced what had been done in `apps/transfer`

Will include with other changes in a follow up PR.

damiannolan: I took a look at this. For `apps/` I don't think we can't use `suite.chainA.QueryServer.` as the query server that exists on the `ibctesting.TestChain` is defined [here](https://github.com/cosmos/ibc-go/blob/main/modules/core/types/query.go) and only encapsulates core. 

The other option which removes the `queryClient` field is to invoke the `InterchainAccountAddress` func directly on the `ICAKeeper`. 

```
res, err := suite.chainA.GetSimApp().ICAKeeper.InterchainAccountAddress(ctx, req)
```

What do you think?

colin-axner: Let's do the second option you outlined since the QueryServer is just the keeper [anyways](https://github.com/cosmos/ibc-go/blob/74182d8d45bca3504ed0190e9cdedb1a8ba45973/testing/chain.go#L108) :+1:

PR #373, opened 2021-09-02
IBC Router Middleware

jackzampolin:


modules/apps/router/module.go
```diff
@@ -0,0 +1,300 @@
...
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			transfertypes.EventTypePacket,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+			sdk.NewAttribute(transfertypes.AttributeKeyReceiver, data.Receiver),
+			sdk.NewAttribute(transfertypes.AttributeKeyDenom, data.Denom),
+			sdk.NewAttribute(transfertypes.AttributeKeyAmount, fmt.Sprintf("%d", data.Amount)),
+			sdk.NewAttribute(transfertypes.AttributeKeyAckSuccess, fmt.Sprintf("%t", ack.Success())),
+		),
+	)
```
AdityaSripal, line 247: we can't emit these meaningfully if we can't successfully parse packet data

modules/apps/router/module.go
```diff
@@ -0,0 +1,300 @@
...
+		} else {
+			if err := am.keeper.OnRecvPacket(ctx, packet, data, receiver, finalDest, port, channel); err != nil {
+				ack = channeltypes.NewErrorAcknowledgement(err.Error())
+			}
+		}
+
+	}
+
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
```
AdityaSripal, line 239: also this should be emitted by transfer module anyway correct?

modules/apps/router/module.go
```diff
@@ -0,0 +1,300 @@
...
+			sdk.NewAttribute(transfertypes.AttributeKeyAckSuccess, fmt.Sprintf("%t", ack.Success())),
+		),
+	)
+
+	return nil
+}
+
+// OnAcknowledgementPacket implements the IBCModule interface
+func (am AppModule) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Packet, acknowledgement []byte, relayer sdk.AccAddress) error {
+	return am.app.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
```
AdityaSripal, line 254: If the acknowledgement of a routed packet comes back as an error, then we should send a packet back to original sender to refund the tokens

modules/apps/router/module.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// OnTimeoutPacket implements the IBCModule interface
+func (am AppModule) OnTimeoutPacket(ctx sdk.Context, packet channeltypes.Packet, relayer sdk.AccAddress) error {
+	return am.app.OnTimeoutPacket(ctx, packet, relayer)
```
AdityaSripal, line 259: ditto we must send new packet here to refund original sender in routing case

modules/apps/router/module.go
```diff
@@ -0,0 +1,300 @@
...
+	// was successfully decoded
+	if ack.Success() {
+		receiver, finalDest, port, channel, err := ParseIncomingTransferField(data.Receiver)
+		if finalDest == "" && port == "" && channel == "" && err == nil {
+			ack = am.app.OnRecvPacket(ctx, packet, relayer)
+		} else if err != nil {
+			ack = channeltypes.NewErrorAcknowledgement("cannot unmarshal ICS-20 transfer packet data")
+		} else {
+			if err := am.keeper.OnRecvPacket(ctx, packet, data, receiver, finalDest, port, channel); err != nil {
+				ack = channeltypes.NewErrorAcknowledgement(err.Error())
```
AdityaSripal, line 232: Replace this with:

```go
ack = am.app.OnRecvPacket
denom := calculateIntermediateDenom(packet)
newPacket := newTransferPacket(denom, packet.Amount, finalDest)
sendPacket(packet)
```

modules/apps/router/keeper/keeper.go
```diff
@@ -0,0 +1,223 @@
...
+		bankKeeper:     bankKeeper,
+	}
+}
+
+// Logger returns a module-specific logger.
+func (k Keeper) Logger(ctx sdk.Context) log.Logger {
+	return ctx.Logger().With("module", "x/"+host.ModuleName+"-"+types.ModuleName)
+}
+
+func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data transfertypes.FungibleTokenPacketData, receiver sdk.AccAddress, finalDest, port, channel string) error {
```
AdityaSripal, line 53: Do not redo the transfer logic here. Instead just use the underlying app, recalculate the denom, and call `ForwardPacket`

AdityaSripal (changes requested):
Great start!

The receive logic in the routing case can be a lot cleaner if you call into the underlying transfer module and just do your own custom logic on top of it.

Acknowledgment and Timeout need to be fixed on the routing case

modules/apps/router/module.go
jackzampolin: Wouldn't the transfer module just handle this?

modules/apps/router/module.go
jackzampolin: This feature would leave the user on the intermediate chain at their refund address in the case where packet times out or fails.

modules/apps/router/keeper/keeper.go
jackzampolin: Ok! I think I know what to do now.

modules/apps/router/module.go
jackzampolin: Ok, I'm understanding this now thank you!

ebuchman:
Is there an ICS for this or docs/spec or a relevant issue? Could we get some text description in the opening comment please?

alexanderbez:
This PR is referenced in the official on-chain governance proposal. It's not clear why this PR was closed. Can we comment what this was replaced with and why?

crodriguezvega:
Being a specialized solution to enable packet routing in the Hub in the short-term (as opposed to a generic, ICS-backed multi-hop solution), we decided that ibc-go was not the most appropriate home for this middleware, so the implementation now lives in a [Strangelove's repo](https://github.com/strangelove-ventures/packet-forward-middleware).

Issue #370, opened 2021-09-02
Allow PacketAcknowledgements to take in packet commitments for filtering


colin-axner:
## Summary

[`UnreceivedAcknowledgements`](https://github.com/cosmos/ibc-go/blob/9d5da1760d3c389e80d3f8af264bb551a0733143/modules/core/04-channel/keeper/grpc_query.go#L420) currently takes in a list of acknowledgements and checks to see if the packet commitment still exists. The list of acknowledgements is obtained via a gRPC which returns all existing acknowledgements (even if the have been acknowledged on the sending chain). That gRPC should take in an optional list of packet commitments and only return acknowledgements for that list of packet commitments. 

## Problem Definition

The issue is acknowledgements live forever so it isn't sustainable to expect a relayer to be able to query for all ack sequences. Currently such a query returns ~64k sequences

## Proposal

Add an additional field to the `QueryPacketAcknowledgementsRequest` which takes in a list of packet commitments. For each packet commitment, if the acknowledgement has been written, that sequence should be added to the array returned in the `QueryPacketAcknowledgementsResponse` 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
@colin-axner This should remain backwards compatible right? If no `commitments` are included in the req, all should be added to the returned array as it works today, correct?

colin-axner:
> @colin-axner This should remain backwards compatible right? If no `commitments` are included in the req, all should be added to the returned array as it works today, correct?

yup!

PR #375, opened 2021-09-03
packet acknowledgment filtering

damiannolan:
## Description

- Adding optional field `commitments` to `QueryPacketAcknowledgementsRequest` to allow filtering

closes: #370

---

write a little note why.


CHANGELOG.md
```diff
@@ -102,7 +107,6 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### Features
 
 * [\#198](https://github.com/cosmos/ibc-go/pull/198) New CLI command `query ibc-transfer escrow-address <port> <channel id>` to get the escrow address for a channel; can be used to then query balance of escrowed tokens
-* [\#372](https://github.com/cosmos/ibc-go/pull/372) New CLI command `query ibc client status <client id>` to get the current activity status of a client
```
damiannolan, line 105: Accidentally placed this under v1.0.0 in previous PR. Moving to Unreleased section

modules/core/04-channel/keeper/grpc_query.go
```diff
@@ -335,8 +335,18 @@ func (q Keeper) PacketAcknowledgements(c context.Context, req *types.QueryPacket
...
-		ack := types.NewPacketState(req.PortId, req.ChannelId, sequence, value)
-		acks = append(acks, &ack)
+		if len(req.Commitments) > 0 {
+			for _, commitment := range req.Commitments {
+				if sequence == commitment.Sequence {
+					ack := types.NewPacketState(req.PortId, req.ChannelId, sequence, value)
+					acks = append(acks, &ack)
+				}
+			}
+		} else {
```
damiannolan, line 345: Assuming we want to preserve previous functionality if no commitments are included in the req

colin-axner: Maybe add a comment for future readers?

modules/core/04-channel/keeper/grpc_query.go
```diff
@@ -335,8 +335,18 @@ func (q Keeper) PacketAcknowledgements(c context.Context, req *types.QueryPacket
...
 
-		ack := types.NewPacketState(req.PortId, req.ChannelId, sequence, value)
-		acks = append(acks, &ack)
+		if len(req.Commitments) > 0 {
+			for _, commitment := range req.Commitments {
+				if sequence == commitment.Sequence {
+					ack := types.NewPacketState(req.PortId, req.ChannelId, sequence, value)
+					acks = append(acks, &ack)
+				}
+			}
```
colin-axner, line 344: You'll want to actually modify the query being done. Since we can assume there to be thousands of acknowledgements written in state, a small pagination limit might actually return the incorrect values since the pagination might stop before getting all the necessary commitments

What we want to do is query for an acknowledgement for each commitment provided. This will look similar to this [code](https://github.com/cosmos/ibc-go/pull/375/files#diff-74764e734e2ec27e29dc1f3e36deef94d0c1fa798c60478cda09f2cd9a1316b4L384) but for acknowledgements

modules/core/04-channel/keeper/grpc_query_test.go
```diff
@@ -1025,6 +1025,34 @@ func (suite *KeeperTestSuite) TestQueryPacketAcknowledgements() {
 			},
 			true,
 		},
+		{
+			"success, filtered res",
+			func() {
+				path := ibctesting.NewPath(suite.chainA, suite.chainB)
+				suite.coordinator.Setup(path)
+
+				for i := uint64(0); i < 10; i++ {
```
colin-axner, line 1034: It might be good to make the amount of acknowledgements in state large and the pagination limit small. This way we can test that we are only querying for acks for the commitments provided. I would try adjusting this so that the current code fails, then fix the code

colin-axner (changes requested):
Great start! Let me know if you have any questions on the comments I left

modules/core/04-channel/keeper/grpc_query.go
damiannolan: Ah okay! So querying for specific packet acks should be done outside the bounds of the paginated query.
I think the link you provided is pointing to tests in this PR, after taking a look I assume it's this code you were referencing (correct me if not):

```
for i, seq := range req.PacketCommitmentSequences {
		if seq == 0 {
			return nil, status.Errorf(codes.InvalidArgument, "packet sequence %d cannot be 0", i)
		}

		// if packet receipt exists on the receiving chain, then packet has already been received
		if _, found := q.GetPacketReceipt(ctx, req.PortId, req.ChannelId, seq); !found {
			unreceivedSequences = append(unreceivedSequences, seq)
		}
}
```

colin-axner: yup :+1: . Since the caller provides a list of sequences we don't need the pagination. We still need the pagination if the caller passes in an empty list of sequences (keep old behaviour)

modules/core/04-channel/keeper/grpc_query.go
damiannolan: Done!

modules/core/04-channel/keeper/grpc_query.go
damiannolan: Updated! 

I generally prefer returning early than having most of the function body wrapped in an if/else. Also omitted the following, but let me know if it's required.

```
if seq == 0 {
	return nil, status.Errorf(codes.InvalidArgument, "packet sequence %d cannot be 0", i)
}
```

modules/core/04-channel/keeper/grpc_query.go
```diff
@@ -327,6 +327,27 @@ func (q Keeper) PacketAcknowledgements(c context.Context, req *types.QueryPacket
 	acks := []*types.PacketState{}
 	store := prefix.NewStore(ctx.KVStore(q.storeKey), []byte(host.PacketAcknowledgementPrefixPath(req.PortId, req.ChannelId)))
 
+	// if a list of packet sequences is provided then query for each specific ack and return
+	// otherwise, maintain previous behaviour and perform paginated query
+	for _, seq := range req.PacketCommitmentSequences {
+		acknowledgementBz, found := q.GetPacketAcknowledgement(ctx, req.PortId, req.ChannelId, seq)
+		if !found || len(acknowledgementBz) == 0 {
+			return nil, status.Error(codes.NotFound, "packet acknowledgement hash not found")
+		}
```
colin-axner, line 336: I think it is ok to just continue and don't append the ack to the list. Relayers want to use the return value to pass it into `UnreceivedAcknowledgements`. This will take a list of acks and check if they still need to be acknowledged on the source chain. So I think the relayers are expecting to pass in a list of commitments and get a len(acks) <= len(commitments) rather than an error

damiannolan: Makes sense! Done

modules/core/04-channel/keeper/grpc_query.go
```diff
@@ -327,6 +327,27 @@ func (q Keeper) PacketAcknowledgements(c context.Context, req *types.QueryPacket
...
+		acknowledgementBz, found := q.GetPacketAcknowledgement(ctx, req.PortId, req.ChannelId, seq)
+		if !found || len(acknowledgementBz) == 0 {
+			continue
+		}
+
+		ack := types.NewPacketState(req.PortId, req.ChannelId, seq, acknowledgementBz)
+		acks = append(acks, &ack)
+	}
+
+	if len(acks) > 0 {
```
colin-axner, line 342: One last change

```suggestion
	if len(req.PacketCommitmentSequences) > 0 {
```

I think we should return an empty array if none of the commitments provided had an acknowledgement written as opposed to returning all acknowledgements. That might be surprising behaviour

damiannolan: Great point! That scenario skipped my mind.
Updated via Github UI suggestion

modules/core/04-channel/keeper/grpc_query.go
colin-axner: I think it is fine to omit this, it shouldn't cause any issues

colin-axner (approved):
Excellent!

Issue #136, opened 2021-04-23
Scalable version queries


colin-axner:
## Summary

Querying for an application version is currently unsupported (during handshakes). My initial thoughts indicate that there are two solutions:
- modifying the module interface to return the supported versions
- supporting gRPC routes for every application version

Both have their draw backs and (2) isn't scalable. I'd like to give this some more thought to determine if there's a way we could abstract it so that it can scale without introducing more constraints required for IBC execution

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
To close this issue, we should modify the [application module interface](https://github.com/cosmos/ibc-go/blob/632d71f593bfc66c472a24378c97a04cb67ba24f/modules/core/05-port/types/module.go#L13) by adding another function. This function should perform application version negotiation and return the version to be used in the OpenTry step. 

The relayer will choose the initial version in the OpenInit step, this will likely be chosen by the user controlling the relayer or by the application which triggers the OpenInit step (interchain accounts)

The version submitted in the OpenInit step will be submitted as an argument to the function being adding to the module interface. The function should look at the version selected in OpenInit and return the matching version to be used for OpenTry. Applications may choose to implement this in however fashion they choose. 

ICS20 would simply return the "ics20-1" or perhaps an error if the counterparty version is unsupported. Assuming a second version of ICS20 eventually arises, we may expect this negotiation only to return a valid version if the counterparty version is supported. So if the counterparty used "ics20-1" in OpenInit, it returns "ics20-1". The same would be true if it used "ics20-2". If the counterparty used an unsupported version, say "ics20-5", then an error should be returned

ICS27 would have similar version negotiation. Assuming ICS27 also decides to interject the host chain account address into the version, then we would expect this function to generate interchain account adddress using the counterparty port ID and append this to the existing version (or use the associated protobuf version)

This function signature may look something like:
```go
func NegotiateAppVersion(ctx sdk.Context, portID string, counterparty channeltypes.Counterparty, counterpartyVersion string) (version string, err error) 
```
The name may be different (I haven't given it to much thought atm). I left off channelID in the function signature since it may not be known in advance by a relayer, but the portID should be known. I'm unsure if we should include the connection hops as well

We should assume some applications may return versions based on current state and thus access to the context may be useful

seantking:
wdyt about adding this to the interchain-accounts milestone?

[...]

Issue #334, opened 2021-08-13
Unify text format for invalid/unexpected value errors


crodriguezvega:
## Summary

The error message for an invalid value is formatted in slightly different ways. Take the following three as an example:

`"expected %s channel, got %s "` (https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/module.go#L207)
`"invalid port: %s, expected %s"` (https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/module.go#L213)
`"got %s, expected %s"` (https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/module.go#L217)

The goal is to format this kind of errors in a consistent way.

The proposal is to do the following:

```
sdkerrors.Wrapf(
  <most specific error type possible>,
  "<optional text description ended by colon and space>expected %s, got %s",
  <value 1>,
  <value 2>
)
```
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

iricehasan:
Hello @crodriguezvega , I would like to do this if this is still needed. Can you assign me this issue?

crodriguezvega:
Sorry for the late reply, @iricehasan. I hope you're still interested in the issue, I will assign it to you now.

iricehasan:
Hi @crodriguezvega, I have just seen your reply. I am still interested.

gjermundgaraba:
We will not do more work on this right now, so closing for now.

Issue #112, opened 2021-04-09
Add gRPC routes for 05-port


colin-axner:
## Summary

05 port doesn't have any gRPC routes, but it might be useful to query the application associated with a port identifier. This can help identify what packets are being sent across certain channels

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Going to close as stale. Can be reopened with port router v2 if it is decided to be useful

PR #384, opened 2021-09-08
feat: scalable version queries

damiannolan:
## Description

- Adding gRPC query interface to `05-port`
- Adding `NegotiateAppVersion` method to `IBCModule` interface

The following adds a new method `NegotiateAppVersion` to the `IBCModule` interface, supported by a gRPC query interface for `05-port` which provides routing of requests to the appropriate application module callback.

Initially opening a draft to raise some questions.

closes: #136

---

write a little note why.


modules/core/05-port/client/cli/query.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/version"
+
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
+	"github.com/cosmos/ibc-go/modules/core/05-port/types"
+	host "github.com/cosmos/ibc-go/modules/core/24-host"
+)
+
+// GetCmdQueryPorts defines the command to query a port
```
damiannolan, line 17: I'm unsure if a CLI cmd is really useful here or not. Thoughts?

proto/ibc/core/port/v1/query.proto
```diff
@@ -0,0 +1,31 @@
...
+package ibc.core.port.v1;
+
+option go_package = "github.com/cosmos/ibc-go/modules/core/05-port/types";
+
+import "ibc/core/channel/v1/channel.proto";
+
+// Query provides defines the gRPC querier service
+service Query {
+  // Port queries an IBC Port.
+  rpc Port(QueryPortRequest) returns (QueryPortResponse) {}
```
damiannolan, line 12: Perhaps the naming here is too generic and would be better suited as something more applicable/specific to port versions. 

Suggestions welcome!

modules/apps/transfer/module.go
```diff
@@ -434,3 +434,17 @@ func (am AppModule) OnTimeoutPacket(
...
+
+// NegotiateAppVersion implements the IBCModule interface
+func (am AppModule) NegotiateAppVersion(
+	ctx sdk.Context,
+	portID string,
+	counterparty channeltypes.Counterparty,
+	counterpartyVersion string,
+) (version string, err error) {
+	if counterpartyVersion != types.Version {
+		return "", sdkerrors.Wrapf(types.ErrInvalidVersion, "invalid counterparty version: got: %s, expected %s", counterpartyVersion, types.Version)
```
damiannolan, line 446: I think we can safely return an error here if there is a version mismatch.
One would imagine that a hypothetical `"ics20-2"` would involve breaking changes

proto/ibc/core/port/v1/query.proto
damiannolan: Another open question here is: should a HTTP interface using grpc gateway sit in front of this endpoint as well?

modules/core/05-port/client/cli/query.go
seantking: Hmm, I think we could probably remove the CLI command for this. I can't think of any reason to use it? I'm in favor of having as little code to maintain as possible if you are. :D 

Can you think of any advantage for having the CLI command as well? @colin-axner

proto/ibc/core/port/v1/query.proto
seantking: wdyt about `PortVersion`?

proto/ibc/core/port/v1/query.proto
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// QueryPortRequest is the request type for the Query/Port RPC method
+message QueryPortRequest {
+  // port unique identifier
+  string port_id = 1;
+  // counterparty channel end
+  ibc.core.channel.v1.Counterparty counterparty = 2;
+  // counterparty version
+  string counterparty_version = 3;
```
seantking, line 22: Does the counterparty version need to be included in the request here?

proto/ibc/core/port/v1/query.proto
seantking: I imagine I'm querying for a resource which is in this case is the `CounterpartyVersion` given a port-id

proto/ibc/core/port/v1/query.proto
```diff
@@ -0,0 +1,31 @@
...
+
+package ibc.core.port.v1;
+
+option go_package = "github.com/cosmos/ibc-go/modules/core/05-port/types";
+
+import "ibc/core/channel/v1/channel.proto";
+
+// Query provides defines the gRPC querier service
+service Query {
+  // Port queries an IBC Port.
```
seantking, line 11: I think we should add some docs here for relayer users

// CounterpartyVersion takes a port-id and routes to the respective application module VersionNegotiation implementation

seantking:
:rocket:

proto/ibc/core/port/v1/query.proto
damiannolan: So this query will be invoked by a relayer on the host chain, post `OnChanOpenInit` on the controller and pre `OnChanOpenTry` on the host, correct? 
We are querying for the `CounterpartyVersion` as from the perspective of the controller, but the chain servicing the request is the counterparty so maybe `PortVersion` could be a better fit. 

Correct me if I'm wrong here! Would be interested to hear @colin-axner's thoughts also on naming conventions

proto/ibc/core/port/v1/query.proto
damiannolan: Yep, definitely! 👍  Planned on improving this when naming is settled

proto/ibc/core/port/v1/query.proto
damiannolan: Yep, it's currently being used in the `transfer` module's `NegotiateAppVersion` implementation, and I think we would also need it for the ICA approach

modules/core/05-port/client/cli/query.go
colin-axner: I am also in favor of keeping a very lean cli. I pushed pretty hard to remove all the handshake cli function :smile: 

I do see benefit of adding a cli function to query ports that are currently binded and/or a cli to query what module belongs to a port. So if I query `transfer` I should get back `ics20`. The app negotiation part definitely does not need a cli. The cli cmds for port could be addressed in a pr for #112 (which this partially addresses)

I'll leave my comments down below as I still need to think about the proposed gRPCs and their usage. Would be beneficial to get input from the Hermes team

damiannolan: Sounds good. I can take on that PR for #112 following on from the ICA work if you like.

modules/core/05-port/types/module.go
```diff
@@ -82,4 +82,11 @@ type IBCModule interface {
 		packet channeltypes.Packet,
 		relayer sdk.AccAddress,
 	) error
+
+	NegotiateAppVersion(
```
colin-axner, line 86: we should add a comment about what the return values means. From my perspective, an error should only be returned if the version negotiation cannot be performed. Maybe this is implicit?

proto/ibc/core/port/v1/query.proto
colin-axner: What about just `AppVersionNegotiation`? This is supposed to be called pre OnChanOpenTry, but really it is just application version negotiation based on some suggested counterparty version

modules/core/05-port/client/cli/query.go
colin-axner: That'd be fantastic honestly. This was one of the first feature requests we had after launching the initial IBC module

modules/core/05-port/types/module.go
```diff
@@ -82,4 +82,11 @@ type IBCModule interface {
...
+		ctx sdk.Context,
+		portID string,
+		counterparty channeltypes.Counterparty,
+		counterpartyVersion string,
```
colin-axner, line 90: Information not provided:
- channel id
- channel ordering
- channel state
- connection hop (connection channel is attached to)

channel id cannot necessarily be known by OnChanOpenTry. Channel state shouldn't matter

Should we include channel ordering and connection hops? We could always break API later to add it? I have no strong opinion except knowing the associated connection may be useful

proto/ibc/core/port/v1/query.proto
colin-axner: Yes this is needed for app version negotiation. We could rename it to `proposed_version` as well, but `counterparty_version` works

proto/ibc/core/port/v1/query.proto
```diff
@@ -0,0 +1,31 @@
...
+  // counterparty channel end
+  ibc.core.channel.v1.Counterparty counterparty = 2;
+  // counterparty version
+  string counterparty_version = 3;
+}
+
+// QueryPortResponse is the response type for the Query/Port RPC method.
+message QueryPortResponse {
+  // port id associated with the request identifiers
+  string port_id = 1;
```
colin-axner, line 28: Do we need this?

proto/ibc/core/port/v1/query.proto
colin-axner: We can add information about this gRPC in our migration docs

colin-axner (commented):
posting comments, haven't finished reviewing but figured my comments might be useful sooner than later. Will continue reviewing when I have time

proto/ibc/core/port/v1/query.proto
seantking: I'm a little confused by this. I thought the point is that at this stage we don't actually know what the `counterparty_version` is?

modules/core/05-port/types/module.go
seantking: I think there is no harm in adding it now. If I'm understanding you correctly, no current module requires these parameters for negotiating the counterparty version but adding them now will let future modules use as required?

proto/ibc/core/port/v1/query.proto
seantking: Kinda nitpicking here but I guess it depends on how we want to structure our overall API. My instinct was to have our query endpoints return a particular resource (and thus have the naming follow that), but seeing as this is not really returning anything from state maybe that doesn't fit. 

I'm happy with whatever you both decide.

proto/ibc/core/port/v1/query.proto
colin-axner: The confusion is what `counterparty` means. This probably means we should change the naming

Happy path:
chainA - ChanOpenInit
chainB - ChanOpenTry

preChanOpenTry, the relayer may need to use this gRPC to determine what version to supply to the channel on chainB (for ChanOpenTry). So if this is called on `chainB` the counterparty is `chainA`. Obviously this could be called on chainA as well. I recommend we use `proposed_version` and title this gRPC `NegotiateAppVersion` - no context of what is the `counterparty` is needed

modules/core/05-port/types/module.go
colin-axner: Correct. To the best of my knowledge, interchain accounts is the first module introducing multiple valid versions. It's hard to know what version negotiation will look like for applications, but I agree probably best to be as flexible as possible

modules/core/05-port/client/cli/cli.go
```diff
@@ -0,0 +1,25 @@
+package cli
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/ibc-go/modules/core/05-port/types"
+)
+
+// GetQueryCmd returns the query commands for IBC channels
```
colin-axner, line 10: godoc needs updating. Could you fix the imports as well?

modules/core/05-port/types/errors.go
```diff
@@ -6,8 +6,9 @@ import (
...
 var (
-	ErrPortExists   = sdkerrors.Register(SubModuleName, 2, "port is already binded")
-	ErrPortNotFound = sdkerrors.Register(SubModuleName, 3, "port not found")
-	ErrInvalidPort  = sdkerrors.Register(SubModuleName, 4, "invalid port")
-	ErrInvalidRoute = sdkerrors.Register(SubModuleName, 5, "route not found")
+	ErrPortExists     = sdkerrors.Register(SubModuleName, 2, "port is already binded")
+	ErrPortNotFound   = sdkerrors.Register(SubModuleName, 3, "port not found")
+	ErrInvalidPort    = sdkerrors.Register(SubModuleName, 4, "invalid port")
+	ErrInvalidRoute   = sdkerrors.Register(SubModuleName, 5, "route not found")
+	ErrInvalidVersion = sdkerrors.Register(SubModuleName, 6, "invalid port version")
```
colin-axner, line 13: What is this error used for? There is no port version

damiannolan: Ah yes! Good spot, added by mistake and didn't end up using this.
Later realised it should be scoped to the application module and `ErrInvalidVersion` already existed for `transfer` 👍

modules/core/05-port/client/cli/cli.go
damiannolan: Sure. Do you want to leave this `cli.go` in place for #112 and remove the cmd for version negotiation? 
Alternatively I can remove the entire `client/cli` and re-add later

proto/ibc/core/port/v1/query.proto
damiannolan: No, it's not required as the caller will already have this, but semantically it may be nice to return a complete response where it's explicit that the following port id and version are coupled. 

I don't feel particularly strongly about this though, I can remove if needs be!

proto/ibc/core/port/v1/query.proto
damiannolan: Agreed, already started to make these changes, will push an update and include other suggestions also!

modules/core/05-port/types/module.go
damiannolan: Added a short comment above!

proto/ibc/core/port/v1/query.proto
seantking: makes sense!

proto/ibc/core/port/v1/query.proto
colin-axner: Nitpicking is good because it is harder to change this later. What about `AppVersion`?

proto/ibc/core/port/v1/query.proto
colin-axner: I don't feel particularly strong about this either. @seantking thoughts?

modules/core/05-port/client/cli/cli.go
colin-axner: lets remove entire cli and re-add later

proto/ibc/core/port/v1/query.proto
damiannolan: Indeed! Naming can be tricky to get right, and it's 100% harder to change after the fact.

Currently named as `NegotiateAppVersion`, I'd be for shorting this to `AppVersion` if you both agree, and request/response would be `QueryAppVersionRequest` and `QueryAppVersionResponse` respectively. 
Does `ChannelVersion` fit the use case here? If so I'd be open to it as an alternative but my instincts are favouring one of `NegotiateAppVersion` or just `AppVersion`

colin-axner: Lets go with `AppVersion` if @seantking agrees. `AppVersion` is the version stored in the channel, so sometimes it is referred to as channel version. It isn't technically referenced as `AppVersion` anywhere (yet), but it is the version for the application so I think it makes sense to head in this direction

modules/core/05-port/types/module.go
```diff
@@ -82,4 +82,14 @@ type IBCModule interface {
...
+
+	// NegotiateAppVersion performs application version negotiation given the provided port ID, counterparty and proposed version.
+	// An error is returned if version negotiation cannot be performed. For example, an application module implementing this interface
+	// may decide to return an error in the event of the proposed version being incompatible with it's own
+	NegotiateAppVersion(
+		ctx sdk.Context,
+		portID string,
+		counterparty channeltypes.Counterparty,
+		proposedVersion string,
+	) (version string, err error)
```
colin-axner, line 94: this still needs to take in the channel order and connection id

modules/apps/transfer/module.go
```diff
@@ -434,3 +434,17 @@ func (am AppModule) OnTimeoutPacket(
...
+
+// NegotiateAppVersion implements the IBCModule interface
+func (am AppModule) NegotiateAppVersion(
+	ctx sdk.Context,
+	portID string,
+	counterparty channeltypes.Counterparty,
+	proposedVersion string,
+) (version string, err error) {
+	if proposedVersion != types.Version {
+		return "", sdkerrors.Wrapf(types.ErrInvalidVersion, "failed to negotiate app version: got: %s, expected %s", proposedVersion, types.Version)
```
colin-axner, line 446:
```suggestion
		return "", sdkerrors.Wrapf(types.ErrInvalidVersion, "failed to negotiate app version: expected %s, got %s", types.Version, proposedVersion)
```
see #334

colin-axner (commented):
Looking great! Just needs the API to include the additional fields I noted and the proto naming changes

modules/apps/transfer/module.go
damiannolan: Noted for future ;)

modules/core/05-port/types/module.go
damiannolan: Sure, I'll add these fields to `QueryAppVersionRequest` also 👍

colin-axner: I'm in favor of `connectionID` with `string`. Connection hops can be confusing and we aren't even sure if multi hop will ever be baked into core IBC. If it is, it'll probably be API breaking anyways so better to force app developers to realize the change has been made

damiannolan: Cool, thank you ❤️

proto/ibc/core/port/v1/query.proto
seantking: I say leave it in, sounds good. :+1:

colin-axner (approved):
Fantastic work!!

All that is left is documentation updates. I think we are still figuring out the organization/flow for keeping docs up to date and where are documentation is stored and what it covers conceptually, so I think it is fine to address this later if needed. 

Normally this sort of feature is one we want to highlight for relayers and thus would include a little more detail in our migration docs. The format for this is still being defined, but I think I have a template in mind, so I'll do the migration docs for this when I handle #390 and then in the future you'll have a better idea of how to add to that document

seantking (approved):

damiannolan:
> Normally this sort of feature is one we want to highlight for relayers and thus would include a little more detail in our migration docs. The format for this is still being defined, but I think I have a template in mind, so I'll do the migration docs for this when I handle #390 and then in the future you'll have a better idea of how to add to that document

Awesome! My next question was going to be about the migration docs, this sounds perfect. Thanks! 💯

Issue #378, opened 2021-09-06
Set interchain account address on controller chain during init: tracking issue


seantking:
## Summary

In the ics27 specification, the following function is defined: 

```
// This helper function is required for the controller chain to get the address of a newly registered interchain account on a host chain.
// Because the registration of an interchain account happens during the channel creation handshake, there is no way for the controller chain to know what the address of the interchain account is on the host chain in advance. 
// This function sends an IBC packet to the host chain, on the owner port + active channel with the sole intention of eventually parsing the interchain account address from the Acknowledgement packet on the controller chain side.
// The OnAcknowledgePacket function on the controller chain will handle the parsing + setting the interchain account address in state.
// The controller chain builds the messages (before sending via IBC in the TrySendTx fn) that the host side will eventually execute. Therefore, the interchain account address must be known by the controller chain.
function GetInterchainAccountAddressFromAck(connectionId: string, counterPartyConnectionId: string, ownerAddress: string) returns (nil){
    // Sends a generic IBC packet to the host chain with the intention of parsing the interchain account address associated with this port/connection/channel from the Acknowledgement packet.
}
```

This step is required for the controller chain to know the address of the interchain account on the host side. 

The problem with this approach is twofold:
1. Poor UX: This is an unnecessary extra step
2. Cost: As this is an additional IBC transaction it will cost extra tokens 

As suggested here: https://github.com/cosmos/ibc/issues/597 we can use the version string during the channel creation handshake to pass the address of the interchain account back to the host chain. 

In order for this to be possible, a relayer will need to query the host chain for an updated version string prior to calling the OnChanOpenAck endpoint. At present this query does not exist. There is an issue here describing the work that needs to be done on `ibc-go`: https://github.com/cosmos/ibc-go/issues/136

Work that needs to be done to achieve this:
- [x] https://github.com/cosmos/ibc-go/issues/136
- [ ] Hermes integration of version string query. At present there is a hardcoded string being used, see: https://github.com/informalsystems/ibc-rs/blob/master/relayer/src/chain.rs#L219
- [x] Once the above work has been done, call `SetInterchainAccount` during the `OnChanOpenAck` callback with the passed in version string (Which will be the address of the registered interchain account)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

seantking:
@adizere

gjermundgaraba:
[...]

Issue #303, opened 2021-07-27
ICA module addresses may conflict


colin-axner:
## Summary

ICA addresses may conflict with other module addresses based on the current construction (A hash of the counterparty portID `ics27-<connectionID>-<counterpartyConnectionID>-<owneraddress>`. This was a concern with ICS20 and there was a lot of [discussion and extra checking](https://github.com/cosmos/cosmos-sdk/issues/7737) that went into ensuring the proposed solutions was safe. We need to revisit this issue for Interchain Accounts. Ideally the changes in v0.43 of SDK allow us to use some sort of prefixed address

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
see [comment](https://github.com/cosmos/cosmos-sdk/issues/7618#issuecomment-897183844)

Each interchain account address created should be a sub account of the module account owned by the interchain module

damiannolan:
Closed by #428

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #410, opened 2021-09-15
feat: ica app version negotiation

damiannolan:
## Description

- Adding NegotiateAppVersion implementation for ica
- Set interchain account address on controller `OnChanOpenAck`

closes: #378 

---

write a little note why.


modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -71,10 +75,12 @@ func (k Keeper) OnChanOpenTry(
...
 	}
-	if version != types.Version {
-		return sdkerrors.Wrapf(types.ErrInvalidVersion, "got: %s, expected %s", version, types.Version)
+
+	if !strings.Contains(version, types.Version) {
+		return sdkerrors.Wrapf(types.ErrInvalidVersion, "invalid version: %s, expected %s", version, types.Version)
 	}
-	if counterpartyVersion != types.Version {
+
+	if !strings.Contains(counterpartyVersion, types.Version) {
```
damiannolan, line 83: With the inclusion of the account address in the version string we now have to deviate from the spec as we cannot explicitly check for equality.

I've updated these checks to use `strings.Contains()`, but perhaps there's some room for discussion on this topic

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -20,9 +20,10 @@ const (
 	ICAPrefix string = "ics-27"
 )
 
-// GenerateAddress returns a truncated SHA256 hash using the provided port string
-func GenerateAddress(port string) []byte {
-	return tmhash.SumTruncated([]byte(port))
+// GenerateAddress returns an sdk.AccAddress using the provided port identifier and derived from the IBC interchainaccounts module account
```
damiannolan, line 23: @colin-axner I've included this as per #303 after checking out the comment you had linked! I'm assuming this is a new approach which has yet to be adopted across the sdk. This seems to me like the correct usage but I'm not 100% as other modules use a different pattern.

For example `transfer` is composed with an `authtypes.AccountKeeper` and accesses the module accounts via that interface which from what I can tell are populated through [app.go maccPerms](https://github.com/cosmos/gaia/blob/7a1b89bc138cb6f24b13b78118de0fc928967f4e/app/app.go#L132)

damiannolan: Also updated this return sig from `[]byte` to `sdk.AccAddress` as it's probably preferred to use that type for bech32 encoding

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -6,22 +6,34 @@ const (
...
+	// ModuleAccountKey defines the key used to construct to the IBC interchainaccounts module account address
+	ModuleAccountKey = []byte(ModuleName)
+
 	// Key to store portID in our store
 	PortKey = "portID"
 
-	QuerierRoute = ModuleName
-
 	// MemStoreKey defines the in-memory store key
 	MemStoreKey = "mem_capability"
```
damiannolan, line 34: `MemStoreKey` and `KeyPrefixRegisteredAccount` below both appear to be used. Any reason they can't be removed?

modules/apps/27-interchain-accounts/types/account.go
seantking: Nice

modules/apps/27-interchain-accounts/module_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	counterparty := &channeltypes.Counterparty{
+		PortId:    counterpartyPortID,
+		ChannelId: path.EndpointB.ChannelID,
+	}
+
+	version, err := cbs.NegotiateAppVersion(suite.chainA.GetContext(), channeltypes.ORDERED, path.EndpointA.ConnectionID, types.PortID, *counterparty, types.Version)
+	suite.Require().NoError(err)
+	suite.Require().True(strings.Contains(version, types.Version))
+	suite.Require().True(strings.Contains(version, types.GenerateAddress(counterpartyPortID).String()))
```
seantking, line 71: Rather than doing a contains here I think it would be better to create a `const` above for the expected format and then doing a strict equal.

modules/apps/27-interchain-accounts/keeper/account.go
```diff
@@ -40,23 +40,28 @@ func (k Keeper) InitInterchainAccount(ctx sdk.Context, connectionID, counterpart
...
+	accAddr, err := sdk.AccAddressFromBech32(address)
+	if err != nil {
+		return err
+	}
 
-	account := k.accountKeeper.GetAccount(ctx, address)
-	if account != nil {
-		return
+	if acc := k.accountKeeper.GetAccount(ctx, accAddr); acc != nil {
+		return nil
```
seantking, line 52: We should probably return a more descriptive error here. wdyt?

seantking: Also, wondering if this check to get an account needs to change at all given we are using the new module acc feature?

modules/apps/27-interchain-accounts/keeper/account.go
```diff
@@ -40,23 +40,28 @@ func (k Keeper) InitInterchainAccount(ctx sdk.Context, connectionID, counterpart
...
 	}
 
 	interchainAccount := types.NewInterchainAccount(
-		authtypes.NewBaseAccountWithAddress(address),
+		authtypes.NewBaseAccountWithAddress(accAddr),
 		portID,
 	)
 
 	k.accountKeeper.NewAccount(ctx, interchainAccount)
 	k.accountKeeper.SetAccount(ctx, interchainAccount)
```
seantking, line 61: I think before we set the account in the state we need to do a check to ensure that the address passed into the version string by the relayer is the one that we expect (This is based on the counterparty (Controller) port-id). 

We want to ensure that a random relayer can not finish the handshake and choose a different address as the relayer address. @colin-axner

seantking: I guess we can do this by calling `GenerateAddress(portID)` here and checking if the return value matches the address in the passed in Version string.

modules/apps/27-interchain-accounts/keeper/handshake.go
seantking: `Contains` may be okay. I'm wondering though should we be more strict about this and have some kind of validate version string helper function to ensure the format we expect is being passed in? 

@colin-axner would like to hear your thoughts

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -20,9 +20,10 @@ const (
...
+func GenerateAddress(portID string) sdk.AccAddress {
```
seantking, line 24: Is it worth adding a test for this?

seantking:
Delighted to see this moving ahead quickly. I left a few preliminary thoughts. I'll review again later. :handshake:

modules/apps/27-interchain-accounts/types/account_test.go
```diff
@@ -44,27 +45,53 @@ func TestTypesTestSuite(t *testing.T) {
...
-		{"invalid connectionID", func() {
-			path.EndpointA.ConnectionID = "connection"
-		}, "", false},
-		{"invalid counterparty connectionID", func() {
-			path.EndpointB.ConnectionID = "connection"
-		}, "", false},
+		{
+			"success",
+			func() {},
+			"ics-27-0-0-testing",
```
seantking, line 60: Can change `testing` to be a mock address that resembles a bech32 address? Just for the sake of readability for newcomers to ICA.

modules/apps/27-interchain-accounts/keeper/keeper_test.go
```diff
@@ -36,7 +37,7 @@ func NewICAPath(chainA, chainB *ibctesting.TestChain) *ibctesting.Path {
 	path.EndpointA.ChannelConfig.Order = channeltypes.ORDERED
 	path.EndpointB.ChannelConfig.Order = channeltypes.ORDERED
 	path.EndpointA.ChannelConfig.Version = types.Version
-	path.EndpointB.ChannelConfig.Version = types.Version
+	path.EndpointB.ChannelConfig.Version = fmt.Sprintf("%s-%s", types.Version, types.GenerateAddress("ics-27-0-0-testing"))
```
seantking, line 40: nit: same here re `testing`. I think we could also maybe have a `const` for this.

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -85,22 +91,33 @@ func (k Keeper) OnChanOpenTry(
...
 	counterpartyVersion string,
 ) error {
-	if counterpartyVersion != types.Version {
+	if !strings.Contains(counterpartyVersion, types.Version) {
 		return sdkerrors.Wrapf(types.ErrInvalidVersion, "invalid counterparty version: %s, expected %s", counterpartyVersion, types.Version)
 	}
 
 	k.SetActiveChannel(ctx, portID, channelID)
 
+	accAddr := strings.TrimPrefix(counterpartyVersion, fmt.Sprintf("%s-", types.Version))
```
seantking, line 118: I think we should call some kind of `ValidateVersionString(types.Version)` fn here that does a validation check for the expected format of a version string depending on the types.Version.

seantking: Once we have validated that the format of the version string is what we expect, we can safely parse the accAddr.

modules/apps/27-interchain-accounts/types/account.go
damiannolan: Yeah for sure, I can add one with the other updates! Should be pretty basic

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: I think this would be a nicer option. Just basic assertion of `${types.Version}-${validateBech32addr}`

modules/apps/27-interchain-accounts/keeper/account.go
damiannolan: Will add an additional check here against `types.GenerateAddress(portID)` and include a test

modules/apps/27-interchain-accounts/keeper/account.go
colin-axner: I think the point is to do a no-op if the account already exists since it could create an attack vector if an error is returned. Or at least that's what comes to mind. It'd be great to have an inline comment explaining the reasoning

modules/apps/27-interchain-accounts/keeper/handshake.go
colin-axner: Yes I'd definitely like stronger validation. The two areas we do validation now is in [24-host](https://github.com/cosmos/ibc-go/blob/main/modules/core/24-host/validate.go) and [identifier parsing](https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/types/keys.go#L45)

I think the address validation should be light weight. I'm not sure it needs to be a valid bech 32 address since this can be address on a different chain. Maybe just a string which doesn't contain a `-`? It'd be nice if the interchain accounts mentioned this validation

I'm also okay doing some arbitrary address length validation and increasing the maximum allowed length if someone runs into an issue 

Adding a validation function also increases the readability of the code

modules/apps/27-interchain-accounts/types/keys.go
colin-axner: Lets remove them. They look outdated

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -6,22 +6,34 @@ const (
 	// ModuleName defines the Interchain Account module name
 	ModuleName = "interchainaccounts"
 
-	// Version defines the current version the IBC tranfer
+	// Version defines the current version the IBC interchainaccounts
 	// module supports
 	Version = "ics27-1"
 
+	// PortID is the default port id that the interchainaccounts module binds to
```
colin-axner, line 13: thanks for adding godoc :heart:

modules/apps/27-interchain-accounts/types/account.go
colin-axner: Nice! Deriving the address looks correct, but we probably need to create the module account in the ICA keeper (in NewKeeper function) and create the sub module account when the address is being generated. I may be mistaken, but typically you have to create a new account in the account keeper since it needs to track account information like sequence

It might be worth doing this in a separate pr. Trying to use this account should test whether everything was setup correctly

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -85,22 +91,33 @@ func (k Keeper) OnChanOpenTry(
 	}
 
 	// Register interchain account if it does not already exist
-	k.RegisterInterchainAccount(ctx, counterparty.PortId)
+	accAddr := strings.TrimPrefix(version, fmt.Sprintf("%s-", types.Version))
+	if err := k.RegisterInterchainAccount(ctx, accAddr, counterparty.PortId); err != nil {
+		return err
+	}
+
```
colin-axner, line 98: We shouldn't use relayer input for generating the account address. We should derive the expected address ourself and use that (while checking it matches that set in the version string)

colin-axner (commented):
Looking good!

modules/apps/27-interchain-accounts/keeper/handshake.go
seantking: exactly this :+1:

modules/apps/27-interchain-accounts/types/account.go
damiannolan: Sure I can do this in a separate PR 👍 

In the `NewKeeper` func for `transfer` I see assertion that the module account exists:

```

```

modules/apps/27-interchain-accounts/keeper/account.go
damiannolan: Is the idea here that there may be an account that already exists and this would return nil (no-op) in the event that a new channel was being established between chains that had already registered an interchain account?

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -6,22 +6,34 @@ const (
...
+	// QuerierRoute is the querier route for IBC interchainaccounts
+	QuerierRoute = ModuleName
+)
+
+var (
+	// ModuleAccountKey defines the key used to construct to the IBC interchainaccounts module account address
+	ModuleAccountKey = []byte(ModuleName)
+
 	// Key to store portID in our store
 	PortKey = "portID"
```
damiannolan, line 31: Should this align with the `transfer` module?

```
PortKey = []byte{0x01}
```

[See here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/keys.go#L36)

modules/apps/27-interchain-accounts/types/validate.go
```diff
@@ -0,0 +1,40 @@
...
+
+// DefaultMinAddrLength defines the default minimum character length used in validation of addresses
+var DefaultMinAddrLength = 32
+
+// IsValidAddr defines a regular expression to check if the provided string consists of
+// strictly alphanumeric characters
+var IsValidAddr = regexp.MustCompile("^[a-zA-Z0-9]*$").MatchString
+
+// ValidateVersion performs basic validation of the provided interchainaccounts version string
+func ValidateVersion(version string) error {
```
damiannolan, line 21: Added the following `ValidateVersion` and tried to maintain flexibility so it can be used for version strings in all handshake callbacks. i.e. supporting version strings where the address is *and* isn't included

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -1,27 +1,36 @@
...
+	// RouterKey is the message route for IBC interchainaccounts
+	RouterKey = ModuleName
 
+	// QuerierRoute is the querier route for IBC interchainaccounts
 	QuerierRoute = ModuleName
 
-	// MemStoreKey defines the in-memory store key
-	MemStoreKey = "mem_capability"
+	// Delimiter is the delimiter used for the interchainaccounts version string
+	Delimiter = "|"
```
damiannolan, line 28: I'm proposing we use a pipe `"|"` as the delimiter between ics-27 version and address as it makes for much simpler parsing/validation due to the string already being hyphenated

modules/apps/27-interchain-accounts/types/errors.go
```diff
@@ -11,9 +11,9 @@ var (
...
 	ErrInvalidOutgoingData         = sdkerrors.Register(ModuleName, 6, "invalid outgoing data")
 	ErrInvalidRoute                = sdkerrors.Register(ModuleName, 7, "invalid route")
-	ErrInterchainAccountNotFound   = sdkerrors.Register(ModuleName, 8, "Interchain Account not found")
-	ErrInterchainAccountAlreadySet = sdkerrors.Register(ModuleName, 9, "Interchain Account is already set")
+	ErrInterchainAccountNotFound   = sdkerrors.Register(ModuleName, 8, "interchain account not found")
+	ErrInterchainAccountAlreadySet = sdkerrors.Register(ModuleName, 9, "interchain account is already set")
 	ErrActiveChannelNotFound       = sdkerrors.Register(ModuleName, 10, "no active channel for this owner")
 	ErrInvalidVersion              = sdkerrors.Register(ModuleName, 11, "invalid interchain accounts version")
-	ErrInvalidOwnerAddress         = sdkerrors.Register(ModuleName, 12, "invalid owner address")
+	ErrInvalidAccountAddress       = sdkerrors.Register(ModuleName, 12, "invalid account address")
```
damiannolan, line 18: Updated this error, discussed with @seantking offline

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: Updated! [See here](https://github.com/cosmos/ibc-go/pull/410#discussion_r710132939)

modules/apps/27-interchain-accounts/types/keys.go
seantking: I don't see any problem with this. @colin-axner

modules/apps/27-interchain-accounts/types/keys.go
seantking: I was wondering this also tbh

modules/apps/27-interchain-accounts/types/validate.go
```diff
@@ -0,0 +1,40 @@
...
+	s := strings.Split(version, Delimiter)
+
```
seantking, line 23:
```suggestion
// When no delimiter is present compare against the expected version
```

modules/apps/27-interchain-accounts/types/validate.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"regexp"
+	"strings"
+
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// DefaultMaxAddrLength defines the default maximum character length used in validation of addresses
+var DefaultMaxAddrLength = 64
```
seantking, line 11: I think we maybe don't need to check for the length of the address here. We may want to leave this flexible for non cosmos sdk account addresses. wdyt @colin-axner

seantking: Is there an upper limit for Version string characters already?

modules/apps/27-interchain-accounts/types/validate.go
```diff
@@ -0,0 +1,40 @@
...
+
+// DefaultMinAddrLength defines the default minimum character length used in validation of addresses
+var DefaultMinAddrLength = 32
+
+// IsValidAddr defines a regular expression to check if the provided string consists of
+// strictly alphanumeric characters
+var IsValidAddr = regexp.MustCompile("^[a-zA-Z0-9]*$").MatchString
+
+// ValidateVersion performs basic validation of the provided interchainaccounts version string
```
seantking, line 20:
```suggestion
// ValidateVersion performs basic validation of the provided ics27 version string 
```

modules/apps/27-interchain-accounts/types/validate.go
seantking: LGTM :+1:

modules/apps/27-interchain-accounts/types/validate_test.go
```diff
@@ -0,0 +1,53 @@
...
+			"ics27-5|abc123",
+			false,
+		},
+		{
+			"invalid account address - 31 chars",
+			"ics27-1|xtignpvthxbwxtmnzyfwhhywobaatlt",
+			false,
+		},
+		{
+			"invalid account address - 65 chars",
```
seantking, line 31: I think we can probably do without this check

modules/apps/27-interchain-accounts/types/validate.go
damiannolan: Good call 👍 Will add this above as part of the func godoc

modules/apps/27-interchain-accounts/types/validate.go
damiannolan: I picked these numbers completely arbitrarily. I think it would be good to still enforce some kind of max length here. If 64 isn't enough we can increase to 128/256

modules/apps/27-interchain-accounts/types/validate_test.go
```diff
@@ -0,0 +1,58 @@
+package types_test
+
+import (
+	"fmt"
+
+	"github.com/cosmos/ibc-go/modules/apps/27-interchain-accounts/types"
+)
+
+func (suite *TypesTestSuite) TestValidateVersion() {
```
colin-axner, line 9: want to add a test case for owner addresses which use the Delimiter, ie len(s) > 2?

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -84,23 +89,37 @@ func (k Keeper) OnChanOpenTry(
 		return err
 	}
 
+	accAddr := types.GenerateAddress(counterparty.PortId)
+	parsedAddr := types.ParseAddressFromVersion(version)
+	if strings.Compare(parsedAddr, accAddr.String()) != 0 {
```
colin-axner, line 94: interesting, how does `strings.Compare` differ from:
```go
if parsedAddr != accAddr.String() {
```

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -84,23 +89,37 @@ func (k Keeper) OnChanOpenTry(
...
+		return sdkerrors.Wrapf(types.ErrInvalidAccountAddress, "invalid account address: %s, expected %s", parsedAddr, accAddr)
```
colin-axner, line 95:
```suggestion
		return sdkerrors.Wrapf(types.ErrInvalidAccountAddress, "version contains invalid account address: expected %s, got %s", parsedAddr, accAddr)
```

colin-axner (commented):
Great work! Left various nits/suggestions

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -20,9 +19,14 @@ const (
 	ICAPrefix string = "ics-27"
```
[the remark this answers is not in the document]
seantking: Agree

modules/apps/27-interchain-accounts/types/validate_test.go
seantking: Fine with me

modules/apps/27-interchain-accounts/types/validate.go
seantking: Makes sense then :+1:

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: > Compare is included only for symmetry with package bytes. It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on.

Originally I was going to use `bytes.Compare` with both `sdk.AccAddress`'s and ended up switching to `strings.Compare`.
I will update actually, I think the "built-in" comparison is a better option.

modules/apps/27-interchain-accounts/types/account.go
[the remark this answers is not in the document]
damiannolan: Agreed, found that confusing at times myself!

modules/apps/27-interchain-accounts/types/keys.go
damiannolan: I think we should opt for consistency either way! I'll update this to align with `transfer` for now. We can revisit and discuss during the internal audit

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -20,9 +19,14 @@ const (
...
-func GenerateAddress(port string) []byte {
-	return tmhash.SumTruncated([]byte(port))
+// GenerateAddress returns an sdk.AccAddress using the provided port identifier
+func GenerateAddress(portID string) sdk.AccAddress {
+	return sdk.AccAddress(tmhash.SumTruncated([]byte(portID)))
+}
+
+// ParseAddressFromVersion trims the interchainaccounts version prefix and returns the associated account address
+func ParseAddressFromVersion(version string) string {
+	return strings.TrimPrefix(version, fmt.Sprint(Version, Delimiter))
```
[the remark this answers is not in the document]
damiannolan: Yeah I can change to `VersionPrefix`. But I agree, it's confusing. Wondering if perhaps a protobuf struct would've been a better option completely... idk something like below where a metadata field could be used for any arbitrary piece of data, in our case an account address string:

```proto
message VersionMetadata {
    string version = 1;
    string metadata = 2;
}
```

The real problem here is with breaking existing APIs though I believe!

modules/apps/27-interchain-accounts/types/account.go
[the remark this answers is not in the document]
colin-axner: Lets do it in a followup, that way we can ensure the spec is fixed as well #423

modules/apps/27-interchain-accounts/types/account.go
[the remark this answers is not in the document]
colin-axner: We can do protobuf, we would just need to encode/decode when passing and interacting with core IBC. The version in the channel is intentionally a string to allow applications the flexibility of deciding what sort of version they want

They main downside of encoding is that the version might look odd when printing out a channel struct, but if we json encode maybe it'd be fine?

modules/apps/27-interchain-accounts/module.go
```diff
@@ -255,6 +255,18 @@ func (am AppModule) OnTimeoutPacket(
...
-func (am AppModule) NegotiateAppVersion(ctx sdk.Context, order channeltypes.Order, connectionID, portID string, counterparty channeltypes.Counterparty, proposedVersion string) (string, error) {
-	return "", nil
+// NegotiateAppVersion implements the IBCModule interface
+func (am AppModule) NegotiateAppVersion(
+	ctx sdk.Context,
+	order channeltypes.Order,
+	connectionID string,
+	portID string,
+	counterparty channeltypes.Counterparty,
+	proposedVersion string,
```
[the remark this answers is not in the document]
damiannolan: Me too! My only argument is that seeing three ignored args with `_` in a vertical line is somewhat unaesthetic... even if they're used 😅 

I'll leave this decision to you! I don't really mind either way

modules/apps/27-interchain-accounts/types/validate.go
```diff
@@ -0,0 +1,41 @@
...
+// ValidateVersion performs basic validation of the provided ics27 version string
+// When no delimiter is present it compares against the expected version
+func ValidateVersion(version string) error {
+	s := strings.Split(version, Delimiter)
+
+	if s[0] != Version {
+		return sdkerrors.Wrapf(ErrInvalidVersion, "invalid version: expected %s, got %s", Version, s[0])
+	}
+
+	if len(s) > 1 {
```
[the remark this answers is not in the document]
damiannolan: Added!

modules/apps/27-interchain-accounts/types/validate_test.go
damiannolan: Done!

modules/apps/27-interchain-accounts/types/account.go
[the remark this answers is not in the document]
damiannolan: So with this approach, the version/counterpartyVersion args to the IBCModule callbacks would remain as strings but we would json encode the structure for ICA and handle encoding within the keeper? 

Let's keep this as an option and we can discuss as part of the audit. If it's something we decide is valuable or a better approach I can implement it later.

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -71,11 +72,13 @@ func (k Keeper) OnChanOpenTry(
...
-		return sdkerrors.Wrapf(types.ErrInvalidVersion, "got: %s, expected %s", version, types.Version)
+
+	if err := types.ValidateVersion(version); err != nil {
+		return err
 	}
-	if counterpartyVersion != types.Version {
-		return sdkerrors.Wrapf(types.ErrInvalidVersion, "invalid counterparty version: %s, expected %s", counterpartyVersion, types.Version)
+
+	if err := types.ValidateVersion(counterpartyVersion); err != nil {
+		return err
```
colin-axner, line 81: This error message should probably indicate that the `counterpartyVersion` failed, same with regular version above. Otherwise we won't be able to distinguish which version was incorrect from a quick glance

colin-axner (approved):
LGTM! Left a couple suggestions. Great work once again!

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -33,8 +33,8 @@ var (
 )
 
 // NewVersion returns a complete version string in the format: VersionPrefix + Delimter + AccAddress
-func NewAppVersion(addr string) string {
-	return fmt.Sprint(VersionPrefix, Delimiter, addr)
+func NewAppVersion(versionPrefix, delimiter, addr string) string {
+	return fmt.Sprint(versionPrefix, delimiter, addr)
```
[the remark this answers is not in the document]
damiannolan: Yep makes sense! I've updated to omit delimiter arg

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: Done!

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -84,23 +87,37 @@ func (k Keeper) OnChanOpenTry(
 		return err
 	}
 
+	accAddr := types.GenerateAddress(counterparty.PortId)
```
seantking, line 90:
```suggestion
        // Check to ensure that the version string contains the expected address generated from the Counterparty portID 
	accAddr := types.GenerateAddress(counterparty.PortId)
```

modules/apps/27-interchain-accounts/types/validate.go
```diff
@@ -0,0 +1,41 @@
...
+
+// DefaultMaxAddrLength defines the default maximum character length used in validation of addresses
+var DefaultMaxAddrLength = 128
+
+// IsValidAddr defines a regular expression to check if the provided string consists of
+// strictly alphanumeric characters
+var IsValidAddr = regexp.MustCompile("^[a-zA-Z0-9]*$").MatchString
+
+// ValidateVersion performs basic validation of the provided ics27 version string
+// When no delimiter is present it compares against the expected version
```
seantking, line 18: I added this comment for line 21 I think rather than the header.

modules/apps/27-interchain-accounts/types/validate.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// DefaultMaxAddrLength defines the default maximum character length used in validation of addresses
+var DefaultMaxAddrLength = 128
+
+// IsValidAddr defines a regular expression to check if the provided string consists of
+// strictly alphanumeric characters
+var IsValidAddr = regexp.MustCompile("^[a-zA-Z0-9]*$").MatchString
+
+// ValidateVersion performs basic validation of the provided ics27 version string
```
seantking, line 17: Maybe put a TODO: for adding a link to the spec (once the spec id is updated?)

seantking: nite: It might be worth adding a very brief comment explaining what we're validating here. I almost forgot for a second myself :laughing:

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -84,23 +87,37 @@ func (k Keeper) OnChanOpenTry(
...
+// Controller Chain
 func (k Keeper) OnChanOpenAck(
 	ctx sdk.Context,
 	portID,
 	channelID string,
 	counterpartyVersion string,
 ) error {
-	if counterpartyVersion != types.Version {
-		return sdkerrors.Wrapf(types.ErrInvalidVersion, "invalid counterparty version: %s, expected %s", counterpartyVersion, types.Version)
+	if err := types.ValidateVersion(counterpartyVersion); err != nil {
```
seantking, line 112: @colin-axner I'm just wondering is there any way that a malicious relayer could pass in an address here to the ack step that isn't the address we validated on the `Try` step? I think we already discussed this but just refreshing my brain after the weekend.

seantking (approved):
Approved (a couple of nits). I would like to resolve the question here before merging -> https://github.com/cosmos/ibc-go/pull/410/files#r712174482

Great work. Thank you! :handshake:

modules/apps/27-interchain-accounts/types/validate.go
damiannolan: Done!

modules/apps/27-interchain-accounts/types/validate.go
damiannolan: I think it's probably more appropriate to include it as part of the godoc rather than inline. [See since updated](https://github.com/cosmos/ibc-go/pull/410/files#diff-094fd8ce1e0b97d12d88f0711bf8bf51c1ad6022dd68b95db8b3b9830bfad99dR17)

modules/apps/27-interchain-accounts/keeper/handshake.go
colin-axner: No, core IBC verifies a proof of the channel state set on the [counterparty](https://github.com/cosmos/ibc-go/blob/2d3181bf37239325002bb6a496a2115b6e88a615/modules/core/04-channel/keeper/handshake.go#L308)

If a relayer changed the version, the proof would fail

seantking: thanks :pray:

modules/apps/27-interchain-accounts/types/validate.go
seantking: great point!

Issue #313, opened 2021-08-05
Implement ICA as middleware on controller side


colin-axner:
## Summary

The controller side of the interchain accounts module is middleware which can be connected to a helper module which might construct the messages to be sent to the counterparty chain

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Starting to look into this. I'm thinking about creating a [router](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/types/router.go) for ICA. I'd like to route to different authentication (helper) modules based on the owner address. The alternative solution is initializing an interchain account module per authentication module

So lets say we have 3 authentication modules:
- gov
- experimental auth 1
- experimental auth 2 (imagine these are zksnarks or some new cryptography stuff)

So in the current design, one might initialize a new interchain accounts module per auth module
```go
NewICAModule(gov)
NewICAModule(auth 1)
NewICAModule(auth 2)
```

This is fine (kinda), but I think we will run into issues. Like which module binds to "interchain accounts" port and what do the others bind to?

Instead I'd like to add a router, so
```go
router := NewRouter()
router.AddRoute("gov", gov)
router.AddRoute("auth 1", auth 1)
router.AddRoute("auth 2", auth 2)
NewICAModule(router)
```

The only problem is then we probably want to map, owner address -> auth module name

Thoughts @seantking ?

seantking:
```
NewICAModule(gov)
NewICAModule(auth 1)
NewICAModule(auth 2)
```

When you say this is the "current design", do you mean the idea you have for middleware implementation or the current ics27 spec? 

Honestly, the more I think about the middleware implementation of ICA, I wonder how much of it is nice to have vs something we should do as a priority.  

The primary upside of having the middleware implementation is the ability to remove the "hooks" design that currently exists, right? 

I think I would be fine with leaving this middleware update of ICA to the very end and deciding then how much value the improvements will add based on us integrating into an app/testing. If only because it's going to deviate from the current design quite a bit and I'd prefer to reach parity with the current spec first (we will need to update the spec + sync with informal on the changes etc). Also, we will probably have a much clearer idea of what we expect from middleware once the 29-fee is being used. 

I am open to opinions on this but I wonder if we can push the current design across the finish line and then begin working on DUX improvements as necessary?

colin-axner:
[...]

PR #417, opened 2021-09-16
ICA Controller Side Middleware

colin-axner:
## Description

closes: #313 

---

write a little note why.


modules/apps/27-interchain-accounts/module.go
```diff
@@ -74,7 +75,9 @@ func (AppModuleBasic) RegisterGRPCGatewayRoutes(clientCtx client.Context, mux *r
 
 type AppModule struct {
 	AppModuleBasic
-	keeper keeper.Keeper
+	keeper       keeper.Keeper
+	scopedKeeper capabilitykeeper.ScopedKeeper
+	app          porttypes.IBCModule
```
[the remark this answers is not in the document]
seantking: You mean the ICA middleware will take a single app rather than the router, right?

seantking:
Nice! Excited to see what this looks like. I'll review properly in the morning :)

modules/apps/27-interchain-accounts/module.go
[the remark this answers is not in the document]
colin-axner: yup, exactly

modules/apps/27-interchain-accounts/module.go
```diff
@@ -147,9 +159,24 @@ func (am AppModule) OnChanOpenInit(
...
+	}
+
+	// issue capability for connected authentication module
+	appCap, err := am.scopedKeeper.NewCapability(ctx, types.AppCapabilityName(portID, channelID))
+	if err != nil {
+		return sdkerrors.Wrap(err, "could not create capability for underlying application")
+	}
+
+	// call underlying app's OnChanOpenInit callback with the appVersion
+	return am.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID,
```
seantking, line 173: Question: If we go with the middleware approach, does it mean that every module that uses interchain accounts (helper/auth module) will have to be an IBC application (i.e. has the callbacks setup)? Currently, the helper modules aren't necessarily IBC applications.

colin-axner: The helper modules will need to implement the IBC App module interface, but they can choose to do no custom logic for these functions

I would consider all helper modules IBC apps even if they aren't directly so, they are causing an IBC packet to be sent and performing logic upon its acknowledgement or timeout

I've been referring to helper modules as authentication modules. Do you have a preferred term, I find helper to be a little vague

seantking: Yeah, I dislike the term `helper` module a lot. Authentication module sounds good for the time being :+1: 

Good point about doing something on ack/timeout for the packets (the hooks approach previously was a little convoluted). So basically when we go with this approach (ica middleware + auth module implementing IBC interface) each ica-auth module can implement custom logic on ack/timeout via the standard IBC App interface?

colin-axner: > So basically when we go with this approach (ica middleware + auth module implementing IBC interface) each ica-auth module can implement custom logic on ack/timeout via the standard IBC App interface?

yup, custom logic can be implemented on:
- channel handshake init and ack
- send packet
- ack
- timeout

(controller side logic only)

modules/apps/27-interchain-accounts/ibc_module.go
```diff
@@ -116,36 +145,25 @@ func (im IBCModule) OnRecvPacket(
...
 ) error {
-	var ack channeltypes.Acknowledgement
-
-	if err := types.ModuleCdc.UnmarshalJSON(acknowledgement, &ack); err != nil {
-		return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal ICS-27 interchain account packet acknowledgment: %v", err)
-	}
-	var data types.IBCAccountPacketData
-	if err := types.ModuleCdc.UnmarshalJSON(packet.GetData(), &data); err != nil {
-		return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "cannot unmarshal ICS-27 interchain account packet data: %s", err.Error())
-	}
```
colin-axner, line 133: do we need this?

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -45,7 +45,20 @@ func (suite *InterchainAccountsTestSuite) SetupTest() {
...
+	// instead of attempting to claim channel capability
+	mockModuleA := suite.chainA.GetSimApp().GetMockModule()
+	mockModuleB := suite.chainB.GetSimApp().GetMockModule()
+	onChanOpenInit := func(ctx sdk.Context, order channeltypes.Order, connectionHops []string,
+		portID, channelID string, chanCap *capabilitytypes.Capability,
+		counterparty channeltypes.Counterparty, version string,
+	) error {
+		// do not claim channel capability
+		return nil
+	}
```
colin-axner, line 59: I'm not sure what the desired functionality is here (for OnChanOpenInit)

Should the controller chain code (the ICA module) claim the channel capability or let the authentication module do that? Should the controller chain code issue an app capability? @AdityaSripal

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -195,7 +208,6 @@ func (suite *InterchainAccountsTestSuite) TestOnChanOpenAck() {
 }
 
 func (suite *InterchainAccountsTestSuite) TestOnChanOpenConfirm() {
-	suite.SetupTest() // reset
```
colin-axner, line 198: suite.SetupTest gets called before every `Test...` function. We need it when doing a for loop on test cases

testing/mock/mock.go
```diff
@@ -94,14 +94,15 @@ type AppModule struct {
 	AppModuleBasic
 	scopedKeeper capabilitykeeper.ScopedKeeper
 	portKeeper   PortKeeper
-	IBCApp       MockIBCApp // base application of an IBC middleware stack
+	IBCApp       *MockIBCApp // base application of an IBC middleware stack
```
colin-axner, line 97: whoops, will open a pr to main

modules/apps/27-interchain-accounts/ibc_module_test.go
AdityaSripal: Since the ICA-Auth is sitting underneath ICA. i.e. ICA-Auth is the base application. You don't even need to claim the capability imo. Just pass it through to the underlying IBC-Auth.

They're both sharing the same scopedKeeper anyway

modules/apps/27-interchain-accounts/ibc_module.go
```diff
@@ -122,35 +145,22 @@ func (im IBCModule) OnRecvPacket(
...
-
-	if err := im.keeper.OnAcknowledgementPacket(ctx, packet, data, ack); err != nil {
-		return err
-	}
-
-	return nil
+	// call underlying app's OnAcknowledgementPacket callback.
+	return im.app.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
```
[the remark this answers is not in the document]
damiannolan: If the bytes were malformed would it fail before it reaches here due to the `VerifyPacketAcknowledgement` step?
Otherwise the underlying app could just return an error if it fails to unmarshal. This is totally fine imo!

modules/apps/27-interchain-accounts/keeper/relay.go
```diff
@@ -3,16 +3,16 @@ package keeper
...
 	"github.com/cosmos/ibc-go/v2/modules/apps/27-interchain-accounts/types"
 	clienttypes "github.com/cosmos/ibc-go/v2/modules/core/02-client/types"
 	channeltypes "github.com/cosmos/ibc-go/v2/modules/core/04-channel/types"
-	host "github.com/cosmos/ibc-go/v2/modules/core/24-host"
 )
 
-// TODO: implement middleware functionality, this will allow us to use capabilities to
-// manage helper module access to owner addresses they do not have capabilities for
-func (k Keeper) TrySendTx(ctx sdk.Context, portID string, icaPacketData types.InterchainAccountPacketData) (uint64, error) {
+// TrySendTx takes in a transaction from a base application and attempts to send the packet
```
seantking, line 13: Can we change the terminology for `base application`? Should we just say `authentication module` or a variant of that?

modules/apps/27-interchain-accounts/types/hook.go
```diff
@@ -1,30 +0,0 @@
-package types
-
-import sdk "github.com/cosmos/cosmos-sdk/types"
```
seantking, line 3: so happy to see this file deleted :pray:

testing/simapp/app.go
```diff
@@ -332,22 +334,24 @@ func NewSimApp(
...
 		app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
 		app.AccountKeeper, scopedICAKeeper, app.MsgServiceRouter(),
 	)
 	icaModule := ica.NewAppModule(app.ICAKeeper)
-	icaIBCModule := ica.NewIBCModule(app.ICAKeeper, nil)
-
-	// NOTE: the IBC mock keeper and application module is used only for testing core IBC. Do
-	// note replicate if you do not need to test core IBC or light clients.
-	mockModule := ibcmock.NewAppModule(scopedIBCMockKeeper, &app.IBCKeeper.PortKeeper)
+	icaIBCModule := ica.NewIBCModule(app.ICAKeeper, icaMockModule)
```
seantking, line 349: Just thinking out loud but is there any quick way to initiate a mock ica-auth (base) module here so we can ensure the integration works with the middleware?

modules/apps/27-interchain-accounts/ibc_module.go
```diff
@@ -40,10 +46,18 @@ func (im IBCModule) OnChanOpenInit(
 	counterparty channeltypes.Counterparty,
 	version string,
 ) error {
-	return im.keeper.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version)
+	if err := im.keeper.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version); err != nil {
```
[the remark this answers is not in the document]
colin-axner: I'm really confused, this is already being [enforced](https://github.com/cosmos/ibc-go/blob/bd10403d679df1a7c23aeea1b676b6711a4112e5/modules/apps/27-interchain-accounts/keeper/handshake.go#L24) by the checks in OnChanOpenInit

testing/simapp/app.go
colin-axner: I could give it a shot, we would need to add functions to ica-auth

testing/simapp/app.go
```diff
@@ -267,6 +268,7 @@ func NewSimApp(
 	// NOTE: the IBC mock keeper and application module is used only for testing core IBC. Do
 	// note replicate if you do not need to test core IBC or light clients.
 	scopedIBCMockKeeper := app.CapabilityKeeper.ScopeToModule(ibcmock.ModuleName)
+	scopedICAMockKeeper := app.CapabilityKeeper.ScopeToModule(ibcmock.ModuleName + icatypes.ModuleName)
```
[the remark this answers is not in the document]
colin-axner: If you don't use different names they get routed to the same scoped keeper/code. Does that make sense? The [docs](https://github.com/cosmos/ibc-go/blob/main/docs/ibc/middleware/integration.md#example-integration) should explain this and we should open an issue if the current explanation isn't sufficient. cc @AdityaSripal

modules/apps/27-interchain-accounts/ibc_module.go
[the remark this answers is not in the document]
colin-axner: Correct!

modules/apps/27-interchain-accounts/ibc_module.go
```diff
@@ -14,7 +14,7 @@ import (
 	ibcexported "github.com/cosmos/ibc-go/v2/modules/core/exported"
 )
 
-// IBCModule implements the ICS26 callbacks for interchain accounts given the
+// IBCModule implements the ICS26 interface for interchain accounts given the
 // interchain account keeper and underlying application.
 type IBCModule struct {
```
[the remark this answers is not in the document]
damiannolan: Opened #516

modules/apps/27-interchain-accounts/module.go
```diff
@@ -24,6 +24,11 @@ var (
 	_ module.AppModule      = AppModule{}
 	_ porttypes.IBCModule   = IBCModule{}
 	_ module.AppModuleBasic = AppModuleBasic{}
+
+	// Middleware must implement types.ChannelKeeper and types.PortKeeper expected interfaces
+	// so that it can wrap IBC channel and port logic for underlying application.
+//	_ types.ChannelKeeper = Keeper{}
```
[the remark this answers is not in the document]
colin-axner: I removed these assertions. ICA is only middleware for ICA authentication modules which must rely on `TrySendTx` function, not `SendPacket` and `WriteAcknowledgement`. The middleware interaction is more restricted than on 29-fee which is very open and thus must fulfill the entire interface

cc @AdityaSripal

testing/simapp/app.go
```diff
@@ -196,6 +196,11 @@ type SimApp struct {
 	ScopedTransferKeeper capabilitykeeper.ScopedKeeper
 	ScopedICAKeeper      capabilitykeeper.ScopedKeeper
 	ScopedIBCMockKeeper  capabilitykeeper.ScopedKeeper
+	ScopedICAMockKeeper  capabilitykeeper.ScopedKeeper
+
+	// make IBC modules public for test purposes
+	// these modules are never directly routed to by the IBC Router
+	ICAAuthModule ibcmock.IBCModule
```
colin-axner, line 203: IBC routes to the top level IBC module of a middleware stack. In this case it would be the ica IBCModule. The ica IBCModule contains this ica auth module as an unexported variable. Thus to access it during testing I made it a public field of the SimApp. This will need to be done for all base middleware applications which cannot standalone (like transfer)

testing/simapp/app.go
colin-axner: ibc_module_test does testing of the ica auth module callbacks. What it does not do is test the flow of calling `TrySendTx`, but I don't really see this as necessary since we already test claiming a capability with an ica auth module correctly routes to the middleware and then the auth module

modules/apps/27-interchain-accounts/ibc_module.go
```diff
@@ -29,7 +29,13 @@ func NewIBCModule(k keeper.Keeper, app porttypes.IBCModule) IBCModule {
 	}
 }
 
-// OnChanOpenInit implements the IBCModule interface
+// OnChanOpenInit implements the IBCModule interface. Interchain Accounts is
+// implemented to act as middleware for connected authentication modules on
+// the controller side. The connected modules may not change the portID or
```
seantking, line 34:
```suggestion
// the controller side. The connected modules may not change the controller side portID or
```

seantking:
Nice work! 

From my understanding, the primary benefit of making ICA middleware on the controller side is to replace the previously existing `hooks` functionality which allowed for custom logic on timeout/recv/ack effectively. 

I'm thinking it would be good to have some tests using the mock module infrastructure you created and test out some custom logic for each of these steps to ensure the middleware is working as intended.

wdyt?

testing/simapp/app.go
seantking: :handshake:  We should make sure to note this in the developer docs. I'll make a note

testing/simapp/app.go
```diff
@@ -267,6 +272,7 @@ func NewSimApp(
 	// NOTE: the IBC mock keeper and application module is used only for testing core IBC. Do
 	// note replicate if you do not need to test core IBC or light clients.
```
seantking, line 273:
```suggestion
	// not replicate if you do not need to test core IBC or light clients.
```

modules/apps/27-interchain-accounts/ibc_module.go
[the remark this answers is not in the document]
seantking: I am confused by this statement also :) Is there any special reason for this? @AdityaSripal

modules/apps/27-interchain-accounts/keeper/relay.go
```diff
@@ -3,16 +3,16 @@ package keeper
...
 	"github.com/cosmos/ibc-go/v2/modules/apps/27-interchain-accounts/types"
 	clienttypes "github.com/cosmos/ibc-go/v2/modules/core/02-client/types"
 	channeltypes "github.com/cosmos/ibc-go/v2/modules/core/04-channel/types"
-	host "github.com/cosmos/ibc-go/v2/modules/core/24-host"
 )
 
-// TODO: implement middleware functionality, this will allow us to use capabilities to
-// manage helper module access to owner addresses they do not have capabilities for
-func (k Keeper) TrySendTx(ctx sdk.Context, portID string, icaPacketData types.InterchainAccountPacketData) (uint64, error) {
+// TrySendTx takes in a transaction from a authentication module and attempts to send the packet
```
damiannolan, line 13:
```suggestion
// TrySendTx takes in a transaction from an authentication module and attempts to send the packet
```

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -0,0 +1,421 @@
...
+
+			tc.malleate()
+
+			// ensure channel on chainA is set in state
+			suite.chainA.GetSimApp().IBCKeeper.ChannelKeeper.SetChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, *channel)
+
+			module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID)
+			suite.Require().NoError(err)
+
+			chanCap, err := suite.chainA.App.GetScopedIBCKeeper().NewCapability(suite.chainA.GetContext(), host.ChannelCapabilityPath(ibctesting.TransferPort, path.EndpointA.ChannelID))
```
seantking, line 168: question: why do we pass the `ibctesting.TransferPort` to `host.ChannelCapabilityPath` ?

testing/simapp/app.go
```diff
@@ -333,24 +339,31 @@ func NewSimApp(
 	transferModule := transfer.NewAppModule(app.TransferKeeper)
 	transferIBCModule := transfer.NewIBCModule(app.TransferKeeper)
 
+	// NOTE: the IBC mock keeper and application module is used only for testing core IBC. Do
+	// note replicate if you do not need to test core IBC or light clients.
```
damiannolan, line 343: Same typo as above! Copy pasta 🍝 

```suggestion
	// not replicate if you do not need to test core IBC or light clients.
```

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -0,0 +1,421 @@
...
+			"success", func() {}, true,
+		},
+		{
+			"ICA OnChanOpenInit fails - UNORDERED channel", func() {
+				channel.Ordering = channeltypes.UNORDERED
+			}, false,
+		},
+		{
+			"ICA auth module callback fails", func() {
+				suite.chainA.GetSimApp().ICAAuthModule.IBCApp.OnChanOpenInit = func(ctx sdk.Context, order channeltypes.Order, connectionHops []string,
```
seantking, line 123: This is a nice way of testing the integration. I like it! :handshake:

testing/simapp/app.go
seantking: Yeah, that's a good point. I didn't see `ibc_module_test` on my first review, github collapsed the file so I missed it. Tests look great :+1: 

I think it might be nice to have the `onRecv/timeout/ack` tests which essentially test the previous hook functionality but we can follow up with another PR :handshake:

modules/apps/27-interchain-accounts/ibc_module_test.go
colin-axner: This is incorrect, it should be the newly binded port. Will fix!

damiannolan (approved):
Super clean, nice work!

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -0,0 +1,421 @@
...
+					portID, channelID string, chanCap *capabilitytypes.Capability,
+					counterparty channeltypes.Counterparty, version string,
+				) error {
+					return fmt.Errorf("mock ica auth fails")
+				}
+			}, false,
```
AdityaSripal, line 132: test handshake in wrong direction fails. ie INIT from host chain

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -0,0 +1,421 @@
...
+					return fmt.Errorf("mock ica auth fails")
+				}
+
+			}, true,
+		},
+		{
+			"ICA callback fails - invalid version", func() {
+				channel.Version = types.VersionPrefix
+			}, false,
+		},
```
AdityaSripal, line 221: test wrong flow. Controller chain calls TRY

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -0,0 +1,421 @@
...
+		{
+			"ICA auth module callback fails", func() {
+				suite.chainA.GetSimApp().ICAAuthModule.IBCApp.OnChanOpenAck = func(
+					ctx sdk.Context, portID, channelID string, counterpartyVersion string,
+				) error {
+					return fmt.Errorf("mock ica auth fails")
+				}
+			}, false,
+		},
```
AdityaSripal, line 301: ditto

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -0,0 +1,421 @@
...
+		{
+			"success: ICA auth module callback returns error", func() {
+				// mock module callback should not be called on host side
+				suite.chainB.GetSimApp().ICAAuthModule.IBCApp.OnChanOpenConfirm = func(
+					ctx sdk.Context, portID, channelID string,
+				) error {
+					return fmt.Errorf("mock ica auth fails")
+				}
+
+			}, true,
```
AdityaSripal, line 359: ditto

modules/apps/27-interchain-accounts/ibc_module_test.go
```diff
@@ -0,0 +1,421 @@
...
+
+	counterpartyPortID, err := types.GeneratePortID(TestOwnerAddress, path.EndpointA.ConnectionID, path.EndpointB.ConnectionID)
+	suite.Require().NoError(err)
+
+	counterparty := &channeltypes.Counterparty{
+		PortId:    counterpartyPortID,
+		ChannelId: path.EndpointB.ChannelID,
+	}
+
+	version, err := cbs.NegotiateAppVersion(suite.chainA.GetContext(), channeltypes.ORDERED, path.EndpointA.ConnectionID, types.PortID, *counterparty, types.VersionPrefix)
```
AdityaSripal, line 504: Is there no error condition to test?

modules/apps/27-interchain-accounts/ibc_module.go
```diff
@@ -1,20 +1,17 @@
 package interchain_accounts
```
damiannolan, line 1: I think we agreed on slack msgs that we'd rename this to `package ica`, right? 
Doesn't necessarily need to be included here, I can do it as part of the pkg split if you want. Just a mental reminder :)

testing/simapp/app.go
```diff
@@ -333,24 +339,31 @@ func NewSimApp(
...
 
-	// NOTE: the IBC mock keeper and application module is used only for testing core IBC. Do
-	// note replicate if you do not need to test core IBC or light clients.
-	mockModule := ibcmock.NewAppModule(scopedIBCMockKeeper, &app.IBCKeeper.PortKeeper)
-	mockIBCModule := ibcmock.NewIBCModule(&ibcmock.MockIBCApp{}, scopedIBCMockKeeper)
+	// initialize ICA module with mock module as the authentication module on the controller side
+	icaAuthModule := ibcmock.NewIBCModule(&ibcmock.MockIBCApp{}, scopedICAMockKeeper)
+	app.ICAAuthModule = icaAuthModule
+
+	icaIBCModule := ica.NewIBCModule(app.ICAKeeper, icaAuthModule)
```
AdityaSripal, line 359: Shouldn't we make the auth module argument optional so we can instantiate an ICA module without enabling the controlling side.

In general, we should make it possible to enable one side of ICA without the other.

AdityaSripal (approved):
Some requests on tests, but nothing crucial. Great work!!

modules/apps/27-interchain-accounts/ibc_module.go
colin-axner: add to the controller/host split issue?

damiannolan: Will do!

testing/simapp/app.go
colin-axner: I don't see a reason to make this optional in our simapp. We should make enabling/disabling one side optional, but I think this will be easier to do after the controller/host split. It should also allow ICA auth modules to be hooked up without being enabled (so gov can enable them dynamically)

modules/apps/27-interchain-accounts/ibc_module_test.go
colin-axner: I agree with the intention, but I think this is out of scope of this pr and I don't think these tests are the place to test this functionality. This pr is replacing the existing hook mechanism with controller side middleware. Testing that the handshake has the correct checks in place to reject the wrong flow isn't what this pr aims to do. Those tests are missing, but the scope of this pr shouldn't be increased to fix that issue

`keeper/handshake_test.go` should test that we have the right checks in place, but we can still add higher level tests. Trying to test wrong flow in these test cases isn't possible, I'd have to add a completely new test (as I'd need to reconstruct the arguments being passed into the callback to make sense - otherwise it'd trivially fail). 

I've opened #522

testing/simapp/app.go
AdityaSripal: agreed, im just saying that for a chain that doesn't want to be a controller chain at all. Why should it even pass in an IBC-Auth module just so it can disable it. I'd like to provide a constructor that doesn't require an auth module.

Doesn't need to be used here, i just think that constructor should be provided

seantking: Great catch @AdityaSripal. I agree that we should be making this optional. There will for sure be chains that don't want to be a controller chain. If it makes sense to do this in a follow-up PR fine with me but if we can make a clear ticket I'd appreciate it :handshake:

modules/apps/27-interchain-accounts/ibc_module_test.go
seantking: I think it's fine to do a follow-up PR :handshake:

seantking (approved):
:handshake:

testing/simapp/app.go
colin-axner: I'm a little confused. A host chain could use:
```go
    icaIBCModule := ica.NewIBCModule(app.ICAKeeper, nil)
```

colin-axner: #523

Issue #416, opened 2021-09-16
ICA: OnChanCloseInit should return error


colin-axner:
## Summary

The current code allows users to close channels. The interchain accounts callback should return an error

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

seantking:
Users are blocked from closing channels, see: https://github.com/cosmos/ibc-go/blob/interchain-accounts/modules/apps/27-interchain-accounts/module.go#L184

I left some commented-out code in `handshake.go` that be confusing.

colin-axner:
Returning nil allows users to close channels

A user sends `CloseChanInit` msg, if the app callback doesn't return an error, core IBC will close the channel. The comment is incorrect

colin-axner:
The close confirm is also incorrect, if the counterparty closes the channel, we should probably allow the channel to be closed

seantking:
Gotcha. Good catch.  :+1: Updating this as part of a PR here - > https://github.com/cosmos/ibc-go/pull/420/commits/61eae324093b1d9edba2878c9aa8182eb2bcd43e

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #420, opened 2021-09-17
test: account/keeper tests for ICA

seantking:
## Description

Doing some housecleaning on the keeper tests + adding tests for the `InterchainAccount` type. 

closes: #416 
partially closes: #279

---

write a little note why.


modules/apps/27-interchain-accounts/keeper/account.go
```diff
@@ -58,16 +58,3 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, portID string) {
 	k.accountKeeper.SetAccount(ctx, interchainAccount)
 	k.SetInterchainAccountAddress(ctx, portID, interchainAccount.Address)
 }
-
-func (k Keeper) GetInterchainAccount(ctx sdk.Context, addr sdk.AccAddress) (types.InterchainAccount, error) {
```
seantking, line 62: This isn't used anywhere. Some leftover code from the previous implementation I think.

modules/apps/27-interchain-accounts/module.go
```diff
@@ -187,15 +187,15 @@ func (am AppModule) OnChanCloseInit(
 	channelID string,
 ) error {
 	// Disallow user-initiated channel closing for interchain account channels
-	return nil
+	return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "user cannot close channel")
```
seantking, line 190: @colin-axner updating chan closing capabilities

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -66,20 +65,24 @@ func NewInterchainAccount(ba *authtypes.BaseAccount, accountOwner string) *Inter
...
 
 // SetSequence - Implements AccountI
-func (InterchainAccount) SetSequence(seq uint64) error {
+func (ia InterchainAccount) SetSequence(seq uint64) error {
 	return fmt.Errorf("not supported for interchain accounts")
 }
 
 func (ia InterchainAccount) Validate() error {
+	if strings.TrimSpace(ia.AccountOwner) == "" {
+		return fmt.Errorf("AccountOwner cannot be empty")
```
colin-axner, line 79: I think this needs to be an sdk registered error otherwise the error message will be lost when it is converted to an sdk internal error

colin-axner: We can only use `fmt.Errorf` for genesis validation or client side code

modules/apps/27-interchain-accounts/types/account_test.go
```diff
@@ -88,3 +95,83 @@ func (suite *TypesTestSuite) TestGeneratePortID() {
 		})
 	}
 }
+
+func (suite *TypesTestSuite) TestInterchainAccount() {
+	pubkey := secp256k1.GenPrivKey().PubKey()
+	addr := sdk.AccAddress(pubkey.Address())
+	baseAcc := authtypes.NewBaseAccountWithAddress(addr)
+	interchainAcc := types.NewInterchainAccount(baseAcc, "account-owner-id")
```
colin-axner, line 103: should we use TestOwner for `account-owner-id`?

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -66,20 +65,24 @@ func NewInterchainAccount(ba *authtypes.BaseAccount, accountOwner string) *Inter
 }
 
 // SetPubKey - Implements AccountI
-func (InterchainAccount) SetPubKey(pubKey crypto.PubKey) error {
+func (ia InterchainAccount) SetPubKey(pubKey crypto.PubKey) error {
 	return fmt.Errorf("not supported for interchain accounts")
```
colin-axner, line 69: ditto, is this what the SDK does for accounts?

modules/apps/27-interchain-accounts/types/account_test.go
```diff
@@ -88,3 +95,83 @@ func (suite *TypesTestSuite) TestGeneratePortID() {
...
+		acc     authtypes.GenesisAccount
+		expPass bool
+	}{
+		{
+			"interchain account with empty AccountOwner field",
+			types.NewInterchainAccount(baseAcc, ""),
+			false,
+		},
+		{
+			"success",
```
colin-axner, line 130: nit, success case first

modules/apps/27-interchain-accounts/types/account_test.go
```diff
@@ -88,3 +95,83 @@ func (suite *TypesTestSuite) TestGeneratePortID() {
...
+
+	bz, err := json.Marshal(interchainAcc)
+	suite.Require().NoError(err)
+
+	bz1, err := interchainAcc.MarshalJSON()
+	suite.Require().NoError(err)
+	suite.Require().Equal(string(bz), string(bz1))
+
+	var a types.InterchainAccount
+	suite.Require().NoError(json.Unmarshal(bz, &a))
```
colin-axner, line 176: should we check if a is equal to interchainAcc?

colin-axner (commented):
LGTM! Mostly just concerned about the `fmt.Errorf` usage

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -66,20 +65,24 @@ func NewInterchainAccount(ba *authtypes.BaseAccount, accountOwner string) *Inter
...
+	if strings.TrimSpace(ia.AccountOwner) == "" {
+		return fmt.Errorf("AccountOwner cannot be empty")
+	}
+
 	return ia.BaseAccount.Validate()
 }
 
-type ibcAccountPretty struct {
+type interchainAccountPretty struct {
 	Address       sdk.AccAddress `json:"address" yaml:"address"`
```
damiannolan, line 86: This field `Address` has the same json tag as `AccountOwner` below, maybe we should update the json tag for `AccountOwner` as they're both using `json:"address"`

modules/apps/27-interchain-accounts/types/account_test.go
```diff
@@ -88,3 +95,83 @@ func (suite *TypesTestSuite) TestGeneratePortID() {
...
+
+func (suite *TypesTestSuite) TestInterchainAccountMarshalYAML() {
+	suite.SetupTest() // reset
+
+	addr := suite.chainA.SenderAccount.GetAddress()
+	ba := authtypes.NewBaseAccountWithAddress(addr)
+
+	interchainAcc := types.NewInterchainAccount(ba, suite.chainB.SenderAccount.GetAddress().String())
+
+	bs, err := yaml.Marshal(interchainAcc)
```
damiannolan, line 155: I think it would be better to test the `MarshalYAML()` func on `InterchainAccount` like you do for `MarshalJSON` below

I think the function signature of `MarshalYAML()` should also be updated.

From:
```
func (ia InterchainAccount) MarshalYAML() (interface{}, error)
```

To:
```
func (ia InterchainAccount) MarshalYAML() ([]byte, error)
```

modules/apps/27-interchain-accounts/types/account.go
seantking: Updated. This code was grandfathered in from the previous work on ICA. I assumed up to now that the override here for `SetPubKey` & `SetSequence` was necessary as an interchain account is controlled by another chains on-chain logic rather than a private key. Perhaps this isn't necessary though? Wdyt?

modules/apps/27-interchain-accounts/types/account_test.go
seantking: updated. thanks

modules/apps/27-interchain-accounts/types/account.go
seantking: Good catch. Updated this.

modules/apps/27-interchain-accounts/types/account_test.go
seantking: Correct. I added a check for this. Thanks

modules/apps/27-interchain-accounts/types/account.go
seantking: Good catch. Updated.

modules/apps/27-interchain-accounts/types/account_test.go
seantking: Updated this. For some reason, I needed to use `InterchainAccountPretty` struct for the YAML test but not for the json.

modules/apps/27-interchain-accounts/types/account_test.go
seantking: I added a reusable test address to the file. Is that what you meant?

colin-axner (approved):

modules/apps/27-interchain-accounts/types/account.go
colin-axner: Lets revisit during the internal audit. I'm wondering how this will interact with module accounts (if we even need this type)

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -70,40 +70,44 @@ func NewInterchainAccount(ba *authtypes.BaseAccount, accountOwner string) *Inter
...
 func (ia InterchainAccount) Validate() error {
+	if strings.TrimSpace(ia.AccountOwner) == "" {
+		return sdkerrors.Wrap(ErrInvalidAccountAddress, "AccountOwner cannot be empty")
+	}
+
 	return ia.BaseAccount.Validate()
 }
 
-type ibcAccountPretty struct {
+type InterchainAccountPretty struct {
```
damiannolan, line 90: Tiny nit: can we make this unexposed/private?

damiannolan (approved):

PR #426, opened 2021-09-20
POC: middleware testing with mock module

colin-axner:
## Description

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/keeper/handshake_test.go
+++ b/modules/apps/27-interchain-accounts/keeper/handshake_test.go
@@ -301,6 +301,7 @@ func (suite *KeeperTestSuite) TestChanOpenAck() {
 	suite.SetupTest() // reset
 	path := NewICAPath(suite.chainA, suite.chainB)
 	owner := "owner"
+	counterpartyVersion := types.Version
 	suite.coordinator.SetupConnections(path)
 
 	err := InitInterchainAccount(path.EndpointA, owner)
@@ -335,7 +336,13 @@ func (suite *KeeperTestSuite) TestChanOpenAck() {
 		return nil
 	}
 
-	err = path.EndpointA.ChanOpenAck()
+	module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), types.PortID)
+	suite.Require().NoError(err)
+
+	cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
+	suite.Require().True(ok)
+
+	err = cbs.OnChanOpenAck(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, counterpartyVersion)
 	suite.Require().Error(err)
 
 	// reset with no middleware
@@ -343,7 +350,7 @@ func (suite *KeeperTestSuite) TestChanOpenAck() {
 	mockModule.Middleware.OnChanOpenAck = nil
 	suite.chainA.GetSimApp().GetModuleManager().Modules["mock"] = mockModule
 
-	err = path.EndpointA.ChanOpenAck()
+	err = cbs.OnChanOpenAck(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, counterpartyVersion)
 	suite.Require().NoError(err)
 
 }
--- a/modules/apps/27-interchain-accounts/keeper/handshake_test.go
+++ b/modules/apps/27-interchain-accounts/keeper/handshake_test.go
@@ -1,12 +1,16 @@
 package keeper_test
 
 import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
 
 	"github.com/cosmos/ibc-go/modules/apps/27-interchain-accounts/types"
 	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
 	host "github.com/cosmos/ibc-go/modules/core/24-host"
 	ibctesting "github.com/cosmos/ibc-go/testing"
+	"github.com/cosmos/ibc-go/testing/mock"
 )
 
 func (suite *KeeperTestSuite) TestOnChanOpenInit() {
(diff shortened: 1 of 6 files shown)


testing/mock/middleware.go
```diff
@@ -0,0 +1,19 @@
...
+
+type MockMiddleware struct {
+	OnChanOpenInit func(ctx sdk.Context, order channeltypes.Order, connectionHops []string, portID string,
+		channelID string, chanCap *capabilitytypes.Capability, counterparty channeltypes.Counterparty, version string,
+	) error
+	OnChanOpenTry func(ctx sdk.Context, order channeltypes.Order, connectionHops []string, portID string,
+		channelID string, chanCap *capabilitytypes.Capability, counterparty channeltypes.Counterparty, version, counterpartyVersion string,
+	) error
+	OnChanOpenAck     func(sdk.Context, string, string, string) error
+	OnChanOpenConfirm func(sdk.Context, string, string) error
```
colin-axner, line 18: This would be extended to entire IBC app interface. Allows applications to test specific functions without having to define all the functions

testing/mock/mock.go
```diff
@@ -178,7 +180,11 @@ func (am AppModule) OnChanOpenTry(
 }
 
 // OnChanOpenAck implements the IBCModule interface.
-func (am AppModule) OnChanOpenAck(sdk.Context, string, string, string) error {
+func (am AppModule) OnChanOpenAck(ctx sdk.Context, portID string, channelID string, counterpartyVersion string) error {
+	if am.Middleware.OnChanOpenAck != nil {
+		return am.Middleware.OnChanOpenAck(ctx, portID, channelID, counterpartyVersion)
+	}
```
colin-axner, line 186: would be added to every mock callback, ie, use developer provided function or do default logic

testing/simapp/app.go
```diff
@@ -658,6 +658,10 @@ func GetMaccPerms() map[string][]string {
 	return dupMaccPerms
 }
 
+func (app *SimApp) GetModuleManager() *module.Manager {
+	return app.mm
+}
+
```
colin-axner, line 664: this will be replaced with a get mock module function

modules/apps/27-interchain-accounts/keeper/handshake_test.go
```diff
@@ -290,3 +295,62 @@ func (suite *KeeperTestSuite) TestOnChanOpenConfirm() {
...
+
+	// add base app
+	mockModule, ok := suite.chainA.GetSimApp().GetModuleManager().Modules["mock"].(mock.AppModule)
+	suite.Require().True(ok)
+	mockModule.Middleware.OnChanOpenAck = func(ctx sdk.Context, portID, channelID, version string) error {
+		if channelID == "channel-0" {
+			return fmt.Errorf("error")
+		}
+		return nil
+	}
```
colin-axner, line 337: main code needed for developers to add their own custom logic to mock callbacks for testing. First line will be made into `GetMockModule` function in simapp

modules/apps/27-interchain-accounts/keeper/handshake_test.go
```diff
@@ -290,3 +295,62 @@ func (suite *KeeperTestSuite) TestOnChanOpenConfirm() {
...
+	cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
+	suite.Require().True(ok)
+
+	err = cbs.OnChanOpenAck(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, counterpartyVersion)
+	suite.Require().Error(err)
+
+	// reset with no middleware
+	mockModule, ok = suite.chainA.GetSimApp().GetModuleManager().Modules["mock"].(mock.AppModule)
+	mockModule.Middleware.OnChanOpenAck = nil
+	suite.chainA.GetSimApp().GetModuleManager().Modules["mock"] = mockModule
```
colin-axner, line 351: this could also be reset by calling `suite.SetupTest()`

modules/apps/27-interchain-accounts/keeper/handshake_test.go
```diff
@@ -290,3 +295,62 @@ func (suite *KeeperTestSuite) TestOnChanOpenConfirm() {
...
+	err = InitInterchainAccount(path.EndpointA, owner)
+	suite.Require().NoError(err)
+
+	err = path.EndpointB.ChanOpenTry()
+	suite.Require().NoError(err)
+
+	// add base app
+	mockModule, ok := suite.chainA.GetSimApp().GetModuleManager().Modules["mock"].(mock.AppModule)
+	suite.Require().True(ok)
+	mockModule.Middleware.OnChanOpenAck = func(ctx sdk.Context, portID, channelID, version string) error {
```
seantking, line 332: Looks cool :-)  I have a few clarifying questions. 

So we are defining a mock module (which in this case is representing an ica-auth module), writing custom logic for the callbacks, and then we are testing if the custom logic for the mock middleware callbacks/ack/timeout gets called correctly when testing the ICA functionality (in this example)?

Question: Are we just testing that the integration points are operating correctly (middleware with ICA) or is the goal to test out a bunch of different middleware use cases? I'm wondering why these tests wouldn't be in the middleware codebase instead.

colin-axner: > So we are defining a mock module (which in this case is representing an ica-auth module), writing custom logic for the callbacks, and then we are testing if the custom logic for the mock middleware callbacks/ack/timeout gets called correctly when testing the ICA functionality (in this example)?

Correct.

We are testing the integration points. I think I'll rename `Middleware` -> `Base Application`. The point is that our module may be middleware and we want to test that it can integrate into another module without having to be dependent on such a module. We may want to implement custom end behaviour depending on our own logic.

In the case of interchain accounts, I'd like to test that OnChanOpenTry is never called by returning errors on those callbacks (ensuring no base application can change the behaviour of this callback). It also can test the successful case. 

Custom parsing logic for the acknowledgement is also an inspiration. The mock module won't know how to generically parse an ack since it is unique for every application

> I'm wondering why these tests wouldn't be in the middleware codebase instead.

I'm not sure I understand. What is the middleware codebase? The above tests were primarily constructed as POC that this is actually calling the defined callbacks

The primary purpose of this is to allow the mock module to be used in an situation an IBC application may need a mock module instead of them requiring them to define their own mock module

seantking: > I'm not sure I understand.

Sorry this question was asked in the case that my assumption of the design was incorrect (it was correct, so you can disregard this)

colin-axner:
implementation of this POC is in #432

Issue #444, opened 2021-09-28
Validate connection id associated with channel matches ids in controller port-id


seantking:
## Summary

We need to do a check during the `OnChanOpenTry` step to validate that the connection-id/counterparty connection id given in the controller side port-id match the connection-ids associated with the current channel. e.g. `ics27-1-<connection-id>-<counterparty-connection-id>-<address>`

You can use the channelkeeper to check for the current connection-id/counterparty connection-id associated with the current channel. We should return an error if the connection-ids do no match.

## Problem Definition

We need to do this check to ensure that a controller chain cannot steal the port-id namespace from another controller chain (which is handled via the connection-ids). 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #446, opened 2021-09-29
Nitpicks: ICA Audit (InitInterchainAccount Flow)


seantking:
## Summary of Issue
The following nits arose as part of the audit. 

- [x] Remove “for address” in ErrPortAlreadyBound
- [x] Update error msg for ordered check and counterparty.Port (OnChanOpenInit/OnChanOpenTry)
- [x] Do basic version string check on OnChanOpenInit/OnChanOpenTry (update ValidateVersion to check full expected version - instead of partial)
- [x] Update claim capability error msg (wrap err instead of err.Error())
- [x] RegisterInterchainAccount API rename portID -> counterpartyPortID
- [x] Wrap parsed Acc check with ErrInvalidVersion instead of using ErrInvalidAccountAddress (OnChanOpenTry)
- [x] Use activeChannel prefix for range queries (something like -> https://github.com/cosmos/ibc-go/blob/interchain-accounts/modules/apps/27-interchain-accounts/types/keys.go#L41)
- [x] Update OnChanOpenConfirm godoc
- [x] Review expected keeper interfaces (double check if we expose anything unnecessarily here)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #454, opened 2021-09-30
chore: ctrl port connection id validation

damiannolan:
## Description

- Updated controller port id delimiter to match version string delimiter of `|`. Allows cleaner parsing for validation
- Adding connection id validation. Ensures connection ids in controller port id match that of the associated channel
- Addressed some minor nits
    - ParseAddressFromVersion uses `strings.Split()` in favour of `strings.TrimPrefix()`
    - Check self port IDs in channel handshake
    - Update error msg for ordered check and counterparty port (OnChanOpenInit/OnChanOpenTry)

closes: #444 

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/keeper/handshake.go
+++ b/modules/apps/27-interchain-accounts/keeper/handshake.go
@@ -1,11 +1,15 @@
 package keeper
 
 import (
+	"strconv"
+	"strings"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
 
 	"github.com/cosmos/ibc-go/v2/modules/apps/27-interchain-accounts/types"
+	connectiontypes "github.com/cosmos/ibc-go/v2/modules/core/03-connection/types"
 	channeltypes "github.com/cosmos/ibc-go/v2/modules/core/04-channel/types"
 	porttypes "github.com/cosmos/ibc-go/v2/modules/core/05-port/types"
 	host "github.com/cosmos/ibc-go/v2/modules/core/24-host"
@@ -31,14 +35,48 @@ func (k Keeper) OnChanOpenInit(
 	version string,
 ) error {
 	if order != channeltypes.ORDERED {
-		return sdkerrors.Wrapf(channeltypes.ErrInvalidChannelOrdering, "invalid channel ordering: %s, expected %s", order.String(), channeltypes.ORDERED.String())
+		return sdkerrors.Wrapf(channeltypes.ErrInvalidChannelOrdering, "expected %s channel, got %s", channeltypes.ORDERED, order)
 	}
+
 	if counterparty.PortId != types.PortID {
-		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "counterparty port-id must be '%s', (%s != %s)", types.PortID, counterparty.PortId, types.PortID)
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "expected %s, got %s", types.PortID, portID)
 	}
 
 	if err := types.ValidateVersion(version); err != nil {
 		return sdkerrors.Wrap(err, "version validation failed")
+	}
+
+	s := strings.Split(portID, types.Delimiter)
+	if len(s) != 4 {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "unexpected format in port ID (%s)", portID)
+	}
+
+	channel, found := k.channelKeeper.GetChannel(ctx, portID, channelID)
+	if !found {
+		return sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	counterpartyHops, found := k.channelKeeper.CounterpartyHops(ctx, channel)
+	if !found {
+		return sdkerrors.Wrap(connectiontypes.ErrConnectionNotFound, channel.ConnectionHops[0])
+	}
+
+	connSeq, err := connectiontypes.ParseConnectionSequence(connectionHops[0])
+	if err != nil {
+		return sdkerrors.Wrap(connectiontypes.ErrInvalidConnection, "failed to parse connection sequence")
+	}
+
+	counterpartyConnSeq, err := connectiontypes.ParseConnectionSequence(counterpartyHops[0])
+	if err != nil {
+		return sdkerrors.Wrap(connectiontypes.ErrInvalidConnection, "failed to parse counterparty connection sequence")
+	}
+
+	if strconv.FormatUint(connSeq, 10) != s[1] {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid connection sequence in port (%s)", s[1])
+	}
+
+	if strconv.FormatUint(counterpartyConnSeq, 10) != s[2] {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid connection sequence in port (%s)", s[2])
 	}
 
 	existingChannelID, found := k.GetActiveChannel(ctx, portID)
@@ -70,7 +108,7 @@ func (k Keeper) OnChanOpenTry(
 	counterpartyVersion string,
 ) error {
 	if order != channeltypes.ORDERED {
-		return sdkerrors.Wrapf(channeltypes.ErrInvalidChannelOrdering, "invalid channel ordering: %s, expected %s", order.String(), channeltypes.ORDERED.String())
+		return sdkerrors.Wrapf(channeltypes.ErrInvalidChannelOrdering, "expected %s channel, got %s", channeltypes.ORDERED, order)
 	}
 
 	if err := types.ValidateVersion(version); err != nil {
@@ -79,6 +117,43 @@ func (k Keeper) OnChanOpenTry(
 
 	if err := types.ValidateVersion(counterpartyVersion); err != nil {
 		return sdkerrors.Wrap(err, "counterparty version validation failed")
+	}
+
+	if portID != types.PortID {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "expected %s, got %s", types.PortID, portID)
+	}
+
+	s := strings.Split(counterparty.PortId, types.Delimiter)
+	if len(s) != 4 {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "unexpected format in port ID (%s)", counterparty.PortId)
+	}
+
+	channel, found := k.channelKeeper.GetChannel(ctx, portID, channelID)
+	if !found {
+		return sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	counterpartyHops, found := k.channelKeeper.CounterpartyHops(ctx, channel)
+	if !found {
+		return sdkerrors.Wrap(connectiontypes.ErrConnectionNotFound, channel.ConnectionHops[0])
+	}
+
+	connSeq, err := connectiontypes.ParseConnectionSequence(connectionHops[0])
+	if err != nil {
+		return sdkerrors.Wrap(connectiontypes.ErrInvalidConnection, "failed to parse connection sequence")
+	}
+
+	counterpartyConnSeq, err := connectiontypes.ParseConnectionSequence(counterpartyHops[0])
+	if err != nil {
+		return sdkerrors.Wrap(connectiontypes.ErrInvalidConnection, "failed to parse counterparty connection sequence")
+	}
+
+	if strconv.FormatUint(connSeq, 10) != s[2] {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid connection sequence in port (%s)", s[2])
+	}
+
+	if strconv.FormatUint(counterpartyConnSeq, 10) != s[1] {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid connection sequence in port (%s)", s[1])
 	}
 
 	// On the host chain the capability may only be claimed during the OnChanOpenTry
--- a/modules/apps/27-interchain-accounts/keeper/keeper_test.go
+++ b/modules/apps/27-interchain-accounts/keeper/keeper_test.go
@@ -21,7 +21,7 @@ var (
 	// TestOwnerAddress defines a reusable bech32 address for testing purposes
 	TestOwnerAddress = "cosmos17dtl0mjt3t77kpuhg2edqzjpszulwhgzuj9ljs"
 	// TestPortID defines a resuable port identifier for testing purposes
-	TestPortID = fmt.Sprintf("%s-0-0-%s", types.VersionPrefix, TestOwnerAddress)
+	TestPortID = fmt.Sprintf("%s|0|0|%s", types.VersionPrefix, TestOwnerAddress)
 	// TestVersion defines a resuable interchainaccounts version string for testing purposes
 	TestVersion = types.NewAppVersion(types.VersionPrefix, TestAccAddress.String())
 )
--- a/modules/apps/27-interchain-accounts/module_test.go
+++ b/modules/apps/27-interchain-accounts/module_test.go
@@ -21,7 +21,7 @@ var (
 	// TestOwnerAddress defines a reusable bech32 address for testing purposes
 	TestOwnerAddress = "cosmos17dtl0mjt3t77kpuhg2edqzjpszulwhgzuj9ljs"
 	// TestPortID defines a resuable port identifier for testing purposes
-	TestPortID = fmt.Sprintf("%s-0-0-%s", types.VersionPrefix, TestOwnerAddress)
+	TestPortID = fmt.Sprintf("%s|0|0|%s", types.VersionPrefix, TestOwnerAddress)
 	// TestVersion defines a resuable interchainaccounts version string for testing purposes
 	TestVersion = types.NewAppVersion(types.VersionPrefix, TestAccAddress.String())
 )
@@ -124,6 +124,9 @@ func (suite *InterchainAccountsTestSuite) TestOnChanOpenInit() {
 		Version:        types.VersionPrefix,
 	}
 
+	// set channel
+	path.EndpointA.SetChannel(*channel)
+
 	module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), types.PortID)
 	suite.Require().NoError(err)
 
@@ -142,7 +145,6 @@ func (suite *InterchainAccountsTestSuite) TestOnChanOpenInit() {
 func (suite *InterchainAccountsTestSuite) TestOnChanOpenTry() {
 	suite.SetupTest() // reset
 	path := NewICAPath(suite.chainA, suite.chainB)
-	counterpartyVersion := types.VersionPrefix
 	suite.coordinator.SetupConnections(path)
 
 	err := InitInterchainAccount(path.EndpointA, TestOwnerAddress)
@@ -158,24 +160,29 @@ func (suite *InterchainAccountsTestSuite) TestOnChanOpenTry() {
 		Version:        types.VersionPrefix,
 	}
 
-	module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), types.PortID)
+	// set channel
+	channelSequence := path.EndpointB.Chain.App.GetIBCKeeper().ChannelKeeper.GetNextChannelSequence(path.EndpointB.Chain.GetContext())
+	path.EndpointB.ChannelID = channeltypes.FormatChannelIdentifier(channelSequence)
+	path.EndpointB.SetChannel(*channel)
+
+	module, _, err := suite.chainB.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainB.GetContext(), types.PortID)
 	suite.Require().NoError(err)
 
-	chanCap, err := suite.chainA.App.GetScopedIBCKeeper().NewCapability(suite.chainA.GetContext(), host.ChannelCapabilityPath(ibctesting.TransferPort, path.EndpointA.ChannelID))
+	chanCap, err := suite.chainB.App.GetScopedIBCKeeper().NewCapability(suite.chainB.GetContext(), host.ChannelCapabilityPath(ibctesting.TransferPort, path.EndpointB.ChannelID))
 	suite.Require().NoError(err)
 
-	cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
+	cbs, ok := suite.chainB.App.GetIBCKeeper().Router.GetRoute(module)
 	suite.Require().True(ok)
 
-	err = cbs.OnChanOpenTry(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
-		path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, chanCap, channel.Counterparty, channel.GetVersion(), counterpartyVersion,
+	err = cbs.OnChanOpenTry(suite.chainB.GetContext(), channel.Ordering, channel.GetConnectionHops(),
+		path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, chanCap, channel.Counterparty, TestVersion, types.VersionPrefix,
 	)
+	suite.Require().NoError(err)
 }
 
 func (suite *InterchainAccountsTestSuite) TestOnChanOpenAck() {
 	suite.SetupTest() // reset
 	path := NewICAPath(suite.chainA, suite.chainB)
-	counterpartyVersion := types.VersionPrefix
 	suite.coordinator.SetupConnections(path)
 
 	err := InitInterchainAccount(path.EndpointA, TestOwnerAddress)
@@ -190,7 +197,7 @@ func (suite *InterchainAccountsTestSuite) TestOnChanOpenAck() {
 	cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
 	suite.Require().True(ok)
 
-	err = cbs.OnChanOpenAck(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, counterpartyVersion)
+	err = cbs.OnChanOpenAck(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, TestVersion)
 	suite.Require().NoError(err)
 }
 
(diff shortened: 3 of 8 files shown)


modules/core/24-host/validate.go
```diff
@@ -25,7 +25,7 @@ var DefaultMaxPortCharacterLength = 128
 // - Alphanumeric
 // - `.`, `_`, `+`, `-`, `#`
 // - `[`, `]`, `<`, `>`
-var IsValidID = regexp.MustCompile(`^[a-zA-Z0-9\.\_\+\-\#\[\]\<\>]+$`).MatchString
+var IsValidID = regexp.MustCompile(`^[a-zA-Z0-9\.\_\+\-\#\|\[\]\<\>]+$`).MatchString
```
damiannolan, line 28: I've added `|` to the valid characters in this regex. I intentionally left this in to raise on the PR. 
Figured as we are using `|` as the delimiter for the version string then it would be appropriate to use it for the controller port ID. For example: `ics27-1|0|0|cosmos17dtl0mjt3t77kpuhg2edqzjpszulwhgzuj9ljs`
This regex is used when ICA attempts to call `BindPort`

This allows cleaner splitting for connection id validation. If there is no issue with adding the pipe char to this regex let me know and I will open a PR with this one liner targeted for main!

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -23,7 +23,7 @@ func GenerateAddress(moduleAccAddr sdk.AccAddress, portID string) sdk.AccAddress
 
 // ParseAddressFromVersion trims the interchainaccounts version prefix and returns the associated account address
 func ParseAddressFromVersion(version string) string {
-	return strings.TrimPrefix(version, fmt.Sprint(VersionPrefix, Delimiter))
+	return strings.Split(version, Delimiter)[1]
```
damiannolan, line 26: The version string validation in `OnChanOpenTry` is such that this should never encounter an index out of bounds but I'm thinking it's probably worth adding a defensive check here regardless

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -81,6 +119,43 @@ func (k Keeper) OnChanOpenTry(
 		return sdkerrors.Wrap(err, "counterparty version validation failed")
 	}
 
+	if portID != types.PortID {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "expected %s, got %s", types.PortID, portID)
+	}
+
+	s := strings.Split(counterparty.PortId, types.Delimiter)
```
seantking, line 126: Maybe we can add the validation check to its own util function just for better readability?

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -31,16 +35,50 @@ func (k Keeper) OnChanOpenInit(
...
+		return sdkerrors.Wrap(connectiontypes.ErrInvalidConnection, "failed to parse connection sequence")
+	}
+
+	counterpartyConnSeq, err := connectiontypes.ParseConnectionSequence(counterpartyHops[0])
+	if err != nil {
+		return sdkerrors.Wrap(connectiontypes.ErrInvalidConnection, "failed to parse counterparty connection sequence")
+	}
+
+	if strconv.FormatUint(connSeq, 10) != s[1] {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid connection sequence in port (%s)", s[1])
```
crodriguezvega, line 75: Should we print in the error message the expect connection sequence? Or would that reveal sensitive information?

modules/apps/27-interchain-accounts/keeper/handshake.go
crodriguezvega: Are these the same checks as in `OnChanOpenInit`?

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -31,16 +35,50 @@ func (k Keeper) OnChanOpenInit(
...
 	if counterparty.PortId != types.PortID {
-		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "counterparty port-id must be '%s', (%s != %s)", types.PortID, counterparty.PortId, types.PortID)
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "expected %s, got %s", types.PortID, portID)
 	}
 
 	if err := types.ValidateVersion(version); err != nil {
 		return sdkerrors.Wrap(err, "version validation failed")
 	}
 
+	s := strings.Split(portID, types.Delimiter)
```
colin-axner, line 49: Validation of the connection id's in the portID should be a helper function. Same for getting the connection sequence (though 04-keeper might expose this function)

modules/core/24-host/validate.go
colin-axner: Hmm, this requires a ICS change. We should open a pr there first for discussion. I'm not sure if it was intentionally left out. Otherwise we should just pick a different character

colin-axner (commented):
Nice start! For future reference, my preference would be always to keep pr's in the smallest scope possible. It makes reviewing changes a lot easier instead of trying to keep the context for why all the changes are necessary

I find reviewing 5 small pr's much faster than reviewing 1 large pr

modules/core/24-host/validate.go
damiannolan: Cool, I'm happy to open a PR for discussion, but I'm open to just using a different character if you think that would be less hassle. But in which case I'd also suggest we change the version string delimiter to align for consistency. 

Aside from a completely hyphenated string which would make parsing awkward, I think dot separation looks best. PortID: `ics27-1.0.0.cosmos17dtl0mjt3t77kpuhg2edqzjpszulwhgzuj9ljs`
Version: `ics27-1.cosmos17dtl0mjt3t77kpuhg2edqzjpszulwhgzuj9ljs`

wdyt? Also tagging @seantking

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -31,16 +35,50 @@ func (k Keeper) OnChanOpenInit(
 	version string,
 ) error {
 	if order != channeltypes.ORDERED {
-		return sdkerrors.Wrapf(channeltypes.ErrInvalidChannelOrdering, "invalid channel ordering: %s, expected %s", order.String(), channeltypes.ORDERED.String())
+		return sdkerrors.Wrapf(channeltypes.ErrInvalidChannelOrdering, "expected %s channel, got %s", channeltypes.ORDERED, order)
```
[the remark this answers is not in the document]
damiannolan: Sure, I can remove. I think I left this in as it was formatted this way in [transfer](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/module.go#L207)

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: > Are these the same checks as in `OnChanOpenInit`?

Essentially yes, but with a subtle difference. In `OnChanOpenInit` the (first-party/self) portID is the encoded string whereas in `OnChanOpenTry` the counterparty portID is encoded with connection sequences. 
As a result in the INIT step, we assert the underlying conn sequence matches index [1] and the counterparty conn sequence matches index [2], however in the TRY step the underlying conn sequence must match index [2] and counterparty conn sequence must match index [1]. 

I've refactored the connection sequence validation to a reusable func `validateConnectionParams`.
Also added two helper funcs to parse controller & host conn sequences from the controller port ID which deals with the swapped indexing

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: Updated, moved to `validateConnectionParams` helper func and added two parsing helper funcs for parsing controller & host conn sequences from port.

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -47,7 +52,29 @@ func GeneratePortID(owner, connectionID, counterpartyConnectionID string) (strin
 		return "", sdkerrors.Wrap(err, "invalid counterparty connection identifier")
 	}
 
-	return fmt.Sprintf("%s-%d-%d-%s", VersionPrefix, connectionSeq, counterpartyConnectionSeq, owner), nil
+	return fmt.Sprint(VersionPrefix, Delimiter, connectionSeq, Delimiter, counterpartyConnectionSeq, Delimiter, owner), nil
+}
+
+// ParseCtrlConnSequence attempts to parse the controller connection sequence from the provided port identifier
+// The port identifier must match the controller chain format outlined in (TODO: link spec), otherwise an empty string is returned
+func ParseCtrlConnSequence(portID string) string {
```
damiannolan, line 60: I can change `Ctrl` -> `Controller` if anyone feels strongly about it. I just left it like this for brevity.

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: I don't think so! I've updated to include

modules/apps/27-interchain-accounts/types/account.go
seantking: I would prefer Controller

modules/core/24-host/validate.go
seantking: Dot separation is fine with me, although I prefer the |

modules/apps/27-interchain-accounts/types/account.go
damiannolan: np, updated

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -130,3 +148,37 @@ func (k Keeper) OnChanOpenConfirm(
...
+	}
+
+	counterpartyHops, found := k.channelKeeper.CounterpartyHops(ctx, channel)
+	if !found {
+		return sdkerrors.Wrap(connectiontypes.ErrConnectionNotFound, channel.ConnectionHops[0])
+	}
+
+	connSeq, err := connectiontypes.ParseConnectionSequence(channel.ConnectionHops[0])
+	if err != nil {
+		return sdkerrors.Wrapf(connectiontypes.ErrInvalidConnectionIdentifier, "failed to parse connection sequence (%s)", channel.ConnectionHops[0])
```
colin-axner, line 167:
```suggestion
		return sdkerrors.Wrapf(err, "failed to parse connection sequence (%s)", channel.ConnectionHops[0])
```

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -130,3 +148,37 @@ func (k Keeper) OnChanOpenConfirm(
...
+	}
+
+	counterpartyConnSeq, err := connectiontypes.ParseConnectionSequence(counterpartyHops[0])
+	if err != nil {
+		return sdkerrors.Wrapf(connectiontypes.ErrInvalidConnectionIdentifier, "failed to parse counterparty connection sequence (%s)", counterpartyHops[0])
```
colin-axner, line 172:
```suggestion
		return sdkerrors.Wrapf(err, "failed to parse counterparty connection sequence (%s)", counterpartyHops[0])
```

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -47,7 +52,29 @@ func GeneratePortID(owner, connectionID, counterpartyConnectionID string) (strin
 		return "", sdkerrors.Wrap(err, "invalid counterparty connection identifier")
 	}
 
-	return fmt.Sprintf("%s-%d-%d-%s", VersionPrefix, connectionSeq, counterpartyConnectionSeq, owner), nil
+	return fmt.Sprint(VersionPrefix, Delimiter, connectionSeq, Delimiter, counterpartyConnectionSeq, Delimiter, owner), nil
+}
+
+// ParseControllerConnSequence attempts to parse the controller connection sequence from the provided port identifier
+// The port identifier must match the controller chain format outlined in (TODO: link spec), otherwise an empty string is returned
+func ParseControllerConnSequence(portID string) string {
```
colin-axner, line 60: My preference would be to return a uint64. Is there a reason for returning a string instead? I guess then you'd need to return an error

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -47,7 +52,29 @@ func GeneratePortID(owner, connectionID, counterpartyConnectionID string) (strin
 		return "", sdkerrors.Wrap(err, "invalid counterparty connection identifier")
 	}
 
-	return fmt.Sprintf("%s-%d-%d-%s", VersionPrefix, connectionSeq, counterpartyConnectionSeq, owner), nil
+	return fmt.Sprint(VersionPrefix, Delimiter, connectionSeq, Delimiter, counterpartyConnectionSeq, Delimiter, owner), nil
```
colin-axner, line 55:
```suggestion
	return fmt.Sprint(
	VersionPrefix, Delimiter, 
	connectionSeq, Delimiter, 
	counterpartyConnectionSeq, Delimiter, 
	owner,
	), nil
```
food for thought

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -21,9 +21,14 @@ func GenerateAddress(moduleAccAddr sdk.AccAddress, portID string) sdk.AccAddress
 	return sdk.AccAddress(sdkaddress.Derive(moduleAccAddr, []byte(portID)))
 }
 
-// ParseAddressFromVersion trims the interchainaccounts version prefix and returns the associated account address
+// ParseAddressFromVersion attempts to extract the associated account address from the provided version string
 func ParseAddressFromVersion(version string) string {
```
colin-axner, line 25: It might be best to explicitly return an error. Otherwise we need something like:
```go
if address == "" {
    return sdkerrors.Wrap(types.ErrInvalidVersion, "could not parse version")
}
```
in most places we call this function

colin-axner, line 25: should we put parsing functions in a separate file? Host/Controller connection parsing doesn't have much to do with `account.go`

`keys.go` is typically where I might put it, open to other ideas

modules/apps/27-interchain-accounts/types/expected_keepers.go
```diff
@@ -31,6 +31,7 @@ type ChannelKeeper interface {
 	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, packet ibcexported.PacketI) error
 	ChanCloseInit(ctx sdk.Context, portID, channelID string, chanCap *capabilitytypes.Capability) error
 	ChanOpenInit(ctx sdk.Context, order channeltypes.Order, connectionHops []string, portID string, portCap *capabilitytypes.Capability, counterparty channeltypes.Counterparty, version string) (string, *capabilitytypes.Capability, error)
+	CounterpartyHops(ctx sdk.Context, channel channeltypes.Channel) ([]string, bool)
```
colin-axner, line 34: All middleware will need to implement these functions (including fees) cc @AdityaSripal

modules/core/24-host/validate.go
colin-axner: Dot separation looks good to me. ICA spec needs to be updated anyways, so lets open a pr there and point out both options. For now, I'd prefer to go with dot separation so we don't have to modify this code

modules/apps/27-interchain-accounts/keeper/keeper_test.go
```diff
@@ -21,7 +21,7 @@ var (
 	// TestOwnerAddress defines a reusable bech32 address for testing purposes
 	TestOwnerAddress = "cosmos17dtl0mjt3t77kpuhg2edqzjpszulwhgzuj9ljs"
 	// TestPortID defines a resuable port identifier for testing purposes
-	TestPortID = fmt.Sprintf("%s-0-0-%s", types.VersionPrefix, TestOwnerAddress)
+	TestPortID = fmt.Sprintf("%s|0|0|%s", types.VersionPrefix, TestOwnerAddress)
```
colin-axner, line 24: should we use generate port id function? Might be good reason to add a `NewPortID` function if we cannot use the generate portID function

modules/apps/27-interchain-accounts/module_test.go
```diff
@@ -21,7 +21,7 @@ var (
 	// TestOwnerAddress defines a reusable bech32 address for testing purposes
 	TestOwnerAddress = "cosmos17dtl0mjt3t77kpuhg2edqzjpszulwhgzuj9ljs"
 	// TestPortID defines a resuable port identifier for testing purposes
-	TestPortID = fmt.Sprintf("%s-0-0-%s", types.VersionPrefix, TestOwnerAddress)
+	TestPortID = fmt.Sprintf("%s|0|0|%s", types.VersionPrefix, TestOwnerAddress)
```
colin-axner, line 24: ditto

modules/apps/27-interchain-accounts/types/account_test.go
```diff
@@ -52,6 +52,27 @@ func (suite *TypesTestSuite) TestParseAddressFromVersion() {
...
+	suite.Require().Empty(addr)
+}
+
+func (suite *TypesTestSuite) TestParseCtrlConnSequence() {
+	portID, err := types.GeneratePortID(TestOwnerAddress, "connection-0", "connection-1")
+	suite.Require().NoError(err)
+
+	connSeq := types.ParseControllerConnSequence(portID)
+	suite.Require().Equal("0", connSeq)
+	suite.Require().Empty(types.ParseControllerConnSequence(types.PortID))
```
colin-axner, line 66: Should this have failed test cases as well?

colin-axner (commented):
The new flow looks good to me! Left lots of suggestions/nits

modules/apps/27-interchain-accounts/keeper/handshake.go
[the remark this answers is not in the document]
crodriguezvega: oh, ok, then no need to remove it. Let's stay consistent.

modules/apps/27-interchain-accounts/types/account.go
damiannolan: I agree, uint64 would be more appropriate here. I originally opted for the `strconv` calls in `validateConnectionParams` to avoid having to do another error check, but it would look more fitting to return uint64! It will just add additional error checking to the handshake handler.

modules/apps/27-interchain-accounts/types/account.go
damiannolan: +1 for readability 👍

modules/apps/27-interchain-accounts/types/account.go
damiannolan: Agreed, will update

damiannolan: Yes, definitely! I was actually going to suggest doing this in a follow up PR next, but I can do it in this. 
I was considering suggesting we that we move the `GeneratePortID` and parsing the conn sequence from port id funcs to `port.go`, and moving the `GenerateAddress` and `ParseAddressFromVersion` to `address.go`, or alternatively we leave those ones here and just move the former to `keys.go`!

modules/apps/27-interchain-accounts/keeper/keeper_test.go
damiannolan: Yeah, I think we can use `GeneratePortID` here, I'll just discard the error

modules/apps/27-interchain-accounts/types/account_test.go
damiannolan: I'll update these when I change the function sig to `(uint64, error)`!

modules/apps/27-interchain-accounts/types/expected_keepers.go
damiannolan: These are included in #446 to review also! For example, `ChanOpenInit` and `ChanCloseInit` above are unused afaik

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -70,7 +86,25 @@ func (k Keeper) OnChanOpenTry(
...
+		return sdkerrors.Wrapf(channeltypes.ErrInvalidChannelOrdering, "expected %s channel, got %s", channeltypes.ORDERED, order)
+	}
+
+	if portID != types.PortID {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "expected %s, got %s", types.PortID, portID)
+	}
+
+	connSequence, err := types.ParseHostConnSequence(counterparty.PortId)
+	if err != nil {
+		return sdkerrors.Wrapf(err, "expected format <app-version%[1]sconn-seq%[1]scounterparty-conn-seq%[1]sowner>, got %s", types.Delimiter, portID)
```
damiannolan, line 98: Format string here may look a little odd using `%[1]s`, [see explicit argument indexing here](https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes). Avoids having to specify `types.Delimiter` multiple times in args.

Suggestions for a better error message are welcome, but I think including the expected format as @crodriguezvega  mentioned would be valuable.

modules/core/24-host/validate.go
damiannolan: I can open a PR over there, I've reverted the change to regex in `24-host` and updated the delimiter to `"."`

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -46,3 +52,45 @@ func KeyActiveChannel(portID string) []byte {
...
+// The port identifier must match the controller chain format outlined in (TODO: link spec), otherwise an empty string is returned
+func ParseHostConnSequence(portID string) (uint64, error) {
+	s := strings.Split(portID, Delimiter)
+	if len(s) != 4 {
+		return 0, sdkerrors.Wrap(porttypes.ErrInvalidPort, "failed to parse port identifier")
+	}
+
+	seq, err := strconv.ParseUint(s[2], 10, 64)
+	if err != nil {
+		return 0, sdkerrors.Wrapf(err, "failed to parse connection sequence (%s)", s[1])
```
colin-axner, line 82:
```suggestion
		return 0, sdkerrors.Wrapf(err, "failed to parse connection sequence (%s)", s[2])
```

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -31,10 +32,25 @@ func (k Keeper) OnChanOpenInit(
...
+	}
+
+	counterpartyConnSequence, err := types.ParseHostConnSequence(portID)
+	if err != nil {
+		return sdkerrors.Wrapf(err, "expected format <app-version%[1]sconn-seq%[1]scounterparty-conn-seq%[1]sowner>, got %s", types.Delimiter, portID)
+	}
+
+	if err := k.validateConnectionParams(ctx, channelID, portID, connSequence, counterpartyConnSequence); err != nil {
+		return sdkerrors.Wrapf(err, "failed to validate controller port (%s)", portID)
```
colin-axner, line 49: maybe the error message should say connection params? Otherwise maybe we should rename the function to `validateControllerPort`?

modules/apps/27-interchain-accounts/keeper/handshake.go
colin-axner: I'd define the expected string as a const and then link the spec in the comment for the expected format. I'd also recommend using `controller-conn-seq` and `host-conn-seq` since counterparty means different things depending on context

modules/apps/27-interchain-accounts/keeper/handshake.go
```diff
@@ -130,3 +172,37 @@ func (k Keeper) OnChanOpenConfirm(
...
+		return sdkerrors.Wrapf(err, "failed to parse connection sequence (%s)", channel.ConnectionHops[0])
+	}
+
+	counterpartyConnSeq, err := connectiontypes.ParseConnectionSequence(counterpartyHops[0])
+	if err != nil {
+		return sdkerrors.Wrapf(err, "failed to parse counterparty connection sequence (%s)", counterpartyHops[0])
+	}
+
+	if connSeq != connectionSeq {
+		return sdkerrors.Wrapf(connectiontypes.ErrInvalidConnection, "sequence mismatch, expected (%d), got (%d)", connSeq, connectionSeq)
```
colin-axner, line 200:
```suggestion
		return sdkerrors.Wrapf(connectiontypes.ErrInvalidConnection, "sequence mismatch, expected %d, got %d", connSeq, connectionSeq)
```
we probably don't need the parenthesis?

modules/apps/27-interchain-accounts/types/expected_keepers.go
colin-axner: We should remove unused interface functions

modules/apps/27-interchain-accounts/types/account.go
colin-axner: I like the change you suggested :+1: but we can do this in a followup if we desire. Using `keys.go` works for me. No strong preference

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: Sure thing! Any preference on this before I update?

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: Probably not, I borrowed this pattern/format from the [transfer module](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L68)

Should we keep it for consistency?

modules/apps/27-interchain-accounts/types/keys.go
damiannolan: Nice catch! Thanks

modules/apps/27-interchain-accounts/types/expected_keepers.go
damiannolan: I can take care of this in a smaller follow up PR 👍

modules/apps/27-interchain-accounts/keeper/handshake.go
colin-axner: I think `validateControllerPort` makes the most sense to me. It will likely align closer with the ICS spec

modules/apps/27-interchain-accounts/keeper/handshake.go
colin-axner: No we should start updating transfer module as well, but that should be done for #334

damiannolan: Updated! Removed all use of parenthesis in errors msgs.

modules/apps/27-interchain-accounts/keeper/handshake.go
damiannolan: Cool, I've updated to `validateControllerPort`. My only gripe with this, and perhaps its overly pedantic, is that we never actually pass the controller port to this function. We pass the self port ID to retrieve the channel which is obviously going to be different in the `OnChainOpenInit` and `OnChanOpenTry` callbacks. 

We could... 🤔 

1. Call it `validateControllerPortParams` and leave the args the same.
2. Move the conn sequence parsing into the func and pass the controller port as an arg

The only problem I see with opt 2 is that having two port ID args may be confusing(?)

Either way I'm happy to leave it as is now if you are 😄

seantking: I vote for `validateControllerPortParams and leave the args the same`

colin-axner (approved):
Excellent work!

seantking (approved):

Issue #449, opened 2021-09-29
Update Packet Data type


seantking:
## Summary
Update  `IBCAccountPacketData` to have the following definition:
```
message InterchainAccountPacketData  {
    bytes data = 1;
    string memo = 2;
    Type  type = 3;
}
```

Investigate if we can remove `IBCTxRaw` & only use `IBCTxBody`. Future versions of ICA may use the data field for use other than executing messages which we need to take into consideration. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
I think the data field should be repeated `Any`s

seantking:
> I think the data field should be repeated `Any`s

I was thinking the same until I considered that depending on the `Type` we may want data to not be a list of messages, which I guess is why it was left to be more arbitrary, to begin with.

colin-axner:
> I was thinking the same until I considered that depending on the `Type` we may want data to not be a list of messages

An [Any](https://pkg.go.dev/google.golang.org/protobuf/types/known/anypb#Any) is:
```go
type Any struct {
    string TypeURL
    []byte Value
}
```

Thus even if the Type isn't `EXECUTE_TX`, the list of `Any`s could be non message types as well

seantking:
Ah, yes that makes sense! I agree let's do `repeated Any`.

PR #456, opened 2021-09-30
Rename IBCAccountPacketData, Remove TxRaw 

colin-axner:
## Description

Doing this issue in two steps. This pr does name changes and trivially remove the TxRaw type. The next pr will modify the data type in packet data to be repeated any's thus removing TxBody type

partially closes: #449

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/keeper/relay.go
+++ b/modules/apps/27-interchain-accounts/keeper/relay.go
@@ -15,7 +15,7 @@ import (
 
 // TODO: implement middleware functionality, this will allow us to use capabilities to
 // manage helper module access to owner addresses they do not have capabilities for
-func (k Keeper) TrySendTx(ctx sdk.Context, portID string, data interface{}) ([]byte, error) {
+func (k Keeper) TrySendTx(ctx sdk.Context, portID string, data interface{}, memo string) ([]byte, error) {
 	// Check for the active channel
 	activeChannelId, found := k.GetActiveChannel(ctx, portID)
 	if !found {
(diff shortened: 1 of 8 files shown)


modules/apps/27-interchain-accounts/keeper/relay.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // TODO: implement middleware functionality, this will allow us to use capabilities to
 // manage helper module access to owner addresses they do not have capabilities for
-func (k Keeper) TrySendTx(ctx sdk.Context, portID string, data interface{}) ([]byte, error) {
+func (k Keeper) TrySendTx(ctx sdk.Context, portID string, data interface{}, memo string) ([]byte, error) {
```
colin-axner, line 18: should we take in the type as well so we don't have to change the api later?

proto/ibc/applications/interchain_accounts/v1/types.proto
```diff
@@ -24,8 +19,9 @@ enum Type {
   TYPE_EXECUTE_TX_UNSPECIFIED = 0 [(gogoproto.enumvalue_customname) = "EXECUTE_TX"];
 }
 
-// Packet data is comprised of raw transaction & type of transaction
-message IBCAccountPacketData {
+// InterchainAccountPacketData is comprised of araw transaction,type of transaction and optional memo field.
```
colin-axner, line 22:
```suggestion
// InterchainAccountPacketData is comprised of a raw transaction, type of transaction and optional memo field.
```

modules/apps/27-interchain-accounts/keeper/relay.go
seantking: Great point. I guess right now it's just hardcoded to always be `EXECUTE`. I guess there is no way to interpret the type based on the data (which the more I think about it may be anything, not just a list of messages). 

I think it probably makes sense to let the caller pass in the type here. :+1:

damiannolan (approved):

seantking (approved):

Issue #259, opened 2021-07-15
Fee Middleware: PayFee, PayFeeAsync + Handler


AdityaSripal:
PayFee escrows fees for the next packet sequence for given channelID and portID

PayFeeAsync escrows fees for a given (channelID, portID, sequence)

Handler escrows the provided fees for (channelID, portID, sequence) with breakdown of recv fee, ack fee, timeout fee

PR #465, opened 2021-10-04
Fee Middleware: Escrow logic

seantking:
## Description

Adding escrow logic for Fee Middleware. 

closes: #259 

---

write a little note why.


proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -15,10 +15,14 @@ service Msg {
...
   // may be called more than once by a relayer, in which case, latest counterparty address is always used.
   rpc RegisterCounterpartyAddress(MsgRegisterCounterpartyAddress) returns (MsgRegisterCounterpartyAddressResponse);
-  // EscrowPacketFee defines a rpc handler method for MsgEscrowPacketFee
-  // EscrowPacketFee is an open callback that may be called by any module/user that wishes to escrow funds in order to
-  // incentivize the relaying of the given packet.
-  rpc EscrowPacketFee(MsgEscrowPacketFee) returns (MsgEscrowPacketFeeResponse);
+  // PayPacketFee defines a rpc handler method for MsgPayPacketFee
+  // PayPacketFee is an open callback that may be called by any module/user that wishes to escrow funds in order to
+  // incentivize the relaying of the packet at the next sequence
+  rpc PayPacketFee(MsgPayPacketFee) returns (MsgPayPacketFeeResponse);
```
seantking, line 21: I updated the naming here to align with the ticket. From the user's perspective, we are trying to Pay for a packet fee so I think the naming makes sense.

Open to changing this back to `Escrow`

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -15,10 +15,14 @@ service Msg {
...
+  // PayPacketFeeAsync defines a rpc handler method for MsgPayPacketFeeAsync
+  // PayPacketFeeAsync is an open callback that may be called by any module/user that wishes to escrow funds in order to
+  // incentivize the relaying of a known packet
+  rpc PayPacketFeeAsync(MsgPayPacketFeeAsync) returns (MsgPayPacketFeeAsyncResponse);
```
seantking, line 25: @AdityaSripal As we discussed I separated the async functionality into a separate endpoint.

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -25,9 +27,51 @@ func (k Keeper) RegisterCounterpartyAddress(goCtx context.Context, msg *types.Ms
...
-// incentivize the relaying of the given packet.
-func (k Keeper) EscrowPacketFee(goCtx context.Context, msg *types.MsgEscrowPacketFee) (*types.MsgEscrowPacketFeeResponse, error) {
-	return &types.MsgEscrowPacketFeeResponse{}, nil
+// PayPacketFee defines a rpc handler method for MsgPayPacketFee
+// PayPacketFee is an open callback that may be called by any module/user that wishes to escrow funds in order to relay the packet with the next sequence
+func (k Keeper) PayPacketFee(goCtx context.Context, msg *types.MsgPayPacketFee) (*types.MsgPayPacketFeeResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// get the next sequence
+	sequence, found := k.GetNextSequenceSend(ctx, msg.SourcePortId, msg.SourceChannelId)
```
seantking, line 36: @AdityaSripal Is this what you had in mind for getting the sequence of the packet for the case whereby we want to pay for a packet that has a bundled 29-fee message + 20-transfer message?

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -25,9 +27,51 @@ func (k Keeper) RegisterCounterpartyAddress(goCtx context.Context, msg *types.Ms
...
+		return nil, err
+	}
+
+	return &types.MsgPayPacketFeeResponse{}, nil
+}
+
+// PayPacketFee defines a rpc handler method for MsgPayPacketFee
+// PayPacketFee is an open callback that may be called by any module/user that wishes to escrow funds in order to
+// incentivize the relaying of a known packet
+func (k Keeper) PayPacketFeeAsync(goCtx context.Context, msg *types.MsgPayPacketFeeAsync) (*types.MsgPayPacketFeeAsyncResponse, error) {
```
seantking, line 63: @colin-axner @AdityaSripal Although the internal logic between `PayPacket` & `PayPacketAsync` is not massively different, I think from an API perspective it makes sense to keep them separate. As far as the user is concerned it is very different functionality. 

In the first case, the user is bundling messages together in a single packet to pay for the packet at the next sequence. In the second case, the user wants to pay for a packet that has been committed already. 

wdyt?

proto/ibc/applications/fee/v1/fee.proto
```diff
@@ -10,15 +10,13 @@ option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
 // See Fee Payment Middleware spec:
 // https://github.com/cosmos/ibc/tree/master/spec/app/ics-029-fee-payment#fee-middleware-contract
 message Fee {
```
seantking, line 12: @AdityaSripal updated Fee structure as discussed. I'm wondering if this is even necessary as we could just have Fee be `Fee sdk.Coins` instead of 3 `sdk.Coin`

modules/apps/29-fee/keeper/msg_server.go
AdityaSripal: The public facing RPCs should definitely be different. but they can both call the same helper function

proto/ibc/applications/fee/v1/fee.proto
AdityaSripal: how would you break it down to what amount is for `recv`, `ack`, `timeout`?

modules/apps/29-fee/keeper/msg_server.go
AdityaSripal: Yes, I believe done this way it requires that the fee msg is placed first in the multimsg tx. This should be well-documented

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,132 @@
...
+	for _, f := range fees {
+		hasBalance := k.bankKeeper.HasBalance(ctx, refundAcc, f)
+		if !hasBalance {
+			return sdkerrors.Wrap(types.ErrBalanceNotFound, fmt.Sprintf("%s", refundAcc.String()))
+		}
+	}
+
+	for _, coin := range fees {
+		// escrow each fee with account module
+		if err := k.bankKeeper.SendCoinsFromAccountToModule(
```
seantking, line 35: @AdityaSripal I think we should probably do these sends atomically, so only send if all three are successful. Otherwise, we run into the same problem as below with the PayFee/PayFeeTimeout

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,132 @@
...
+	feeModuleAccAddr := k.authKeeper.GetModuleAddress(types.ModuleName)
+
+	// send ack fee to forward relayer
+	if feeInEscrow.AckFee != nil {
+		err := k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, refundAcc, sdk.Coins{*feeInEscrow.AckFee})
+		if err != nil {
+			return sdkerrors.Wrap(types.ErrFeeNotFound, fmt.Sprintf("%s", refundAcc.String()))
+		}
+		feeInEscrow.AckFee = nil
+		k.SetFeeInEscrow(ctx, feeInEscrow, refundAcc.String(), packetID.ChannelId, packetID.Sequence)
```
seantking, line 108: @AdityaSripal Same issue here when we return an error on SendCoin. We need to keep track of the state of the Fee (say in the case where AckFee is sent but the TimeoutFee fails, we don't want to lose track of the fact that AckFee was escrowed). 

I think the solution is to only do the send here in the case that all are successful. I was thinking we could use the cacheContext similar to how we are doing this with interchain accounts?

modules/apps/29-fee/keeper/escrow.go
AdityaSripal: You don't have to worry about that with the SDK. SDK transactions are already atomic (including multimsg txs). If you return an error during tx execution, all state changes made during tx execution get reverted

modules/apps/29-fee/keeper/escrow.go
AdityaSripal: checkout `baseapp.go` in the SDK. `runTx` already wraps entire tx execution in cache context.

What we did in ibc-go was because we wanted the tx to pass (and thus commit IBC state) while reverting any changes in callbacks. That's not necessary here. You already get atomicity out of the box with SDK

seantking: Ahhh very good. That makes things much simpler. Thank you :handshake:

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -34,14 +38,36 @@ message MsgRegisterCounterpartyAddress {
 message MsgRegisterCounterpartyAddressResponse {}
 
 // MsgEscrowPacketFee defines the request type EscrowPacketFee RPC
```
charleenfei, line 40: I think this description needs to get updated to reflect the new message name

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -16,18 +18,67 @@ var _ types.MsgServer = Keeper{}
...
+
+	packetId := &channeltypes.PacketId{
+		PortId:    msg.SourcePortId,
+		ChannelId: msg.SourceChannelId,
+		Sequence:  sequence,
+	}
+
+	refundAccAddr, err := sdk.AccAddressFromBech32(msg.RefundAccount)
+	if err != nil {
+		return &types.MsgPayPacketFeeResponse{}, err
```
seantking, line 54: same here re: test cov

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,236 @@
...
+		ackFee     *sdk.Coin
+		receiveFee *sdk.Coin
+		timeoutFee *sdk.Coin
+	)
+
+	// refundAcc does not have balance for the following Coin
+	invalidCoin := &sdk.Coin{Denom: "cosmos", Amount: sdk.NewInt(8000000000)}
+	validChannelId := "channel-0"
+
+	testCases := []struct {
```
AdityaSripal, line 22: Test cases where `RecvFee+AckFee > TimoutFee` and vice versa after implementing capital efficient escrow

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,236 @@
...
+		reverseRelayer sdk.AccAddress
+		forwardRelayer sdk.AccAddress
+		refundAcc      sdk.AccAddress
+	)
+
+	// refundAcc does not have balance for the following Coin
+	validChannelId := "channel-0"
+	validSeq := uint64(1)
+
+	testCases := []struct {
```
AdityaSripal, line 107: Test cases where `RecvFee+AckFee > TimoutFee` and vice versa after implementing capital efficient escrow

AdityaSripal, line 107: Test cases where `RecvFee+AckFee > TimoutFee` and vice versa after implementing capital efficient escrow

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,236 @@
...
+		packetId       *channeltypes.PacketId
+		reverseRelayer sdk.AccAddress
+		refundAcc      sdk.AccAddress
+	)
+
+	// refundAcc does not have balance for the following Coin
+	validChannelId := "channel-0"
+	validSeq := uint64(1)
+
+	testCases := []struct {
```
AdityaSripal, line 181: Test cases where `RecvFee+AckFee > TimoutFee` and vice versa after implementing capital efficient escrow

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	for _, tc := range testCases {
+		tc := tc
+
+		suite.Run(tc.name, func() {
+			suite.SetupTest() // reset
+
+			// setup
+			refundAcc = suite.chainA.SenderAccount.GetAddress()
+			reverseRelayer = suite.chainA.SenderAccount.GetAddress()
```
AdityaSripal, line 205: Use different address and check their balance

AdityaSripal, line 205:
```suggestion
			timeoutRelayer = suite.chainA.SenderAccount.GetAddress()
```

AdityaSripal: nit

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -97,9 +107,9 @@ func (k Keeper) IsFeeEnabled(ctx sdk.Context, portID, channelID string) bool {
 
 // SetCounterpartyAddress maps the destination chain relayer address to the source relayer address
 // The receiving chain must store the mapping from: address -> counterpartyAddress for the given channel
-func (k Keeper) SetCounterpartyAddress(ctx sdk.Context, address, counterpartyAddress string) {
+func (k Keeper) SetCounterpartyAddress(ctx sdk.Context, address string, counterpartyAddress sdk.AccAddress) {
```
AdityaSripal, line 110: This should remain a string. You don't know that the counterpartyAddress will be an sdk.AccAddress.

All you know is that address is `sdk.AccAddress`

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -113,3 +123,57 @@ func (k Keeper) GetCounterpartyAddress(ctx sdk.Context, address sdk.AccAddress)
...
+	fee := k.MustUnmarshalFee(bz)
+
+	// if the returned Fee is empty return false
+	if (types.Fee{}) == *fee.Fee {
+		return false
+	}
+
+	return true
+}
+
```
AdityaSripal, line 166: Create `DeleteFeeInEscrowMethod` and call `store.Delete(key)` in it

modules/apps/29-fee/types/keys.go
```diff
@@ -37,3 +40,8 @@ func FeeEnabledKey(portID, channelID string) []byte {
 func KeyRelayerAddress(address string) []byte {
 	return []byte(fmt.Sprintf("%s/%s", RelayerAddressKeyPrefix, address))
 }
+
+// KeyFeeInEscrow returns the key for escrowed fees
+func KeyFeeInEscrow(channelId string, sequenceId uint64) []byte {
+	return []byte(fmt.Sprintf("%s/%s/packet/%d", FeeInEscrowPrefix, channelId, sequenceId))
+}
```
AdityaSripal, line 47: This should include the portID in the key as well just to remain consistent with rest of IBC key structure.

modules/apps/29-fee/types/msgs.go
```diff
@@ -41,3 +43,52 @@ func (msg MsgRegisterCounterpartyAddress) GetSigners() []sdk.AccAddress {
...
+		SourcePortId:    sourcePortId,
+		SourceChannelId: sourceChannelId,
+		RefundAccount:   refundAccount,
+		Relayers:        relayers,
+	}
+}
+
+// ValidateBasic performs a basic check of the MsgPayPacketFee fields
+func (msg MsgPayPacketFee) ValidateBasic() error {
+	return nil
```
AdityaSripal, line 60: Check that:

- Fee not empty (some fields may be empty but not all of them)
- PortID and channelID are valid. See 24-host
- Refund account not empty
- Enforce relayers list is nil for now. We can remove that check once we implement relayer-specific fees

modules/apps/29-fee/types/msgs.go
```diff
@@ -41,3 +43,52 @@ func (msg MsgRegisterCounterpartyAddress) GetSigners() []sdk.AccAddress {
...
+		Fee:           fee,
+		PacketId:      packetId,
+		RefundAccount: refundAccount,
+		Relayers:      relayers,
+	}
+}
+
+// ValidateBasic performs a basic check of the MsgPayPacketFeeAsync fields
+func (msg MsgPayPacketFeeAsync) ValidateBasic() error {
+	return nil
```
AdityaSripal, line 84: ditto above checks.

Also check that sequence is not 0

proto/ibc/applications/fee/v1/fee.proto
```diff
@@ -10,15 +10,14 @@ option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
...
-  repeated cosmos.base.v1beta1.Coin amount = 1
-      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
+  cosmos.base.v1beta1.Coin receive_fee = 1 [(gogoproto.moretags) = "yaml:\"receive_fee\""];
```
AdityaSripal, line 13: Should these be `sdk.Coins` instead of `Coin`? cc: @colin-axner

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,118 @@
...
+// EscrowPacketFee sends the packet fee to the 29-fee module account to hold in escrow
+func (k Keeper) EscrowPacketFee(ctx sdk.Context, refundAcc sdk.AccAddress, identifiedFee types.IdentifiedPacketFee) error {
+	// check if the refund account exists
+	hasRefundAcc := k.authKeeper.GetAccount(ctx, refundAcc)
+	if hasRefundAcc == nil {
+		return sdkerrors.Wrap(types.ErrRefundAccNotFound, fmt.Sprintf("Account with address: %s not found", refundAcc.String()))
+	}
+
+	fees := sdk.Coins{
+		*identifiedFee.Fee.AckFee, *identifiedFee.Fee.ReceiveFee, *identifiedFee.Fee.TimeoutFee,
```
AdityaSripal, line 22: THis is wrong anyway. Since i think it will fail if the fees share a denom.

There's a helper function in `sdk/types` to add Coins together correctly.

Check out this file to help you: https://github.com/cosmos/cosmos-sdk/blob/master/types/coin.go

proto/ibc/applications/fee/v1/fee.proto
seantking: I was wondering this myself. Unsure what is the best practice here.

modules/apps/29-fee/keeper/escrow.go
seantking: > Since i think it will fail if the fees share a denom.

Tests seem to pass with this, interestingly. I was actually just adding the individual fee's to this Coins array to better loop over the fee's and thus reduce code. 

> Check out this file to help you: https://github.com/cosmos/cosmos-sdk/blob/master/types/coin.go

thanks :pray:

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	for _, coin := range fees {
+		// escrow each fee with account module
+		if err := k.bankKeeper.SendCoinsFromAccountToModule(
+			ctx, refundAcc, types.ModuleName, sdk.Coins{coin},
+		); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
seantking: Agree with @colin-axner :+1:

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -33,15 +37,34 @@ message MsgRegisterCounterpartyAddress {
...
+// MsgPayPacketFeeResponse defines the response type for Msg/PayPacketFee
+message MsgPayPacketFeeResponse {}
+
+// MsgPayPacketFeeAsync defines the request type PayPacketFeeAsync RPC
+message MsgPayPacketFeeAsync {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  // packet to pay fee for
+  ibc.applications.fee.v1.IdentifiedPacketFee identified_packet_fee = 1;
```
colin-axner, line 64: yaml

modules/apps/29-fee/types/keys.go
```diff
@@ -25,6 +29,9 @@ const (
 
 	// RelayerAddressKeyPrefix is the key prefix for relayer address mapping
 	RelayerAddressKeyPrefix = "relayerAddress"
+
+	// RelayerAddressKeyPrefix is the key prefix for relayer address mapping
```
colin-axner, line 33: godoc

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+func (k Keeper) EscrowPacketFee(ctx sdk.Context, refundAcc sdk.AccAddress, identifiedFee *types.IdentifiedPacketFee) error {
+	// check if the refund account exists
+	hasRefundAcc := k.authKeeper.GetAccount(ctx, refundAcc)
+	if hasRefundAcc == nil {
+		return sdkerrors.Wrap(types.ErrRefundAccNotFound, fmt.Sprintf("Account with address: %s not found", refundAcc.String()))
+	}
+
+	coins := identifiedFee.Fee.AckFee
+	coins = coins.Add(identifiedFee.Fee.ReceiveFee...)
+	coins = coins.Add(identifiedFee.Fee.TimeoutFee...)
```
colin-axner, line 24: can't you just used `coins := identifiedFee.Fee`?

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	err := k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, reverseRelayer, feeInEscrow.Fee.AckFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, fmt.Sprintf("error sending fee to reverse relayer"))
+	}
+
+	// send receive fee to forward relayer
+	err = k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, forwardRelayer, feeInEscrow.Fee.ReceiveFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, fmt.Sprintf("error sending fee to forward relayer"))
+	}
```
colin-axner, line 59: nit, I'd do the receive fee first. Naturally makes most sense to me since this was the first relay

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	// send ack fee to reverse relayer
+	err := k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, reverseRelayer, feeInEscrow.Fee.AckFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, fmt.Sprintf("error sending fee to reverse relayer"))
+	}
+
+	// send receive fee to forward relayer
+	err = k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, forwardRelayer, feeInEscrow.Fee.ReceiveFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, fmt.Sprintf("error sending fee to forward relayer"))
```
colin-axner, line 58: no fmt.Sprintf needed

proto/ibc/applications/fee/v1/tx.proto
seantking: updated :+1:

modules/apps/29-fee/types/keys.go
seantking: updated

modules/apps/29-fee/keeper/escrow.go
seantking: `identifiedFee.Fee` is not of type` sdk.Coins` it is three separate `sdk.Coins`

modules/apps/29-fee/keeper/escrow.go
seantking: done

modules/apps/29-fee/keeper/escrow.go
seantking: updated

modules/apps/29-fee/types/msgs.go
AdityaSripal: still need to check fee is not empty

modules/apps/29-fee/types/msgs.go
AdityaSripal: still need to check fee not empty

modules/apps/29-fee/types/msgs.go
```diff
@@ -41,3 +44,109 @@ func (msg MsgRegisterCounterpartyAddress) GetSigners() []sdk.AccAddress {
...
+	if err != nil {
+		return sdkerrors.Wrap(err, "failed to convert msg.Signer into sdk.AccAddress")
+	}
+
+	// enforce relayer is nil
+	if msg.IdentifiedPacketFee.Relayers != nil {
+		return ErrRelayersNotNil
+	}
+
+	// enforce relayer is nil
```
AdityaSripal, line 129:
```suggestion
	// ensure sequence is not 0
```

modules/apps/29-fee/types/msgs_test.go
```diff
@@ -34,3 +39,205 @@ func TestMsgRegisterCountepartyAddressValidation(t *testing.T) {
 		}
 	}
 }
+
+// TestRegisterCounterpartyAddressGetSigners tests GetSigners
+func TestRegisterCountepartyAddressGetSigners(t *testing.T) {
+	addr := sdk.AccAddress(secp256k1.GenPrivKey().PubKey().Address())
+	// build message
+	msg := NewMsgRegisterCounterpartyAddress(addr.String(), addr.String())
```
AdityaSripal, line 47:
```suggestion
	msg := NewMsgRegisterCounterpartyAddress(addr.String(), "counterparty")
```

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -33,15 +37,38 @@ message MsgRegisterCounterpartyAddress {
 // MsgRegisterCounterpartyAddressResponse defines the Msg/RegisterCounterypartyAddress response type
 message MsgRegisterCounterpartyAddressResponse {}
 
-// MsgEscrowPacketFee defines the request type EscrowPacketFee RPC
-message MsgEscrowPacketFee {
+// MsgPayPacketFee defines the request type EscrowPacketFee RPC
```
AdityaSripal, line 40:
```suggestion
// MsgPayPacketFee defines the request type PayPacketFee RPC
```

AdityaSripal (changes requested):
Great work!! Most are nits, only a few necessary changes

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -16,18 +17,66 @@ var _ types.MsgServer = Keeper{}
 func (k Keeper) RegisterCounterpartyAddress(goCtx context.Context, msg *types.MsgRegisterCounterpartyAddress) (*types.MsgRegisterCounterpartyAddressResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	counterpartyAddress, err := sdk.AccAddressFromBech32(msg.CounterpartyAddress)
+	if err != nil {
+		return &types.MsgRegisterCounterpartyAddressResponse{}, err
+	}
```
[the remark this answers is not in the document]
AdityaSripal: necessary

proto/ibc/applications/fee/v1/tx.proto
seantking: done

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,265 @@
...
+			suite.SetupTest() // reset
+
+			// setup
+			refundAcc = suite.chainA.SenderAccount.GetAddress()
+			reverseRelayer = sdk.AccAddress(secp256k1.GenPrivKey().PubKey().Address())
+			forwardRelayer = sdk.AccAddress(secp256k1.GenPrivKey().PubKey().Address())
+
+			ackFee = validCoins
+			receiveFee = validCoins
+			timeoutFee = validCoins
```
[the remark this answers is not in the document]
seantking: done :+1

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,265 @@
...
+		suite.Run(tc.name, func() {
+			suite.SetupTest() // reset
+
+			// setup
+			refundAcc = suite.chainA.SenderAccount.GetAddress()
+			timeoutRelayer := sdk.AccAddress(secp256k1.GenPrivKey().PubKey().Address())
+
+			ackFee = validCoins
+			receiveFee = validCoins
+			timeoutFee = validCoins
```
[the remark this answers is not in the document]
seantking: done

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
+package keeper
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
```
colin-axner, line 7:
```suggestion
```

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
+)
+
+// EscrowPacketFee sends the packet fee to the 29-fee module account to hold in escrow
+func (k Keeper) EscrowPacketFee(ctx sdk.Context, refundAcc sdk.AccAddress, identifiedFee *types.IdentifiedPacketFee) error {
```
colin-axner, line 15: Does identifiedFee need to be a pointer?

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	// check if the refund account exists
+	hasRefundAcc := k.authKeeper.GetAccount(ctx, refundAcc)
+	if hasRefundAcc == nil {
+		return sdkerrors.Wrap(types.ErrRefundAccNotFound, fmt.Sprintf("Account with address: %s not found", refundAcc))
```
colin-axner, line 19:
```suggestion
		return sdkerrors.Wrapf(types.ErrRefundAccNotFound, "account with address: %s not found", refundAcc)
```

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	coins = coins.Add(identifiedFee.Fee.TimeoutFee...)
+
+	if err := k.bankKeeper.SendCoinsFromAccountToModule(
+		ctx, refundAcc, types.ModuleName, coins,
+	); err != nil {
+		return err
+	}
+
+	// Store fee in state for reference later
+	// feeInEscrow/<port-id>/<channel-id>/packet/<sequence-id>/ -> Fee (timeout, ack, onrecv)
```
colin-axner, line 33: This can be removed imo. If we change the storage path, we will probably forget to remove this. We can indicate the storage path in the `SetFeeInEscrow` godoc if necessary

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	k.SetFeeInEscrow(ctx, identifiedFee)
+	return nil
+}
+
+// DistributeFee pays the acknowledgement fee & receive fee for a given packetId while refunding the timeout fee to the refund account associated with the Fee
+func (k Keeper) DistributeFee(ctx sdk.Context, refundAcc, forwardRelayer, reverseRelayer sdk.AccAddress, packetID *channeltypes.PacketId) error {
```
colin-axner, line 39: ditto on PacketId being a pointer?

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
 	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
+
```
colin-axner, line 13:
```suggestion
```

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -70,6 +75,11 @@ func (k Keeper) GetNextSequenceSend(ctx sdk.Context, portID, channelID string) (
 	return k.channelKeeper.GetNextSequenceSend(ctx, portID, channelID)
 }
 
+// GetFeeAccount returns the ICS29 Fee ModuleAccount address
```
colin-axner, line 78:
```suggestion
// GetFeeModuleAddress returns the ICS29 Fee ModuleAccount address
```

proto/ibc/applications/fee/v1/tx.proto
```diff
@@ -33,15 +37,38 @@ message MsgRegisterCounterpartyAddress {
...
 }
 
-// MsgEscrowPacketFeeResponse defines the response type for Msg/EscrowPacketFee
-message MsgEscrowPacketFeeResponse {}
+// MsgPayPacketFeeResponse defines the response type for Msg/PayPacketFee
+message MsgPayPacketFeeResponse {}
+
+// MsgPayPacketFeeAsync defines the request type PayPacketFeeAsync RPC
+// This Msg can be used to pay for a packet at a specified sequence (instead of the next sequence send)
+message MsgPayPacketFeeAsync {
```
colin-axner, line 62: should this message specify relayers?

modules/apps/29-fee/types/keys.go
```diff
@@ -37,3 +44,8 @@ func FeeEnabledKey(portID, channelID string) []byte {
 func KeyRelayerAddress(address string) []byte {
 	return []byte(fmt.Sprintf("%s/%s", RelayerAddressKeyPrefix, address))
 }
+
+// KeyFeeInEscrow returns the key for escrowed fees
+func KeyFeeInEscrow(packetID *channeltypes.PacketId) []byte {
```
colin-axner, line 49: I don't think packetID needs to be a pointer

modules/apps/29-fee/types/msgs.go
```diff
@@ -41,3 +44,135 @@ func (msg MsgRegisterCounterpartyAddress) GetSigners() []sdk.AccAddress {
...
+	}
+}
+
+// ValidateBasic performs a basic check of the MsgPayPacketFee fields
+func (msg MsgPayPacketFee) ValidateBasic() error {
+	// validate channelId
+	err := host.ChannelIdentifierValidator(msg.SourceChannelId)
+	if err != nil {
+		return err
+	}
```
colin-axner, line 64: nit/suggestion (and applied when necessary):
```suggestion
	if err := host.ChannelIdentifierValidator(msg.SourceChannelId); err != nil {
		return err
	}
```

modules/apps/29-fee/types/msgs.go
```diff
@@ -41,3 +44,135 @@ func (msg MsgRegisterCounterpartyAddress) GetSigners() []sdk.AccAddress {
...
+		return sdkerrors.ErrInvalidCoins
+	}
+
+	if !msg.Fee.ReceiveFee.IsValid() {
+		return sdkerrors.ErrInvalidCoins
+	}
+
+	if !msg.Fee.TimeoutFee.IsValid() {
+		return sdkerrors.ErrInvalidCoins
+	}
```
colin-axner, line 94: Should we check that one of the coins is non-zero?

modules/apps/29-fee/types/msgs.go
```diff
@@ -41,3 +44,135 @@ func (msg MsgRegisterCounterpartyAddress) GetSigners() []sdk.AccAddress {
...
+// GetSigners implements sdk.Msg
+func (msg MsgPayPacketFeeAsync) GetSigners() []sdk.AccAddress {
+	signer, err := sdk.AccAddressFromBech32(msg.Signer)
+	if err != nil {
+		panic(err)
+	}
+	return []sdk.AccAddress{signer}
+}
+
+func NewIdentifiedPacketFee(packetId *channeltypes.PacketId, fee Fee, relayers []string) *IdentifiedPacketFee {
```
colin-axner, line 171: maybe we should add a validate basic check for identified packet fee (and then do `msg.IdentifiedPacketFee.ValidateBasic()` in async msg

We could also add a ValidateBasic to Fee which checks that the coins are valid and at least one of the coins is non empty

modules/apps/29-fee/keeper/msg_server_test.go
```diff
@@ -0,0 +1,126 @@
...
+	testCases := []struct {
+		name     string
+		expPass  bool
+		malleate func()
+	}{
+		{
+			"success",
+			true,
+			func() {},
+		},
```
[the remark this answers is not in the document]
seantking: done :+1:

modules/apps/29-fee/types/msgs.go
colin-axner: could be done in a follow if we think this is desirable

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	// check if there is a Fee in escrow for the given packetId
+	feeInEscrow, found := k.GetFeeInEscrow(ctx, packetID)
+	if !found {
+		return sdkerrors.Wrap(types.ErrFeeNotFound, fmt.Sprintf("with channelID: %s, sequenceId %d", packetID.ChannelId, packetID.Sequence))
```
colin-axner, line 43:
```suggestion
		return sdkerrors.Wrapf(types.ErrFeeNotFound, "with channelID %s, sequence %d", packetID.ChannelId, packetID.Sequence)
```

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// get module accAddr
+	feeModuleAccAddr := k.authKeeper.GetModuleAddress(types.ModuleName)
+
+	// send receive fee to forward relayer
+	err := k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, forwardRelayer, feeInEscrow.Fee.ReceiveFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, "error sending fee to forward relayer")
```
colin-axner, line 52:
```suggestion
		return sdkerrors.Wrap(err, "failed to send fee to forward relayer")
```

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// send ack fee to reverse relayer
+	err = k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, reverseRelayer, feeInEscrow.Fee.AckFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, "error sending fee to reverse relayer")
```
colin-axner, line 58:
```suggestion
		return sdkerrors.Wrap(err, "failed to send fee to reverse relayer")
```

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// refund timeout fee to refundAddr
+	err = k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, refundAcc, feeInEscrow.Fee.TimeoutFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, "error refunding timeout fee")
```
colin-axner, line 64:
```suggestion
		return sdkerrors.Wrap(err, "failed to refund timeout fee")
```

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	return nil
+}
+
+// DistributeFeeTimeout pays the timeout fee for a given packetId while refunding the acknowledgement fee & receive fee to the refund account associated with the Fee
+func (k Keeper) DistributeFeeTimeout(ctx sdk.Context, refundAcc, timeoutRelayer sdk.AccAddress, packetID *channeltypes.PacketId) error {
+	// check if there is a Fee in escrow for the given packetId
+	feeInEscrow, hasFee := k.GetFeeInEscrow(ctx, packetID)
+	if !hasFee {
+		return sdkerrors.Wrap(types.ErrFeeNotFound, fmt.Sprintf("%s", refundAcc.String()))
+	}
```
colin-axner, line 79:
```suggestion
	feeInEscrow, found := k.GetFeeInEscrow(ctx, packetID)
	if !found {
		return sdkerrors.Wrap(types.ErrFeeNotFound, refundAcc)
	}
```
I think not having the `.String()` should work

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// get module accAddr
+	feeModuleAccAddr := k.authKeeper.GetModuleAddress(types.ModuleName)
+
+	// refund the receive fee
+	err := k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, refundAcc, feeInEscrow.Fee.ReceiveFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, "error refunding receive fee")
```
colin-axner, line 87:
```suggestion
		return sdkerrors.Wrap(err, "failed to refund receive fee")
```

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// refund the ack fee
+	err = k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, refundAcc, feeInEscrow.Fee.AckFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, "error refunding ack fee")
```
colin-axner, line 93:
```suggestion
		return sdkerrors.Wrap(err, "failed to refund ack fee")
```

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// pay the timeout fee to the reverse relayer
+	err = k.bankKeeper.SendCoins(ctx, feeModuleAccAddr, timeoutRelayer, feeInEscrow.Fee.TimeoutFee)
+	if err != nil {
+		return sdkerrors.Wrap(err, "error sending fee to timeout relayer")
```
colin-axner, line 99:
```suggestion
		return sdkerrors.Wrap(err, "failed to send fee to timeout relayer")
```

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -103,13 +113,63 @@ func (k Keeper) SetCounterpartyAddress(ctx sdk.Context, address, counterpartyAdd
...
+}
+
+// Deletes the fee associated with the given packetId
+func (k Keeper) DeleteFeeInEscrow(ctx sdk.Context, packetId *channeltypes.PacketId) {
+	store := ctx.KVStore(k.storeKey)
+	key := types.KeyFeeInEscrow(packetId)
+	store.Delete(key)
+}
+
+// GetFeeInEscrow returns true if there is a Fee still to be escrowed for a given packet
```
colin-axner, line 155:
```suggestion
// HasFeeInEscrow returns true if there is a Fee in escrow for a given packet
```

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,282 @@
+package keeper_test
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/ibc-go/modules/apps/29-fee/types"
+	channeltypes "github.com/cosmos/ibc-go/modules/core/04-channel/types"
```
colin-axner, line 6: nit import separation

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,282 @@
...
+			if tc.expPass {
+				feeInEscrow, _ := suite.chainA.GetSimApp().IBCFeeKeeper.GetFeeInEscrow(suite.chainA.GetContext(), packetId)
+				// check if the escrowed fee is set in state
+				suite.Require().True(feeInEscrow.Fee.AckFee.IsEqual(fee.AckFee))
+				suite.Require().True(feeInEscrow.Fee.ReceiveFee.IsEqual(fee.ReceiveFee))
+				suite.Require().True(feeInEscrow.Fee.TimeoutFee.IsEqual(fee.TimeoutFee))
+				// check if the fee is escrowed correctly
+				hasBalance := suite.chainA.GetSimApp().BankKeeper.HasBalance(suite.chainA.GetContext(), suite.chainA.GetSimApp().IBCFeeKeeper.GetFeeModuleAddress(), sdk.Coin{Denom: sdk.DefaultBondDenom, Amount: sdk.NewInt(600)})
+				suite.Require().True(hasBalance)
+				expectedBal := originalBal.Amount.Sub(sdk.NewInt(600))
```
colin-axner, line 92: I'd replace `600` with the value calculated from the fee coins

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -0,0 +1,282 @@
...
+				// check if the refund acc has sent the fee
+				hasBalance = suite.chainA.GetSimApp().BankKeeper.HasBalance(suite.chainA.GetContext(), refundAcc, sdk.Coin{Denom: sdk.DefaultBondDenom, Amount: expectedBal})
```
colin-axner, line 94: nit: use `sdk.NewCoin`

colin-axner (approved):
Excellent work!! Left many more nits/suggestions, but most can be addressed in a followup. 

My mains concern is that I don't think we should pass around pointers for IdentifiedPacketFee and PacketId. If it isn't necessary (we only use pointers for types which implement an interface because of Any's) then I don't think we should use pointers. 

I think code coverage can also be improved, but this pr is already massive so we should totally address code coverage concerns in a followup 

Approving as I think the code is good enough for now. Any unaddressed/missed concerns I have I'll mention during the internal audit

AdityaSripal (approved):
Great work!! The only thing that is needed is to check that **at least one** of the fees in `PayFeeMsg` is non-zero as @colin-axner mentioned.

But approving now, so that you can add this in quickly and then merge

modules/apps/29-fee/types/msgs.go
AdityaSripal: agreed this is needed for sync and async messages

seantking: good catch. I've updated this & added tests.

AdityaSripal:
So we should check that at least one of the fees is positive, but **all** fees should be non-negative. Otherwise, I don't know if there is an upstream check that prevents people from escrowing "negative" fees and stealing money

modules/apps/29-fee/types/msgs.go
```diff
@@ -80,8 +80,13 @@ func (msg MsgPayPacketFee) ValidateBasic() error {
 		return ErrRelayersNotNil
 	}
 
-	// if all three fee's are invalid return an error
-	if !msg.Fee.AckFee.IsValid() && !msg.Fee.ReceiveFee.IsValid() && !msg.Fee.TimeoutFee.IsValid() {
+	// if any of the fee's are invalid return an error
+	if !msg.Fee.AckFee.IsValid() || !msg.Fee.ReceiveFee.IsValid() || !msg.Fee.TimeoutFee.IsValid() {
+		return sdkerrors.ErrInvalidCoins
```
AdityaSripal, line 85: nit: wrap error with more description of what went wrong

AdityaSripal (approved):

Issue #451, opened 2021-09-29
Update portkey to include portID


seantking:
## Summary of Issue
We need to ensure that new ports that get bound are all stored under unique keys so they can be correctly exported/imported for genesis. We can achieve this by updating the `portID` mapping key to use `portKey + portID` instead of just the [portKey](https://github.com/cosmos/ibc-go/issues/new/choose?milestone=Interchain+Accounts).

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
Closed by #467

Issue #448, opened 2021-09-29
Update ICA genesis functionality + type


seantking:
## Summary
Update `genesis.go` functions to include the `owner address -> interchain-acc address` mapping &` active-channel` mapping. 

The genesis type (+ [proto definition](https://github.com/cosmos/ibc-go/blob/interchain-accounts/proto/ibc/applications/interchain_accounts/v1/genesis.proto) ) will also need updating. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #467, opened 2021-10-05
chore: update portkey to include port ID

damiannolan:
## Description

- Updating port keys to include port ID
- Separate key prefixes to vars

This was required as a precursor to the issue for implementing `InitGenesis` & `ExportGenesis`.
This should allow range queries using a prefix iterator. It's likely that `GetPort` should be replaced by `GetAllPorts` when tackling the GenesisState work.

closes: #451 

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/genesis.go
+++ b/modules/apps/27-interchain-accounts/genesis.go
@@ -22,7 +22,7 @@ func InitGenesis(ctx sdk.Context, keeper keeper.Keeper, state types.GenesisState
 
 // ExportGenesis exports transfer module's portID into its geneis state
 func ExportGenesis(ctx sdk.Context, keeper keeper.Keeper) *types.GenesisState {
-	portID := keeper.GetPort(ctx)
+	portID := keeper.GetPort(ctx, types.PortID)
 
 	return &types.GenesisState{
 		PortId: portID,
--- a/modules/apps/27-interchain-accounts/keeper/keeper.go
+++ b/modules/apps/27-interchain-accounts/keeper/keeper.go
@@ -97,15 +97,15 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 }
 
 // GetPort returns the portID for the interchain accounts module. Used in ExportGenesis
-func (k Keeper) GetPort(ctx sdk.Context) string {
+func (k Keeper) GetPort(ctx sdk.Context, portID string) string {
 	store := ctx.KVStore(k.storeKey)
-	return string(store.Get([]byte(types.PortKey)))
+	return string(store.Get(types.KeyPort(portID)))
 }
 
 // BindPort stores the provided portID and binds to it, returning the associated capability
 func (k Keeper) BindPort(ctx sdk.Context, portID string) *capabilitytypes.Capability {
 	store := ctx.KVStore(k.storeKey)
-	store.Set([]byte(types.PortKey), []byte(portID))
+	store.Set(types.KeyPort(portID), []byte(portID))
 
 	return k.portKeeper.BindPort(ctx, portID)
 }
--- a/modules/apps/27-interchain-accounts/keeper/keeper_test.go
+++ b/modules/apps/27-interchain-accounts/keeper/keeper_test.go
@@ -110,7 +110,7 @@ func (suite *KeeperTestSuite) TestIsBound() {
 }
 
 func (suite *KeeperTestSuite) TestGetPort() {
-	port := suite.chainA.GetSimApp().ICAKeeper.GetPort(suite.chainA.GetContext())
+	port := suite.chainA.GetSimApp().ICAKeeper.GetPort(suite.chainA.GetContext(), types.PortID)
 	suite.Require().Equal(types.PortID, port)
 }
 
--- a/modules/apps/27-interchain-accounts/types/keys.go
+++ b/modules/apps/27-interchain-accounts/types/keys.go
@@ -48,7 +48,7 @@ func KeyActiveChannel(portID string) []byte {
 	return []byte(fmt.Sprintf("%s/%s", ActiveChannelKeyPrefix, portID))
 }
 
-// KeyOwnerAccount creates and returns a new key used for owner account store operations
+// KeyOwnerAccount creates and returns a new key used for interchain account store operations
 func KeyOwnerAccount(portID string) []byte {
 	return []byte(fmt.Sprintf("%s/%s", OwnerKeyPrefix, portID))
 }
--- a/modules/apps/27-interchain-accounts/types/keys.go
+++ b/modules/apps/27-interchain-accounts/types/keys.go
@@ -28,8 +28,14 @@ const (
 )
 
 var (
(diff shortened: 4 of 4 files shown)


modules/apps/27-interchain-accounts/keeper/keeper.go
```diff
@@ -97,15 +97,15 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
 	store := ctx.KVStore(k.storeKey)
-	return string(store.Get([]byte(types.PortKey)))
+	return string(store.Get(types.KeyPort(portID)))
 }
 
 // BindPort stores the provided portID and binds to it, returning the associated capability
 func (k Keeper) BindPort(ctx sdk.Context, portID string) *capabilitytypes.Capability {
 	store := ctx.KVStore(k.storeKey)
-	store.Set([]byte(types.PortKey), []byte(portID))
+	store.Set(types.KeyPort(portID), []byte(portID))
```
colin-axner, line 108:
```suggestion
	store.Set(types.KeyPort(portID), []byte{0x01})
```
I think we should just store a single byte. The mapping can be confusing otherwise

modules/apps/27-interchain-accounts/keeper/keeper.go
```diff
@@ -97,15 +97,15 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 }
 
 // GetPort returns the portID for the interchain accounts module. Used in ExportGenesis
-func (k Keeper) GetPort(ctx sdk.Context) string {
+func (k Keeper) GetPort(ctx sdk.Context, portID string) string {
```
colin-axner, line 100: I don't think we need this function? Maybe `HasPort` instead?

modules/apps/27-interchain-accounts/genesis.go
```diff
@@ -22,7 +22,7 @@ func InitGenesis(ctx sdk.Context, keeper keeper.Keeper, state types.GenesisState
 
 // ExportGenesis exports transfer module's portID into its geneis state
 func ExportGenesis(ctx sdk.Context, keeper keeper.Keeper) *types.GenesisState {
-	portID := keeper.GetPort(ctx)
+	portID := keeper.GetPort(ctx, types.PortID)
```
colin-axner, line 25: Export Genesis should now be doing a range query on the prefix key and exporting all portIDs. Maybe we can add a todo and link the genesis ica issue?

modules/apps/27-interchain-accounts/keeper/keeper.go
damiannolan: Sure, I agree it definitely is confusing! Is there any reason we can't set the value as nil as if it were a set? Not much of a difference I guess ¯\_(ツ)_/¯

modules/apps/27-interchain-accounts/keeper/keeper.go
damiannolan: Correct, we don't need this! I left it in to satisfy the compiler and was planning on removing/adapting to `GetAllPorts` in a follow up for the genesis work.

modules/apps/27-interchain-accounts/genesis.go
damiannolan: Yep np! Done

seantking (approved):

modules/apps/27-interchain-accounts/keeper/keeper.go
AdityaSripal: Seems a bit wasteful to store the portID under

`ports/{portID} -> {portID}`

You could just do:

`ports/{portID} -> []byte{0x01}`

And then retrieve the portID from the key if the value is non-nil. Just a nit

AdityaSripal (changes requested):
Small nit, feel free to ignore

modules/apps/27-interchain-accounts/keeper/keeper.go
damiannolan: Yep, agreed. The value is being stored a single byte `[]byte{0x01}` on line 108. 

`GetPort` is a little redundant at the moment as the only place it is used is for the genesis state, but for #448 we'll want to export all ports so I will update in a follow up with the genesis state work to something like:

```
func (k Keeper) GetAllPorts(ctx sdk.Context) []string {
	iterator := sdk.KVStorePrefixIterator(store, []byte(types.PortKeyPrefix))

	var ports []string
		keySplit := strings.Split(string(iterator.Key()), "/")

		ports = append(ports, keySplit[1])
	}

	return ports
}
```

modules/apps/27-interchain-accounts/genesis.go
```diff
@@ -22,9 +22,10 @@ func InitGenesis(ctx sdk.Context, keeper keeper.Keeper, state types.GenesisState
 
 // ExportGenesis exports transfer module's portID into its geneis state
 func ExportGenesis(ctx sdk.Context, keeper keeper.Keeper) *types.GenesisState {
-	portID := keeper.GetPort(ctx)
+	// TODO: Using a range query with KVStorePrefixIterator export all port IDs
+	// See https://github.com/cosmos/ibc-go/issues/448
 
 	return &types.GenesisState{
-		PortId: portID,
+		PortId: types.PortID,
```
AdityaSripal, line 29: THis should be `PortIds` correct?

modules/apps/27-interchain-accounts/genesis.go
```diff
@@ -22,9 +22,10 @@ func InitGenesis(ctx sdk.Context, keeper keeper.Keeper, state types.GenesisState
 
 // ExportGenesis exports transfer module's portID into its geneis state
 func ExportGenesis(ctx sdk.Context, keeper keeper.Keeper) *types.GenesisState {
-	portID := keeper.GetPort(ctx)
+	// TODO: Using a range query with KVStorePrefixIterator export all port IDs
+	// See https://github.com/cosmos/ibc-go/issues/448
```
AdityaSripal, line 26: Looks like you implemented it below so you can use it here?

modules/apps/27-interchain-accounts/keeper/keeper_test.go
```diff
@@ -108,9 +108,17 @@ func (suite *KeeperTestSuite) TestIsBound() {
...
+	suite.SetupTest()
+	path := NewICAPath(suite.chainA, suite.chainB)
+	suite.coordinator.SetupConnections(path)
+
+	err := SetupICAPath(path, TestOwnerAddress)
+	suite.Require().NoError(err)
+
+	ports := suite.chainA.GetSimApp().ICAKeeper.GetAllPorts(suite.chainA.GetContext())
+	suite.Require().Contains(ports, types.PortID)
+	suite.Require().Contains(ports, TestPortID)
```
AdityaSripal, line 121: Try testing with multiple portIds in store

AdityaSripal (dismissed):
These can be addressed in a subsequent PR if you want to merge as-is

AdityaSripal (approved):

PR #474, opened 2021-10-07
ICA: Rename TxBody, Remove serialization logic from controller, introduce CosmosTx type

colin-axner:
## Description

closes: #449

---

write a little note why.


modules/apps/27-interchain-accounts/keeper/keeper.go
```diff
@@ -57,28 +57,18 @@ func NewKeeper(
 	}
 }
 
-// SerializeCosmosTx marshals data to bytes using the provided codec
-func (k Keeper) SerializeCosmosTx(cdc codec.BinaryCodec, msgs []sdk.Msg) ([]byte, error) {
-	msgAnys := make([]*codectypes.Any, len(msgs))
+// SerializeCosmosTx packs a slice of sdk.Msg into a slice of Any's.
+func (k Keeper) SerializeCosmosTx(cdc codec.BinaryCodec, msgs []sdk.Msg) (msgAnys []*codectypes.Any, err error) {
```
colin-axner, line 61: I need to update this to return the CosmosTx bytes. Should be called by ica-auth modules

proto/ibc/applications/interchain_accounts/v1/types.proto
```diff
@@ -25,3 +20,9 @@ message InterchainAccountPacketData {
   bytes data = 2;
```
colin-axner, line 20: should we rename to something else? Data is a little vague?

proto/ibc/applications/interchain_accounts/v1/types.proto
```diff
@@ -25,3 +20,9 @@ message InterchainAccountPacketData {
...
   string memo = 3;
 }
+
+// CosmosTx contains a list of sdk.Msg's. It should be used when sending transactions to an SDK host chain.
+message CosmosTx {
```
colin-axner, line 25: When trying to embed repeated `Any`s into the Packet Data type, I realized JSON marshalling will fail since the proto json codec will try to unpack the Any into its interface using the UnpackInterfaces implementation. This would require specifying what type of Any the messages are (ie sdk.Msg)

Requiring the controller chain to marshal the packet data also requires the controller chain to register every type sent to the host chain on its codec. This may become a large burden over time

Instead, I think the better solution is to push serialization logic onto the ica-auth module. This means each ica-auth module will need to be able to marshal the types they would like to send to the host chain. They pass in the ICA PacketData in the API of TrySendTx. The controller chain does very basic validation of this packet data and then passes this along after some basic checks

This prevents the controller chain from needing to register every type each host chain it is talking to has registered. Each ica-auth module only needs to register the types it uses and we will expose a serialization function ica-auth modules can all to get the data for the ica packet data

CosmosTx will only be used by ica-auth modules sending to SDK chains. The controller code should not need to understand this type (except to expose the helper function to serialize a CosmosTx)

modules/apps/27-interchain-accounts/keeper/relay.go
```diff
@@ -36,27 +36,10 @@ func (k Keeper) createOutgoingPacket(
...
-	}
-
-	var (
-		txBytes []byte
-		err     error
-	)
-
-	switch data := data.(type) {
-	case []sdk.Msg:
-		txBytes, err = k.SerializeCosmosTx(k.cdc, data)
```
colin-axner, line 53: ica-auth now need to call this function and format the ica packet data properly

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -1,9 +1,40 @@
...
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&iapd))
+}
+
+// GetBytes returns the JSON marshalled interchain account CosmosTx.
+func (ct CosmosTx) GetBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&ct))
+}
+
+// UnpackInterfaces implements UnpackInterfacesMessage.UnpackInterfaces
+func (ct CosmosTx) UnpackInterfaces(unpacker codectypes.AnyUnpacker) error {
```
colin-axner, line 31: If you type contains an Any or a struct that contains an Any, this function must be implemented or JSON marshal/unmarshal will fail (the compiler won't catch this, requires a test)

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -1,9 +1,40 @@
...
 
-func (packet InterchainAccountPacketData) GetBytes() []byte {
-	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&packet))
+// ValidateBasic performs basic validation of the interchain account packet data.
+// The memo may be empty.
+func (iapd InterchainAccountPacketData) ValidateBasic() error {
+	if iapd.Data == nil {
+		return sdkerrors.Wrap(ErrInvalidOutgoingData, "packet data cannot be empty")
+	}
+	// TODO: add type validation when data type enum supports unspecified type
```
colin-axner, line 15: should we add type validation? Might be better not to for future compatibility purposes

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -1,9 +1,40 @@
...
 import (
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-func (packet InterchainAccountPacketData) GetBytes() []byte {
-	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&packet))
+// ValidateBasic performs basic validation of the interchain account packet data.
+// The memo may be empty.
```
colin-axner, line 10: Should we enforce a max memo length?

damiannolan: Sounds like a good idea! What were you thinking, something like 256 chars?

proto/ibc/applications/interchain_accounts/v1/types.proto
damiannolan: Hmm! Maybe `tx_bytes`, `tx_payload`... I'm not sure 🤔

modules/apps/27-interchain-accounts/types/packet.go
seantking: > I think there's only a single value at the moment which is somewhere between an unspecified and specified enum TYPE_EXECUTE_TX_UNSPECIFIED laughing

Thought I updated this in another PR :thinking: We should definitely update this.

proto/ibc/applications/interchain_accounts/v1/types.proto
seantking: I like this approach & I think the idea to expose the deserialize/serial fns to the `ica-auth` module makes for a clean API. Nice work :+1:

seantking (approved):

modules/apps/27-interchain-accounts/types/packet.go
colin-axner: > This `ValidateBasic` happens on the `host` chain, right? Or is this validate function called manually?

Called on controller and host.

modules/apps/27-interchain-accounts/types/packet.go
colin-axner: Hmm, no idea. I'll do 256 for now and open an issue on cosmos/ibc

modules/apps/27-interchain-accounts/types/packet.go
damiannolan: Created #484

damiannolan (approved):

Issue #447, opened 2021-09-29
Nitpicks: ICA Audit (TrySendTxFlow)


seantking:
## Summary of Issue
The following nits arose as part of the audit.

- [x] TrySendTx (relay.go) channel not found should return for which port id
- [ ] ~~Return nil instead of []byte{} in all returns for keeper/relay.go~~
- [x] Require that we pass in an array of sdk.Msg instead of single sdk.Msg (keeper/keeper.go)
- [ ] ~~Add channel/port id in capability not found error (createOutgoingPacket - keeper/relay.go)~~
- [x] Add channel/port id in get next send sequence not found error (createOutgoingPacket - keeper/relay.go)
- [x] Potentially panic on errors that indicate bugs in code? (keeper/keeper.go)
- [x] ErrUnkownPacketData - ErrUnknownDataType (errors.go)
- [x] relay.go: AuthenticateTx: Get Interchain address first then loop through signers and return an error if the expected signer is not ICA address
- [x] relay.go: ExectureTx reduce code by returning when error occurs on executeMsg. Add comment for how cache context is functioning (atomic execution)
- [ ] ~~Fix error type and wrapping in module.go AcknowledgePacket~~

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #490, opened 2021-10-13
ica: AuthenticateTx/executeTx clean up

damiannolan:
## Description

Further clean up of nits from #447 

- relay.go: AuthenticateTx: Get Interchain address first then loop through signers and return an error if the expected signer is not ICA address
- relay.go: ExectureTx reduce code by returning when error occurs on executeMsg. Add comment for how cache context is functioning (atomic execution)

partially closes: #447 

---

write a little note why.


modules/apps/27-interchain-accounts/keeper/relay.go
```diff
@@ -99,60 +99,46 @@ func (k Keeper) DeserializeCosmosTx(_ sdk.Context, data []byte) ([]sdk.Msg, erro
...
 		return sdkerrors.ErrUnauthorized
 	}
 
-	for _, signer := range signers {
-		if interchainAccountAddr != signer.String() {
-			return sdkerrors.ErrUnauthorized
+	for _, msg := range msgs {
+		for _, signer := range msg.GetSigners() {
+			if interchainAccountAddr != signer.String() {
+				return sdkerrors.ErrUnauthorized
```
seantking, line 113: Good point about the errors not being wrapped. Maybe it would be nice to have an ICA-specific message here as well?

seantking (approved):

modules/apps/27-interchain-accounts/keeper/relay.go
damiannolan: Added error wrapping to `AuthenticateTx` 👍

modules/apps/27-interchain-accounts/keeper/relay.go
```diff
@@ -75,60 +75,46 @@ func (k Keeper) createOutgoingPacket(
...
+		return sdkerrors.Wrapf(types.ErrInterchainAccountNotFound, "failed to retrieve interchain account on port %s", portID)
 	}
 
-	for _, signer := range signers {
-		if interchainAccountAddr != signer.String() {
-			return sdkerrors.ErrUnauthorized
+	for _, msg := range msgs {
+		for _, signer := range msg.GetSigners() {
+			if interchainAccountAddr != signer.String() {
+				return sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "failed to authenticate interchain account transaction")
```
colin-axner, line 89: should we print the expected signer?

colin-axner (approved):

modules/apps/27-interchain-accounts/keeper/relay.go
seantking: My instinct is no as it could potentially allow a malicious party to find out what the associated address they're trying to access is. 

Maybe that is overly paranoid though?

colin-axner: This information is in the state machine, so it is public already

seantking: Yeah, I can't think of anything either, but unless there is a clear benefit to logging it out I usually air on the side of caution for these things. Is there any reason why knowing the associated address would be helpful?

Just an instinctual feeling :thinking:

damiannolan: It seems fine to me to include it. If anything it will help debug an issue if it comes to it. But in reality something has gone fundamentally wrong if we ever reach this error, in terms of handshake flow and data stored in state. 

I'll update to the following:
```go
return sdkerrors.Wrapf(sdkerrors.ErrUnauthorized, "unexpected signer address: expected %s, got %s", interchainAccountAddr, signer.String())
```

seantking: Fine with me then :+1: 

> But in reality something has gone fundamentally wrong if we ever reach this error
Won't this error be reached if a malicious controller chain tries to put a signer address that isn't an interchain account they own?

damiannolan: Yeah, good point! 💯

Issue #347, opened 2021-08-26
Support hooks in Transfer app


thomas-nguy:
## Summary

We would like to have the ability to define post-processing logic after an IBC transfer.   

## Problem Definition

Supporting hook would be quite beneficial for app developper that wants to include post-processing logic after an IBC transfer. For example, 
- Transform the voucher denomination to a more friendly coin denomination
- Wrap the voucher to a non ICS20 standard token (ex ERC20 for Ethermint)
- Consume the voucher & performs additional operation
- Add more tracking and logs
- Etc..

The design should be flexible and should not impact the current implementation logic.

## Proposal

Based on the hook implementation in GravityBridge, we propose a hook interface supporting those following methods 

```
type TransferHooks interface {
	AfterSendTransfer(
		ctx sdk.Context,
		sourcePort, sourceChannel string,
		token sdk.Coin,
		sender sdk.AccAddress,
		receiver string,
		isSource bool)
	AfterRecvTransfer(
		ctx sdk.Context,
		destPort, destChannel string,
		token sdk.Coin,
		receiver string,
		isSource bool)
	AfterRefundTransfer(
		ctx sdk.Context,
		sourcePort, sourceChannel string,
		token sdk.Coin,
		sender string,
		isSource bool)
}

```

Those hooks will be triggered after a Send, Receive and Refund events respectively.

Submitting a draft in a moment, It is open to discussion

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #533, opened 2021-11-10
feat: middleware support for ICS20

thomas-nguy:
## Description

Based on interchains-account implementation (https://github.com/cosmos/ibc-go/pull/417/files)

closes: #347

---

write a little note why.


modules/apps/transfer/keeper/msg_server.go
```diff
@@ -19,8 +22,14 @@ func (k Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types.
 	if err != nil {
 		return nil, err
 	}
+
+	channelCap, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(msg.SourcePort, msg.SourceChannel))
```
thomas-nguy, line 26: @colin-axner I am not totally sure how I should handle the channel cap here. I moved it to the msg_server level. Is there anything else that needs to be done?

colin-axner: This looks good to me, although I'm not sure it matters if ics20 continues to claim the channel capability in the channel handshake callbacks. 

@AdityaSripal Do we want to allow applications using ics20 as middleware to claim the capability?

fedekunze:
what's the status of this PR? @thomas-nguy

modules/apps/transfer/ibc_module.go
```diff
@@ -60,6 +61,11 @@ func ValidateTransferChannelParams(
 	return nil
 }
 
+// SetMiddleware sets ICS30 middleware
+func (im IBCModule) SetMiddleware(app porttypes.IBCModule) {
+	im.app = app
```
fedekunze, line 66: should this panic if already set?

fedekunze (approved):
LGTM, @AdityaSripal can you reply to the comment above?

@thomas-nguy can you add a changelog entry?

modules/apps/transfer/ibc_module.go
thomas-nguy: done

colin-axner:
To help push this along, I'm going to pull out some of the changes into a separate pr, primarily allowing transfer to hook up to a middleware like ics29. This change is very small and doesn't change any ics20 functionality (simply adding the ics4Wrapper interface)

I like the direction of this pr (excluding OnRecv from supported callbacks). I'd like to resolve the channel capability discussion with @AdityaSripal once he returns to work. I'm leaning towards only allowing applications connecting to ics20 from below to act as logging for now. I think logic that requires returning of acknowledgements (such as triggers ie sending an IBC packet) require more thoughts as we may need to change the ics20 acknowledgement format to wrap the returned acknowledgements

modules/apps/transfer/ibc_module.go
```diff
@@ -130,7 +145,12 @@ func (im IBCModule) OnChanOpenAck(
 	if counterpartyVersion != types.Version {
 		return sdkerrors.Wrapf(types.ErrInvalidVersion, "invalid counterparty version: %s, expected %s", counterpartyVersion, types.Version)
 	}
-	return nil
+
+	if im.app == nil {
+		return nil
+	}
+	// call underlying app's OnChanOpenAck callback with the counterparty app version.
+	return im.app.OnChanOpenAck(ctx, portID, channelID, counterpartyVersion)
```
colin-axner, line 153: What's the reasoning for now adding callbacks for OnChanOpenTry and OnChanOpenConfirm?

modules/apps/transfer/keeper/msg_server.go
colin-axner: @thomas-nguy Is it desirable for your use case for the connected application to be capable of sending packets? Otherwise I'm thinking maybe we should leave all the capability management as it was. Essentially only allowing the connected application to get information about the ICS20 transfer, but not act upon it

modules/apps/transfer/ibc_module.go
thomas-nguy: not much reason beside adding more flexibility, might be useful for the wired app? if not it should return nil anyways?

modules/apps/transfer/keeper/msg_server.go
thomas-nguy: I think I did it to preserve the original behavior.

Before it was claimed at "SendTransfer" level, but this line needs to be removed since the capability is now passed by parameter

AdityaSripal (changes requested):
For middleware architecture, the underlying base app must be the one to claim capability.

So in this case, we should have some sort of conditional logic that only claims capability if `im.app` is nil. This applies to both port and channnel capabilities.

Can we have some usecases presented to see how this will be used in real life so we can see if that is workable in those cases? cc: @fedekunze 

Also, please wrap all callbacks to allow maximum flexibility to underlying apps

modules/apps/transfer/keeper/msg_server.go
thomas-nguy:
```
So in this case, we should have some sort of conditional logic that only claims capability if im.app is nil. This applies to both port and channnel capabilities.
```

Do you refer to this section @AdityaSripal ?

The capability here is claim by the msg_server which is outside the middleware layer. Not sure to understand which check you want to do

fedekunze:
> Can we have some use cases presented to see how this will be used in real life so we can see if that is workable in those cases?

Our main use case is to enable IBC hooks that can execute trigger different functionality (eg: claiming airdrop tokens)

colin-axner:
@thomas-nguy sorry that this is just being brought up now, but have you thought about making your hooks be the middleware application?

I had suggested to @fedekunze to do this for claiming airdrop tokens and @AdityaSripal pointed out that this is also possible for post processing hooks. 

The middleware would simply call the ICS20 application callbacks and then would perform any custom logic as desired. Acting as a pass through module (not modifying acknowledgements and errors)

This would render this pr unnecessary. I think logically it makes sense that ICS20 is a base application that would never need to call additional logic. Any custom logic should simply wrap ICS20 (either performing the logic before or after calling ICS20). Does this make sense?

thomas-nguy:
@colin-axner thanks for your feedback.

No issue, I can close this PR if it become unnecessary?

Dumb question though, would it be possible to do post processing logic after a "SendTransfer" without this line change?
https://github.com/cosmos/ibc-go/pull/533/files#diff-ed533db5ed79787367690f88eacfb58bf5225e25a39cdebc40fc5eae26d7f1d6L180

I would requires to wrap the channel keeper for additional logic but that does not seems clean?

colin-axner:
> Dumb question though, would it be possible to do post processing logic after a "SendTransfer" without this line change?
https://github.com/cosmos/ibc-go/pull/533/files#diff-ed533db5ed79787367690f88eacfb58bf5225e25a39cdebc40fc5eae26d7f1d6L180

Not a dumb question. It is required. The ICS4Wrapper is the interface for middleware applications when we are doing logic which originates from the application and ends up at core IBC. 

In IBC there are two flows:
app -> core 
core -> app

app -> mw -> core utilizes ICS4Wrapper to achieve this (SendPacket/WriteAcknowledgement)
core -> mw -> app utilizes implementation of the 05-port IBC Module to achieve this (all On callbacks, OnRecv, OnAck etc)

When you hook up your middleware to ICS20, when the line for SendTransfer uses `ics4Wrapper.SendPacket`, then your function `SendPacket` will be called. Your function does its logic then call its ics4Wrapper (most likely the actual IBC channel Keeper). Does this make sense?

The change from calling `ChannelKeeper.SendPacket` to `ics4Wrapper.SendPacket` will be included in v3 (its API breaking so we cannot backport)

thomas-nguy:
thanks for the explanation, yes it is clear.

If the change from calling `ChannelKeeper.SendPacket` to `ics4Wrapper.SendPacket` is already done in another PR then that would work for me 👌 

I am closing this PR

colin-axner:
yup https://github.com/cosmos/ibc-go/pull/675 thanks again for starting this discussion

Issue #509, opened 2021-11-01
Utility to open new channel for an interchain-account


seantking:
## Summary
At the moment we do not have any built-in mechanism to re-open a channel and regain access to an `interchain-account` after a channel closes (for any reason). The spec outlines the following fn definition for re-opening a channel. 

```
// Opens a new channel on a particular port given a connection
// This is a helper function to open a new channel 
// This is a safety function in case of a channel closing and the controller chain needs to regain access to an interchain account on the host chain 
function InitChannel(portId: string, connectionId: string, counterpartyConnectionId) returns (nil){
  // An `OnChannelOpenInit` event is emitted which can be picked up by an offchain process such as a relayer which will finish the channel opening handshake
}
````
## Problem Definition

Each interchain account is assigned a single unique channel. If this channel closes it is not possible to access the account without opening a new channel and setting that channel as active for the particular account/owner pair (on both controller / host chains). We need to add the outlined helper function to create a new channel for a particular account. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

seantking:
@damiannolan Not sure how this got left out of the milestone as it has been in the spec for a while. 

@damiannolan @colin-axner the stub defined above is just a template. I'm open to ideas for naming/functionality here.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #534, opened 2021-11-11
feat: adding InitChannel util function

seantking:
## Description
Adding a helper function to open a new channel in the case that a channel closes. This is necessary so that a controller chain can regain access to an ICA. 

closes: #509 

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/keeper/account.go
+++ b/modules/apps/27-interchain-accounts/keeper/account.go
@@ -56,3 +56,16 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, accAddr sdk.AccAddres
 	k.accountKeeper.SetAccount(ctx, interchainAccount)
 	k.SetInterchainAccountAddress(ctx, controllerPortID, interchainAccount.Address)
 }
+
+// Opens a new channel on a particular port given a connection
+// This is a helper function to open a new channel
+// in case of a channel closing and the controller chain needs to regain access to an interchain account on the host chain
+func (k Keeper) InitChannel(ctx sdk.Context, portID, connectionID string) error {
+	msg := channeltypes.NewMsgChannelOpenInit(portID, types.VersionPrefix, channeltypes.ORDERED, []string{connectionID}, types.PortID, types.ModuleName)
+	handler := k.msgRouter.Handler(msg)
+	if _, err := handler(ctx, msg); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/modules/apps/27-interchain-accounts/keeper/account_test.go
+++ b/modules/apps/27-interchain-accounts/keeper/account_test.go
@@ -2,6 +2,8 @@ package keeper_test
 
 import (
 	"github.com/cosmos/ibc-go/v2/modules/apps/27-interchain-accounts/types"
+	channeltypes "github.com/cosmos/ibc-go/v2/modules/core/04-channel/types"
+	host "github.com/cosmos/ibc-go/v2/modules/core/24-host"
 	ibctesting "github.com/cosmos/ibc-go/v2/testing"
 )
 
@@ -53,7 +55,6 @@ func (suite *KeeperTestSuite) TestInitInterchainAccount() {
 			owner = TestOwnerAddress // must be explicitly changed
 			path = NewICAPath(suite.chainA, suite.chainB)
 			suite.coordinator.SetupConnections(path)
-
 			tc.malleate() // explicitly change fields in channel and testChannel
 
 			err = suite.chainA.GetSimApp().ICAKeeper.InitInterchainAccount(suite.chainA.GetContext(), path.EndpointA.ConnectionID, path.EndpointB.ConnectionID, owner)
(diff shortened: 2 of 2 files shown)


modules/apps/27-interchain-accounts/keeper/account.go
```diff
@@ -56,3 +56,16 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, accAddr sdk.AccAddres
 	k.accountKeeper.SetAccount(ctx, interchainAccount)
 	k.SetInterchainAccountAddress(ctx, controllerPortID, interchainAccount.Address)
 }
+
+// Opens a new channel on a particular port given a connection
+// This is a helper function to open a new channel
+// in case of a channel closing and the controller chain needs to regain access to an interchain account on the host chain
+func (k Keeper) InitChannel(ctx sdk.Context, portID, connectionID string) error {
```
seantking, line 63: Not sure about the naming here. I would like some opinions on this.

modules/apps/27-interchain-accounts/keeper/account.go
```diff
@@ -56,3 +56,16 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, accAddr sdk.AccAddres
...
+	msg := channeltypes.NewMsgChannelOpenInit(portID, types.VersionPrefix, channeltypes.ORDERED, []string{connectionID}, types.PortID, types.ModuleName)
```
seantking, line 64: I was initially considering doing a check to see if there is an active channel currently set and returning an err if so but on second thought I liked the idea of this being more of an  'admin' function capable of overriding the active channel and just creating a new channel (and thus setting a new active channel). 

I'm happy to add in this check and return an error if we think it's necessary. I'm leaning more towards putting this check in, but would like to hear what others think.

modules/apps/27-interchain-accounts/keeper/account_test.go
```diff
@@ -67,3 +68,74 @@ func (suite *KeeperTestSuite) TestInitInterchainAccount() {
...
+		tc := tc
+
+		suite.Run(tc.name, func() {
+			suite.SetupTest() // reset
+			path = NewICAPath(suite.chainA, suite.chainB)
+			suite.coordinator.SetupConnections(path)
+			portID, _ = types.GeneratePortID(TestOwnerAddress, path.EndpointA.ConnectionID, path.EndpointA.Counterparty.ConnectionID)
+
+			// bind port and claim capability
+			cap := suite.chainA.GetSimApp().ICAKeeper.BindPort(suite.chainA.GetContext(), portID)
```
seantking, line 103: Also not entirely sure what the best method for testing this out is. I initially made a call to `InitInterchainAccount` and finished the handshake, and then tried to call `InitChannel` and finish a second handshake, but I ran into verification issues. This is probs overkill anyway. 

Basically, I'm just binding the port fresh here and then checking that we can finish a handshake.

modules/apps/27-interchain-accounts/keeper/account_test.go
```diff
@@ -67,3 +68,74 @@ func (suite *KeeperTestSuite) TestInitInterchainAccount() {
...
+			// Init new channel
+			err = suite.chainA.GetSimApp().ICAKeeper.InitChannel(suite.chainA.GetContext(), portID, path.EndpointA.ConnectionID)
+
+			if tc.expPass {
+				suite.Require().NoError(err)
+
+				// finish the channel handshake
+
+				// commit state changes for proof verification
+				path.EndpointA.Chain.App.Commit()
```
seantking, line 121: maybe overkill finishing the handshake here?

seantking:
@AdityaSripal @colin-axner @damiannolan I'm actually questioning now if this is even necessary at all and maybe should be removed from the spec as well. 

In the case that a channel closes, an ICA-auth module could just call the `OnChanOpenInit` function directly I imagine. Or just use a relayer to re-open a channel. Although perhaps this util will make things a bit easier.

modules/apps/27-interchain-accounts/keeper/account.go
seantking: the onChanOpenInit function checks this anyway, so it's sort of a moot point actually

AdityaSripal:
Hmm it seems like the way you have it here is the cleanest option.

If you want to reopen the channel directly, you could do

```go
icaKeeper.channelKeeper.ChanOpenInit(...)
// callback doesn't get called by channelKeeper's ChanOpenInit function
// it gets called by the msg server
icaModule.OnChanOpenInit(...)
```

But sending the message to IBC's msgserver, just abstracts the above lines for you since that is basically what the message server does. And it seems like easier maintanence to me, if core IBC ever refactors its message handling, then you don't have to reimplement the channel msg server function within ICA

modules/apps/27-interchain-accounts/keeper/account.go
colin-axner: ReopenChannel? I'm not sure either, but I wouldn't want external users to start using `InitChannel` in place of `InitInterchainAccount`

modules/apps/27-interchain-accounts/keeper/account.go
```diff
@@ -56,3 +56,16 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, accAddr sdk.AccAddres
 	k.accountKeeper.SetAccount(ctx, interchainAccount)
 	k.SetInterchainAccountAddress(ctx, controllerPortID, interchainAccount.Address)
 }
+
+// Opens a new channel on a particular port given a connection
+// This is a helper function to open a new channel
+// in case of a channel closing and the controller chain needs to regain access to an interchain account on the host chain
```
colin-axner, line 62: godoc should begin with the name of the function

modules/apps/27-interchain-accounts/keeper/account_test.go
```diff
@@ -67,3 +68,74 @@ func (suite *KeeperTestSuite) TestInitInterchainAccount() {
 		})
 	}
 }
+
+func (suite *KeeperTestSuite) TestInitChannel() {
```
colin-axner, line 72: this should test that when a channel is closed and the portID is set to inactive, a new channel can be opened

modules/apps/27-interchain-accounts/keeper/account_test.go
colin-axner: I have no strong preference, but might be better to split this off into its own test, ie one test doing table tests and another finishing the handshake in the happy case

colin-axner:
@seantking I also question whether we need this. My hesitancy is that we don't fully understand when this might be needed. If the function is just creating a `MsgChannelOpenInit` then I wonder if it is best to let the caller handle this complexity (since it isn't very much). I don't have a preference though

modules/apps/27-interchain-accounts/keeper/account.go
seantking: Yeah, good point.

crodriguezvega:
Not sure if this is feasible, but would there be any way of using the `InitInterchainAccount` for this use case as well? Maybe the function needs to be renamed to something else more descriptive, but if we have only one function to initialize the interchain account whether the channel is open or not, then I guess the API is simplified for the user. For example, if the channel is closed and the caller calls `InitInterchainAccount` again (assuming with the same arguments as for the first time it was invoked), then you skip the port binding and do only the opening of the channel.

colin-axner:
One point of consideration is that this change could be released in a minor release. Perhaps it is best to not add it for now and add it later if it becomes desirable? Relayers could always just send the MsgChanOpenInit themselves?

seantking:
Closing this for the time being as we are deciding to leave this out of the API for the time being. @AdityaSripal We may want to remove this from the spec if we decide to leave it out of the release.

Issue #522, opened 2021-11-05
Add integration test for wrong handshake/packet initialition flows for interchain accounts


colin-axner:
## Summary

Add top level sanity tests which ensure that initiating a handshake or packet send on the wrong side should fail. This should include a basic sanity test, but it might also be possible to construct test cases which ensure we have all the right checks in place (think like a malicious attacker)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #538, opened 2021-11-15
ica: wrong handshake flow tests

colin-axner:
## Description

closes: #550 
closes: #522 

---

write a little note why.


modules/apps/27-interchain-accounts/interchain_accounts_test.go
```diff
@@ -0,0 +1,119 @@
...
+	suite.SetupTest() // reset
+	path := NewICAPath(suite.chainA, suite.chainB)
+	suite.coordinator.SetupConnections(path)
+
+	// use chainB (host) for ChanOpenInit
+	msg := channeltypes.NewMsgChannelOpenInit(path.EndpointB.ChannelConfig.PortID, types.VersionPrefix, channeltypes.ORDERED, []string{path.EndpointB.ConnectionID}, path.EndpointA.ChannelConfig.PortID, types.ModuleName)
+	handler := suite.chainB.GetSimApp().MsgServiceRouter().Handler(msg)
+	_, err := handler(suite.chainB.GetContext(), msg)
+
+	suite.Require().Error(err)
```
seantking, line 21: Am I correct in saying this only fails because we don't have a Path setup for Chain B -> Chain A?

modules/apps/27-interchain-accounts/interchain_accounts_test.go
```diff
@@ -0,0 +1,119 @@
...
+	suite.coordinator.SetupConnections(path)
+
+	err := InitInterchainAccount(path.EndpointA, TestOwnerAddress)
+	suite.Require().NoError(err)
+
+	// chainB also creates a controller port
+	err = InitInterchainAccount(path.EndpointB, TestOwnerAddress)
+	suite.Require().NoError(err)
+
+	path.EndpointA.UpdateClient()
```
seantking, line 39: Question: Why do we need to make use of this util here? I run into client issues sometimes when writing tests so would like to better understand. :handshake: 

Is this essentially just updating Chain A with the latest state from chain B?

modules/apps/27-interchain-accounts/interchain_accounts_test.go
```diff
@@ -0,0 +1,119 @@
...
+	path := NewICAPath(suite.chainA, suite.chainB)
+	suite.coordinator.SetupConnections(path)
+
+	err := InitInterchainAccount(path.EndpointA, TestOwnerAddress)
+	suite.Require().NoError(err)
+
+	err = path.EndpointB.ChanOpenTry()
+	suite.Require().NoError(err)
+
+	// chainA maliciously sets channel to TRYOPEN
```
seantking, line 64: Is the idea that by doing this a relayer will try and call `OnChanOpenAck` on the host chain as the next step? What would be the benefit of this to the malicious chain A? 

Just trying to better understand what we are testing here (other than the sanity check) :)

seantking (approved):

modules/apps/27-interchain-accounts/interchain_accounts_test.go
colin-axner: This fails because OpenInit is called on the host chain port ID

modules/apps/27-interchain-accounts/interchain_accounts_test.go
colin-axner: MsgChannelOpenTry is being delivered to chainA with proofs from chainB

We made state changes on chainB so in order for those proofs to pass on chainA, we need to update the light client on chainA to include a consensus state which can verify the new state changes

modules/apps/27-interchain-accounts/interchain_accounts_test.go
colin-axner: > Is the idea that by doing this a relayer will try and call `OnChanOpenAck` on the host chain as the next step?

The idea is that the relayer could call OnChanOpenAck on the controller chain using the host port ID. 

>  What would be the benefit of this to the malicious chain A?

The host portID would be marked as an active channel. I'm not sure if there is any benefit. This is mostly just a sanity check, the existing logic makes it difficult to confuse a host or controller chain

colin-axner:
> Nice! If you want to go ahead and merge this I can try port these tests to #541. Or if you think it's easier to wait until after we can do that too :)

I will wait!

colin-axner:
Updated to add invalid flow tests to both controller/host where applicable. Only Ack/Timeout needs code coverage in controller now

modules/apps/27-interchain-accounts/controller/ibc_module_test.go
```diff
@@ -189,6 +190,54 @@ func (suite *InterchainAccountsTestSuite) TestOnChanOpenInit() {
...
+	err := InitInterchainAccount(path.EndpointA, TestOwnerAddress)
+	suite.Require().NoError(err)
+
+	// chainB also creates a controller port
+	err = InitInterchainAccount(path.EndpointB, TestOwnerAddress)
+	suite.Require().NoError(err)
+
+	path.EndpointA.UpdateClient()
+	channelKey := host.ChannelKey(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
+	proofInit, proofHeight := path.EndpointB.Chain.QueryProof(channelKey)
```
damiannolan, line 213: So this `QueryProof` on the testchain just calls into ABCI for a proof that this channelKey is set in the store? 
Do I understand that correctly? 

What happens if it doesn't exist? Do you get an empty `[]byte` and height of -1 or something? 

Just asking out of my own curiosity :)

damiannolan (approved):
Awesome! Really appreciate you holding off on this for the submodules too, many thanks 🙌

modules/apps/27-interchain-accounts/controller/ibc_module_test.go
colin-axner: > So this QueryProof on the testchain just calls into ABCI for a proof that this channelKey is set in the store?
Do I understand that correctly?

Correct! That the key is set in store with the expected value

> What happens if it doesn't exist? Do you get an empty []byte and height of -1 or something?

It returns a non existence proof

damiannolan:
It looks like there's a failing test in the CI actually.

modules/apps/27-interchain-accounts/host/ibc_module_test.go
```diff
@@ -337,6 +458,114 @@ func (suite *InterchainAccountsTestSuite) TestOnRecvPacket() {
...
+
+			path := NewICAPath(suite.chainA, suite.chainB)
+			suite.coordinator.SetupConnections(path)
+
+			err := SetupICAPath(path, TestOwnerAddress)
+			suite.Require().NoError(err)
+
+			tc.malleate() // malleate mutates test data
+
+			module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID)
```
damiannolan, line 487: I think this fixes the failing test

```suggestion
			module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), path.EndpointB.ChannelConfig.PortID)
```

modules/apps/27-interchain-accounts/controller/ibc_module_test.go
damiannolan: Ah okay, great! So if you were to use the non existence proof then as argument to some other call, an error would eventually be propagated back to you I assume?

colin-axner: correct, the non existence proof should be rejected by the tendermint client

colin-axner:
Clicking auto-merge. The only non test code change has been there since the beginning (2 approvals). The test code changes seem fine to be merged with 1 approval

Issue #290, opened 2021-07-22
Create host and controller submodules


colin-axner:
## Summary

It can be difficult to understand which handlers belong to host and which to controller. The keeper package should have two files, host.go and controller.go. For example, `InitInterchainAccount` should go in `controller.go`. I'm fine with moving or leaving the handshakes as well

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
[Rename app root pkg definition also](https://github.com/cosmos/ibc-go/pull/417#discussion_r743541950)

AdityaSripal:
Agreed with this approach. I do the same thing for CCV. The parent and child modules are separated rather than keeping it in the same ibc module.

https://github.com/cosmos/interchain-security

damiannolan:
closed by #541

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #541, opened 2021-11-15
ICA controller/host submodules

damiannolan:
## Description

The following PR splits the interchain accounts functionality into two distinct submodules - `controller` and `host` with a shared `types` package maintained for common code. This also includes the splitting of `GenesisState` into both:
- `ControllerGenesisState`
- `HostGenesisState`

Unfortunately this is a very large PR due to the nature of the work. If there is an alternative approach we can take, then I can do my best to make that happen. It might be useful to explore the code outside the diff context by viewing the branch directly.

Splitting the code into their respective submodules uncovered some issues, the most concerning being the registration of shared types which I have tried to address in this PR with the recommendation of @AdityaSripal. However, the approach taken doesn't come without its own issues, a global package level var `IsRegistered` must be exported due to the way in which package initialisation works in Go. The runtime initialises and loads a package once, and only once. This becomes an issue for instantiation of multiple test chains in the ibc-go testing package. If a testchain `chainA` registers a number of types and sets `IsRegistered` equal to `true`, then a subsequent testchain `chainB` will skip this step. With this in mind, exporting the `IsRegistered` bool provides an opportunity to reset this flag (see testing/coordinator.go:42), and allows testchain `chainB` to register the appropriate types. In my opinion, this is not really an appropriate solution. However, for the purposes of our testing package it may be required. 
This problem also exists for the [interchain-security `ccv`](https://github.com/cosmos/interchain-security) module.

A number of subtle tweaks were required to the majority of tests throughout both submodules to accommodate the package splitting. These were mostly to do with module naming and capabilities, and ensuring the correct chain (chainA or chainB) was being invoked to make assertions on state changes.

The remaining work is primarily around separating some `controller` and `host` specific code in the shared `types` package into their respective nested `types` packages. 

closes: #290 

---

write a little note why.


damiannolan:
I'd like to note that I've disabled the ability to invoke `NegotiateAppVersion` on controller chains by simply returning an appropriate error. 

I've also noticed that we can probably remove the module account in the `controller` submodule also, as it's only really required inside of the `host` submodule afaik.

testing/simapp/app.go
```diff
@@ -133,7 +136,8 @@ var (
 		evidence.AppModuleBasic{},
 		transfer.AppModuleBasic{},
 		ibcmock.AppModuleBasic{},
-		ica.AppModuleBasic{},
+		icacontroller.AppModuleBasic{},
+		icahost.AppModuleBasic{},
```
colin-axner, line 140: What about having a single ICA AppModule but 2 ICA IBC modules? This should fix the registration issue as the codec registration is conflicting due to the AppModule using the same RegisterInterfaces function (I'm fairly certain). The genesis would become shared (I think this is fine)

damiannolan: That may be a perfect solution to the problem! I can look into this and follow up with you if I find any questions.

modules/apps/27-interchain-accounts/controller/ibc_module.go
```diff
@@ -66,16 +63,15 @@ func (im IBCModule) OnChanOpenTry(
 	version,
 	counterpartyVersion string,
 ) error {
-	return im.keeper.OnChanOpenTry(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version, counterpartyVersion)
+	return sdkerrors.Wrap(types.ErrInvalidChannelFlow, "channel handshake must be initiated by controller chain")
```
damiannolan, line 66: Controller submodule returns errors for both `OnChanOpenTry` and `OnChanOpenConfirm` handshake callbacks

modules/apps/27-interchain-accounts/controller/ibc_module.go
```diff
@@ -121,31 +115,15 @@ func (im IBCModule) OnChanCloseConfirm(
...
-	if ack.Success() {
-		err := im.keeper.OnRecvPacket(ctx, packet)
-		if err != nil {
-			ack = channeltypes.NewErrorAcknowledgement(err.Error())
-		}
-	}
-
-	// NOTE: acknowledgement will be written synchronously during IBC handler execution.
-	return ack
+	return channeltypes.NewErrorAcknowledgement("cannot receive packet on host controller chain")
```
damiannolan, line 123: Controller submodule returns a `channeltypes.NewErrorAcknowledgement()` for `OnRecvPacket`

modules/apps/27-interchain-accounts/controller/ibc_module.go
```diff
@@ -180,5 +156,5 @@ func (im IBCModule) NegotiateAppVersion(
 	counterparty channeltypes.Counterparty,
 	proposedVersion string,
 ) (string, error) {
-	return im.keeper.NegotiateAppVersion(ctx, order, connectionID, portID, counterparty, proposedVersion)
+	return "", sdkerrors.Wrap(types.ErrInvalidChannelFlow, "ICS-27 app version negotiation is unsupported on controller chains")
```
damiannolan, line 159: Controller submodule returns an error for invoking `NegotiateAppVersion`, for ics27 this should only be provided by the `host` submodule

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -0,0 +1,222 @@
...
+	return ctx.Logger().With("module", fmt.Sprintf("x/%s-%s", host.ModuleName, controllertypes.ModuleName))
+}
+
+// InitInterchainAccount is the entry point to registering an interchain account.
+// It generates a new port identifier using the owner address, connection identifier,
+// and counterparty connection identifier. It will bind to the port identifier and
+// call 04-channel 'ChanOpenInit'. An error is returned if the port identifier is
+// already in use. Gaining access to interchain accounts whose channels have closed
+// cannot be done with this function. A regular MsgChanOpenInit must be used.
+func (k Keeper) InitInterchainAccount(ctx sdk.Context, connectionID, counterpartyConnectionID, owner string) error {
```
damiannolan, line 71: The previous account.go file contained two functions:
- `InitInterchainAccount`
- `RegisterInterchainAccount`

As `InitInterchainAccount` belongs to the `controller` submodule and `RegisterInterchainAccount` to the `host`, I decided to move each to their respective keeper.go files. In my opinion this is a little cleaner, but I don't feel particularly strongly about it.

modules/apps/27-interchain-accounts/controller/types/keys.go
```diff
@@ -0,0 +1,9 @@
+package types
+
+const (
+	// ModuleName defines the interchain accounts controller module name
+	ModuleName = "icacontroller"
```
damiannolan, line 5: Here a new `ModuleName` is defined for the `controller` package. I'm open to making this `SubmoduleName` if anyone thinks it suits better, but I think this is fine. 

This is also used below for `StoreKey` as it is in other ibc and sdk modules. This is necessary to namespace the stores. Otherwise if both submodule stores were namespaced under the root module name, then an exported genesis calling into `GetAllPorts` would return the default host port - "ibcaccount" ("interchainaccount" when updated), and when a chain attempts to restart it would panic due to the capabilities with "port already bound".
It's important to note this wouldn't happen if a chain started with DefaultGenesis values, the issue would only occur with a custom/exported genesis file.

modules/apps/27-interchain-accounts/host/ibc_module.go
```diff
@@ -0,0 +1,143 @@
...
+	ctx sdk.Context,
+	order channeltypes.Order,
+	connectionHops []string,
+	portID string,
+	channelID string,
+	chanCap *capabilitytypes.Capability,
+	counterparty channeltypes.Counterparty,
+	version string,
+) error {
+	return sdkerrors.Wrap(types.ErrInvalidChannelFlow, "channel handshake must be initiated by controller chain")
```
damiannolan, line 37: Host submodule returns an error for both the `OnChanOpenInit` and `OnChanOpenAck` handshake callbacks

modules/apps/27-interchain-accounts/host/ibc_module.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// OnAcknowledgementPacket implements the IBCModule interface
+func (im IBCModule) OnAcknowledgementPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	acknowledgement []byte,
+	relayer sdk.AccAddress,
+) error {
+	return sdkerrors.Wrap(types.ErrInvalidChannelFlow, "cannot receive acknowledgement on host port, host chain does not send packet over channel")
```
damiannolan, line 121: Host submodule also returns an error for both `OnAcknowledgementPacket` and `OnTimeoutPacket`

modules/apps/27-interchain-accounts/host/ibc_module.go
```diff
@@ -0,0 +1,143 @@
...
+func (im IBCModule) OnRecvPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	_ sdk.AccAddress,
+) ibcexported.Acknowledgement {
+	ack := channeltypes.NewResultAcknowledgement([]byte{byte(1)})
+
+	// only attempt the application logic if the packet data
+	// was successfully decoded
+	if ack.Success() {
```
damiannolan, line 103: This was here previously but to me looks like a redundant check. 

We create the result acknowledgement above with `ack := channeltypes.NewResultAcknowledgement([]byte{byte(1)})` so it will always return `true` for `Success()`, right?

Let me know if I can remove this check and simplify to:

```
if err := im.keeper.OnRecvPacket(ctx, packet); err != nil {
	ack = channeltypes.NewErrorAcknowledgement(err.Error())
}
```

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
```diff
@@ -104,41 +106,51 @@ func TestKeeperTestSuite(t *testing.T) {
...
 	suite.coordinator.SetupConnections(path)
 
 	err := SetupICAPath(path, TestOwnerAddress)
 	suite.Require().NoError(err)
 
-	expectedPorts := []string{types.PortID, TestPortID}
+	expectedPorts := []string{types.PortID}
 
-	ports := suite.chainA.GetSimApp().ICAKeeper.GetAllPorts(suite.chainA.GetContext())
+	ports := suite.chainB.GetSimApp().ICAHostKeeper.GetAllPorts(suite.chainB.GetContext())
```
damiannolan, line 132: Many small updates to tests, a lot of which are subtle and require functions to just be called on the appropriate chain (chainB for the host submodule in most cases)

modules/apps/27-interchain-accounts/host/keeper/relay.go
```diff
@@ -0,0 +1,89 @@
...
+	return nil
+}
+
+func (k Keeper) executeTx(ctx sdk.Context, sourcePort, destPort, destChannel string, msgs []sdk.Msg) error {
+	if err := k.AuthenticateTx(ctx, msgs, sourcePort); err != nil {
+		return err
+	}
+
+	for _, msg := range msgs {
+		if err := msg.ValidateBasic(); err != nil {
```
damiannolan, line 36: I can move this for loop as we discussed yesterday on call or leave it for a follow up PR. Any preference?

modules/apps/27-interchain-accounts/types/account.go
```diff
@@ -11,71 +10,52 @@ import (
...
+	authtypes.AccountI
 }
 
-// GeneratePortID generates the portID for a specific owner
-// on the controller chain in the format:
-//
-// 'ics-27-<connectionSequence>-<counterpartyConnectionSequence>-<owner-address>'
-// https://github.com/seantking/ibc/tree/sean/ics-27-updates/spec/app/ics-027-interchain-accounts#registering--controlling-flows
-// TODO: update link to spec
-func GeneratePortID(owner, connectionID, counterpartyConnectionID string) (string, error) {
```
damiannolan, line 30: - Moved `GeneratePortID` to port.go
- Moved interface/struct definitions to top of file
- Added/updated godocs

modules/apps/27-interchain-accounts/types/errors.go
```diff
@@ -8,7 +8,7 @@ var (
 	ErrUnknownDataType             = sdkerrors.Register(ModuleName, 2, "unknown data type")
 	ErrAccountAlreadyExist         = sdkerrors.Register(ModuleName, 3, "account already exist")
 	ErrPortAlreadyBound            = sdkerrors.Register(ModuleName, 4, "port is already bound")
-	ErrUnsupportedChain            = sdkerrors.Register(ModuleName, 5, "unsupported chain")
+	ErrInvalidChannelFlow          = sdkerrors.Register(ModuleName, 5, "invalid message sent to channel end")
```
damiannolan, line 11: `ErrUnsupportedChain` was unused, so I replaced with the new error type which I copied from ccv as reference

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -31,10 +25,6 @@ const (
 
 	// Delimiter is the delimiter used for the interchain accounts version string
 	Delimiter = "."
-
-	// ControllerPortFormat is the expected port identifier format to which controller chains must conform
-	// See (TODO: Link to spec when updated)
-	ControllerPortFormat = "<app-version>.<controller-conn-seq>.<host-conn-seq>.<owner>"
```
damiannolan, line 37: Moved to port.go

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -48,11 +38,6 @@ var (
 	PortKeyPrefix = "port"
 )
 
-// NewVersion returns a complete version string in the format: VersionPrefix + Delimter + AccAddress
-func NewAppVersion(versionPrefix, accAddr string) string {
```
damiannolan, line 52: Moved `NewAppVersion` and `ParseAddressFromVersion` to version.go

modules/apps/27-interchain-accounts/types/keys.go
```diff
@@ -67,45 +52,3 @@ func KeyOwnerAccount(portID string) []byte {
 func KeyPort(portID string) []byte {
 	return []byte(fmt.Sprintf("%s/%s", PortKeyPrefix, portID))
 }
-
-// ParseControllerConnSequence attempts to parse the controller connection sequence from the provided port identifier
-// The port identifier must match the controller chain format outlined in (TODO: link spec), otherwise an empty string is returned
-func ParseControllerConnSequence(portID string) (uint64, error) {
```
damiannolan, line 73: Moved both `ParseControllerConnSequence` and `ParseHostConnSequence` to port.go

modules/apps/27-interchain-accounts/types/port.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return "", sdkerrors.Wrap(err, "invalid connection identifier")
+	}
+
+	counterpartyConnectionSeq, err := connectiontypes.ParseConnectionSequence(counterpartyConnectionID)
+	if err != nil {
+		return "", sdkerrors.Wrap(err, "invalid counterparty connection identifier")
+	}
+
+	return fmt.Sprint(
```
damiannolan, line 40: Maybe we should add a `NewControllerPortID` helper func?

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -0,0 +1,216 @@
...
+	portKeeper    types.PortKeeper
+	accountKeeper types.AccountKeeper
+
+	scopedKeeper capabilitykeeper.ScopedKeeper
+
+	msgRouter *baseapp.MsgServiceRouter
+}
+
+// NewKeeper creates a new interchain accounts controller Keeper instance
+func NewKeeper(
```
damiannolan, line 37: The module account is only required within the `host` submodule so I've removed the following check from `NewKeeper()` for the `controller`:

```
if addr := accountKeeper.GetModuleAddress(types.ModuleName); addr == nil {
	panic("the Interchain Accounts module account has not been set")
}
```

modules/apps/27-interchain-accounts/module.go
```diff
@@ -37,60 +43,70 @@ func (AppModuleBasic) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
...
 	AppModuleBasic
-	keeper keeper.Keeper
+	controllerKeeper *controllerkeeper.Keeper
+	hostKeeper       *hostkeeper.Keeper
 }
 
-// NewAppModule creates an interchain accounts app module.
-func NewAppModule(k keeper.Keeper) AppModule {
+// NewAppModule creates a new IBC interchain accounts module
+func NewAppModule(controllerKeeper *controllerkeeper.Keeper, hostKeeper *hostkeeper.Keeper) AppModule {
```
damiannolan, line 88: Both keepers are pointer types to allow for nil checking below in `InitGenesis` and `ExportGenesis`. 
You mentioned it may be okay to keep these as value types @colin-axner and control the enabling/disabling of controller/host via gov params, right?

colin-axner (approved):
Fantastic!!! :100: 

Amazing work!

I left many suggestions, but I'd prefer to have this merged as soon as possible, so all can be done in followups (if necessary)

modules/apps/27-interchain-accounts/controller/ibc_module.go
damiannolan: Good spot, copy pasta strikes again!

proto/ibc/applications/interchain_accounts/v1/genesis.proto
```diff
@@ -8,11 +8,24 @@ import "gogoproto/gogo.proto";
...
+message ControllerGenesisState {
   repeated ActiveChannel               active_channels     = 1 [(gogoproto.moretags) = "yaml:\"active_channels\""];
   repeated RegisteredInterchainAccount interchain_accounts = 2 [(gogoproto.moretags) = "yaml:\"interchain_accounts\""];
   repeated string                      ports               = 3;
 }
 
+// HostGenesisState defines the interchain accounts host genesis state
+message HostGenesisState {
+  repeated ActiveChannel               active_channels     = 1 [(gogoproto.moretags) = "yaml:\"active_channels\""];
+  repeated RegisteredInterchainAccount interchain_accounts = 2 [(gogoproto.moretags) = "yaml:\"interchain_accounts\""];
```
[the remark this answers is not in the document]
damiannolan: Cool, sounds good to me. This can be done using the `[(gogoproto.nullable) = false]` feature, right?

modules/apps/27-interchain-accounts/host/keeper/keeper.go
```diff
@@ -58,10 +57,27 @@ func NewKeeper(
 
 // Logger returns the application logger, scoped to the associated module
 func (k Keeper) Logger(ctx sdk.Context) log.Logger {
-	return ctx.Logger().With("module", fmt.Sprintf("x/%s-%s", host.ModuleName, types.ModuleName))
+	return ctx.Logger().With("module", fmt.Sprintf("x/%s-%s", host.ModuleName, hosttypes.ModuleName))
```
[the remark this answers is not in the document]
damiannolan: Sure thing! Dropping the alias will result in this being as evaluated as `"x/ibc-interchainaccounts"` as opposed to `"x/ibc-icahost"`. I think that's fine tho, I mean the logger isn't actually being used in here anyways! Will update both

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -0,0 +1,133 @@
...
+) error {
+
+	k.DeleteActiveChannelID(ctx, portID)
+
+	return nil
+}
+
+// validateControllerPortParams asserts the provided connection sequence and counterparty connection sequence
+// match that of the associated connection stored in state
+func (k Keeper) validateControllerPortParams(ctx sdk.Context, channelID, portID string, connectionSeq, counterpartyConnectionSeq uint64) error {
```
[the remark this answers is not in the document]
damiannolan: Agree! I was wondering how we could approach this to move some of the code to `types` without having to pass a channel keeper

modules/apps/27-interchain-accounts/host/ibc_module_test.go
```diff
@@ -104,93 +106,12 @@ func SetupICAPath(path *ibctesting.Path, owner string) error {
 	return nil
 }
 
-func (suite *InterchainAccountsTestSuite) TestOnChanOpenInit() {
```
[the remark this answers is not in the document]
damiannolan: 💯 

Figured these were better left until after

testing/simapp/app.go
damiannolan: @AdityaSripal maybe this approach can be used for `ccv` also! :)

modules/apps/27-interchain-accounts/module.go
```diff
@@ -100,22 +116,44 @@ func (am AppModule) LegacyQuerierHandler(legacyQuerierCdc *codec.LegacyAmino) sd
...
+
 	return []abci.ValidatorUpdate{}
 }
 
+// ExportGenesis returns the exported genesis state as raw bytes for the interchain accounts module
 func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.RawMessage {
-	gs := ExportGenesis(ctx, am.keeper)
+	var (
+		controllerGenesisState *types.ControllerGenesisState
+		hostGenesisState       *types.HostGenesisState
```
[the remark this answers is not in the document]
damiannolan: Ah yes! You're 100% right here. In fact, I think this could cause some problems if it marshalled using `nil` for one or the other and a chain attempted to restart with a genesis file provided. In that scenario the `DefaultGenesis` wouldn't be loaded upon restart, right?

modules/apps/27-interchain-accounts/module.go
seantking: @colin-axner does that mean a chain would have to pass a governance proposal in order to disable either host/controller functionality (and not be able to set this at module init)? 

I feel like it needs to be possible to set this at init time and then have the ability to change after the fact via on-chain params/gov prop.

modules/apps/27-interchain-accounts/host/ibc_module.go
seantking: super nit: Error message should probs be `cannot receive acknowledgment on host channel end, host chain does not send a packet over channel`. Also spelling error in acknowledgment :)

modules/apps/27-interchain-accounts/host/ibc_module.go
```diff
@@ -0,0 +1,138 @@
...
+	return sdkerrors.Wrap(types.ErrInvalidChannelFlow, "cannot receive acknowledgement on host port, host chain does not send packet over channel")
+}
+
+// OnTimeoutPacket implements the IBCModule interface
+func (im IBCModule) OnTimeoutPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	relayer sdk.AccAddress,
+) error {
+	return sdkerrors.Wrap(types.ErrInvalidChannelFlow, "cannot timeout packet on host port, host chain does not send packet over channel")
```
seantking, line 125: nit: `cannot cause a packet timeout on host channel end, a host chain does not send a packet over the channel`

modules/apps/27-interchain-accounts/host/keeper/relay.go
```diff
@@ -0,0 +1,89 @@
...
+			return err
+		}
+	}
+
+	writeCache()
+
+	return nil
+}
+
+// It tries to get the handler from router. And, if router exites, it will perform message.
```
seantking, line 55:
```suggestion
// Attempts to get the message handler from the router and if found will then execute the message
```

modules/apps/27-interchain-accounts/types/port.go
```diff
@@ -0,0 +1,78 @@
...
+const (
+	// ControllerPortFormat is the expected port identifier format to which controller chains must conform
+	// See (TODO: Link to spec when updated)
+	ControllerPortFormat = "<app-version>.<controller-conn-seq>.<host-conn-seq>.<owner>"
+)
+
+// GeneratePortID generates an interchain accounts controller port identifier for the provided owner
+// in the following format:
+//
+// 'ics-27-<connectionSequence>-<counterpartyConnectionSequence>-<owner-address>'
```
seantking, line 22:
```suggestion
// 'ics-27-1.<connectionSequence>.<counterpartyConnectionSequence>.<owner-address>'
```

seantking (approved):
Beautiful <3 Everything looks good to me. So much cleaner :1st_place_medal: 

I left a few nits for comments that stood out.

modules/apps/27-interchain-accounts/types/port.go
damiannolan: Will leave this one out as I want to remove it completely and link to the spec when merged. The format is also defined as a const at the top of file so will be fine to remove it and point to the spec

proto/ibc/applications/interchain_accounts/v1/genesis.proto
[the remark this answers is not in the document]
colin-axner: yup!

modules/apps/27-interchain-accounts/module.go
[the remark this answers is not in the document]
colin-axner: correct!

Issue #475, opened 2021-10-07
Fee Middleware: Closing Handshake


AdityaSripal:
Closing handshake still needs to be implemented once escrow logic is implemented

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #551, opened 2021-11-19
Fee Closing Handshake

AdityaSripal:
## Description

Refund all escrowed fees on the channel during channel closing
Panic if module account has insufficient funds

closes: #475 

---

write a little note why.


modules/apps/29-fee/ibc_module.go
```diff
@@ -130,8 +130,13 @@ func (im IBCModule) OnChanCloseInit(
 	portID,
 	channelID string,
 ) error {
-	// TODO: Unescrow all remaining funds for unprocessed packets
+	// delete fee enabled on channel
+	// and refund any remaining fees escrowed on channel
 	im.keeper.DeleteFeeEnabled(ctx, portID, channelID)
+	err := im.keeper.RefundFeesOnChannel(ctx, portID, channelID)
+	if err != nil {
+		panic(err)
```
AdityaSripal, line 138: This should only happen if there was a bug in the code and the module account has insufficient funds to refund all escrowed fees. So I panic here, open to a different suggestion on what to do

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -282,3 +282,53 @@ func (suite *KeeperTestSuite) TestDistributeTimeoutFee() {
...
+
+	ackFee := validCoins
+	receiveFee := validCoins2
+	timeoutFee := validCoins3
+
+	for i := 0; i < 5; i++ {
+		packetId := &channeltypes.PacketId{ChannelId: "channel-0", PortId: types.PortKey, Sequence: uint64(i)}
+		fee := types.Fee{receiveFee, ackFee, timeoutFee}
+
+		identifiedPacketFee := types.IdentifiedPacketFee{PacketId: packetId, Fee: fee, RefundAddress: refundAcc.String(), Relayers: []string{}}
```
seantking, line 301: does it matter that we're using the same packetId each time?

modules/apps/29-fee/ibc_module.go
seantking: Seems logical to me. Is there any way that the fee's could be refunded just before the channel closes? If so this would panic and we wouldn't call `im.app.OnChanCloseInit`

seantking (approved):

modules/apps/29-fee/ibc_module.go
charleenfei: why panic and not just return the error without closing the channel?

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -108,3 +108,40 @@ func (k Keeper) DistributeFeeTimeout(ctx sdk.Context, refundAcc, timeoutRelayer
 
 	return nil
 }
+
+func (k Keeper) RefundFeesOnChannel(ctx sdk.Context, portID, channelID string) error {
```
charleenfei, line 112: nit: the body of this function is really similar to the code in DistributeFee in escrow.go, maybe a way to extract the redundant code to a helper function

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -144,14 +144,29 @@ func (k Keeper) GetFeeInEscrow(ctx sdk.Context, packetId *channeltypes.PacketId)
 	return fee, true
 }
 
+// IterateChannelFeesInEscrow iterates over all the fees on the given channel currently escrowed and calls the provided callback
+// if the callback returns true, then iteration is stopped.
+func (k Keeper) IterateChannelFeesInEscrow(ctx sdk.Context, portID, channelID string, cb func(identifiedFee types.IdentifiedPacketFee) (stop bool)) {
```
charleenfei, line 149: nit: maybe a way to refactor GetFeeInEscrow or this function so they aren't as redundant

modules/apps/29-fee/types/keys.go
```diff
@@ -47,5 +47,10 @@ func KeyRelayerAddress(address string) []byte {
 
 // KeyFeeInEscrow returns the key for escrowed fees
 func KeyFeeInEscrow(packetID *channeltypes.PacketId) []byte {
-	return []byte(fmt.Sprintf("%s/%s/%s/packet/%d", FeeInEscrowPrefix, packetID.PortId, packetID.ChannelId, packetID.Sequence))
+	return []byte(fmt.Sprintf("%s/%d", KeyFeeInEscrowChannelPrefix(packetID.PortId, packetID.ChannelId), packetID.Sequence))
+}
+
+// KeyFeeInEscrowChannelPrefix returns the key prefix for escrowed fees on the given channel
+func KeyFeeInEscrowChannelPrefix(portID, channelID string) []byte {
```
charleenfei, line 54: 👍

charleenfei (approved):

modules/apps/29-fee/ibc_module.go
```diff
@@ -130,8 +130,20 @@ func (im IBCModule) OnChanCloseInit(
...
+	err := im.keeper.RefundFeesOnChannel(ctx, portID, channelID)
+	// error should only be non-nil if there is a bug in the code
+	// that causes module account to have insufficient funds to refund
+	// all escrowed fees on the channel.
+	// Disable all channels to allow for coordinated fix to the issue
+	// and mitigate/reverse damage.
+	// NOTE: Underlying application's packets will still go through, but
+	// fee module will be disabled for all channels
+	if err != nil {
+		im.keeper.DisableAllChannels(ctx)
```
AdityaSripal, line 145: Disabling all channels will effectively "turn off" the fee module. I added a check for `IsFeeEnabled` on `EscrowFee`.

Also, `OnAckPacket` and `OnTimeoutPacket` will check if fee is enabled on the channel before running middleware logic. So if we disable all channels, any further escrows will get rejected and all distributions will be bypassed. Thus, we can turn off the fee functionality without affecting underlying applications. IBC can still run even if fee module gets disabled due to invalid state

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -12,6 +12,10 @@ import (
 
 // EscrowPacketFee sends the packet fee to the 29-fee module account to hold in escrow
 func (k Keeper) EscrowPacketFee(ctx sdk.Context, identifiedFee *types.IdentifiedPacketFee) error {
+	if !k.IsFeeEnabled(ctx, identifiedFee.PacketId.PortId, identifiedFee.PacketId.ChannelId) {
```
AdityaSripal, line 15: Prevents further escrows if channel gets disabled

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -64,13 +55,14 @@ func (suite *KeeperTestSuite) TestEscrowPacketFee() {
...
 			suite.SetupTest() // reset
+			suite.coordinator.Setup(suite.path)
 
 			// setup
 			refundAcc = suite.chainA.SenderAccount.GetAddress()
 			ackFee = validCoins
 			receiveFee = validCoins2
 			timeoutFee = validCoins3
-			packetId := &channeltypes.PacketId{ChannelId: validChannelId, PortId: types.PortID, Sequence: uint64(1)}
+			packetId := &channeltypes.PacketId{ChannelId: suite.path.EndpointA.ChannelID, PortId: transfertypes.PortID, Sequence: uint64(1)}
```
AdityaSripal, line 65: Removed `types.PortID` since it was a holdover from earlier design. Now, the fee module does not own or bind to any ports. It is responsibility of an underlying app like transfer to bind to unique port

AdityaSripal, line 65: Switched to suite path's channelID, since it already is setup for us with fee enabled during channel handshake

AdityaSripal (commented):
Switched to disabling fee module if it enters invalid state rather than panic

Testing cleanup

modules/apps/29-fee/keeper/escrow_test.go
seantking: We should probs change this elsewhere also

modules/apps/29-fee/ibc_module.go
seantking: It would be nice to have `IsFeeEnabled` as a param we can change via governance. That way if a bug was discovered a chain could potentially shut down a channel. Although, maybe the proposal would take too long. 

Either way, I think this approach works better. Nice :+1:

modules/apps/29-fee/keeper/escrow_test.go
AdityaSripal: I removed it completely so it won't be anywhere else

modules/apps/29-fee/ibc_module.go
AdityaSripal: Potentially we could have a global enabled param, rather than a per-channel one. But that should be in a separate PR

modules/apps/29-fee/keeper/keeper_test.go
```diff
@@ -43,37 +52,62 @@ func (suite *KeeperTestSuite) SetupTest() {
...
+	ackFee := validCoins
+	receiveFee := validCoins2
+	timeoutFee := validCoins3
+	fee := types.Fee{ReceiveFee: receiveFee, AckFee: ackFee, TimeoutFee: timeoutFee}
 
-	if err := path.EndpointB.ChanOpenConfirm(); err != nil {
-		return err
+	// set some fees
+	for i := 1; i < 6; i++ {
+		packetId := channeltypes.NewPacketId(fmt.Sprintf("channel-1"), transfertypes.PortID, uint64(i))
```
seantking, line 67:
```suggestion
		packetId := channeltypes.NewPacketId(fmt.Sprintf(ibctesting.FirstChanneID), transfertypes.PortID, uint64(i))
```

seantking: Can do everywhere else `channel-1` is used also

modules/core/04-channel/types/packet.go
```diff
@@ -11,6 +11,11 @@ import (
 	"github.com/cosmos/ibc-go/modules/core/exported"
 )
 
+// NewPacketId returns a new instance of PacketId
```
seantking, line 14: I think @colin-axner will likely flag this as needing to be a separate PR into main with a changelog. We should probs add the validation fn here as well.

seantking: I'll move it over in my validation PR for now as well

seantking (approved):
Looks good. I like the changes. Nice work

modules/apps/29-fee/ibc_module.go
seantking: Agreed :+1:

modules/core/04-channel/types/packet.go
seantking: Actually, disregard that entirely. For some reason, I thought we made this change on main and not in the fee middleware branch -> https://github.com/cosmos/ibc-go/pull/272/files#diff-fc6caa5ba0a543336f3c02d081f09660a36e55412cca742e619e94ba959bf4c5R135

I must have been mixing this up with another change

PR #564, opened 2021-11-26
feat: add incentivised ack proto

charleenfei:
## Description

for use in packet callbacks and WriteAcknowledgement

closes: #XXXX

---

write a little note why.


proto/ibc/applications/fee/v1/ack.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package ibc.applications.fee.v1;
+import "gogoproto/gogo.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// IncentivizedAcknowledgement is the acknowledgement format to be used by applications wrapped in the fee middleware
+// It contains the raw acknowledgement bytes, as well as the forward relayer address
+message IncentivizedAcknowledgement {
+  bytes  result                  = 1;
```
seantking, line 10: Is this the underlying apps acknowledgment bytes? If so wondering if this is the right naming. 

This will need to be wrapped again in the generic acknowledgment type, right? Maybe I'm not understanding it properly.

proto/ibc/applications/fee/v1/ack.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package ibc.applications.fee.v1;
+import "gogoproto/gogo.proto";
+option go_package = "github.com/cosmos/ibc-go/modules/apps/29-fee/types";
+
+// IncentivizedAcknowledgement is the acknowledgement format to be used by applications wrapped in the fee middleware
+// It contains the raw acknowledgement bytes, as well as the forward relayer address
+message IncentivizedAcknowledgement {
```
colin-axner, line 9:
```suggestion
message Acknowledgement {
```
Perhaps maybe just use `Acknowledgement`? This type will always be referenced under its import path "github.com/cosmos/ibc-go/modules/apps/29-fee/types", so it won't collide with the channeltypes naming

proto/ibc/applications/fee/v1/ack.proto
charleenfei: just spoken to Aditya: confirmed that this does not need to be wrapped bc the version negotiation will be able differentiate which ack to use/return.

proto/ibc/applications/fee/v1/ack.proto
charleenfei: what would be the benefit of using Acknowledgement here? this is a custom Ack type which we will be using only if the versioning determines that the fee middleware is being used, so i think it would make sense to still keep it as a custom ack type

AdityaSripal: I think he just means naming here, its already a custom type since its in the `fee.proto` file. But I feel like its fine to explicitly call it `IncentivizedAcknowledgement` or `IncentivizedAck` so that its very clear what its purpose is

AdityaSripal (approved):

Issue #563, opened 2021-11-26
Deduplicate ICA controller/host keeper code


colin-axner:
## Summary

After splitting host/controller code into 2 submodules (amazing work by @damiannolan), we have some duplicated code. We also have 2 keepers. 2 keepers is nice for dividing storage space so the 2 submodules don't overlap. 

**Edit:** I tried the proposed solution and ran into a couple issues that would require increasing the complexity of the implementation

I propose we combine the keepers into 1 keepers and use 2 key prefixes (host and controller) and allow host/controller to use these prefixed stores. This way we can deduplicate a lot of code (functions like GetAllActiveChannels), we require less code in the app.go file (only 1 keeper needs to be created, no pointer needed), and shared functionality like on-chain parameters can be set at the top level of the ica module. 

The controller/host IBC modules will remained split (of course). The ICA keeper would instantiate a controller/host keeper. So in the app.go file, the changes would look something like:
```diff
diff --git a/testing/simapp/app.go b/testing/simapp/app.go
index 1955558..702162a 100644
--- a/testing/simapp/app.go
+++ b/testing/simapp/app.go
@@ -352,27 +352,21 @@ func NewSimApp(
        mockModule := ibcmock.NewAppModule(scopedIBCMockKeeper, &app.IBCKeeper.PortKeeper)
        mockIBCModule := ibcmock.NewIBCModule(&ibcmock.MockIBCApp{}, scopedIBCMockKeeper)
 
-       app.ICAControllerKeeper = icacontrollerkeeper.NewKeeper(
+       app.ICAKeeper = icakeeper.NewKeeper(
                appCodec, keys[icacontrollertypes.StoreKey],
                app.IBCKeeper.ChannelKeeper, // may be replaced with middleware such as ics29 fee
                app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
-               app.AccountKeeper, scopedICAControllerKeeper, app.MsgServiceRouter(),
+               app.AccountKeeper, scopedICAKeeper, app.MsgServiceRouter(),
        )
 
-       app.ICAHostKeeper = icahostkeeper.NewKeeper(
-               appCodec, keys[icahosttypes.StoreKey],
-               app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
-               app.AccountKeeper, scopedICAHostKeeper, app.MsgServiceRouter(),
-       )
-
[...]

PR #565, opened 2021-11-26
Introduce ICA Keeper (deduplicating host/controller code)

colin-axner:
## Description

Introduces ICAKeeper to deduplicate code within controller and host. The controller and host modules will be given a prefixed store from the ICAKeeper. Chains not wishing to run controller or host code do not need to create the respective keepers for those modules

This code allows for multiple controller sub modules to be created on the same chain!

closes: #563 

---

write a little note why.


modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -93,91 +69,52 @@ func (k Keeper) ClaimCapability(ctx sdk.Context, cap *capabilitytypes.Capability
 	return k.scopedKeeper.ClaimCapability(ctx, cap, name)
 }
 
-// GetActiveChannelID retrieves the active channelID from the store keyed by the provided portID
-func (k Keeper) GetActiveChannelID(ctx sdk.Context, portID string) (string, bool) {
-	store := ctx.KVStore(k.storeKey)
-	key := types.KeyActiveChannel(portID)
+// GetAllPorts returns all ports to which the interchain accounts controller module is bound. Used in ExportGenesis
+func (k Keeper) GetAllPorts(ctx sdk.Context) []string {
```
colin-axner, line 73: All duplicate keeper funcs in controller/host are now wrappers

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -32,17 +32,17 @@ type Keeper struct {
...
-	cdc codec.BinaryCodec, key sdk.StoreKey,
+	cdc codec.BinaryCodec, storePrefix string, icaKeeper icakeeper.Keeper,
 	ics4Wrapper types.ICS4Wrapper, channelKeeper types.ChannelKeeper, portKeeper types.PortKeeper,
-	accountKeeper types.AccountKeeper, scopedKeeper capabilitykeeper.ScopedKeeper, msgRouter *baseapp.MsgServiceRouter,
+	scopedKeeper capabilitykeeper.ScopedKeeper, msgRouter *baseapp.MsgServiceRouter,
 ) Keeper {
 	return Keeper{
-		storeKey:      key,
 		cdc:           cdc,
+		storePrefix:   storePrefix,
```
colin-axner, line 41: prefix used to create a prefixed store from the ICAKeeper

So, one store, but each ica submodule has its own namespace. This is actual quite useful. Now we can create many controller modules (currently require a new keeper for each)

modules/apps/27-interchain-accounts/keeper/keeper.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// prefixStore returns isolated prefix store for each ICA submodule so they can read/write in separate
+// namespace without being able to read/write to each other's data.
+func (k Keeper) prefixStore(ctx sdk.Context, storePrefix string) sdk.KVStore {
+	return prefix.NewStore(ctx.KVStore(k.storeKey), []byte(storePrefix))
+}
+
+// GetAllPorts returns all ports to which a interchain accounts submodule is bound. Used in ExportGenesis
+func (k Keeper) GetAllPorts(ctx sdk.Context, storePrefix string) []string {
```
colin-axner, line 48: I didn't change any logic in these functions except the first line. Replacing the ctx.KVStore(k.storeKey) with the prefixStore call

modules/apps/27-interchain-accounts/keeper/keeper_test.go
```diff
@@ -0,0 +1,233 @@
+package keeper_test
```
colin-axner, line 1: This test file is copied from controller keeper. Didn't change any logic except calling ICAKeeper directly with icatypes.ControllerModuleName

colin-axner (commented):
I think this is a nice improvement and cuts down on technical debt.

There's one more change necessary to enable support for many ica auth modules (each namespaced). To remove the pointer in the InitGenesis/ExportGenesis logic, we should replace the individual host/controller init/export with a single function at the ica keeper level. The genesis should contain a repeated list of ica submodule genesis, each of which contains the existing fields + the store prefix

My main concern about the store prefix is name collisions. We might want to have some sort of registration function to ensure there are no name collisions. Thoughts?

colin-axner:
After some more reflection, I worry a little too much about introducing reliance on store prefixes not colliding. Chains should be able to register multiple controller modules since they can create multiple IBC controller modules and the keeper functions all use portID for mappings, thus owner addresses could be equivalent across controller modules (assuming we modify the portID to namespace channelID as well). 

Thus, the main benefit would be code deduplication. I don't think this trade off is worth it right now given that the SDK is making significant changes to how the stores will work. Once the SDK moves to SMT, we should revisit this discussion. Closing this pr and I will update the issue with my thoughts

seantking:
Thanks for exploring this regardless. I learned a few things :)

Issue #523, opened 2021-11-05
Allow chains to disable controller or host functionality in interchain accounts


colin-axner:
## Summary

There should exist governance parameters which enable or disable host or controller chain functionality. If the controller chain functionality is enabled, the controller side code should ensure the controller authentication module is not nil. Passing in a nil authentication application to `ica.NewIBCModule` should be considered valid (this overrides the on chain parameter as no authentication module is provided)

Code changes needed to close issue:
- [x] adding host/controller parameters (#566)
- [ ] Add gRPC/cli for host/controller parameters (#569)
- [x] Add if statement checks to disable host or controller usage on handshake and packet callbacks
- [x] Validate params in genesis

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #566, opened 2021-11-26
add ica params

colin-axner:
## Description

closes: #523 

---

write a little note why.


modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -9,21 +9,24 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	capabilitykeeper "github.com/cosmos/cosmos-sdk/x/capability/keeper"
 	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 	"github.com/tendermint/tendermint/libs/log"
 
-	"github.com/cosmos/ibc-go/v2/modules/apps/27-interchain-accounts/types"
+	"github.com/cosmos/ibc-go/v2/modules/apps/27-interchain-accounts/controller/types"
+	icatypes "github.com/cosmos/ibc-go/v2/modules/apps/27-interchain-accounts/types"
```
colin-axner, line 16: this should be applied to all of controller/host in a separate pr

modules/apps/27-interchain-accounts/controller/types/params.go
```diff
@@ -0,0 +1,59 @@
...
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+)
+
+const (
+	// DefaultControllerEnabled is the default value for the controller param (set to true)
+	DefaultControllerEnabled = true
+)
+
+var (
+	// KeyControllerEnabled is store's key for ControllerEnabled Params
```
damiannolan, line 15: tiny nit:

```suggestion
	// KeyControllerEnabled is the store key for ControllerEnabled Params
```

modules/apps/27-interchain-accounts/controller/types/params.go
```diff
@@ -0,0 +1,59 @@
...
+	KeyControllerEnabled = []byte("ControllerEnabled")
+)
+
+// ParamKeyTable type declaration for parameters
+func ParamKeyTable() paramtypes.KeyTable {
+	return paramtypes.NewKeyTable().RegisterParamSet(&Params{})
```
damiannolan, line 21: So there's no registration issues here between controller/host as `NewKeyTable()` is creating separate instances, correct?

modules/apps/27-interchain-accounts/host/types/params.go
```diff
@@ -0,0 +1,59 @@
...
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+)
+
+const (
+	// DefaultHostEnabled is the default value for the host param (set to true)
+	DefaultHostEnabled = true
+)
+
+var (
+	// KeyHostEnabled is store's key for HostEnabled Params
```
damiannolan, line 15: Same nit here: 

```suggestion
	// KeyHostEnabled is the store key for HostEnabled Params
```

proto/ibc/applications/interchain_accounts/v1/types.proto
```diff
@@ -29,3 +29,4 @@ message InterchainAccountPacketData {
 message CosmosTx {
   repeated google.protobuf.Any messages = 1;
 }
+
```
damiannolan, line 32: Side note, but I think we should rename this file to `packet.proto`. In my opinion it makes more sense and is more consistent with transfer for example. 

What do you think?

modules/apps/27-interchain-accounts/controller/keeper/params.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/cosmos/ibc-go/v2/modules/apps/27-interchain-accounts/controller/types"
+)
+
+// GetControllerEnabled retrieves the host enabled boolean from the paramstore
+func (k Keeper) GetControllerEnabled(ctx sdk.Context) bool {
+	var res bool
+	k.paramSpace.Get(ctx, types.KeyControllerEnabled, &res)
```
damiannolan, line 12: Interesting API in the params module to pass a pointer instead of returning bytes. Is this purely to abstract the unmarshalling of types from the caller?

modules/apps/27-interchain-accounts/host/types/params.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+)
+
+const (
+	// DefaultHostEnabled is the default value for the host param (set to true)
+	DefaultHostEnabled = true
```
damiannolan, line 11: So host is enabled by default and controller must be enabled via gov proposal?

damiannolan (approved):
Awesome, super clean work and very much appreciate this as a starting point for the host msg whitelist! LGTM and follow up on the icatypes import alias :)

One last question, is there a default params proposal handler in the sdk that allows us to change any given module's params, or does one need to be implemented on a per module basis? 

We can also probably create an issue for CLI queries for controller/host params.

colin-axner:
> One last question, is there a default params proposal handler in the sdk that allows us to change any given module's params, or does one need to be implemented on a per module basis?

Yes there is! A parameter change proposal. Can be accessed via cli `simd tx gov submit-proposal param-change`

> We can also probably create an issue for CLI queries for controller/host params.

Yea, might be nice to have in separate pr's. I'll modify the existing issue to include check boxes for the things that need to be done

modules/apps/27-interchain-accounts/controller/types/params.go
colin-axner: Correct! The param types are also not equal

proto/ibc/applications/interchain_accounts/v1/types.proto
colin-axner: Sounds good to me! I didn't like that vagueness of `types.proto`

modules/apps/27-interchain-accounts/controller/keeper/params.go
colin-axner: No idea, but it is an interesting design decision. The reasoning for this decision is likely buried deep in the SDK (if ever documented)

modules/apps/27-interchain-accounts/host/types/params.go
colin-axner: Turns out ics20 is enabled by default so I copied that. Both host/controller are currently enabled by default. We could set this to false. We will need to add some code into our tests so that we don't use the default value (if we set it to false)

modules/apps/27-interchain-accounts/controller/types/params.go
damiannolan: Ah yes! I noticed after I wrote this comment the separation of host/controller subpackages in the protos :)

modules/apps/27-interchain-accounts/types/genesis_test.go
```diff
@@ -85,7 +87,7 @@ func (suite *TypesTestSuite) TestValidateControllerGenesisState() {
 					},
 				}
 
-				genesisState = types.NewControllerGenesisState(activeChannels, []types.RegisteredInterchainAccount{}, []string{})
+				genesisState = types.NewControllerGenesisState(activeChannels, []types.RegisteredInterchainAccount{}, []string{}, controllertypes.DefaultParams())
```
colin-axner, line 90: Will add genesis validation tests in a followup pr

modules/apps/27-interchain-accounts/types/genesis.go
```diff
@@ -80,6 +80,10 @@ func (gs ControllerGenesisState) Validate() error {
 		}
 	}
 
+	if err := gs.Params.Validate(); err != nil {
```
colin-axner, line 83: these technically cannot return an error. Validate only checks that params contain a bool as its field (the compiler would break if it didn't)

Keeping here for when we add more params which do validation

seantking (approved):
Nice and clean and I learned something new. Great :handshake:

modules/apps/27-interchain-accounts/types/genesis.go
damiannolan: No problem, I'll likely have something ready today. I'll just need to pull in your latest changes. Ack the additions :)

PR #568, opened 2021-11-29
use local clock time as reference time for timeout timestamp if later than consensus state timestamp

crodriguezvega:
## Description

In the IBC transfer cli command use the local clock time as reference time to calculate the timeout timestamp on the counter party chain if it is later than the consensus state timestamp of the counter party chain, otherwise still use consensus state timestamp as reference.

closes: #508

---

write a little note why.

### DIFF
--- a/modules/apps/transfer/client/cli/tx.go
+++ b/modules/apps/transfer/client/cli/tx.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -104,8 +105,7 @@ corresponding to the counterparty channel. Any timeout set to 0 is disabled.`),
 							timeoutTimestamp = consensusStateTimestamp + timeoutTimestamp
 						}
 					} else {
-						// rare case: if local clock time is equal to or earlier than Jan 1st, 1970 12:00 AM
-						timeoutTimestamp = consensusStateTimestamp + timeoutTimestamp
+						return errors.New("local clock time is not greater than Jan 1st, 1970 12:00 AM")
 					}
 				}
 			}
--- a/modules/apps/transfer/client/cli/tx.go
+++ b/modules/apps/transfer/client/cli/tx.go
@@ -103,6 +103,7 @@ to the counterparty channel. Any timeout set to 0 is disabled.`),
 							timeoutTimestamp = consensusStateTimestamp + timeoutTimestamp
 						}
 					} else {
+						// rare case: if local clock time is equal to or earlier than Jan 1st, 1970 12:00 AM
 						timeoutTimestamp = consensusStateTimestamp + timeoutTimestamp
 					}
 				}
--- a/modules/apps/transfer/client/cli/tx.go
+++ b/modules/apps/transfer/client/cli/tx.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -89,7 +90,21 @@ to the counterparty channel. Any timeout set to 0 is disabled.`),
 				}
 
 				if timeoutTimestamp != 0 {
-					timeoutTimestamp = consensusState.GetTimestamp() + timeoutTimestamp
+					// use local clock time as reference time if it is later than the
+					// consensus state timestamp of the counter party chain, otherwise
+					// still use consensus state timestamp as reference
+					now := time.Now().UnixNano()
+					consensusStateTimestamp := consensusState.GetTimestamp()
+					if now > 0 {
+						now := uint64(now)
+						if now > consensusStateTimestamp {
+							timeoutTimestamp = now + timeoutTimestamp
+						} else {
+							timeoutTimestamp = consensusStateTimestamp + timeoutTimestamp
+						}
+					} else {
+						timeoutTimestamp = consensusStateTimestamp + timeoutTimestamp
+					}
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


modules/apps/transfer/client/cli/tx.go
```diff
@@ -89,7 +90,21 @@ to the counterparty channel. Any timeout set to 0 is disabled.`),
 				}
 
 				if timeoutTimestamp != 0 {
-					timeoutTimestamp = consensusState.GetTimestamp() + timeoutTimestamp
+					// use local clock time as reference time if it is later than the
+					// consensus state timestamp of the counter party chain, otherwise
+					// still use consensus state timestamp as reference
+					now := time.Now().UnixNano()
+					consensusStateTimestamp := consensusState.GetTimestamp()
+					if now > 0 {
```
AdityaSripal, line 98: why would now be 0?

AdityaSripal (approved):

modules/apps/transfer/client/cli/tx.go
```diff
@@ -89,7 +90,21 @@ to the counterparty channel. Any timeout set to 0 is disabled.`),
 				}
 
 				if timeoutTimestamp != 0 {
-					timeoutTimestamp = consensusState.GetTimestamp() + timeoutTimestamp
+					// use local clock time as reference time if it is later than the
+					// consensus state timestamp of the counter party chain, otherwise
+					// still use consensus state timestamp as reference
+					now := time.Now().UnixNano()
```
damiannolan, line 96: I think its probably safe to convert to `uint64` here:

```suggestion
					now := uint64(time.Now().UnixNano())
```

modules/apps/transfer/client/cli/tx.go
crodriguezvega: I was being very defensive. It could potentially be 0 if the clock of the machine is (for who knows what reason) on Jan 1st 1970 12:00 AM. And it could be negative if the clock is set before that date time.

modules/apps/transfer/client/cli/tx.go
crodriguezvega: I know it's very unlikely, but if the clock of the machine is set to a date time before Jan 1st 1970 12:00 AM, then it could be negative and the conversion would give the wrong time, right?

damiannolan: Sure, I guess its better to be overly defensive, although I don't think it would affect the chain, just panic the cli. 
I wonder if the datetime on the machine was that badly out of sync if the node would have run into problems earlier, perhaps at runtime 🤔

damiannolan (approved):

modules/apps/transfer/client/cli/tx.go
crodriguezvega: Yes, I guess if the clock so much out of sync there would have been other problems already before.

modules/apps/transfer/client/cli/tx.go
seantking: Perhaps we can leave a comment explaining this? :)

crodriguezvega: I will do that, yes.

colin-axner: Maybe we should return an error if now is < 0?

colin-axner (approved):

modules/apps/transfer/client/cli/tx.go
crodriguezvega: Initially I also thought of returning an error, but then I realized that it was a situation where we could still do our best effort to fulfill the request by using the consensus state timestamp.

colin-axner: Yes, but this could cause problems in other areas of the codebase?

crodriguezvega: All right, I can change it. No problem.

crodriguezvega: @colin-axner I pushed this change. Do you approve it the way it is now? If yes, then I will merge.

wattsjohnny2020:
client_id: 07-tendermint-18
client_state:
  '@type': /ibc.lightclients.tendermint.v1.ClientState
  allow_update_after_expiry: true
  allow_update_after_misbehaviour: true
  chain_id: mars
  frozen_height:
    revision_height: "0"
    revision_number: "0"
  latest_height:
    revision_height: "2207"
    revision_number: "0"
  max_clock_drift: 600s
  proof_specs:
  - inner_spec:
      child_order:
      - 0
      - 1
      child_size: 33
      empty_child: null
      hash: SHA256
      max_prefix_length: 12
      min_prefix_length: 4
    leaf_spec:
      hash: SHA256
      length: VAR_PROTO
      prefix: AA==
      prehash_key: NO_HASH
      prehash_value: SHA256
    max_depth: 0
    min_depth: 0
  - inner_spec:
      child_order:
      - 0
      - 1
      child_size: 32
      empty_child: null
      hash: SHA256
      max_prefix_length: 1
      min_prefix_length: 1
    leaf_spec:
      hash: SHA256
      length: VAR_PROTO
      prefix: AA==
      prehash_key: NO_HASH
      prehash_value: SHA256
    max_depth: 0
    min_depth: 0
  trust_level:
    denominator: "3"
    numerator: "1"
  trusting_period: 1209600s
  unbonding_period: 1814400s
  upgrade_path:
  - upgrade
  - upgradedIBCState

colin-axner (approved):
thanks @crodriguezvega!

modules/apps/transfer/client/cli/tx.go
```diff
@@ -89,7 +92,21 @@ to the counterparty channel. Any timeout set to 0 is disabled.`),
...
+					consensusStateTimestamp := consensusState.GetTimestamp()
+					if now > 0 {
+						now := uint64(now)
+						if now > consensusStateTimestamp {
+							timeoutTimestamp = now + timeoutTimestamp
```
peterbourgon, line 103: If the system clock is set far in the future, does this not create an effectively infinite timeout?

PR #715, opened 2022-01-12
Fix ambiguity in TimeoutTimestamp docs

assafmo:
This PR fixes an ambiguity in the docs of `TimeoutTimestamp` parameter. The docs say `Timeout timestamp (in nanoseconds) relative to the current block timestamp.`, but as evident from the source code, `TimeoutTimestamp` is a timeout in nanoseconds since Unix epoch:
1. https://github.com/cosmos/ibc-go/blob/57679db9dfea53aa83f5c9b12def3605dfd837e0/modules/core/04-channel/keeper/packet.go#L94-L99
2. https://github.com/cosmos/ibc-go/blob/57679db9dfea53aa83f5c9b12def3605dfd837e0/modules/core/04-channel/keeper/packet.go#L212-L218
3. https://github.com/cosmos/ibc-go/blob/57679db9dfea53aa83f5c9b12def3605dfd837e0/modules/core/04-channel/keeper/timeout.go#L62-L72

As an anecdote, https://wrap.scrt.network uses only `TimeoutTimestamp` and only as an absolute timestamp in nanoseconds since Unix epoch.


colin-axner (approved):
Thanks!

assafmo:
@colin-axner - updated.

crodriguezvega (approved):

PR #700, opened 2022-01-07
refactor: move ica connection identifiers from port to version metadata bytestring

damiannolan:
## Description

The following PR refactors the interchain accounts handshake flow to move the connection identifier pair from controller port ID to a JSON encoded metadata type encapsulated within the ICS004 version bytestring.
See: https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#Versioning

Apologies for the large diff as these changes affected a decent chunk of the test suite.

**NOTE:** Changes will be required to /docs/app_modules/interchain_accounts/ica_auth.md when this is merged. I can make the changes in a follow up PR to not bloat this any further. 

closes: #615

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#pr-targeting))
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [x] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [x] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Review `Codecov Report` in the comment section below once CI passes

PR #729, opened 2022-01-13
refactor: reusable metadata validation

damiannolan:
## Description

Followup as mentioned in #700 
Refactors metadata validation into a reusable function.

closes: #XXXX

---

write a little note why.


modules/apps/27-interchain-accounts/types/metadata.go
```diff
@@ -9,3 +16,59 @@ func NewMetadata(version, controllerConnectionID, hostConnectionID, accAddress s
 		Address:                accAddress,
 	}
 }
+
+// ValidateControllerMetadata performs validation of the provided ICS27 controller metadata parameters
+func ValidateControllerMetadata(ctx sdk.Context, channelKeeper ChannelKeeper, connectionHops []string, metadata Metadata) error {
```
damiannolan, line 21: Originally I had a single `ValidateMetadata` function but depending on whether your in `OnChanOpenInit` or `OnChanOpenTry` the `connectionHops` will obviously correspond to opposing connection ends, and it would be comparing against the incorrect `metadata` field for controller/host. 

This can be consolidated to a single `ValidateMetadata` function when the connections IDs are removed from  `metadata`

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -84,12 +79,13 @@ func (k Keeper) OnChanOpenAck(
...
-	if err := icatypes.ValidateAccountAddress(metadata.Address); err != nil {
-		return err
+	channel, found := k.channelKeeper.GetChannel(ctx, portID, channelID)
+	if !found {
+		return sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "failed to retrieve channel %s on port %s", channelID, portID)
 	}
 
-	if metadata.Version != icatypes.Version {
-		return sdkerrors.Wrapf(icatypes.ErrInvalidVersion, "expected %s, got %s", icatypes.Version, metadata.Version)
+	if err := icatypes.ValidateControllerMetadata(ctx, k.channelKeeper, channel.ConnectionHops, metadata); err != nil {
```
colin-axner, line 87: should the controller reject the metadata if the host didn't set the address?

modules/apps/27-interchain-accounts/types/metadata.go
colin-axner: we could make a private function which takes in the controller connection ID and the host connection ID. ie just making the `ValidateControllerMetdata` get the connection and call the private function. Ditto for host. Then you deduplicate code

modules/apps/27-interchain-accounts/types/metadata_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+	var metadata types.Metadata
+
+	testCases := []struct {
+		name     string
+		malleate func()
+		expPass  bool
+	}{
+		{
+			"success",
```
colin-axner, line 18: add success with empty address test case

modules/apps/27-interchain-accounts/types/metadata.go
```diff
@@ -9,3 +16,59 @@ func NewMetadata(version, controllerConnectionID, hostConnectionID, accAddress s
...
+
+	if metadata.ControllerConnectionId != connectionHops[0] {
+		return sdkerrors.Wrapf(connectiontypes.ErrInvalidConnection, "expected %s, got %s", connectionHops[0], metadata.ControllerConnectionId)
+	}
+
+	if metadata.HostConnectionId != connection.GetCounterparty().GetConnectionID() {
+		return sdkerrors.Wrapf(connectiontypes.ErrInvalidConnection, "expected %s, got %s", connection.GetCounterparty().GetConnectionID(), metadata.HostConnectionId)
+	}
+
+	if metadata.Address != "" {
```
seantking, line 35: If the address is an empty string should we be returning an error?

colin-axner: No on OnChanOpenAck

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
damiannolan: It's a good point. Although we can't really put it inside the `ValidateControllerMetadata` func as it won't be set on INIT, only on ACK.

modules/apps/27-interchain-accounts/types/metadata.go
damiannolan: Great suggestion. Updated!

modules/apps/27-interchain-accounts/types/metadata_test.go
damiannolan: Done!

modules/apps/27-interchain-accounts/types/metadata.go
seantking: Right, I guess this is what your comment above was about as well

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
colin-axner: yea I'm thinking we should add a check in ACK that the address is non empty (outside validate controller metadata)

damiannolan: Added an additional check and test case 👍

modules/apps/27-interchain-accounts/types/metadata_test.go
```diff
@@ -0,0 +1,202 @@
...
```
colin-axner, line 122: success with empty address test case?

damiannolan: Added! My bad, missed adding this in the second func :)

PR #774, opened 2022-01-19
Fix for Evmos tests

ramacarlucho:
## Description

When using ibctesting suite on evmos some tests where failing due to conflicts with the ChainID defined by the suite. This pr fixes these issues

---

write a little note why.


testing/coordinator.go
```diff
@@ -174,7 +173,7 @@ func (coord *Coordinator) GetChain(chainID string) *TestChain {
 
 // GetChainID returns the chainID used for the provided index.
 func GetChainID(index int) string {
-	return ChainIDPrefix + strconv.Itoa(index)
+	return ChainIDPrefix + strconv.Itoa(index+1)
```
fedekunze, line 176: Why +1?

fedekunze (commented):
LGTM, one comment only

fedekunze:
@ramacarlucho can you change the branch to `main`?

testing/coordinator.go
fedekunze: can you only +1 if `index` is 0 then?

fedekunze (approved):
ACK. Pending comment

testing/coordinator.go
ramacarlucho: There will be duplicated ids for 0 and 1. And the coordinator uses a map with key chainid

docs/roadmap/roadmap.md
```diff
@@ -94,4 +94,4 @@ Scope is still TBD.
 
 #### H2 April
 
-- [`v5.0.0`](https://github.com/cosmos/ibc-go/milestone/21): Final release that includes the update of Cosmos SDK from `v0.46` to [v1.0](https://github.com/cosmos/cosmos-sdk/milestone/52) and that will support the migration to SMT storage.
\ No newline at end of file
+- [`v5.0.0`](https://github.com/cosmos/ibc-go/milestone/21): Final release that includes the update of Cosmos SDK from `v0.46` to [v1.0](https://github.com/cosmos/cosmos-sdk/milestone/52) and that will support the migration to SMT storage.
```
crodriguezvega, line 97: I believe you don't need to push any changes to the roadmap file, right?

fedekunze: I think the code editor just adds a newline at the end of the file

testing/coordinator.go
colin-axner: Instead, bump the starting index in the [for loop](https://github.com/cosmos/ibc-go/pull/774/files#diff-c18b7943741f428b34ce8ef78da1978fb3a372d0218b2fd5b2bd17346c7d454bR36) to 1 and make it `<= n`. It doesn't make sense that `GetChainID(3)` would return `testchain-4`

colin-axner (commented):
Thanks for the contribution. The pr needs a changelog, this is API breaking. Otherwise LGTM

testing/coordinator.go
ramacarlucho: The current solution only affects key generation and doesnt require changes on any tests.
If I change the starting index on the for loop, I will need to update every test that uses `GetChainID` to bump +1 on every call. For example every instance of `GetChainId(0)` would fail.

I can make the change if necessary.

fedekunze: > It doesn't make sense that GetChainID(3) would return testchain-4

agree, I think we need to update the `for` loop @ramacarlucho

colin-axner: There aren't too many instances for the bump not to be viable. The alternative option is to create a variable for the starting index, let it default to 0 and then Evmos can override it. This seems kinda unnecessary though to avoid bumping the `GetChainID` argument

colin-axner (approved):
Thanks for the contribution and updating all the tests!

Issue #768, opened 2022-01-19
Host chain allows overwriting of the active channel


colin-axner:
## Summary

The host chain assumes the controller chain will never allow a channel to OPEN if it is not in the active channel mapping. 

## Problem Definition

[`OnChanOpenConfirm`](https://github.com/cosmos/ibc-go/blob/89ffaafb5956a5ea606e1f1bf249c880bea802ed/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L77) will overwrite an active channel. This is fine for creating a new active channel after the previous one closed, but if the controller chain allows two channels to be opened for the same port (by incorrectly accounting for active channels), then this will overwrite an existing active channel

## Proposal

Be defensive. If an active channel exist, return an error. Please add a test case for this scenario

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
@AdityaSripal I think this should be updated in the spec

AdityaSripal:
But if the controller chain rewrites the active channel, shouldn't the host chain just use the latest active channel? I fail to see an attack here

seantking:
> The host chain assumes the controller chain will never allow a channel to OPEN if it is not in the active channel mapping.

I don't understand this point. The current architecture will allow a controller chain to initialize a new channel as many times as it likes. I don't understand why this would be a security issue. 

The only potential vulnerability I see here is in the case where a relayer could spam channel creation after the first `InitInterchainAccount` and potentially stop packets from reaching the interchain account (by overwriting the active channel by spamming channel creation on the bound port for a specific interchain account).

colin-axner:
> But if the controller chain rewrites the active channel, shouldn't the host chain just use the latest active channel? I fail to see an attack here

There's a race condition here though right? If chainA ACKs channel-1 and channel-2 on the same port. A relayer could decide which becomes the active channel on chainB by submitting the OpenConfirm in a certain order. We could end up in a state where the controller active channel does not equal the host active channel (and it cannot be fixed)

I think the host chain should defensively consider overwrites of an active channel on the controller chain as byzantine behaviour (not necessarily attacking, but something is going wrong)

colin-axner:
Action items:
- [ ] Ensure only 1 active channel can be created on ACK by checking if active channel exists + and is open
- [ ] document that host (in OpenConfirm) assumes controller only allows 1 channel to be ACK'd
[...]

PR #785, opened 2022-01-25
Defensive checks for active channel

seantking:
## Description

closes: #768

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/controller/keeper/handshake.go
+++ b/modules/apps/27-interchain-accounts/controller/keeper/handshake.go
@@ -79,6 +79,10 @@ func (k Keeper) OnChanOpenAck(
 		return sdkerrors.Wrapf(icatypes.ErrUnknownDataType, "cannot unmarshal ICS-27 interchain accounts metadata")
 	}
 
+	if activeChannelID, found := k.GetOpenActiveChannel(ctx, portID); found {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "existing active channel %s for portID %s", activeChannelID, portID)
+	}
+
 	channel, found := k.channelKeeper.GetChannel(ctx, portID, channelID)
 	if !found {
 		return sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "failed to retrieve channel %s on port %s", channelID, portID)
--- a/modules/apps/27-interchain-accounts/host/keeper/handshake.go
+++ b/modules/apps/27-interchain-accounts/host/keeper/handshake.go
@@ -9,6 +9,7 @@ import (
 
 	icatypes "github.com/cosmos/ibc-go/v3/modules/apps/27-interchain-accounts/types"
 	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+	porttypes "github.com/cosmos/ibc-go/v3/modules/core/05-port/types"
 	host "github.com/cosmos/ibc-go/v3/modules/core/24-host"
 )
 
@@ -45,6 +46,10 @@ func (k Keeper) OnChanOpenTry(
 
 	if err := icatypes.ValidateHostMetadata(ctx, k.channelKeeper, connectionHops, metadata); err != nil {
 		return "", err
+	}
+
+	if activeChannelID, found := k.GetOpenActiveChannel(ctx, portID); found {
+		return "", sdkerrors.Wrapf(porttypes.ErrInvalidPort, "existing active channel %s for portID %s", activeChannelID, portID)
 	}
 
 	// On the host chain the capability may only be claimed during the OnChanOpenTry
--- a/modules/apps/27-interchain-accounts/host/keeper/handshake_test.go
+++ b/modules/apps/27-interchain-accounts/host/keeper/handshake_test.go
@@ -111,6 +111,17 @@ func (suite *KeeperTestSuite) TestOnChanOpenTry() {
 			},
 			false,
 		},
+		{
+			"active channel already set",
+			func() {
+				// create a new channel and set it in state
+				ch := channeltypes.NewChannel(channeltypes.OPEN, channeltypes.ORDERED, channeltypes.NewCounterparty(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID), []string{path.EndpointA.ConnectionID}, ibctesting.DefaultChannelVersion)
+				suite.chainB.GetSimApp().GetIBCKeeper().ChannelKeeper.SetChannel(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, ch)
+
+				// set the active channelID in state
+				suite.chainB.GetSimApp().ICAHostKeeper.SetActiveChannelID(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
+			}, false,
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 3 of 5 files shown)


seantking:
>  Document that host (in OpenConfirm) assumes controller only allows 1 channel to be ACK'd

@colin-axner where are you thinking of adding this documentation?

modules/apps/27-interchain-accounts/host/keeper/handshake.go
```diff
@@ -47,6 +48,10 @@ func (k Keeper) OnChanOpenTry(
 		return "", err
 	}
 
+	if activeChannelID, found := k.GetOpenActiveChannel(ctx, portID); found {
```
damiannolan, line 51: Doesn't this effectively restrict host chains to a single interchain account? In `OnChanOpenTry` the `portID` argument is always `"interchain-account"`.

The first handshake to take place will set an active channel in `OnChanOpenConfirm` using the `portID` argument (`"interchain-account"`) as the key, and all subsequent handshakes will then check for an active channel with this portID in `OnChanOpenTry` and fail with an error.

Should we rather use the controller (counterparty) port ID when setting the active channel in `OnChanOpenConfirm`?

colin-axner: Nice catch! That's a :bug: . It should be the counterparty port ID

colin-axner: You'll note that [ica registration](https://github.com/cosmos/ibc-go/pull/785/files#diff-0af39e5e94b598423a42ea2db6d5611a6c3ab78d653d048bf0a89e9cb2aa7bf1L59) uses the counterparty port ID

colin-axner:
> @colin-axner where are you thinking of adding this documentation?

In the code. When finding this issue, I just found it so odd we set the active channel without doing any checks. For the future reader, it might be nice to explain why

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -79,6 +79,10 @@ func (k Keeper) OnChanOpenAck(
 		return sdkerrors.Wrapf(icatypes.ErrUnknownDataType, "cannot unmarshal ICS-27 interchain accounts metadata")
 	}
 
+	if activeChannelID, found := k.GetOpenActiveChannel(ctx, portID); found {
+		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "existing active channel %s for portID %s", activeChannelID, portID)
```
colin-axner, line 83: I think the base error could be something else? Not sure the port is technically invalid

modules/apps/27-interchain-accounts/host/keeper/handshake.go
```diff
@@ -47,6 +48,10 @@ func (k Keeper) OnChanOpenTry(
...
+		return "", sdkerrors.Wrapf(porttypes.ErrInvalidPort, "existing active channel %s for portID %s", activeChannelID, portID)
```
colin-axner, line 52: ditto

modules/apps/27-interchain-accounts/host/keeper/handshake_test.go
```diff
@@ -111,6 +111,17 @@ func (suite *KeeperTestSuite) TestOnChanOpenTry() {
...
 		},
+		{
+			"active channel already set",
+			func() {
+				// create a new channel and set it in state
+				ch := channeltypes.NewChannel(channeltypes.OPEN, channeltypes.ORDERED, channeltypes.NewCounterparty(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID), []string{path.EndpointA.ConnectionID}, ibctesting.DefaultChannelVersion)
+				suite.chainB.GetSimApp().GetIBCKeeper().ChannelKeeper.SetChannel(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, ch)
+
+				// set the active channelID in state
+				suite.chainB.GetSimApp().ICAHostKeeper.SetActiveChannelID(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
```
colin-axner, line 122: this needs to be changed as well to use EndpointA portID

colin-axner (commented):
LGTM, just needs the fix @damiannolan pointed out

modules/apps/27-interchain-accounts/host/keeper/handshake.go
seantking: done. great catch

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -52,7 +51,7 @@ func (k Keeper) OnChanOpenInit(
 
 	activeChannelID, found := k.GetOpenActiveChannel(ctx, connectionHops[0], portID)
 	if found {
-		return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "existing active channel %s for portID %s", activeChannelID, portID)
+		return sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s", activeChannelID, portID)
```
damiannolan, line 54: Should we include the connectionID in the error msg?

modules/apps/27-interchain-accounts/host/keeper/keeper.go
```diff
@@ -102,6 +103,22 @@ func (k Keeper) GetActiveChannelID(ctx sdk.Context, connectionID, portID string)
 	return string(store.Get(key)), true
 }
 
+// GetOpenActiveChannel retrieves the active channelID from the store, keyed by the provided connectionID and portID & checks if the channel in question is in state OPEN
+func (k Keeper) GetOpenActiveChannel(ctx sdk.Context, connectionID, portID string) (string, bool) {
```
damiannolan, line 107: Should we just change the `GetActiveChannelID` function to use this implementation (add the state check)? And use that throughout.

Afaik, the only place `GetActiveChannelID` is used is controller/relay.go in `SendTx`.
It seems most of the active channel checks are now using this approach with the additional state check, which kind of makes `GetActiveChanelID` redundant. 

Thoughts?

damiannolan (approved):

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
colin-axner: not necessary since the active channel is returned, but might be nice to have

modules/apps/27-interchain-accounts/host/keeper/handshake.go
```diff
@@ -78,7 +82,8 @@ func (k Keeper) OnChanOpenConfirm(
 		return sdkerrors.Wrapf(channeltypes.ErrChannelNotFound, "failed to retrieve channel %s on port %s", channelID, portID)
 	}
 
-	k.SetActiveChannelID(ctx, channel.ConnectionHops[0], portID, channelID)
+	// There can only be one active channel per connectionID/portID so we can overwrite if active channel exists in state already
```
colin-axner, line 85:
```suggestion
	// It is assumed the controller chain will not allow multiple active channels to be created for the same connectionID/portID
	// If the controller chain does allow multiple active channels to be created for the same connectionID/portID, 
	// disallowing overwriting the current active channel guarantees the channel can no longer be used as the controller
	// and host will disagree on what the currently active channel is 
```

modules/apps/27-interchain-accounts/host/keeper/keeper.go
colin-axner: I like the separation of concerns. Usually `Get` function retrieve values from the store. I'd find it odd if `GetActiveChannelID` also did a state check

colin-axner (approved):

Issue #767, opened 2022-01-19
Host chain doesn't account for multiple controller chains using overlapping namespace


colin-axner:
## Summary

The host chain currently assumes that the controller port ID is unique amongst all its connected chains. This is immediately evident when viewing its setting of the interchain account. It uses only the [controller port](https://github.com/cosmos/ibc-go/blob/89ffaafb5956a5ea606e1f1bf249c880bea802ed/modules/apps/27-interchain-accounts/host/keeper/account.go#L25) as the key. Thus two controller chains using the same format to generate the port address will collide. This seems highly probable as chains will likely reuse the same ICA auth modules. 

Previously the connection sequences were included in the portID thus making this assumption okay since a chains binded ports must be unique. Thus, there would never be two ownerA ports on connection 0. But there may be ownerA on connection 0 and ownerA on connection 1. 

## Problem Definition

The interchain account mapping on the host chain uses the controller port as the key. Controller ports are a [prefix and account owner address](https://github.com/cosmos/ibc-go/blob/89ffaafb5956a5ea606e1f1bf249c880bea802ed/modules/apps/27-interchain-accounts/types/port.go#L16). There is no uniqueness properties in the controller ports anymore. 

## Proposal

We should use either the connection sequence or channel sequence in the key for the interchain account address mapping. Since we still have the active channel mapping, I think it makes sense to use the connection sequence, otherwise you'd need to migrate the mapping when opening a new active channel. You'd also need to account for the channelID in ICA address generation. It is possible a channel is created on the controller chain which will never become an active channel (but the current controller handshake code might create an interchain account anyways). I don't think we need the controller chain connection sequence? 

Action items (separate pr's):
- [ ] use the host chain connection sequence in the ica address mapping (key is now connectionSeq/controllerPort) + test cases
- [ ] use host chain connection sequence when [generating ica address](https://github.com/cosmos/ibc-go/blob/89ffaafb5956a5ea606e1f1bf249c880bea802ed/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L56) + test cases

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

AdityaSripal:
Nice catch, until we use ORDERED channels that don't close on timeout, I think we'll need to use the connection sequence as you mentioned.

seantking:
Yeah good catch, makes sense to me. Agree with the proposal.

[...]

PR #791, opened 2022-01-25
chore: use connection ID in interchain account store keys

damiannolan:
## Description

- Use host chain connection ID in interchain account address store key

needs: #790 
ref: #767 

---

write a little note why.


damiannolan:
As mentioned in description, needs #790. I've written a test case which sets up two paths with three chains: `A(controller) -> B(host)` and `C(controller) -> B(host)` and registers two interchain accounts successfully. However I need to get both PRs together in order to add the test case successfully. I can do that in a follow up PR.

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -163,10 +163,10 @@ func (k Keeper) IsActiveChannel(ctx sdk.Context, portID string) bool {
 	return ok
 }
 
-// GetInterchainAccountAddress retrieves the InterchainAccount address from the store keyed by the provided portID
-func (k Keeper) GetInterchainAccountAddress(ctx sdk.Context, portID string) (string, bool) {
+// GetInterchainAccountAddress retrieves the InterchainAccount address from the store associated with the provided connectionID and portID
+func (k Keeper) GetInterchainAccountAddress(ctx sdk.Context, connectionID, portID string) (string, bool) {
 	store := ctx.KVStore(k.storeKey)
-	key := icatypes.KeyOwnerAccount(portID)
+	key := icatypes.KeyOwnerAccount(connectionID, portID)
```
colin-axner, line 169: technically not necessary for the controller chain since we are guaranteed to have a unique port, but it keeps the code cleaner so I don't think it hurts. Is this the controller or host connection id?

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -93,7 +93,7 @@ func (k Keeper) OnChanOpenAck(
 	}
 
 	k.SetActiveChannelID(ctx, portID, channelID)
-	k.SetInterchainAccountAddress(ctx, portID, metadata.Address)
+	k.SetInterchainAccountAddress(ctx, metadata.HostConnectionId, portID, metadata.Address)
```
colin-axner, line 96: controller docs should specify this is the host connection ID (currently says "associated" connection)

colin-axner (approved):
Excellent! 

Curious if @AdityaSripal has a preference on whether to use host or connection ID in controller ica mapping

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
damiannolan: Yeah, I figured it was better to keep consistent as both submodules are using the shared genesis types also. 

This is currently using the host connection ID on the controller side, but I would also be interested to hear @AdityaSripal's thoughts. Maybe it makes more sense to use the controller connection ID when setting controller side state?

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
damiannolan: Agreed. I'll update the godoc to be more explicit when its decided whether to use the controller or host connection ID here when setting state.

damiannolan: Updated to use controller connection ID here as agreed

colin-axner (approved):

seantking (approved):
clean

Issue #803, opened 2022-01-27
Controller chain should use connection sequence when using portID as a key in a mapping


colin-axner:
## Summary

same changes needed as #767. This will allow owner addresses to be reused for different connections instead of requiring ica auth modules to account for this by modifying the owner address in some fashion to account for the connection sequence. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
Closed by #807

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #807, opened 2022-01-27
chore: use connection ID in active channel store keys

damiannolan:
## Description

ref: #767 & #803 

This PR addresses including the correct connection IDs in the active channel store key on both controller and host
I've targeted this to be merged to #791 so that the inclusion of connection IDs in account address store keys can be merged together and we can close both issues based off that. 

closes: #XXXX

---

write a little note why.


modules/apps/27-interchain-accounts/controller/keeper/relay.go
```diff
@@ -15,16 +15,10 @@ import (
 // If the base application has the capability to send on the provided portID. An appropriate
 // absolute timeoutTimestamp must be provided. If the packet is timed out, the channel will be closed.
 // In the case of channel closure, a new channel may be reopened to reconnect to the host chain.
-func (k Keeper) SendTx(ctx sdk.Context, chanCap *capabilitytypes.Capability, portID string, icaPacketData icatypes.InterchainAccountPacketData, timeoutTimestamp uint64) (uint64, error) {
-	// Check for the active channel
-	activeChannelID, found := k.GetActiveChannelID(ctx, portID)
```
colin-axner, line 20: hmm, shouldn't this check that the channel is in the active channel mapping? Optimistic sends aren't strictly disabled (except via checking active channel mapping)

colin-axner (commented):
Excellent! I think there is just one small issue

modules/apps/27-interchain-accounts/controller/keeper/relay.go
damiannolan: To get the channel capability, an auth module will need to call `GetActiveChannelID(ctx, connectionID, portID)`. Do we still need to enforce the check here? If so I can just pass the connectionID as arg here as opposed to the channelID.

colin-axner: Yes I think so, it doesn't hurt to be defensive. What happens if the auth module allows the channel ID to be specified somehow (bypassing GetActiveChannelIID)

damiannolan: This is true! I'll update

colin-axner (approved):
Wahoo!

docs/ibc/proto-docs.md
```diff
@@ -311,11 +311,12 @@ An InterchainAccount is defined as a BaseAccount & the address of the account ow
 <a name="ibc.applications.interchain_accounts.v1.ActiveChannel"></a>
 
 ### ActiveChannel
-ActiveChannel contains a pairing of port ID and channel ID for an active interchain accounts channel
+ActiveChannel contains a connection ID, port ID and associated active channel ID
```
seantking, line 314: In some places we use the terminology `connectionID` & others `connection sequence`. When should we use one or the other? cc @AdityaSripal @colin-axner

seantking (approved):
:100:

docs/ibc/proto-docs.md
damiannolan: Anywhere the connection ID is used in the string format `"connection-{sequence}"` eg. `"connection-0"`, then `connectionID` is used, otherwise `connectionSeq` when it's just the parsed uint64 sequence value associated with the connection ID.

Issue #802, opened 2022-01-27
Modify RegisterInterchainAccount to allow creating active channels on multiple connections using the same owner address


colin-axner:
## Summary

`RegisterInterchainAccount` should:
- not error if port is already bound
- error if active channel exists

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #814, opened 2022-01-28
refactor: RegisterInterchainAccount

seantking:
## Description

closes: #802 

---

write a little note why.


modules/apps/27-interchain-accounts/controller/keeper/account.go
```diff
@@ -21,13 +21,17 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, connectionID, owner s
...
+	// if there is an active channel for this portID / connectionID return an error
+	activeChannelID, found := k.GetOpenActiveChannel(ctx, connectionID, portID)
+	if found {
+		return sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s on connection %s for owner %s", activeChannelID, portID, connectionID, owner)
 	}
 
-	cap := k.BindPort(ctx, portID)
-	if err := k.ClaimCapability(ctx, cap, host.PortPath(portID)); err != nil {
-		return sdkerrors.Wrap(err, "unable to bind to newly generated portID")
+	if !k.IsBound(ctx, portID) {
```
seantking, line 30: I switched this to use our custom keeper fn `IsBound` which also ensures the capability has been claimed for this portID. Otherwise, we would panic below (the tests picked this up).

modules/apps/27-interchain-accounts/controller/keeper/account.go
```diff
@@ -21,13 +21,17 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, connectionID, owner s
 		return err
 	}
 
-	if k.portKeeper.IsBound(ctx, portID) {
-		return sdkerrors.Wrap(icatypes.ErrPortAlreadyBound, portID)
+	// if there is an active channel for this portID / connectionID return an error
+	activeChannelID, found := k.GetOpenActiveChannel(ctx, connectionID, portID)
```
seantking, line 25: By checking this we're ensuring that an owner address can only have one interchain account per connection. Ideally, an owner could have many accounts per connection. I think this is fine for v1 though. 

There's still nothing stopping ica-auth dev from using owner input parameter however they please though (increment sequence appended or w/e), so it's not a total blocker anyway. 

cc @AdityaSripal @colin-axner @damiannolan

modules/apps/27-interchain-accounts/controller/keeper/account.go
```diff
@@ -21,13 +21,17 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, connectionID, owner s
...
+	if found {
+		return sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s on connection %s for owner %s", activeChannelID, portID, connectionID, owner)
 	}
 
-	cap := k.BindPort(ctx, portID)
-	if err := k.ClaimCapability(ctx, cap, host.PortPath(portID)); err != nil {
-		return sdkerrors.Wrap(err, "unable to bind to newly generated portID")
+	if !k.IsBound(ctx, portID) {
+		cap := k.BindPort(ctx, portID)
+		if err := k.ClaimCapability(ctx, cap, host.PortPath(portID)); err != nil {
```
seantking, line 32: Test coverage is down as we're not covering this line. Not sure what the best way to test this is. We're already checking above to ensure the capability is not claimed.

modules/apps/27-interchain-accounts/controller/keeper/account.go
```diff
@@ -21,13 +21,17 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, connectionID, owner s
...
+			return sdkerrors.Wrap(err, "unable to bind to newly generated portID")
```
damiannolan, line 33: Could be useful to include portID in this error msg, unless its indicated in the wrapped `err`

damiannolan (approved):

modules/apps/27-interchain-accounts/controller/keeper/account.go
colin-axner: This can still panic. The purpose of using the `portKeeper.IsBound` was that ports claimed outside of ics27 (by another IBC module) would cause a panic here since you are only checking the portID's claimed by the controller keeper and not the portID's binded to via 05-port. 

The code needs to be more nuanced:
```go
switch:
case portkeeper.IsBound() && !k.IsBound():
    return err (controller keeper doesn't own capability)
    
case !portKeeper.IsBound():
    k.BindPort
    k.ClaimCapability
    
}
```
Or something like that

colin-axner: On second thought, It might be better to panic on the case that the controller doesn't own the capability since this means another module on the chain claimed a capability in the namespace of ics27. No real difference since a panic just results in a failed transaction anyways. A panic might just be more likely to surface the problem (we should likely investigate closer if that error ever did occur)

colin-axner: we should still use port keeper `IsBound()` though

seantking: So, leave it as is then?

colin-axner: No, the current code unintentionally has ok behaviour. If the port is bounded by a different moudle, it will panic on `BindPort()`, but as far as I can tell this is unintentional. I'd prefer to restructure the code such that we have clearly defined expected paths such that if anything changes in the future, we aren't resting the correctness of our code on assumptions (what if BindPort returns an error instead of panic'ing)

I would probably use my switch above, but panic instead of returning an error

modules/apps/27-interchain-accounts/types/errors.go
```diff
@@ -13,8 +13,8 @@ var (
 	ErrInvalidRoute                = sdkerrors.Register(ModuleName, 7, "invalid route")
 	ErrInterchainAccountNotFound   = sdkerrors.Register(ModuleName, 8, "interchain account not found")
 	ErrInterchainAccountAlreadySet = sdkerrors.Register(ModuleName, 9, "interchain account is already set")
-	ErrActiveChannelNotFound       = sdkerrors.Register(ModuleName, 10, "no active channel for this owner")
-	ErrActiveChannelAlreadySet     = sdkerrors.Register(ModuleName, 11, "active channel already set for this owner")
+	ErrActiveChannelAlreadySet     = sdkerrors.Register(ModuleName, 10, "active channel already set for this owner")
+	ErrActiveChannelNotFound       = sdkerrors.Register(ModuleName, 11, "no active channel for this owner")
```
colin-axner, line 17: as a side note, this is state machine breaking (ABCI code changes), so I'd generally refrain from reorganizing error code ordering. It is hard to come up with any logical ordering when the code are grouped linearly

modules/apps/27-interchain-accounts/controller/keeper/account.go
seantking: Alright, gotcha. Thanks for making it clear :+1:

AdityaSripal (commented):
The current diffs look good.

But I think we should change the key format of active channel. Currently it is `connectionID/portID`.

I think it makes more sense to have it be: `portID/connectionID`. That way its easy to iterate over the store and figure out which channels are owned by a given owner address.

cc: @colin-axner

modules/apps/27-interchain-accounts/types/errors.go
seantking: That is an excellent point

modules/apps/27-interchain-accounts/controller/keeper/account.go
```diff
@@ -21,13 +21,20 @@ func (k Keeper) RegisterInterchainAccount(ctx sdk.Context, connectionID, owner s
...
+	if found {
+		return sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s on connection %s for owner %s", activeChannelID, portID, connectionID, owner)
 	}
 
-	cap := k.BindPort(ctx, portID)
-	if err := k.ClaimCapability(ctx, cap, host.PortPath(portID)); err != nil {
-		return sdkerrors.Wrap(err, "unable to bind to newly generated portID")
+	switch {
+	case k.portKeeper.IsBound(ctx, portID) && !k.IsBound(ctx, portID):
+		return sdkerrors.Wrapf(icatypes.ErrPortAlreadyBound, "another module has claimed capability for and bound port with portID: %s", portID)
```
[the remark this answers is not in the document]
colin-axner: #828 - not to be done now, but thought this change might help in the future

colin-axner (approved):
Great work! Thanks for following up with my code nit

colin-axner:
It would be nice if there was a test though using the same owner address on different connections

damiannolan:
bad bot @mergify

Issue #795, opened 2022-01-26
ICA controller chain must verify old active channel fields against newly created channel


colin-axner:
## Summary

Upon `OnChanOpenInit`, the controller chain needs to verify that all channel params are exactly the same as the old channel, otherwise relayers could modify channel params such as the version, causing potentially different behaviour in relation to the registered account. Currently it is possible to change which connection the channel is associated with.

The following check should be added to OnChanOpenInit, OnChanOpenTry:
- compare for equality the old metadata against the new metadata

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #847, opened 2022-02-03
chore: add defensive check to ensure metadata does not change when reopening an active channel

colin-axner:
## Description

closes: #795 

---

write a little note why.


modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -49,9 +50,20 @@ func (k Keeper) OnChanOpenInit(
...
+			panic(fmt.Sprintf("active channel mapping set for %s but channel does not exist in channel store", activeChannelID))
+		}
+
+		if channel.State == channeltypes.OPEN {
+			return sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s", activeChannelID, portID)
+		}
+
+		if channel.Version != version {
+			return sdkerrors.Wrap(icatypes.ErrInvalidVersion, "previous active channel metadata does not match provided version")
+		}
```
colin-axner, line 66: this requires the full metadata to be passed in when reopening a channel

modules/apps/27-interchain-accounts/controller/keeper/handshake_test.go
```diff
@@ -30,6 +30,24 @@ func (suite *KeeperTestSuite) TestOnChanOpenInit() {
...
+			func() {
+				suite.chainA.GetSimApp().ICAControllerKeeper.SetActiveChannelID(suite.chainA.GetContext(), ibctesting.FirstConnectionID, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+
+				counterparty := channeltypes.NewCounterparty(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
+				channel := channeltypes.Channel{
+					State:          channeltypes.CLOSED,
+					Ordering:       channeltypes.ORDERED,
+					Counterparty:   counterparty,
+					ConnectionHops: []string{path.EndpointA.ConnectionID},
+					Version:        TestVersion,
```
damiannolan, line 44: Hmm, shouldn't the version set in state on the closed channel also include the populated address field? 

This would mean that you cannot re-open a channel using `RegisterInterchainAccount` tho as the metadata built explicitly with an address: https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/keeper/account.go#L45-L53

damiannolan: (whoops, didn't see marked as draft)

colin-axner:
I reworked the code, checks now added for controller and host. The check added cannot be reached in tests since we don't support multiple application versions.

Running into an issue with the reopening a channel after closure test. Will debug monday, feel free to push to this branch if you see the fix

modules/apps/27-interchain-accounts/controller/keeper/handshake_test.go
colin-axner: I had opened it as r4r and then decided to change it up

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -49,9 +50,20 @@ func (k Keeper) OnChanOpenInit(
...
+		channel, found := k.channelKeeper.GetChannel(ctx, portID, activeChannelID)
+		if !found {
+			panic(fmt.Sprintf("active channel mapping set for %s but channel does not exist in channel store", activeChannelID))
+		}
+
+		if channel.State == channeltypes.OPEN {
+			return sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s", activeChannelID, portID)
+		}
+
+		if !icatypes.IsPreviousMetadataEqual(version, metadata) {
```
crodriguezvega, line 64: I might be missing something, but the `version` that is passed here, isn't it the same that is unmarshalled to `metadata` above? So here:

```
var metadata icatypes.Metadata
if err := icatypes.ModuleCdc.UnmarshalJSON([]byte(version), &metadata); err != nil {
    return sdkerrors.Wrapf(icatypes.ErrUnknownDataType, "cannot unmarshal ICS-27 interchain accounts metadata")
}
```

If that's the case, wouldn't this check here just redundant?

Oh, wait... Or shouldn't you be using here `channel.Version` instead?

modules/apps/27-interchain-accounts/types/metadata.go
```diff
@@ -27,6 +27,21 @@ func NewMetadata(version, controllerConnectionID, hostConnectionID, accAddress,
 	}
 }
 
+// IsMetadataEqual compares a metadata to a pevious version string set in a channel struct.
+// It ensure all fields are equal except the Address string
+func IsPreviousMetadataEqual(previousVersion string, metadata Metadata) bool {
```
crodriguezvega, line 32: Suggestion...

We already have code to unmarhsall the version string [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L43-L46) and the same code is found in [`OnChannelOpenTry`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L41-L44) and [`OnOpenChannelAck`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L76-L79). And now we also have it inside this function. So we could make an extra constructor function and remove the duplication of code in all these places. So basically to create a function with a signature like this: `func NewMetadata(version string) (Metadata, error)`.

And if you do that then you could also change this function so that it has the `Metadata` receiver type and change also the name and signature, so something like this: `func (Metadata m) AreEqual(metadata Metadata) bool`. I leave it up to you if you prefer a pointer receiver type.

The you can invoke these functions like:

```
previousMetadata, err := icatypes.NewMetadata(previousVersion)
if !err {
    return "", err
}
if !metadata.AreEqual(previousMetadata) {
    ...   
}

```

modules/apps/27-interchain-accounts/host/keeper/handshake.go
```diff
@@ -47,8 +48,20 @@ func (k Keeper) OnChanOpenTry(
 		return "", err
 	}
 
-	if activeChannelID, found := k.GetOpenActiveChannel(ctx, connectionHops[0], counterparty.PortId); found {
-		return "", sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s", activeChannelID, portID)
+	activeChannelID, found := k.GetActiveChannelID(ctx, connectionHops[0], counterparty.PortId)
```
crodriguezvega, line 51: If this whole block of code is the same as the one in `OnChannelOpenInit`, could we extract it into a helper function?

modules/apps/27-interchain-accounts/host/keeper/handshake.go
```diff
@@ -47,8 +48,20 @@ func (k Keeper) OnChanOpenTry(
...
+	if found {
+		channel, found := k.channelKeeper.GetChannel(ctx, portID, activeChannelID)
+		if !found {
+			panic(fmt.Sprintf("active channel mapping set for %s but channel does not exist in channel store", activeChannelID))
+		}
+
+		if channel.State == channeltypes.OPEN {
+			return "", sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s", activeChannelID, portID)
```
AdityaSripal, line 59:
```suggestion
			return "", sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s is already OPEN", activeChannelID, portID)
```

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
colin-axner: nice catch, it should be channel.Version :+1:

modules/apps/27-interchain-accounts/host/keeper/handshake.go
colin-axner: I think it is fine to leave as is/be explicit. There's still a lot of duplicate code between OpenInit and OpenTry

modules/apps/27-interchain-accounts/types/metadata.go
colin-axner: I like your suggestion. `NewMetadata` would need to be named something else to avoid name collision with the traditional usage of `NewMetadata`. `IsEqual` is also a good idea, but we will remove all this logic with the removal of active channels, so I'm leaning towards leaving as is. It makes the equality check explicitly to be used for active channel security and nothing else (otherwise external users may start depending on that API which should be removed in the future)

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
colin-axner: fixed

modules/apps/27-interchain-accounts/host/keeper/handshake.go
colin-axner: updated on controller and host

modules/apps/27-interchain-accounts/types/metadata.go
```diff
@@ -27,6 +27,21 @@ func NewMetadata(version, controllerConnectionID, hostConnectionID, accAddress,
 	}
 }
 
+// IsMetadataEqual compares a metadata to a pevious version string set in a channel struct.
+// It ensure all fields are equal except the Address string
```
damiannolan, line 31:
```suggestion
// IsPreviousMetadataEqual compares a metadata to a previous version string set in a channel struct.
// It ensures all fields are equal except the Address string
```

damiannolan (approved):

modules/apps/27-interchain-accounts/types/metadata.go
damiannolan: I prefer @crodriguezvega's suggestion too, but given this will be removed in a couple of months hopefully, then I think we should just leave it as is.

seantking: agree with @damiannolan

seantking (approved):
:handshake:

colin-axner:
> Looks great! I'm assuming there isn't error cases because it's not possible for it to even pass at the moment with some other metadata?

correct

modules/apps/27-interchain-accounts/controller/keeper/handshake_test.go
```diff
@@ -30,6 +30,24 @@ func (suite *KeeperTestSuite) TestOnChanOpenInit() {
 			},
 			true,
 		},
+		{
+			"success - previous active channel closed",
```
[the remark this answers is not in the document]
colin-axner: Unreachable, but I might as well add one

Issue #854, opened 2022-02-07
Change `WriteAcknowledgement` API to take in the `exported.Acknowledgement` interface instead of []byte


crodriguezvega:

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #749, opened 2022-01-17
refactor: update IncentivzedAck + add channeltype helpers for Ack

seantking:
## Description

- [x] Rename result in IncentivizedAcknowledgement to ack or app_acknowledgement or underlying_acknowledgement 
- [x] Change IncentivizedAcknowledgement ack field to be an Any (implement Acknowledgement interface)
- [x] Change the Success() function for IncentivizedAcknowledgement to take into account the success of the underlying acknowledgment

closes: #XXXX

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #882, opened 2022-02-07
refactor: WriteAcknowledgement API

seantking:
## Description
Updating WriteAcknowledgement API to take `exported.Acknowledgement` type instead of `[ ] bytes` 

closes: #854 

---

write a little note why.

### DIFF
--- a/docs/migrations/v3-to-v4.md
+++ b/docs/migrations/v3-to-v4.md
@@ -11,7 +11,7 @@ There are four sections based on the four potential user groups of this document
 
 **Note:** ibc-go supports golang semantic versioning and therefore all imports must be updated to bump the version number on major releases.
 ```go
-github.com/cosmos/ibc-go/v2 -> github.com/cosmos/ibc-go/v3
+github.com/cosmos/ibc-go/v3 -> github.com/cosmos/ibc-go/v4
 ```
 
 No genesis or in-place migrations required when upgrading from v1 or v2 of ibc-go.
--- a/docs/migrations/v3-to-v4.md
+++ b/docs/migrations/v3-to-v4.md
@@ -1,0 +1,26 @@
+# Migrating from ibc-go v2 to v3
+
+This document is intended to highlight significant changes which may require more information than presented in the CHANGELOG.
+Any changes that must be done by a user of ibc-go should be documented here.
+
+There are four sections based on the four potential user groups of this document:
+- Chains
+- IBC Apps
+- Relayers
+- IBC Light Clients
+
+**Note:** ibc-go supports golang semantic versioning and therefore all imports must be updated to bump the version number on major releases.
+```go
+github.com/cosmos/ibc-go/v2 -> github.com/cosmos/ibc-go/v3
+```
+
+No genesis or in-place migrations required when upgrading from v1 or v2 of ibc-go.
+
+## Chains
+
+### IS04 - Channel 
+
+The `WriteAcknowledgement` API now takes the `exported.Acknowledgement` type instead of passing in the acknowledgement byte array directly. 
+This is an API breaking change and as such IBC application developers will have to update any calls to `WriteAcknowledgement`. 
+
+
--- a/modules/core/04-channel/keeper/packet.go
+++ b/modules/core/04-channel/keeper/packet.go
@@ -343,6 +343,9 @@ func (k Keeper) WriteAcknowledgement(
 	}
 
 	bz := acknowledgement.Acknowledgement()
+	if len(bz) == 0 {
+		return sdkerrors.Wrap(types.ErrInvalidAcknowledgement, "acknowledgement cannot be empty")
+	}
 
 	// set the acknowledgement so that it can be verified on the other side
 	k.SetPacketAcknowledgement(
--- a/modules/core/04-channel/keeper/packet.go
+++ b/modules/core/04-channel/keeper/packet.go
@@ -312,7 +312,7 @@ func (k Keeper) WriteAcknowledgement(
 	ctx sdk.Context,
 	chanCap *capabilitytypes.Capability,
 	packet exported.PacketI,
-	acknowledgement []byte,
+	acknowledgement exported.Acknowledgement,
 ) error {
 	channel, found := k.GetChannel(ctx, packet.GetDestPort(), packet.GetDestChannel())
 	if !found {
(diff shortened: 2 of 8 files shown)


modules/core/04-channel/keeper/packet.go
```diff
@@ -342,14 +342,12 @@ func (k Keeper) WriteAcknowledgement(
 		return types.ErrAcknowledgementExists
 	}
 
-	if len(acknowledgement) == 0 {
```
damiannolan, line 345: How come the length check is removed? Should this not be kept?

colin-axner (commented):
Nice!

The ack length check needs to be added back. A changelog entry is needed and an entry into the migration docs for IBC app developers. We could backport to v3, but it isn't needed for transfer/ica so maybe best to leave for v4 instead of introducing changes for IBC app developers to implement (so make a new migration doc, you can copy the template)

seantking:
> Nice!
> 
> The ack length check needs to be added back. A changelog entry is needed and an entry into the migration docs for IBC app developers. We could backport to v3, but it isn't needed for transfer/ica so maybe best to leave for v4 instead of introducing changes for IBC app developers to implement (so make a new migration doc, you can copy the template)

Just added back the ack length check (I wrongly assumed an Ack wouldn't have empty bytes). 

I'll do the changelog and the docs update. Good catch, thanks :+1:

modules/core/04-channel/keeper/packet.go
seantking: Yeah thanks, good catch. I was working under the assumption that an Acknowledgement wouldn't have empty ack bytes

damiannolan (approved):

seantking:
@AdityaSripal @colin-axner So I've ran into an annoying issue with this. 

We aren't able to leave in the check to check for empty byte array i.e. `len(ackbytes) == 0` as calling `ack.Acknowledgement()` will always have len > 0 due to how we're serializing [into JSON](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/acknowledgement.go#L61) (I guess there is extra padding because of the JSON). 

Am I correct in saying the main reason we want to do this update to the API is for [code cleanliness](https://github.com/cosmos/ibc-go/pull/749/files#diff-60916badc8c21c745d39164a632c72516e493179a3a7bb3707c3aa682df6e019R26 ). I guess we have to pack and then unpack the ack bytes if we don't have this change, which is a bit ugly?

modules/core/04-channel/keeper/packet.go
```diff
@@ -342,14 +342,15 @@ func (k Keeper) WriteAcknowledgement(
...
+	bz := acknowledgement.Acknowledgement()
+	if len(bz) == 0 {
 		return sdkerrors.Wrap(types.ErrInvalidAcknowledgement, "acknowledgement cannot be empty")
```
AdityaSripal, line 347: Just change this check to `if acknowledgement == nil`

seantking: So this is what I had originally. 

Can we go with this @damiannolan @colin-axner?

testing/mock/ack.go
```diff
@@ -0,0 +1,23 @@
+package mock
+
+// MockEmptyAcknowledgement implements the exported.Acknowledgement interface and always returns an empty byte string as Response
+type MockEmptyAcknowledgement struct {
```
seantking, line 4: @colin-axner Open to suggestions with this. I don't want to add technical debt unnecessarily.

seantking:
@AdityaSripal @colin-axner I added a `MockEmptyAcknowledgement` type which implements the `exported.Acknowledgement` interface. We can use this in the tests to ensure we test the check for returning an error if ack bytes length == 0. 

Open to ideas for alternative workarounds.

testing/mock/ack.go
colin-axner: No this is great. None of the mock code has added any technical debt so far. I think it'll be useful to keep around

colin-axner (approved):
Excellent!

docs/migrations/v3-to-v4.md
```diff
@@ -0,0 +1,26 @@
...
+There are four sections based on the four potential user groups of this document:
+- Chains
+- IBC Apps
+- Relayers
+- IBC Light Clients
+
+**Note:** ibc-go supports golang semantic versioning and therefore all imports must be updated to bump the version number on major releases.
+```go
+github.com/cosmos/ibc-go/v2 -> github.com/cosmos/ibc-go/v3
+```
```
AdityaSripal, line 15: Shouldn't this be v3->v4?

modules/core/04-channel/keeper/packet.go
```diff
@@ -342,14 +342,15 @@ func (k Keeper) WriteAcknowledgement(
 		return types.ErrAcknowledgementExists
 	}
 
-	if len(acknowledgement) == 0 {
+	bz := acknowledgement.Acknowledgement()
+	if len(bz) == 0 {
```
AdityaSripal, line 346:
```suggestion
	if acknowledgement == nil || len(acknowledgement.Acknowledgement()) == 0 {
```

modules/core/04-channel/keeper/packet.go
```diff
@@ -342,14 +342,15 @@ func (k Keeper) WriteAcknowledgement(
 		return types.ErrAcknowledgementExists
 	}
 
-	if len(acknowledgement) == 0 {
+	bz := acknowledgement.Acknowledgement()
```
AdityaSripal, line 345: You should still check if acknowledgement is nil first or it will panic

AdityaSripal (changes requested):
Still need a nil check. Everything else looks great!

docs/migrations/v3-to-v4.md
seantking: yep

modules/core/04-channel/keeper/packet.go
seantking: Good call, thank you

modules/core/04-channel/keeper/packet.go
seantking: Added the check as a separate if statement. Just my personal preference.

AdityaSripal (approved):

colin-axner:
I think we should backport this to v3. Does anyone disagree?

Issue #876, opened 2022-02-07
Add GetTimestampAtHeight to client state interface


colin-axner:
## Summary

The solo machine cannot currently process timeouts because it does not have a consensus state associated with the timeout height until after processing the timeout. Core IBC assumes a consensus state exists in state for the timeout height, but solo machines stores the consensus state in the client state. 

We should add a function `GetTimestampAtHeight` to the client state function which returns the timestamp at a given consensus state height. 

#### PR 1

Add the `GetTimetstampAtHeight` to each client state + unit tests
- [x] 06-solomachine
- [x] 07-tendermint
- [x] 09-localhost

#### PR 2

Modify connection keeper [`GetTimestampAtHeight`](https://github.com/cosmos/ibc-go/blob/c378ff3b9e637dbae171d2c5129d966b468a2305/modules/core/03-connection/keeper/keeper.go#L88) to use the client state interface function. See if it is possible to write a test for solo machines to send a successful timeout msg

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged/assigned

crodriguezvega:
[I have asked today if it's possible to merge Colin's PR](https://github.com/polymerdao/ibc-go/pull/1#issuecomment-1167003791) to move forward with PR1 of this issue.

PR #888, opened 2022-02-09
chore: add GetTimestampAtHeight to client state

notbdu:
## Description

Add the GetTimetstampAtHeight to each client state + unit tests

- [x] 06-solomachine
- [x] 07-tendermint
- [x] 09-localhost

PR 1/2 for: #876 

---

write a little note why.


modules/light-clients/06-solomachine/types/client_state.go
```diff
@@ -39,6 +39,20 @@ func (cs ClientState) GetLatestHeight() exported.Height {
 	return clienttypes.NewHeight(0, cs.Sequence)
 }
 
+// GetTimestampAtHeight returns 0.
```
crodriguezvega, line 42: Is this comment correct? The implementation seems to indicate that a non-zero timestamp could also be returned...

modules/light-clients/06-solomachine/types/client_state.go
```diff
@@ -39,6 +39,20 @@ func (cs ClientState) GetLatestHeight() exported.Height {
...
+func (cs ClientState) GetTimestampAtHeight(
+	_ sdk.Context,
+	clientStore sdk.KVStore,
+	cdc codec.BinaryCodec,
+	height exported.Height,
+) (uint64, error) {
+	consensusState, err := GetConsensusState(clientStore, cdc, height)
+	if err != nil {
+		return 0, err
```
crodriguezvega, line 51: Is there a reason not to return the error wrapped like you do for the tendermint client (i.e. `sdkerrors.Wrapf(err, "height (%s)", height,)`)?

crodriguezvega (commented):
Thanks for this PR, @notbdu!

modules/light-clients/06-solomachine/types/client_state.go
notbdu: Ah, forgot to update the comment here... :man_facepalming:

colin-axner:
Thanks for the pr's @notbdu. I might have forgotten to mention that we plan to merge all the 02-client changes into a separate feature branch. It still needs to be created, but I can make it now. Would you mind changing the target branch to it?

modules/light-clients/06-solomachine/types/client_state.go
```diff
@@ -39,6 +39,20 @@ func (cs ClientState) GetLatestHeight() exported.Height {
...
+	_ sdk.Context,
+	clientStore sdk.KVStore,
+	cdc codec.BinaryCodec,
+	height exported.Height,
+) (uint64, error) {
+	consensusState, err := GetConsensusState(clientStore, cdc, height)
+	if err != nil {
+		return 0, sdkerrors.Wrapf(err, "height (%s)", height)
+	}
+	return consensusState.GetTimestamp(), nil
```
colin-axner, line 53: solo machine client state has access to the consensus state, no need to call `GetConsensusState`. This is actually why timeouts don't work for solo machines (they don't store a consensus state for each height they have a consensus state). Only the latest consensus state is tracked/needed

modules/light-clients/06-solomachine/types/store.go
```diff
@@ -0,0 +1,38 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	clienttypes "github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
+	host "github.com/cosmos/ibc-go/v3/modules/core/24-host"
+	"github.com/cosmos/ibc-go/v3/modules/core/exported"
+)
+
+// GetConsensusState retrieves the consensus state from the client prefixed
+// store. An error is returned if the consensus state does not exist.
+func GetConsensusState(store sdk.KVStore, cdc codec.BinaryCodec, height exported.Height) (*ConsensusState, error) {
```
colin-axner, line 15: delete, it isn't needed

modules/light-clients/06-solomachine/types/client_state_test.go
```diff
@@ -845,3 +846,57 @@ func (suite *SoloMachineTestSuite) TestVerifyNextSeqRecv() {
...
+			},
+			expValue: suite.solomachine.ClientState().ConsensusState.Timestamp,
+			expPass:  true,
+		},
+		{
+			name:        "get timestamp at height not exists",
+			clientState: suite.solomachine.ClientState(),
+			malleate:    func() {},
+		},
+	}
```
colin-axner, line 878: test should ensure the timestamp returned matches the consensus state timestamp (stored in the client state)

modules/light-clients/07-tendermint/types/client_state.go
```diff
@@ -58,6 +58,21 @@ func (cs ClientState) GetLatestHeight() exported.Height {
...
 }
 
+// GetTimestampAtHeight returns the timestamp in nanoseconds of the consensus state at the given height.
+func (cs ClientState) GetTimestampAtHeight(
+	ctx sdk.Context,
+	clientStore sdk.KVStore,
+	cdc codec.BinaryCodec,
+	height exported.Height,
+) (uint64, error) {
+	// get latest consensus state from clientStore to check for expiry
```
colin-axner, line 68: comment is incorrect

modules/light-clients/09-localhost/types/client_state.go
```diff
@@ -42,6 +42,16 @@ func (cs ClientState) GetLatestHeight() exported.Height {
...
 }
 
+// GetTimestampAtHeight returns 0. Localhost client has no consensus state.
+func (cs ClientState) GetTimestampAtHeight(
+	_ sdk.Context,
+	_ sdk.KVStore,
+	_ codec.BinaryCodec,
+	_ exported.Height,
+) (uint64, error) {
+	return 0, nil
```
colin-axner, line 52: This should probably return an error for now. The localhost doesn't work atm

modules/light-clients/09-localhost/types/client_state_test.go
```diff
@@ -527,3 +527,45 @@ func (suite *LocalhostTestSuite) TestVerifyNextSeqRecv() {
 		})
 	}
 }
+
+func (suite *LocalhostTestSuite) TestGetTimestampAtHeight() {
```
colin-axner, line 531: Probably should just ensure this always returns an error

modules/light-clients/06-solomachine/types/client_state.go
notbdu: I added a check to see if `height` passed in matches latest height or return an error if that makes sense?

notbdu:
This is ready for another pass btw!

seantking (approved):
LGTM but would like a review from @colin-axner. 

Thanks for the great work :handshake:

modules/light-clients/06-solomachine/types/client_state.go
```diff
@@ -39,6 +39,19 @@ func (cs ClientState) GetLatestHeight() exported.Height {
...
+// GetTimestampAtHeight returns the timestamp in nanoseconds of the consensus state at the given height.
+func (cs ClientState) GetTimestampAtHeight(
+	_ sdk.Context,
+	clientStore sdk.KVStore,
+	cdc codec.BinaryCodec,
+	height exported.Height,
+) (uint64, error) {
+	if !cs.GetLatestHeight().EQ(height) {
+		return 0, sdkerrors.Wrapf(ErrInvalidSequence, "not latest height (%s)", height)
+	}
```
colin-axner, line 51: this can be removed. The solo machine only has a single consensus state (the latest consensus state). This function is used when checking if a packet has timed out, so we want to compare the packet timestamp against the latest consensus state, as this would be the consensus state used process the timeout proof

colin-axner (changes requested):
Fantastic work! Really appreciated the clean tests. There's small fix regarding solo machines. 

Could you add a changelog entry as well? Thanks again!

notbdu:
Thanks for the review! I just rebased and pushed.

modules/light-clients/06-solomachine/types/client_state.go
```diff
@@ -39,6 +39,19 @@ func (cs ClientState) GetLatestHeight() exported.Height {
...
 }
 
+// GetTimestampAtHeight returns the timestamp in nanoseconds of the consensus state at the given height.
+func (cs ClientState) GetTimestampAtHeight(
+	_ sdk.Context,
+	clientStore sdk.KVStore,
+	cdc codec.BinaryCodec,
+	height exported.Height,
+) (uint64, error) {
+	if !cs.GetLatestHeight().Increment().EQ(height) {
```
colin-axner, line 49:
```suggestion
	if !cs.GetLatestHeight().EQ(height) {
```
Sorry I was incorrect! 

The solo machine stores the current sequence of the next proof it'll verify. In the timeout case it'll either be verifying absence of next seq recv (ordered channels) or packet receipt absence (unordered channels). I'm not sure why I got mixed up, but the sequence is incremented after verifying and this function is always called before verifying. So the height passed in should match the current sequence. The only time it won't is during connection handshake verification which doesn't need to rely on `GetTimestampAtHeight`

colin-axner:
closed by #1659

Issue #762, opened 2022-01-18
Emit Events on EscrowPacketFee


AdityaSripal:
Emit the latest fee for relaying a packet whenever `EscrowPacket` successfully executes. This will allow relayers to monitor the latest rewards for relaying a packet and decide to relay accordingly

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #648, opened 2021-12-20
Move event emissions to functions


colin-axner:
## Summary

For all core IBC event emissions, we should put them in functions within their respective keepers, like [so](https://github.com/cosmos/ibc-go/blob/a8ebc3f8a2675321903403f861eb4814e1e77f69/modules/core/04-channel/keeper/events.go#L15)

It increases readability of the critical code, provides self documentation, and allows us to ensure consistency between event emission

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

nir1218:
@colin-axner @crodriguezvega Hey, I would like to work on this issue, kindly assign it to me.

crodriguezvega:
hey @nir1218, glad to see that you want to work on another one. Thanks for your dedicating your time to it!

colin-axner:
@nir1218 let us know if you have any questions

nir1218:
@crodriguezvega @colin-axner  I noticed that in msg_server.go in the upgrade and update client methods there is an emit event of a message type , and then in the client keeper there is an emit of the operation type itself. So my question is, should I consolidate both emissions in one method and invoke it from the client keeper as done in events.go under channel module?

crodriguezvega:
Yes, I think that makes sense. So, for example, for the case of update client to have a method that emits both [this event](https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/keeper/client.go#L141-L149) and [this event](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/msg_server.go#L70-L75), right?

Do you agree, @colin-axner?

colin-axner:
Yes that makes sense to me

nir1218:
@crodriguezvega @colin-axner Will you be closing the issue and the PR? Is there anything else I can do on this?

PR #914, opened 2022-02-11
ics29:feat: emit event escrow

seantking:
## Description

closes: #762 
---

write a little note why.


modules/apps/29-fee/keeper/escrow.go
```diff
@@ -38,6 +39,17 @@ func (k Keeper) EscrowPacketFee(ctx sdk.Context, identifiedFee types.IdentifiedP
 
 	// Store fee in state for reference later
 	k.SetFeeInEscrow(ctx, identifiedFee)
+
+	// Emit event so that relayers know an incentivized packet is ready to be relayed
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeSendIncentivizedPacket,
+			sdk.NewAttribute(channeltypes.AttributeKeyPortID, identifiedFee.PacketId.PortId),
```
seantking, line 47: Should maybe put the relayer address here as well?

damiannolan: Sounds like a good idea.
Should we add an `events.go` and move the event emissions funcs there? [Similarly to 04-channel](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/events.go)

AdityaSripal: I actually think its unnecessary. Right now, relayer list is guaranteed to be nil, so no point in emitting it.

And once it does become non-nil, we can add it to events in a non-breaking manner

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -38,6 +39,17 @@ func (k Keeper) EscrowPacketFee(ctx sdk.Context, identifiedFee types.IdentifiedP
...
+			sdk.NewAttribute(channeltypes.AttributeKeyChannelID, identifiedFee.PacketId.ChannelId),
+			sdk.NewAttribute(channeltypes.AttributeKeySequence, fmt.Sprint(identifiedFee.PacketId.Sequence)),
```
AdityaSripal, line 49: We need to emit the fee for each relayer msg:

`RecvFee => recv_coins`
`AckFee => ack_coins`
`TimeoutFee => timeout_coins`

AdityaSripal (changes requested):

modules/apps/29-fee/types/events.go
```diff
@@ -1,4 +1,6 @@
 package types
 
 // 29-fee events
-const ()
+const (
+	EventTypeSendIncentivizedPacket = "send_incentivized_packet"
```
colin-axner, line 5: The packet is already sent, the packet is just now being incentivized (or further incentivized)

modules/apps/29-fee/keeper/escrow.go
colin-axner: > we can add it to events in a non-breaking manner

this won't necessarily be true by the time the change is made (as it does seem there is desire to eventually include events in consensus), but I agree we can leave it off for now

colin-axner (changes requested):
Can you move the event emission into a function, like outlined in #648?

modules/apps/29-fee/keeper/events.go
```diff
@@ -0,0 +1,24 @@
...
+func EmitIncentivizedPacket(ctx sdk.Context, identifiedFee types.IdentifiedPacketFee) {
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeIncentivizedPacket,
+			sdk.NewAttribute(channeltypes.AttributeKeyPortID, identifiedFee.PacketId.PortId),
+			sdk.NewAttribute(channeltypes.AttributeKeyChannelID, identifiedFee.PacketId.ChannelId),
+			sdk.NewAttribute(channeltypes.AttributeKeySequence, fmt.Sprint(identifiedFee.PacketId.Sequence)),
+			sdk.NewAttribute(types.AttributeKeyRecvFee, fmt.Sprint(identifiedFee.Fee.RecvFee)),
+			sdk.NewAttribute(types.AttributeKeyAckFee, fmt.Sprint(identifiedFee.Fee.AckFee)),
+			sdk.NewAttribute(types.AttributeKeyTimeoutFee, fmt.Sprint(identifiedFee.Fee.TimeoutFee)),
```
AdityaSripal, line 21: Question: Does it make sense to emit the cumulative fee for each type here?

It would be useful for now, but we may have to revert back if relayer list becomes used.

cc: @colin-axner

AdityaSripal (approved):

modules/apps/29-fee/keeper/events.go
colin-axner: It might be hard to do with the way we plan to iterate over the identified fees. I don't think validators should be running extra logic so events are presented. Is there a clean way of aggregating the fees?

modules/apps/29-fee/keeper/events.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/ibc-go/v3/modules/apps/29-fee/types"
+	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
+)
+
+// Emit event so that relayers know an incentivized packet is ready to be relayed
```
colin-axner, line 11: godoc needs to start with function name

modules/apps/29-fee/keeper/events.go
```diff
@@ -0,0 +1,24 @@
...
+func EmitIncentivizedPacket(ctx sdk.Context, identifiedFee types.IdentifiedPacketFee) {
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeIncentivizedPacket,
+			sdk.NewAttribute(channeltypes.AttributeKeyPortID, identifiedFee.PacketId.PortId),
+			sdk.NewAttribute(channeltypes.AttributeKeyChannelID, identifiedFee.PacketId.ChannelId),
+			sdk.NewAttribute(channeltypes.AttributeKeySequence, fmt.Sprint(identifiedFee.PacketId.Sequence)),
+			sdk.NewAttribute(types.AttributeKeyRecvFee, fmt.Sprint(identifiedFee.Fee.RecvFee)),
```
colin-axner, line 19:
```suggestion
			sdk.NewAttribute(types.AttributeKeyRecvFee, identifiedFee.Fee.RecvFee.String()),
```
thoughts on using .String() instead?

colin-axner (approved):

modules/apps/29-fee/keeper/events.go
```diff
@@ -0,0 +1,24 @@
+package keeper
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/ibc-go/v3/modules/apps/29-fee/types"
```
colin-axner, line 7: separate imports

modules/apps/29-fee/keeper/events.go
AdityaSripal: Not really, I think it requires some iteration. Though only one retrieval from store

colin-axner: But in the case that the relayers are different?

seantking: If we do this I think we should do this in a separate PR for adding some helper functions for aggregating fees & it also gives us time to think if we want it.

modules/apps/29-fee/keeper/events.go
seantking: Good idea. I've updated in the other spots also.

PR #939, opened 2022-02-16
ibctesting: custom voting power reduction for testing

fedekunze:
Adds support for chains with power reduction different than 6 to test IBC (eg: Evmos uses 18)

Use case: test transfers between chains with 6 and 18 as their power reduction values


testing/app.go
```diff
@@ -58,16 +58,20 @@ func SetupTestingApp() (TestingApp, map[string]json.RawMessage) {
...
+	app.AppCodec().MustUnmarshalJSON(genesisState[authtypes.ModuleName], authGenesis)
+
+	authGenesis = authtypes.NewGenesisState(authGenesis.Params, genAccs)
 	genesisState[authtypes.ModuleName] = app.AppCodec().MustMarshalJSON(authGenesis)
 
 	validators := make([]stakingtypes.Validator, 0, len(valSet.Validators))
 	delegations := make([]stakingtypes.Delegation, 0, len(valSet.Validators))
 
-	bondAmt := sdk.NewInt(1000000)
+	bondAmt := sdk.TokensFromConsensusPower(1, powerReduction)
```
AdityaSripal, line 74: Would it work for you if we just made this a global variable in the tests that you could change? Seems odd to add this as a function argument.

fedekunze: I could use the default consensus power (global from the SDK), but then it wouldn't support testing between chains with different power values

fedekunze: We want tests that involve powers such as

ChainA: 10e6
ChainB: 10e18

testing/chain.go
```diff
@@ -51,7 +49,7 @@ type SenderAccount struct {
 // is used for delivering transactions through the application state.
 // NOTE: the actual application uses an empty chain-id for ease of testing.
 type TestChain struct {
-	t *testing.T
```
fedekunze, line 54: had to make this public in order to support the custom test chain from a new constructor

testing/app.go
colin-axner: When chains change the power reduction, do they just change the global in the SDK, i.e. there is no on-chain parameter? I'm wondering if there's a way we can allow this information to be specified in `DefaultTestingAppInit`

testing/chain.go
colin-axner: Can you point me to the issue? Not sure I follow

fedekunze: This line:

https://github.com/tharsis/evmos/pull/282/files#diff-e1b6b9bd60ea81ed06c40d66a72dfba81bb74c77f3fee03c99197cbd1249f083R16

testing/app.go
fedekunze: > When chains change the power reduction, do they just change the global in the SDK, i.e. there is no on-chain parameter?

yes, there's no chain parameter. If we just change the parameters, the 2 chains from the tests will share the same power. That's why I'm adding it as a field so that we can tests txs between chains with different power values

testing/chain.go
colin-axner: Fair enough

fedekunze:
@colin-axner @AdityaSripal reverted the staking changes

colin-axner (approved):

fedekunze:
@crodriguezvega @AdityaSripal can you look at this again, we need it for testing purposes

AdityaSripal (approved):

Issue #850, opened 2022-02-04
Update client tries to use pruned consensus states


timlind:
## Summary of Bug

After upgrading to ibc-go v.2, so that we could use the update-client proposal, it appears a migration has pruned expired consensus states (lookup on the cli also fails), so the update-client process is unable to find an expired state, and results in the following error: 

```
raw_log: 'failed to execute message; message index: 0: client cannot be updated with
  proposal: unable to update light client: invalid proposal content'
```

## Version

v2.0.0

## Steps to Reproduce

Upgrade past v1.

sifnoded tx gov submit-proposal update-client 07-tendermint-42 07-tendermint-79 ...
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Hi @timlind. That's unfortunate. Thanks for opening the issue. Nice catch!

The issue here is the status of the client is [Unknown](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/client_state.go#L81), but this is not something we [account for](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/proposal_handle.go#L44-#L58)

Potential solutions:
- classify client states with consensus states as expired (since we can assume it was pruned)
- address #607 and allow Unknown states to be updated (only preventing Active states from being overwritten)
- allow Unknown status to be counted as Expired in the proposal handle

We can do a minor release for whatever fix is put in place.

timlind:
@colin-axner I've submitted a PR to treat unknown as expired. I don't see any obvious way write a test, but we'll be doing manual testing to see if this resolves our issue.

timlind:
@colin-axner this looks good after some testing. Do you have a timeframe for a release?

colin-axner:
Hey @timlind! We decided we actually want to change the return value from [Unknown to Expired](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/client_state.go#L81). Our reasoning is that if the consensus state cannot be retrieved for the latest height, then it must be expired (the latest height got pruned)

With regards to timeframe, we plan to do a minor backport release within the next few weeks  (subject to change) so we can include this fix in that release 

[...]

PR #941, opened 2022-02-17
fix: classify client states without consensus states as expired

timlind:
## Description

closes: #850 

---

write a little note why.

### DIFF
--- a/modules/light-clients/07-tendermint/types/client_state.go
+++ b/modules/light-clients/07-tendermint/types/client_state.go
@@ -78,7 +78,7 @@ func (cs ClientState) Status(
 	// get latest consensus state from clientStore to check for expiry
 	consState, err := GetConsensusState(clientStore, cdc, cs.GetLatestHeight())
 	if err != nil {
-		return exported.Unknown
+		return exported.Expired
 	}
 
 	if cs.IsExpired(consState.Timestamp, ctx.BlockTime()) {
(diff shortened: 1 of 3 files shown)


modules/light-clients/07-tendermint/types/client_state.go
```diff
@@ -78,7 +78,7 @@ func (cs ClientState) Status(
 	// get latest consensus state from clientStore to check for expiry
 	consState, err := GetConsensusState(clientStore, cdc, cs.GetLatestHeight())
 	if err != nil {
-		return exported.Unknown
+		return exported.Expired
```
colin-axner, line 81:
```suggestion
                // if the client state does not have an associated consensus state for its latest height
                // then it must be expired
		return exported.Expired
```

colin-axner: might need to commit locally, I don't know how to format properly with gh UI

colin-axner (approved):

AdityaSripal (approved):

colin-axner:
Hey @timlind, there appears to be a broken test, I think the expected value just needs to be changed. Can you add a changelog as well in the state machine breaking section?

colin-axner:
The branch needs to be brought up to date with main and then should automerge. Thanks again @timlind!

marshsif:
Hi @colin-axner could you or someone on the team give us a sense of when this might be included in the next 2.x release? We'd like to be able to deploy the patch on our end at least, and see how things look. Thanks

crodriguezvega:
Hi @marshsif. It is very likely that this will be released next week in v2.1.0.

crodriguezvega:
Hi @marshsif. If v2.1.0 imports SDK 0.45.1, would that be a problem for you?

marshsif:
No, I think that would be ok. At this point, we're going to deploy a forked version of IBCGO with the patch, and then when we do the v.45 upgrade on our end, we can include 2.1.

marshsif:
Thanks again for keeping us updated on this, really appreciate it @crodriguezvega!

Issue #953, opened 2022-02-21
Update existing grpc queries to handle multiple packet fees


crodriguezvega:
## Summary

- [x] Update `IncentivizedPackets` query to return packet fees for the given paginated request under the `FeesInEscrowPrefix` key
- [x] Update `IncentivizedPacket` query to return `[]PacketFee` and associated `PacketId` 

The above relies on models changed expected in #760 

```
// PacketFee contains the relayer fee, refund address and an optional list of relayers that are permitted to receive the
// fee
message PacketFee {
  Fee             fee            = 1 [(gogoproto.nullable) = false];
  string          refund_address = 2 [(gogoproto.moretags) = "yaml:\"refund_address\""];
  repeated string relayers       = 3;
}

// PacketFees contains a list of type PacketFee
message PacketFees {
  repeated PacketFee packet_fees = 1 [(gogoproto.moretags) = "yaml:\"packet_fees\"", (gogoproto.nullable) = false];
}

// IdentifiedPacketFees contains a PacketFree and associated PacketId
message IdentifiedPacketFees {
  ibc.core.channel.v1.PacketId packet_id = 1
      [(gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"packet_id\""];
  repeated PacketFee packet_fees = 2 [(gogoproto.moretags) = "yaml:\"packet_fees\"", (gogoproto.nullable) = false];
}
```

damiannolan:
Updated with relevant content!

PR #967, opened 2022-02-22
chore: update grpc queries to handle multiple fees

damiannolan:
## Description

Addresses updating of grpc queries and tests following model changes

closes: #953 

---

write a little note why.


colin-axner (commented):
LGTM! Will approve once upstream dependencies are merged.

seantking (approved):

proto/ibc/applications/fee/v1/query.proto
```diff
@@ -51,5 +51,5 @@ message QueryIncentivizedPacketRequest {
 // QueryIncentivizedPacketsResponse is the response type for the incentivized packet RPC
 message QueryIncentivizedPacketResponse {
   // Incentivized_packet
-  ibc.applications.fee.v1.IdentifiedPacketFee incentivized_packet = 1;
+  ibc.applications.fee.v1.IdentifiedPacketFees incentivized_packet = 1 [(gogoproto.nullable) = false];
```
AdityaSripal, line 54: This query doesn't make much sense to me. Why include packet id in response when it is already in request?

Makes more sense to have a PacketFee RPC that takes packetID and returns the packet fees for it.

Also I think it makes more sense to have an RPC that just aggregates all packet fees for a given packetID.

Something like

`QueryRecvFee(packetID)` => `total receive fees earned for this packet`

and similar for ack and timeout

AdityaSripal (approved):
Looks good, but I think we should add other queries that will be more useful in future prs

proto/ibc/applications/fee/v1/query.proto
damiannolan: I totally agree, my main goal with this PR was to just fix what was there w.r.t. model changes that were made. 
I think we should definitely follow up on fleshing out the query api. I really like the idea of:
`QueryRecvFee(packetID)` => `total receive fees earned for this packet`
`QueryAckFee(packetID)` ...
`QueryTimeoutFee(packetID)` ...

If you want I can update this response type to include just `PacketFees` (the list of packet fees without the packet ID) instead of the identified type.

modules/apps/29-fee/keeper/grpc_query_test.go
```diff
@@ -40,117 +23,116 @@ func (suite *KeeperTestSuite) TestQueryIncentivizedPacketI() {
...
+
+			for i := 0; i < 3; i++ {
+				// escrow three packet fees for the same packet
+				err := suite.chainA.GetSimApp().IBCFeeKeeper.EscrowPacketFee(suite.chainA.GetContext(), packetID, packetFee)
+				suite.Require().NoError(err)
+			}
+
+			req = &types.QueryIncentivizedPacketRequest{
+				PacketId:    packetID,
+				QueryHeight: 0,
```
colin-axner, line 124: haha I just thought to do the same on #983

modules/apps/29-fee/keeper/grpc_query.go
```diff
@@ -24,22 +27,30 @@ func (k Keeper) IncentivizedPackets(c context.Context, req *types.QueryIncentivi
...
+	var identifiedPackets []types.IdentifiedPacketFees
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), []byte(types.FeesInEscrowPrefix))
+	_, err := query.Paginate(store, req.Pagination, func(key, value []byte) error {
+		keySplit := strings.Split(string(key), "/")
+
+		channelID, portID := keySplit[2], keySplit[1]
+		seq, err := strconv.ParseUint(keySplit[3], 10, 64)
+		if err != nil {
+			panic(err)
+		}
```
colin-axner, line 39: Is this correct? From my investigation, pagination doesn't include the prefix iterated on in the return value. I guess there's a trailing "/" leading the key, which is subtle

So I think the key is something like "/transfer/channel-0/5"

colin-axner (approved):

modules/apps/29-fee/keeper/grpc_query.go
colin-axner: I'm kinda wondering if we should just reconstruct the full key and then use the parse functions in keys.go

damiannolan: Yeah, correct, so a paginated query doesn't include the prefix, but because of the leading `"/"` when we split using the slash separator the first element is an empty string. So with your example, `keySplit` ends up as a 4 element slice: 
`[]string{"", "transfer", "channel-0", "5"}`.

I do think since you've added the parse function we should try to leverage it here. It should be something like below, right?
```
packetID, err := types.ParseKeyFeesInEscrow(types.FeesInEscrowPrefix + string(key))
```

Issue #860, opened 2022-02-07
Fix RefundAllFees


AdityaSripal:
Modify RefundFeesOnChannel in OnChanCloseInit to iterate over all incentivized packets, before refunding for each packet, it should check the escrow address for sufficient balance to refund the packet. If the escrow balance would be negative, this implies a severe bug and thus all fee channels should become disabled. If the refund fails for another reason (blocked address) we should continue iteration. RefundFeesOnChannel can be renamed. DisableAllChannels should be called from within the logic of detecting the negative escrow balanceWhen we refund packet upon channel closure we need to remove the incentivized packets from the mapping (since they have been refunded). More discussion is needed to determine the best approach since there exist potential edge case scenarios. 

https://github.com/cosmos/ibc-go/blob/13f77de949ce84704db1c7307d31ba8c6e16842e/modules/apps/29-fee/keeper/escrow.go#L106

This should not be taken up until discussion in #821 is resolved.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #821, opened 2022-02-01
Find Elegant Solution if Fee Escrow account is out of balance


AdityaSripal:
The current way to deal with the worst-case scenario of when escrow account is out of balance is to disable the fee module on all channels.

However, this will only happen on our channel end. The other channel end will continue with the fee module enabled since there isn't a good way to communicate that our side has disabled fees once the channel is already open.

This will lead to an invalid state where one side has fee enabled and the other side does not. 

We should determine a better way to handle this case.

Perhaps we keep fee enabled, but just do not distribute the fee on our side.

Thanks to @soareschen for pointing this out

colin-axner:
Should we close the channel? That would allow the counterparty to close as well

colin-axner:
I'm working on this in smaller pieces, but in general:
- when escrow balance will be negative -> disable all channels
- when writing an acknowledgement, if the channel is disabled + the fee version is set -> return ics29 ack

PR #1031, opened 2022-02-28
feat: allow ability to lock fee module in presence of severe bug

colin-axner:
## Description

ref: #821, #860 

if the escrow balance may become negative, we want to lock the fee module such that it doesn't do escrows or distribute from the escrow account

closes: #XXXX

---

write a little note why.


modules/apps/29-fee/keeper/escrow.go
```diff
@@ -96,6 +96,10 @@ func (k Keeper) DistributePacketFeesOnTimeout(ctx sdk.Context, timeoutRelayer sd
 // If the distribution fails for any reason (such as the receiving address being blocked),
 // the state changes will be discarded.
 func (k Keeper) distributeFee(ctx sdk.Context, receiver sdk.AccAddress, fee sdk.Coins) {
+	if k.IsLocked(ctx) {
```
colin-axner, line 99: I'm not sure how much extra computation this does. It is a little annoying we need to check if the fee module is locked for each single fee distribution, but otherwise we have to rework the API of all our other functions to properly abort a transaction without returning an error

AdityaSripal (commented):
LGTM so far

modules/apps/29-fee/ibc_module.go
```diff
@@ -253,7 +253,7 @@ func (im IBCModule) OnTimeoutPacket(
 	packet channeltypes.Packet,
 	relayer sdk.AccAddress,
 ) error {
-	if !im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
+	if !im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) || im.keeper.IsLocked(ctx) {
```
AdityaSripal, line 256: Document why we're just passing through to underlying app in case of locked module

modules/apps/29-fee/ibc_module.go
```diff
@@ -221,7 +221,7 @@ func (im IBCModule) OnAcknowledgementPacket(
 	acknowledgement []byte,
 	relayer sdk.AccAddress,
 ) error {
-	if !im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) {
+	if !im.keeper.IsFeeEnabled(ctx, packet.SourcePort, packet.SourceChannel) || im.keeper.IsLocked(ctx) {
```
AdityaSripal, line 224: This will fail because the counterparty will still be sending an incentivized ack. We should unmarshal the ack and just send to underlying without distributing fees

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -611,6 +611,16 @@ func (suite *FeeTestSuite) TestOnAcknowledgementPacket() {
 			},
 			true,
 		},
+		{
+			"success: fee module is disabled, skip fee logic",
+			func() {
+				lockFeeModule(suite.chainA)
+
+				ack = ibcmock.MockAcknowledgement.Acknowledgement()
```
AdityaSripal, line 619: This is not true. The ack coming from counterparty will still be incentivized.

We just want to expect that escrow is not paid out

AdityaSripal (changes requested):
Acknowledgement logic is incorrect since the incentivized ack will still be sent

modules/apps/29-fee/ibc_module.go
```diff
@@ -230,6 +230,13 @@ func (im IBCModule) OnAcknowledgementPacket(
 		return sdkerrors.Wrapf(err, "cannot unmarshal ICS-29 incentivized packet acknowledgement: %v", ack)
 	}
 
+	if im.keeper.IsLocked(ctx) {
+		// if the fee keeper is locked then fee logic should be skipped
+		// this may occur in the presence of a severe bug which leads invalid state
+		// the fee keeper will be unlocked after manual intervention
+		return im.app.OnAcknowledgementPacket(ctx, packet, ack.Result, relayer)
```
AdityaSripal, line 237: Let's also document why we're still unmarshalling the acknowledgement because the counterparty is still sending the incentivized ack over

modules/apps/29-fee/keeper/escrow.go
AdityaSripal: It's an extra state read. From a efficiency point of view, this will probs be cached so i wouldn't worry about it. But it is additional gas

AdityaSripal, line 99: Is this line even necessary? I believe in both Ack and Timeout we're checking locked before we even call this function.

Why don't we just add it as CONTRACT that we must check `locked` before calling this function?

colin-axner: When we distribute fees, we do it one by one. It is possible distributing one of the fees causes the escrow balance to go negative at which point we want to stop any distribution. 

For example, the escrow balance has 10 coins. 
recv fee - 12 coins
ack fee - 5 coins
timeout fee - 5 coins

We start with recv fee, see the balance would be negative - we want to stop distribution. If I don't have this call, then ack fee and timeout fee would be paid out, unless I brought the check up to both `DistributePacketFees` and `DistributePacketFeesOnTimeout` - each distribution event would need this check and would need to return a nil error which can be confusing from a code point of view

colin-axner: I also generally don't like the idea of function contracts. They are unenforceable and more prone to misuse

colin-axner: Yes it is. The module may become locked during fee distribution. I'd have to add checks  around every call to distribute fee. If the balance becomes negative due to locking, I'd need to return immediately

I find this to be much easier to reason about from a code point of view. I also don't like the idea of function contracts since they are unenforceable and prone to misuse

colin-axner: I think I should actually go with a solution similar to this https://github.com/cosmos/ibc-go/pull/1029/commits/2b9aa5d69d2ead8c3027d7a05df9a5982143bb19

Since we now have multiple fees per ID, we probably want to revert all state changes if a negative balance would be reached and then return immediately

colin-axner: Thoughts?

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -57,6 +61,10 @@ func (k Keeper) PayPacketFee(goCtx context.Context, msg *types.MsgPayPacketFee)
 func (k Keeper) PayPacketFeeAsync(goCtx context.Context, msg *types.MsgPayPacketFeeAsync) (*types.MsgPayPacketFeeAsyncResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	if k.IsLocked(ctx) {
+		return nil, types.ErrFeeModuleLocked
+	}
+
 	if err := k.EscrowPacketFee(ctx, msg.PacketId, msg.PacketFee); err != nil {
```
colin-axner, line 68: Should we move the `IsFeeEnabled` check from EscrowPacketFee to here and make EscrowPacketFee private?

modules/apps/29-fee/keeper/escrow.go
AdityaSripal: Agreed, I think caching and discarding a partially executed tx is much cleaner.

We should cache the context and if the balance goes negative. Let's discard state changes and commit just the write to the locked flag

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -124,22 +124,51 @@ func (suite *KeeperTestSuite) TestDistributeFee() {
...
 
 	validSeq := uint64(1)
 
 	testCases := []struct {
-		name     string
-		malleate func()
-		expPass  bool
+		name      string
+		malleate  func()
+		expResult func()
```
colin-axner, line 137: I made these changes to test `distributeFee`. No longer necessary for this pr, but will be for future pr's (implementing the locking functionality) so I will leave the changes

colin-axner:
superseded by #1239

PR #1042, opened 2022-03-02
add function EscrowAccountHasBalance

colin-axner:
## Description

To be used for checking for a potential negative balance after fee distribution

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/apps/29-fee/keeper/keeper.go
+++ b/modules/apps/29-fee/keeper/keeper.go
@@ -77,6 +77,17 @@ func (k Keeper) GetFeeModuleAddress() sdk.AccAddress {
 	return k.authKeeper.GetModuleAddress(types.ModuleName)
 }
 
+// EscrowAccountHasBalance verifies if the escrow account has the provided fee.
+func (k Keeper) EscrowAccountHasBalance(ctx sdk.Context, fee types.Fee) bool {
+	for _, coin := range fee.Total() {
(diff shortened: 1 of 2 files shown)


modules/apps/29-fee/keeper/keeper.go
```diff
@@ -77,6 +77,17 @@ func (k Keeper) GetFeeModuleAddress() sdk.AccAddress {
 	return k.authKeeper.GetModuleAddress(types.ModuleName)
 }
 
+// EscrowAccountHasBalance verifies if the escrow account has the provided fee.
+func (k Keeper) EscrowAccountHasBalance(ctx sdk.Context, fee types.Fee) bool {
```
colin-axner, line 81: The API changes the usage of this. If I swap fee to be sdk.Coins then I could call this within `distributeFee`. If I leave as is then I need to call twice, one for DistributePacketFees and one for DistributePacketFeesOnTimeout. 

I'm kinda leaning towards just calling with distributeFee. Might be best to isolate all fee distribution/locking logic to a single funciton. Thoughts?

damiannolan: I agree here, I think it would be cleaner as `sdk.Coins` and call in `distributeFee`. Like you say, that way everything is encapsulated inside a single function, and it's easier to reason about 👍

damiannolan (approved):

Issue #879, opened 2022-02-07
Add 668 functions to 07-tendermint


colin-axner:
In preparation for the additional functions to the client state interface, #668, add:
- VerifyHeader
- CheckForMisbehaviour
- UpdateStateOnMisbehaviour
- UpdateState

for tendermint client

### PR 1

~~Rename [`Misbehaviour`](https://github.com/cosmos/ibc-go/blob/c378ff3b9e637dbae171d2c5129d966b468a2305/modules/light-clients/07-tendermint/types/misbehaviour.go#L21) to `DuplicateHeightHeader` or some name like that (we can discuss on pr)~~

We decided to leave as Misbehaviour, see [comment](https://github.com/cosmos/ibc-go/pull/1090#discussion_r822876705)

### PR 2

Rename [`checkValidity`](https://github.com/cosmos/ibc-go/blob/c378ff3b9e637dbae171d2c5129d966b468a2305/modules/light-clients/07-tendermint/types/update.go#L87) to `VerifyHeader`. Code may need to be modified to fulfill the expected functionality of `VerifyHeader`, including changing API or pulling in/removing code from the existing function. 

Add a switch
```go
switch header.(type) {
case NormalHeader:
   // verify one TM header
  
case DuplicateHeightHeader:
    // verify both TM headers
   
default:
    return error
}
```

Additional code may need to be added to the normal header or duplicate height header. Code should be deduplicated from `CheckMisbehaviourAndUpdateState` as well

### PR 3

CheckForMisbehaviour should check for duplicate height misbehaviour and BFT Time violation misbehaviour. See existing [code](https://github.com/cosmos/ibc-go/blob/c378ff3b9e637dbae171d2c5129d966b468a2305/modules/light-clients/07-tendermint/types/update.go#L91-#L111). Code should be deduplicated from `CheckMisbehaviourAndUpdateState` as well. Ensure duplicate header misbehaviour is checked by seeing if a consensus state already exists for that height

#### PR 4

~UpdateStateOnMisbehaviour set client state in state to be [frozen](https://github.com/cosmos/ibc-go/blob/c378ff3b9e637dbae171d2c5129d966b468a2305/modules/light-clients/07-tendermint/types/update.go#L109)~

Closed by: #1168 

### PR 5

[...]

PR #1090, opened 2022-03-09
chore: rename 07-tendermint type Misbehaviour to ConflictingHeaders

colin-axner:
## Description

ref: #879 

---

write a little note why.


proto/ibc/lightclients/tendermint/v1/tendermint.proto
```diff
@@ -75,9 +75,9 @@ message ConsensusState {
   ];
 }
 
-// Misbehaviour is a wrapper over two conflicting Headers
-// that implements Misbehaviour interface expected by ICS-02
-message Misbehaviour {
+// ConflictingHeaders is a wrapper over two conflicting Headers
+// that implements Header interface expected by ICS-02
+message ConflictingHeaders {
```
colin-axner, line 80: @AdityaSripal thoughts on the naming here?

We could potentially leave this as `Misbehaviour`?

I'm still have some concerns about merging the Misbehaviour and Header interface, mostly around naming. It'd be nice if we could align with the spec as well

We know a normal `Header` can lead to a valid client update, or be evidence of misbehaviour. We also know there can be another structure which can show misbehaviour (in this case duplicate header), but I imagine a light client could have multiple of these structures. My main concern is that `Header` is usually associated with the header produced by a blockchain, but the interface we are using expects potentially one or more headers to be passed in to update the IBC client (either to a valid state or to freeze it)

colin-axner: Oh that's great. I'm in favor!

colin-axner:
Leaving as misbehaviour as @AdityaSripal suggests

Issue #284, opened 2021-07-22
Modify CheckHeaderAndUpdateState to set client state/consensus states in IBC light client implementations


colin-axner:
## Summary

02-client should not set client states and consensus states as generalizing IBC client implementations is difficult. Better to give IBC clients more control and flexibility. This will already be useful for solo machines which don't need to store consensus states for each update height

This design is already specified in ICS, our implementation didn't follow those changes. See [comment](https://github.com/cosmos/ibc/issues/569#issuecomment-884783641)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Revisiting this decision many months later, I'm trying to piece together the exact benefits. 

This decision was sparked by the motivation of allowing the client state definition to be upgraded over time. The idea was that the light client code internally use a different type than what is stored in the client store. But now I see that it would also be possible to create an entirely new client state definition for the breaking versions of the same client. That is, if we modified the client state definition of the IBC tendermint client, the new definition should be v2 and treated as an entirely different client rather than a continuation of v1. Note, this upgrade path still requires all counterparty chains to understand v2 before upgrading. 

There do remain subtle benefits, such as the consensus state not necessarily needing to be set for every consensus state (solo machines), or removal of the `GetHeight` function from the header interface. The removal of the `GetHeight` function from the header interface was requested as part of the Wasm client implementation, but was also justified by the proposed changes. It might be possible to find a solution which allows for the `GetHeight` removal without completely handing over all responsibility to the light client developer

The changes would also align with the spec, but increases the responsibility of light client developers. 

One possible adjustment that comes to mind is only removing the setting of consensus states from 02-client. This would require light client developers to set the consensus states themselves (allowing header.GetHeight to be removed and stopping unnecessary state storage for solo machines). Although at that point, I'm not sure why they shouldn't be expected to update the client state as well

@mkaczanowski do you have any insights into whether this change would positively or negatively benefit the Wasm client? 

cc @AdityaSripal @cwgoes any thoughts on how other light client types might be affected by this decision?

colin-axner:
[...]

PR #1144, opened 2022-03-21
chore: add solomachine client storage

damiannolan:
## Description

- Adds storage operations to `UpdateState` and `UpdateStateOnMisbehaviour`

closes: #XXXX

---

write a little note why.


modules/light-clients/06-solomachine/types/update.go
```diff
@@ -112,6 +123,49 @@ func (cs ClientState) UpdateState(ctx sdk.Context, cdc codec.BinaryCodec, client
...
+	// set default consensus height with header height
+	consensusHeight := smHeader.GetHeight()
+	if cs.ClientType() != exported.Localhost {
+		clientStore.Set(host.ConsensusStateKey(consensusHeight), types.MustMarshalConsensusState(cdc, consensusState))
+	} else {
+		consensusHeight = types.GetSelfHeight(ctx)
+	}
+
+	// TODO: Should be telemetry and events be emitted from 02-client?
+	// The clientID would need to be passed an additional arg here
```
damiannolan, line 138: I think we need to make a decision on whether telemetry and events should remain to be emitted from `02-client` or from within the `ClientState` implementation. We may need to pass the client ID as an additional arg if we want to do this in the lightclient implementation, or otherwise we may need to return the consensus height. Any thoughts on this? cc. @colin-axner @seantking

colin-axner: We should not require light client implementations to emit events. This decision informed the proposed design. We should allow light client to optionally emit their own events and telemetry. This is technically possible by constructing a `ClientMessage` which contains the clientID, but I'm not opposed to adding the clientID as an argument to the function since 02-client already has that information. It doesn't provide any usefulness outside of event emission/telemetry since the client store is prefixed by the clientID

modules/light-clients/06-solomachine/types/update.go
```diff
@@ -1,11 +1,17 @@
...
 import (
+	"encoding/hex"
+
+	"github.com/armon/go-metrics"
 	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 
+	"github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
```
colin-axner, line 12: alias should be clienttypes. Any `types` package not from this submodule should have an alias to distinguish which types package is being used. (ie clienttypes, connectiontypes, channeltypes, etc)

modules/light-clients/06-solomachine/types/update.go
```diff
@@ -112,6 +123,49 @@ func (cs ClientState) UpdateState(ctx sdk.Context, cdc codec.BinaryCodec, client
...
+		telemetry.IncrCounterWithLabels(
+			[]string{"ibc", "client", "update"},
+			1,
+			[]metrics.Label{
+				telemetry.NewLabel(types.LabelClientType, cs.ClientType()),
+				telemetry.NewLabel(types.LabelClientID, "clientID"), // TODO: Should clientID be passed as an arg
+				telemetry.NewLabel(types.LabelUpdateType, "msg"),
+			},
+		)
+	}()
```
colin-axner, line 149: can be removed, this should occur in 02-client

modules/light-clients/06-solomachine/types/update.go
```diff
@@ -112,6 +123,49 @@ func (cs ClientState) UpdateState(ctx sdk.Context, cdc codec.BinaryCodec, client
...
+			sdk.NewAttribute(types.AttributeKeyClientID, "clientID"),
+			sdk.NewAttribute(types.AttributeKeyClientType, cs.ClientType()),
+			sdk.NewAttribute(types.AttributeKeyConsensusHeight, consensusHeight.String()),
+			sdk.NewAttribute(types.AttributeKeyHeader, headerStr),
+		),
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+		),
+	})
```
colin-axner, line 167: ditto

modules/light-clients/06-solomachine/types/update.go
```diff
@@ -112,6 +123,49 @@ func (cs ClientState) UpdateState(ctx sdk.Context, cdc codec.BinaryCodec, client
...
 	cs.ConsensusState = consensusState
+
+	clientStore.Set(host.ClientStateKey(), types.MustMarshalClientState(cdc, &cs))
+
+	// set default consensus height with header height
+	consensusHeight := smHeader.GetHeight()
+	if cs.ClientType() != exported.Localhost {
+		clientStore.Set(host.ConsensusStateKey(consensusHeight), types.MustMarshalConsensusState(cdc, consensusState))
+	} else {
+		consensusHeight = types.GetSelfHeight(ctx)
```
colin-axner, line 134: where is this coming from? Solo machines don't store consensus states

modules/light-clients/06-solomachine/types/update.go
```diff
@@ -126,9 +180,12 @@ func (cs ClientState) CheckForMisbehaviour(_ sdk.Context, _ codec.BinaryCodec, _
...
 // as it does not perform any misbehaviour checks.
-func (cs ClientState) UpdateStateOnMisbehaviour(
-	_ sdk.Context, _ codec.BinaryCodec, _ sdk.KVStore, // prematurely include args for self storage of consensus state
-) (*ClientState, exported.ConsensusState, error) {
+func (cs ClientState) UpdateStateOnMisbehaviour(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore) (*ClientState, exported.ConsensusState, error) {
 	cs.IsFrozen = true
+
+	clientStore.Set(host.ClientStateKey(), clienttypes.MustMarshalClientState(cdc, &cs))
+
+	// TODO: Telemetry and events
```
colin-axner, line 188: can be removed

modules/light-clients/06-solomachine/types/update.go
colin-axner: To expand further on why we should not require light client implementations to emit events:

Relayers make use of the events we have in 02-client. By passing this to light client implementations, we lose standardization and risk light client implementators making mistakes. Since the functionality called by 02-client is very verbose now, 02-client can emit appropriate events when doing updates or updates on misbehaviour

damiannolan: Okay, great, that makes a lot of sense! Thanks, I'll remove the events and telemetry from here.

modules/light-clients/06-solomachine/types/update.go
damiannolan: Slightly confused here, isn't [this code](https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/keeper/client.go#L100-L106) run for solomachines too?

colin-axner: It is, but it shouldn't be. This was another justification for why setting the consensus state within the light client implementation is useful. I thought I documented this clearly, but this was all I could find:

>  This would require light client developers to set the consensus states themselves (allowing header.GetHeight to be removed and stopping unnecessary state storage for solo machines). 

[ref](https://github.com/cosmos/ibc-go/issues/284#issuecomment-988767293)

Solo machines shouldn't be setting a consensus state ever. It has a single consensus state which is constantly updated and thus it is stored within the client state. The code linked above was an inefficiency, performing unnecessary storage

damiannolan: Ahh, okay! Tbh I was confused why `ConsensusState` existed as a field on the solomachine `ClientState` and the extra storage was happening.
This should all be much easier to reason about when the refactor is done. Thank you for clarifying this and providing context, much appreciated!

Updated and removed this code :)

modules/light-clients/06-solomachine/types/update.go
```diff
@@ -112,6 +118,15 @@ func (cs ClientState) UpdateState(ctx sdk.Context, cdc codec.BinaryCodec, client
...
 	cs.Sequence++
 	cs.ConsensusState = consensusState
+
+	clientStore.Set(host.ClientStateKey(), clienttypes.MustMarshalClientState(cdc, &cs))
+
+	// set default consensus height with header height
+	consensusHeight := smHeader.GetHeight()
+	if cs.ClientType() != exported.Localhost {
+		clientStore.Set(host.ConsensusStateKey(consensusHeight), clienttypes.MustMarshalConsensusState(cdc, consensusState))
+	}
```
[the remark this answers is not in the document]
damiannolan: done

modules/light-clients/06-solomachine/types/update_test.go
```diff
@@ -615,7 +623,7 @@ func (suite *SoloMachineTestSuite) TestUpdateState() {
 						suite.Require().Equal(consensusState, cs.(*types.ClientState).ConsensusState)
```
colin-axner, line 623: can you add checks on the successful case that the correct client state is set in the store?

modules/light-clients/06-solomachine/types/update.go
```diff
@@ -126,9 +135,10 @@ func (cs ClientState) CheckForMisbehaviour(_ sdk.Context, _ codec.BinaryCodec, _
 
 // UpdateStateOnMisbehaviour updates state upon misbehaviour. This method should only be called on misbehaviour
 // as it does not perform any misbehaviour checks.
-func (cs ClientState) UpdateStateOnMisbehaviour(
-	_ sdk.Context, _ codec.BinaryCodec, _ sdk.KVStore, // prematurely include args for self storage of consensus state
-) (*ClientState, exported.ConsensusState, error) {
+func (cs ClientState) UpdateStateOnMisbehaviour(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore) (*ClientState, exported.ConsensusState, error) {
```
colin-axner, line 138: Is there a test for this function? Can you check that the correct client state is set in the store?

colin-axner: Sweet, I just figure better to do now so when we remove the return values we don't have to add testing code

damiannolan: Definitely! Agreed 💯

seantking (approved):

PR #1146, opened 2022-03-21
update migration document for ica to make it clearer

faddat:
## Description

New users may not initially understand that an ica host can support any sdk message so I've made the migration docs more instructive. 

closes: #XXXX

---

write a little note why.


docs/migrations/v2-to-v3.md
```diff
@@ -42,11 +42,38 @@ app.UpgradeKeeper.SetUpgradeHandler("v3",
...
+        // create ICS27 Host submodule params 
         hostParams := icahosttypes.Params{
             HostEnabled: true, 
-            AllowMessages: []string{"/cosmos.bank.v1beta1.MsgSend", ...}, 
-        }
+			hostParams := icahosttypes.Params{
+					HostEnabled: true,
+                    // here you can add any transaction supported by your chain, and ica controllers will be able to make
+                    // those transactions on your chain.
+					AllowMessages: []string{
```
crodriguezvega, line 52: I agree that adding an extra explanation could be helpful. Maybe also adding the link to the section of the docs that explains the [`AllowMessages` parameter](https://github.com/cosmos/ibc-go/blob/main/docs/apps/interchain-accounts/parameters.md#allowmessages).
But I don't think we should add a long list of SDK messages here; with one or two as example, I think it's more than enough.

crodriguezvega (changes requested):
Thanks for this improvement, @faddat! 
I personally think that adding such a long list of SDK messages is not necessary, though.

docs/migrations/v2-to-v3.md
```diff
@@ -42,11 +42,38 @@ app.UpgradeKeeper.SetUpgradeHandler("v3",
...
 
-        // create ICS27 Host submodule params
+        // create ICS27 Host submodule params 
         hostParams := icahosttypes.Params{
             HostEnabled: true, 
-            AllowMessages: []string{"/cosmos.bank.v1beta1.MsgSend", ...}, 
-        }
+			hostParams := icahosttypes.Params{
+					HostEnabled: true,
+                    // here you can add any transaction supported by your chain, and ica controllers will be able to make
```
seantking, line 50:
```suggestion
                    // here you can add any transaction supported by your host chain, and ica controllers will be able to make
```

docs/migrations/v2-to-v3.md
```diff
@@ -42,11 +42,38 @@ app.UpgradeKeeper.SetUpgradeHandler("v3",
...
+                    // those transactions on your chain.
```
seantking, line 51:
```suggestion
                    // those transactions on your host chain.
```

docs/migrations/v2-to-v3.md
colin-axner: Thanks for taking the time to add all these messages. I appreciate the improvement to the migration docs!

I do agree that explicitly providing an example will create a sort of "default" within the ecosystem, but I think it's important to respect the purpose of the `AllowMessages` parameter. The intent of adding such a parameter was to limit the functionality a controller chain could perform on the host chain since the messages executed by interchain account transactions does not follow the exact same authentication path as regular transactions (although in theory they should be equivalent). Thus, in my opinion, providing few examples and explaining that any message could be added is perhaps the best way to get each chain to actually consider the implications here. We shouldn't be providing a default. The default allowed messages should be empty (which is why this needs to be set during an upgrade handler)

colin-axner:
Hey! Are there still plans to apply the [requested changes](https://github.com/cosmos/ibc-go/pull/1146#discussion_r831050786)? Or should we close this pr?

colin-axner:
Hey! Are there still plans to apply the [requested changes](https://github.com/cosmos/ibc-go/pull/1146#discussion_r831050786)? Or should we close this pr?

crodriguezvega:
After receiving feedback from different teams, we are going to [implement the whitelist all functionality](https://github.com/cosmos/ibc-go/issues/1479), so maybe we can close this PR then?

faddat:
Makes sense to close it in that case ❤️

PR #1163, opened 2022-03-22
refactor: adding CheckForMisbehaviour to 07-tendermint client

damiannolan:
## Description

- Adds `CheckForMisbehaviour` to `07-tendermint` 

ref: #879 PR3

---

write a little note why.

### DIFF
--- a/modules/light-clients/07-tendermint/types/update.go
+++ b/modules/light-clients/07-tendermint/types/update.go
@@ -261,3 +261,71 @@ func update(ctx sdk.Context, clientStore sdk.KVStore, clientState *ClientState, 
 
 	return clientState, consensusState
 }
+
+// CheckForMisbehaviour -
+func (cs ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore, msg exported.ClientMessage) bool {
+	switch msg := msg.(type) {
+	case *Header:
+		tmHeader := msg
+		consState := tmHeader.ConsensusState()
+
+		// Check if the Client store already has a consensus state for the header's height
+		// If the consensus state exists, and it matches the header then we return early
+		// since header has already been submitted in a previous UpdateClient.
+		prevConsState, _ := GetConsensusState(clientStore, cdc, tmHeader.GetHeight())
+		if prevConsState != nil {
+			// This header has already been submitted and the necessary state is already stored
+			// in client store, thus we can return early without further validation.
+			// if reflect.DeepEqual(prevConsState, tmHeader.ConsensusState()) {
+			// 	return false
+			// }
+
+			// A consensus state already exists for this height, but it does not match the provided header.
+			// The assumption is that Header has already been validated. Thus we can return true as misbehaviour is present
+			// return true
+			return !reflect.DeepEqual(prevConsState, tmHeader.ConsensusState())
+		}
+
+		// Check that consensus state timestamps are monotonic
+		prevCons, prevOk := GetPreviousConsensusState(clientStore, cdc, tmHeader.GetHeight())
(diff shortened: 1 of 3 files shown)


damiannolan:
I think there is some more test scenarios that need to be covered from looking at `misbehaviour_handle_test.go`, but I wanted to share this so I know I'm moving in the right direction. I've updated the return sig from `(bool, error)` to only return a `bool`, and adapted the logic.
Just want to confirm logic is correct before moving forward with more test cases. cc. @colin-axner @seantking

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -261,3 +261,70 @@ func update(ctx sdk.Context, clientStore sdk.KVStore, clientState *ClientState,
...
+			// Ensure that Commit Hashes are different
+			if !bytes.Equal(blockID1.Hash, blockID2.Hash) {
+				return true
+			}
+		} else {
+			// Header1 is at greater height than Header2, therefore Header1 time must be less than or equal to
+			// Header2 time in order to be valid misbehaviour (violation of monotonic time).
+			if tmMisbehaviour.Header1.SignedHeader.Header.Time.Before(tmMisbehaviour.Header2.SignedHeader.Header.Time) {
+				return true
+			}
```
colin-axner, line 325: I think we can just return true here. I think this should be done in `VerifyClientMessage`. If the type is `Misbehaviour` it should be considered invalid if it isn't evidence of misbehaviour

damiannolan: Okay, I get you! So if the type is indeed `Misbehaviour` then it should be validated as such in `VerifyClientMessage` and `CheckForMisbehaviour` can just directly return true for that type. So `CheckForMisbehaviour` should only check for time monotonicity of `Header` types as well as the previous consensus state checks, correct?

I can go ahead and remove this code and the tests for `Misbehaviour` types.

colin-axner: correct! Does this API still make sense? It does seem kinda odd that developers need to know how these functions are used. But at the same time, I guess the perspective is "this is misbehaviour, thus I return true" and "this is a header, does it happen to be evidence of misbehaviour?"

modules/light-clients/07-tendermint/types/update.go
```diff
@@ -261,3 +261,48 @@ func update(ctx sdk.Context, clientStore sdk.KVStore, clientState *ClientState,
...
+		if prevConsState != nil {
+			// This header has already been submitted and the necessary state is already stored
+			// in client store, thus we can return early without further validation.
+			if reflect.DeepEqual(prevConsState, tmHeader.ConsensusState()) {
+				return false
+			}
+
+			// A consensus state already exists for this height, but it does not match the provided header.
+			// The assumption is that Header has already been validated. Thus we can return true as misbehaviour is present
+			return true
```
damiannolan, line 285: My linter here suggests this should be simplified to: 
```
return !reflect.DeepEqual(prevConsState, tmHeader.ConsensusState())
```

Any objections to changing this and I can adjust the comments as necessary?

Mehmetbatmaz (approved):

modules/light-clients/07-tendermint/types/update.go
seantking: I actually find this suggestion harder to parse than what exists currently.

damiannolan: Same tbh, it always is when it's the negation of some boolean expression. Happy to leave as is, I wonder if we can suppress the static-check linter here.

colin-axner: I'd prefer to leave as is

colin-axner (approved):
Nice work :100:

seantking (approved):
nice!

PR #1169, opened 2022-03-24
test: add UpgradeChain and UpgradeClient endpoint test functions

colin-axner:
## Description

Adds:
- UpgradeChain
- UpgradeClient

`UpgradeChain` is a helper function to test upgrading a chain to a different revision number. It will change the testing chain's revision number and set the associated client/consensus states in state. It does not use upgrade plan's, though we can modify this behaviour in the future

`UpgradeClient` submits a `MsgUpgradeClient` using the upgraded client and consensus states set on the counterparty chain

closes: #XXXX

---

write a little note why.

### DIFF
--- a/testing/endpoint.go
+++ b/testing/endpoint.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 	"github.com/stretchr/testify/require"
 
@@ -496,14 +497,21 @@ func (endpoint *Endpoint) TimeoutOnClose(packet channeltypes.Packet) error {
 // UpgradeChain performs an IBC client upgrade using the provided client state.
 // The chainID within the client state will have its revision number incremented by 1.
 // The counterparty client will be upgraded if it exists.
-func (endpoint *Endpoint) UpgradeChain(clientState *ibctmtypes.ClientState) error {
+func (endpoint *Endpoint) UpgradeChain(clientState *ibctmtypes.ClientState) (err error) {
 	if endpoint.Counterparty.ClientID == "" {
 		return fmt.Errorf("cannot upgrade chain if there is no counterparty client")
 	}
 
+	// the upgrade will be perfromed in 2 blocks
+	// the current block will commit the upgradeClientState into state
+	// the next block will commit the upgradeConsensusState into state via begin blocker
+	// the upgrade height will be used to update the counterparty client and perform the upgrade
+	upgradeHeight := endpoint.Chain.GetContext().BlockHeight() + 2
+
 	// increment revision number in chainID
-	revisionNumber := clienttypes.ParseChainID(clientState.ChainId)
-	newChainID, err := clienttypes.SetRevisionNumber(clientState.ChainId, revisionNumber+1)
+	oldChainID := clientState.ChainId
+	revisionNumber := clienttypes.ParseChainID(oldChainID)
+	newChainID, err := clienttypes.SetRevisionNumber(oldChainID, revisionNumber+1)
 	if err != nil {
 		// current chainID is not in revision format
 		newChainID = clientState.ChainId + "-1"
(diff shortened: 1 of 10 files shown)


testing/endpoint.go
```diff
@@ -492,6 +493,85 @@ func (endpoint *Endpoint) TimeoutOnClose(packet channeltypes.Packet) error {
 	return endpoint.Chain.sendMsgs(timeoutOnCloseMsg)
 }
 
+// UpgradeChain performs a IBC client upgrade using the provided client state.
```
seantking, line 496: nit: `an IBC`

testing/endpoint.go
```diff
@@ -492,6 +493,85 @@ func (endpoint *Endpoint) TimeoutOnClose(packet channeltypes.Packet) error {
...
+	// commit changes so they are stored in state
+	endpoint.Chain.NextBlock()
+
+	endpoint.Chain.ChainID = newChainID
+	endpoint.Chain.CurrentHeader.ChainID = newChainID
+
+	if endpoint.Counterparty.ClientID != "" {
+
+	}
+
```
colin-axner, line 532: whoops, forgot to add `endpoint.UpgradeClient` call

testing/endpoint.go
```diff
@@ -492,6 +493,85 @@ func (endpoint *Endpoint) TimeoutOnClose(packet channeltypes.Packet) error {
...
+	clientState.ChainId = newChainID
+
+	// set upgraded client state
+	bz := clienttypes.MustMarshalClientState(endpoint.Chain.App.AppCodec(), clientState)
+	endpoint.Chain.GetSimApp().UpgradeKeeper.SetUpgradedClient(endpoint.Chain.GetContext(), endpoint.Chain.GetContext().BlockHeight(), bz)
+
+	// set upgraded consensus state
+	consensusState := &ibctmtypes.ConsensusState{
```
damiannolan, line 515: Does this to include a root/app hash?

damiannolan (approved):
Nice! Super useful, just left one question :)

testing/endpoint.go
```diff
@@ -492,6 +493,85 @@ func (endpoint *Endpoint) TimeoutOnClose(packet channeltypes.Packet) error {
 	return endpoint.Chain.sendMsgs(timeoutOnCloseMsg)
 }
 
+// UpgradeChain performs a IBC client upgrade using the provided client state.
+// The chainID within the client state will have its revision number incremented by 1.
+// The counterparty client will be upgraded if it exists.
+func (endpoint *Endpoint) UpgradeChain(clientState *ibctmtypes.ClientState) error {
```
seantking, line 499: It might be useful to pass revisionID as a parameter rather than incrementing :thinking:

testing/endpoint.go
colin-axner: Nope, the consensus state is set for the next block to be produced under the new chain id so the root hash cannot be known before hand

colin-axner:
It took me a while, but I mostly understand the issue here. It is very nuanced.

ref https://github.com/cosmos/cosmos-sdk/pull/8187/files#discussion_r545757275

Note: `chain.App.Commit()` != a commit to a block

So tendermint process is:
```
vote
precommit (checkTx happening somewhere around here)
commit to block (height x)
beginblock (height x)
delivertx
deilvertx
delivertx
endblock
app.Commit
```

the testing package does things out of order
```
begin (height x)
deliver
end
app.Commit 
commit to block (height x)
begin block (height x + 1)
```

in most situations it doesn't matter because we never verify functionality between `commit to block` and `beginblock` - that is until now.

Normally I'd fix this issue in my specific instance by actually doing tm process for the upgrade
That is

```
begin block (upgradeHeight - 1)
app.Commit()
// swap to tm process
commit to block height x
// perform client upgrades
```

The problem is `app.Commit()` wipes the `deliverState` (which is beginblock is always called first). Every call to `GetContext()` panics (since `GetContext()` says to create a context for a deliver tx)

In other words, the testing package in order to be useful needs to be able to modify state during block execution (beginblock must be called).

We could be more explicit with checktx vs delivertx usage for context (all queries use checktx)

Quickly modifying my code to do this, I still run into

```
failed to execute message; message index: 0: cannot upgrade client with ID 07-tendermint-0: client state proof failed. Path: /upgrade/upgradedIBCState/6/upgradedClient: proof did not commit to expected root: 6DF8EF50366AA416EF9A8A1676D5CABFC499499C18D143B9CE9EC29538EDD1E6, got: 565FD18B44E72D4CE76E18B3995569DC3C913B5DC6BE4DB37DE60E09086C92A1. Please ensure proof was submitted with correct proofHeight and to the correct chain.: invalid proof
```

colin-axner:
I fixed the tests, but it currently requires a few hacks:
- mocking behaviour of `SendMsgs` minus call to begin block on all chains
- usage of CheckTx Context after `app.Commit` but before `BeginBlock`

I think it might be best to change the testing package to more elegantly mock the TM process before merging these changes, but I think this pr is very useful in understanding how things work. 

In the short term, on the 02-client-refactor branch we can add a testing helper function which simply sets the lastest client/consensus state to be after the upgrade (bypassing the actual verification of such an upgrade)

Issue #1140, opened 2022-03-17
Update 07-tendermint GetConsensusState to return a bool instead of an error


damiannolan:
## Summary

The helper function [`GetConsensusState`](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/store.go#L55) in `07-tendermint` currently returns `(*ConsensusState, error)`.
This deviates from the standard pattern of using `(type, bool)` as the return signature for state accessor functions used throughout ibc-go and the cosmos-sdk. 

## Proposal

- Consider updating the return values of [`GetConsensusState`](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/store.go#L55) to return `(*ConsensusState, error)`
- Panic in the event of malformed bytes (i.e. the stored bytes cannot be unmarshalled to the desired type)

ref: [PR comment](https://github.com/cosmos/ibc-go/pull/1117/files#r828873134)
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #1180, opened 2022-03-28
chore: update 07-tendermint GetConsensusState to return bool over error

damiannolan:
## Description

- Updates the `07-tendermint` helper `GetConsensusState` to return `(*ConsensusState, bool)` in favour of `(*ConsensusState, error)`

closes: #1140 

---

write a little note why.

### DIFF
--- a/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
+++ b/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
@@ -85,7 +85,7 @@ func (cs *ClientState) verifyMisbehaviour(ctx sdk.Context, clientStore sdk.KVSto
 	// Get consensus bytes from clientStore
 	tmConsensusState1, found := GetConsensusState(clientStore, cdc, misbehaviour.Header1.TrustedHeight)
 	if !found {
-		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header1 at TrustedHeight: %s", misbehaviour.Header1)
+		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header1 at TrustedHeight: %s", misbehaviour.Header1.TrustedHeight)
 	}
 
 	// Get consensus bytes from clientStore
--- a/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
+++ b/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
@@ -80,15 +80,11 @@ func (cs *ClientState) verifyMisbehaviour(ctx sdk.Context, clientStore sdk.KVSto
 	// Regardless of the type of misbehaviour, ensure that both headers are valid and would have been accepted by light-client
 
 	// Retrieve trusted consensus states for each Header in misbehaviour
-	// and unmarshal from clientStore
-
-	// Get consensus bytes from clientStore
 	tmConsensusState1, found := GetConsensusState(clientStore, cdc, misbehaviour.Header1.TrustedHeight)
 	if !found {
 		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header1 at TrustedHeight: %s", misbehaviour.Header1.TrustedHeight)
 	}
 
-	// Get consensus bytes from clientStore
 	tmConsensusState2, found := GetConsensusState(clientStore, cdc, misbehaviour.Header2.TrustedHeight)
 	if !found {
 		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header2 at TrustedHeight: %s", misbehaviour.Header2.TrustedHeight)
--- a/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
+++ b/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
@@ -91,7 +91,7 @@ func (cs *ClientState) verifyMisbehaviour(ctx sdk.Context, clientStore sdk.KVSto
 	// Get consensus bytes from clientStore
 	tmConsensusState2, found := GetConsensusState(clientStore, cdc, misbehaviour.Header2.TrustedHeight)
 	if !found {
-		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header2 at TrustedHeight: %s", misbehaviour.Header2)
+		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header2 at TrustedHeight: %s", misbehaviour.Header2.TrustedHeight)
 	}
 
 	// Check the validity of the two conflicting headers against their respective
--- a/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
+++ b/modules/light-clients/07-tendermint/types/misbehaviour_handle.go
@@ -64,15 +64,15 @@ func (cs ClientState) CheckMisbehaviourAndUpdateState(
 	// and unmarshal from clientStore
 
 	// Get consensus bytes from clientStore
-	tmConsensusState1, err := GetConsensusState(clientStore, cdc, tmMisbehaviour.Header1.TrustedHeight)
-	if err != nil {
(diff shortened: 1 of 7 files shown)


damiannolan:
Happy to leave this on the back burner until the higher priority PRs are merged. Should minimise disruptions with merge conflicts..etc

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -59,29 +58,23 @@ func SetConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensus
...
-			clienttypes.ErrConsensusStateNotFound,
-			"consensus state does not exist for height %s", height,
-		)
+		return nil, false
 	}
 
 	consensusStateI, err := clienttypes.UnmarshalConsensusState(cdc, bz)
 	if err != nil {
-		return nil, sdkerrors.Wrapf(clienttypes.ErrInvalidConsensus, "unmarshal error: %v", err)
+		return nil, false
```
colin-axner, line 69: should we panic?

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -59,29 +58,23 @@ func SetConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensus
...
 	}
 
 	consensusState, ok := consensusStateI.(*ConsensusState)
 	if !ok {
-		return nil, sdkerrors.Wrapf(
-			clienttypes.ErrInvalidConsensus,
-			"invalid consensus type %T, expected %T", consensusState, &ConsensusState{},
-		)
+		return nil, false
```
colin-axner, line 74: ditto?

modules/light-clients/07-tendermint/types/store.go
damiannolan: Wondered that too myself, was considering putting it in this PR, but I figured I'd wait to discuss here. 

I think its reasonable enough to panic if unmarshalling or type assertion fails. It *should* be unreachable code anyway, but in the event it fails to unmarshal for example something has clearly gone wrong..

Happy to panic here if you are. Any thoughts @seantking ?

seantking: I think panic here makes sense. I don't see a reason to err gracefully in this situation :thinking:

seantking (approved):
LGTM once @colin-axner suggestion is applied.

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -59,29 +58,18 @@ func SetConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensus
...
-		return nil, sdkerrors.Wrapf(clienttypes.ErrInvalidConsensus, "unmarshal error: %v", err)
-	}
-
-	consensusState, ok := consensusStateI.(*ConsensusState)
-	if !ok {
-		return nil, sdkerrors.Wrapf(
-			clienttypes.ErrInvalidConsensus,
-			"invalid consensus type %T, expected %T", consensusState, &ConsensusState{},
-		)
+		panic(err)
```
houshengbo, line 69: When this functions returns an error, at least something descriptive was returned, and that error is not fatal to use panic.
Could you elaborate more on why panic is being used here. Is the error here fatal to crash the system?
If malformed bytes is unlikely to happen, do we have validation on cdc before calling UnmarshalConsensusState? I think this validation can make sure we do not run into this fatal error.

colin-axner: The SDK gracefully handles panic during transaction execution. The error returned for the transaction usually wipes the original error message but indicates that something very unexpected occurred. In this case, for the panic to occur, there must be a bug in the existing code (we must of set malformed bytes in state)

modules/light-clients/07-tendermint/types/misbehaviour_handle.go
```diff
@@ -80,14 +80,18 @@ func (cs *ClientState) verifyMisbehaviour(ctx sdk.Context, clientStore sdk.KVSto
 	// Regardless of the type of misbehaviour, ensure that both headers are valid and would have been accepted by light-client
 
 	// Retrieve trusted consensus states for each Header in misbehaviour
-	tmConsensusState1, err := GetConsensusState(clientStore, cdc, misbehaviour.Header1.TrustedHeight)
-	if err != nil {
-		return sdkerrors.Wrapf(err, "could not get trusted consensus state from clientStore for Header1 at TrustedHeight: %s", misbehaviour.Header1.TrustedHeight)
+	// and unmarshal from clientStore
+
+	// Get consensus bytes from clientStore
```
colin-axner, line 85: comment can be removed, below as well

modules/light-clients/07-tendermint/types/misbehaviour_handle.go
```diff
@@ -80,14 +80,18 @@ func (cs *ClientState) verifyMisbehaviour(ctx sdk.Context, clientStore sdk.KVSto
...
+	tmConsensusState1, found := GetConsensusState(clientStore, cdc, misbehaviour.Header1.TrustedHeight)
+	if !found {
+		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header1 at TrustedHeight: %s", misbehaviour.Header1)
```
colin-axner, line 88:
```suggestion
		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header1 at TrustedHeight: %s", misbehaviour.Header1.TrustedHeight)
```

modules/light-clients/07-tendermint/types/misbehaviour_handle.go
```diff
@@ -80,14 +80,18 @@ func (cs *ClientState) verifyMisbehaviour(ctx sdk.Context, clientStore sdk.KVSto
...
 	}
 
-	tmConsensusState2, err := GetConsensusState(clientStore, cdc, misbehaviour.Header2.TrustedHeight)
-	if err != nil {
-		return sdkerrors.Wrapf(err, "could not get trusted consensus state from clientStore for Header2 at TrustedHeight: %s", misbehaviour.Header2.TrustedHeight)
+	// Get consensus bytes from clientStore
+	tmConsensusState2, found := GetConsensusState(clientStore, cdc, misbehaviour.Header2.TrustedHeight)
+	if !found {
+		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header2 at TrustedHeight: %s", misbehaviour.Header2)
```
colin-axner, line 94:
```suggestion
		return sdkerrors.Wrapf(clienttypes.ErrConsensusStateNotFound, "could not get trusted consensus state from clientStore for Header2 at TrustedHeight: %s", misbehaviour.Header2.TrustedHeight)
```

modules/light-clients/07-tendermint/types/store.go
```diff
@@ -59,29 +58,14 @@ func SetConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensus
 
 // GetConsensusState retrieves the consensus state from the client prefixed
 // store. An error is returned if the consensus state does not exist.
```
colin-axner, line 60: godoc needs to be updated. Should we make this function private?

modules/light-clients/07-tendermint/types/store_test.go
```diff
@@ -64,16 +65,26 @@ func (suite *TendermintTestSuite) TestGetConsensusState() {
 
 			tc.malleate() // change vars as necessary
 
+			if tc.expPanic {
```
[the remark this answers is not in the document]
colin-axner: lgtm

modules/light-clients/07-tendermint/types/store.go
damiannolan: The panic will essentially bubble up the callstack of the `UpdateClient` handler and the [grpc recovery interceptor](https://github.com/cosmos/cosmos-sdk/blob/master/baseapp/grpcserver.go#L81) in the cosmos sdk will catch and recover 👍

modules/light-clients/07-tendermint/types/store.go
damiannolan: Good spot

damiannolan: Making private makes a bit more awkward for testing. We would actually need to remove the test function for this.
`getConsensusState` will still get coverage from a number of other tests but testing the panic scenarios then isn't really possible.

What do you think? I don't really have any strong feelings about making it public/private

colin-axner: lets leave for now

colin-axner (approved):
Wahoo, nice!

Issue #1089, opened 2022-03-09
Remove localhost client 


colin-axner:
## Summary

The localhost client is currently non-functional. We should either fix it or remove it so there is no confusion. I propose we remove it. The existing code is minimal and I think it'd be easy enough to reimplement it from scratch if we so desire. There was an existing proposed [ADR](https://github.com/cosmos/ibc-go/pull/75) on how to fix the implementation, but it's still very unclear when a use case for the localhost will arise (thus far no one has asked for it to be functional)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Can we use this as a learning exercise? Is there anything valuable about IBC and/or light clients that could be learnt from fixing/re-implementing this?

PR #1187, opened 2022-03-29
refactor: remove localhost client implementation

seantking:
## Description

closes: #1089

---

write a little note why.


modules/core/02-client/abci.go
```diff
@@ -1,42 +0,0 @@
...
-			upgradedConsState := &ibctmtypes.ConsensusState{
-				Timestamp:          ctx.BlockTime(),
-				NextValidatorsHash: ctx.BlockHeader().NextValidatorsHash,
-			}
-			bz := k.MustMarshalConsensusState(upgradedConsState)
-
-			k.SetUpgradedConsensusState(ctx, plan.Height, bz)
-		}
-	}
-
```
colin-axner, line 32: this is not related to the localhost

modules/core/02-client/genesis.go
```diff
@@ -65,7 +60,6 @@ func ExportGenesis(ctx sdk.Context, k keeper.Keeper) types.GenesisState {
 		ClientsMetadata:    clientsMetadata,
 		ClientsConsensus:   k.GetAllConsensusStates(ctx),
 		Params:             k.GetParams(ctx),
-		CreateLocalhost:    false,
```
colin-axner, line 68: we might actually want to leave this param in. I'm not sure what the migration process is like for removing on chain parameters

colin-axner (changes requested):

modules/core/02-client/abci.go
seantking: https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/abci.go#L11

This looks to me like it's only being used by the localhost client? (comments also suggest as much)

seantking: https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/abci.go#L11

This looks to me like it's only being used by the localhost client? (comments also suggest as much)

modules/core/02-client/genesis.go
seantking: I've added it back in and also added it to the `AppModule` struct again (although I'm not sure if this is necessary). From what I can tell we don't actually use the field in `InitGenesis` at all.

modules/core/02-client/abci.go
colin-axner: No the godoc is incorrect. This code is required to perform IBC client upgrades, see in code documentation

modules/core/02-client/abci.go
```diff
@@ -1,42 +0,0 @@
...
-	_, found = k.GetClientState(ctx, exported.Localhost)
-	if !found {
-		return
-	}
-
-	// update the localhost client with the latest block height
-	if err := k.UpdateClient(ctx, exported.Localhost, nil); err != nil {
-		panic(err)
-	}
```
colin-axner, line 41: these lines can be deleted

modules/core/02-client/genesis.go
colin-axner: we can remove the `AppModule` field :+1:

modules/core/02-client/keeper/client.go
```diff
@@ -98,8 +95,7 @@ func (k Keeper) UpdateClient(ctx sdk.Context, clientID string, header exported.C
 	// Else the update was proof of misbehaviour and we must emit appropriate misbehaviour events.
 	if status := newClientState.Status(ctx, clientStore, k.cdc); status != exported.Frozen {
 		// if update is not misbehaviour then update the consensus state
-		// we don't set consensus state for localhost client
-		if header != nil && clientID != exported.Localhost {
+		if header != nil {
 			k.SetClientConsensusState(ctx, clientID, header.GetHeight(), newConsensusState)
 		} else {
 			consensusHeight = types.GetSelfHeight(ctx)
```
colin-axner, line 101: this line can be removed

modules/core/02-client/keeper/client_test.go
```diff
@@ -25,7 +24,6 @@ func (suite *KeeperTestSuite) TestCreateClient() {
 		expPass     bool
 	}{
 		{"success", ibctmtypes.NewClientState(testChainID, ibctmtypes.DefaultTrustLevel, trustingPeriod, ubdPeriod, maxClockDrift, testClientHeight, commitmenttypes.GetSDKSpecs(), ibctesting.UpgradePath, false, false), true},
-		{"client type not supported", localhosttypes.NewClientState(testChainID, clienttypes.NewHeight(0, 1)), false},
```
colin-axner, line 28: refactor test case instead of removing

modules/core/02-client/keeper/keeper_test.go
```diff
@@ -177,11 +169,6 @@ func (suite *KeeperTestSuite) TestValidateSelfClient() {
 			ibctmtypes.NewClientState(suite.chainA.ChainID, ibctmtypes.DefaultTrustLevel, trustingPeriod, ubdPeriod, maxClockDrift, testClientHeight, commitmenttypes.GetSDKSpecs(), nil, false, false),
 			true,
 		},
-		{
-			"invalid client type",
-			localhosttypes.NewClientState(suite.chainA.ChainID, testClientHeight),
-			false,
-		},
```
colin-axner, line 184: refactor test case

modules/core/02-client/abci.go
seantking: Ah right, got it thanks. I'll update the godoc :+1:

modules/core/module.go
```diff
@@ -164,7 +164,6 @@ func (AppModule) ConsensusVersion() uint64 { return 2 }
 
 // BeginBlock returns the begin blocker for the ibc module.
 func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
-	ibcclient.BeginBlocker(ctx, am.keeper.ClientKeeper)
```
colin-axner, line 167: add back

modules/core/02-client/types/genesis_test.go
```diff
@@ -141,33 +134,13 @@ func (suite *TypesTestSuite) TestValidateGenesis() {
 			),
 			expPass: false,
 		},
-		{
-			name: "invalid client",
```
colin-axner, line 145: why was this test case removed?

colin-axner (changes requested):
Great work! Migration code is needed to remove any existing localhost from state otherwise removing all the `if clientID == localhost` checks could be problematic. It may even result in a panic if the clientID is ever referenced from store (we would no longer be able to unmarshal the client state)

modules/core/02-client/keeper/keeper_test.go
seantking: I thought the point here was that localhost is the invalid client type?

colin-axner: yea but so is the solo machine. `ValidateSelfClient` must be called with a tendermint client

modules/core/02-client/types/genesis_test.go
seantking: I assumed the light client was invalid because it was a localhost client in this case. 

I can update the test case name to be more descriptive (to not confuse others) and put it back in.

modules/core/02-client/keeper/keeper_test.go
seantking: OK yeah makes sense.

modules/core/02-client/types/genesis_test.go
colin-axner: Hmm, that might be true, but maybe we can use it for `!ps.Params.AllowedClient` check

seantking: Updated :handshake:

modules/core/02-client/abci_test.go
```diff
@@ -30,65 +22,8 @@ func (suite *ClientTestSuite) SetupTest() {
...
-			client.BeginBlocker(suite.chainA.GetContext(), suite.chainA.App.GetIBCKeeper().ClientKeeper)
-		}, "BeginBlocker shouldn't panic")
-
-		localHostClient = suite.chainA.GetClientState(exported.Localhost)
-		suite.Require().Equal(prevHeight.Increment(), localHostClient.GetLatestHeight())
-		prevHeight = localHostClient.GetLatestHeight().(types.Height)
-	}
-}
-
-func (suite *ClientTestSuite) TestBeginBlockerConsensusState() {
```
colin-axner, line 66: add back test

modules/core/02-client/keeper/client.go
```diff
@@ -98,11 +95,8 @@ func (k Keeper) UpdateClient(ctx sdk.Context, clientID string, header exported.C
 	// Else the update was proof of misbehaviour and we must emit appropriate misbehaviour events.
 	if status := newClientState.Status(ctx, clientStore, k.cdc); status != exported.Frozen {
 		// if update is not misbehaviour then update the consensus state
-		// we don't set consensus state for localhost client
-		if header != nil && clientID != exported.Localhost {
+		if header != nil {
```
colin-axner, line 98: remove if statement, header was only nil for localhost

colin-axner (commented):
As a note, these changes are not correct until #1201 is addressed, so we should keep that in mind if we merge into 02-client-refactor

modules/core/02-client/keeper/client.go
seantking: Done

modules/core/02-client/abci_test.go
seantking: Done :+1: good catch

colin-axner (approved):
ACK

damiannolan (approved):

Issue #880, opened 2022-02-07
Route MsgSubmitMisbehaviour to UpdateClient function


colin-axner:
## Summary

Following #875, MsgUpdateClient will be capable of handling any misbehaviour updates making `MsgSubmitMisbehaviour` redundant. We will we want to leave around for legacy reasons and just route `MsgSubmitMisbheaviour` to the client keeper `UpdateClient` function

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #1143, opened 2022-03-18
fix: prefix ResponseResultType enums for proto linting

damiannolan:
## Description

Running `make proto-all` fails with protobuf lint error

- Fixes proto lint error for enum prefixing

```
make proto-all
Formatting Protobuf files
{"path":"proto/ibc/core/channel/v1/tx.proto","start_line":50,"start_column":3,"end_line":50,"end_column":30,"type":"ENUM_VALUE_PREFIX","message":"Enum value name \"RESPONSE_RESULT_UNSPECIFIED\" should be prefixed with \"RESPONSE_RESULT_TYPE_\"."}
{"path":"proto/ibc/core/channel/v1/tx.proto","start_line":52,"start_column":3,"end_line":52,"end_column":23,"type":"ENUM_VALUE_PREFIX","message":"Enum value name \"RESPONSE_RESULT_NOOP\" should be prefixed with \"RESPONSE_RESULT_TYPE_\"."}
{"path":"proto/ibc/core/channel/v1/tx.proto","start_line":54,"start_column":3,"end_line":54,"end_column":26,"type":"ENUM_VALUE_PREFIX","message":"Enum value name \"RESPONSE_RESULT_SUCCESS\" should be prefixed with \"RESPONSE_RESULT_TYPE_\"."}
make: *** [proto-lint] Error 100
```

closes: #1142

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #1188, opened 2022-03-29
refactor: routing MsgSubmitMisbehaviour to UpdateClient keeper fn

seantking:
## Description

closes: #880

---

write a little note why.

### DIFF
--- a/modules/core/keeper/msg_server.go
+++ b/modules/core/keeper/msg_server.go
@@ -79,6 +79,8 @@ func (k Keeper) UpgradeClient(goCtx context.Context, msg *clienttypes.MsgUpgrade
 }
 
 // SubmitMisbehaviour defines a rpc handler method for MsgSubmitMisbehaviour.
+// Warning: Legacy code
+// This handler is redudant as `MsgUpdateClient` is now capable of handling both a Header and a Misbehaviour
 func (k Keeper) SubmitMisbehaviour(goCtx context.Context, msg *clienttypes.MsgSubmitMisbehaviour) (*clienttypes.MsgSubmitMisbehaviourResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
@@ -87,8 +89,8 @@ func (k Keeper) SubmitMisbehaviour(goCtx context.Context, msg *clienttypes.MsgSu
 		return nil, err
 	}
 
-	if err := k.ClientKeeper.CheckMisbehaviourAndUpdateState(ctx, msg.ClientId, misbehaviour); err != nil {
-		return nil, sdkerrors.Wrap(err, "failed to process misbehaviour for IBC client")
+	if err = k.ClientKeeper.UpdateClient(ctx, msg.ClientId, misbehaviour); err != nil {
+		return nil, err
 	}
 
 	return &clienttypes.MsgSubmitMisbehaviourResponse{}, nil


colin-axner (commented):
Needs changelog. Also would be good to update the `MsgSubmitMisbehaviour` as deprecated as well

modules/core/keeper/msg_server.go
```diff
@@ -79,6 +79,8 @@ func (k Keeper) UpgradeClient(goCtx context.Context, msg *clienttypes.MsgUpgrade
 }
 
 // SubmitMisbehaviour defines a rpc handler method for MsgSubmitMisbehaviour.
+// Warning: DEPRECATED
```
seantking, line 82: Is that what you meant @colin-axner ?

damiannolan: I think it would be good to add a deprecation notice to the proto files for `MsgSubmitBehaviour`.

Unfortunately protobuf only seems to have the field option [`[deprecated = true]`](https://developers.google.com/protocol-buffers/docs/proto3#options), there isn't any concrete deprecation flags for an entire message type.

modules/core/keeper/msg_server.go
```diff
@@ -87,8 +89,8 @@ func (k Keeper) SubmitMisbehaviour(goCtx context.Context, msg *clienttypes.MsgSu
 		return nil, err
 	}
 
-	if err := k.ClientKeeper.CheckMisbehaviourAndUpdateState(ctx, msg.ClientId, misbehaviour); err != nil {
-		return nil, sdkerrors.Wrap(err, "failed to process misbehaviour for IBC client")
+	if err = k.ClientKeeper.UpdateClient(ctx, msg.ClientId, misbehaviour); err != nil {
+		return nil, err
```
damiannolan, line 93: Should we continue to wrap the `err` here?

modules/core/keeper/msg_server.go
seantking: Yeah great point, thanks.

modules/core/keeper/msg_server.go
colin-axner: No preference. Preferably `UpdateClient` provides a pretty descriptive error message

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -47,11 +47,11 @@ enum ResponseResultType {
   option (gogoproto.goproto_enum_prefix) = false;
 
   // Default zero value enumeration
-  RESPONSE_RESULT_UNSPECIFIED = 0 [(gogoproto.enumvalue_customname) = "UNSPECIFIED"];
+  RESPONSE_RESULT_TYPE_UNSPECIFIED = 0 [(gogoproto.enumvalue_customname) = "UNSPECIFIED"];
```
seantking, line 50: I copied over the changes that are on main. There is some linting issues for channel protos on the `02-client-refactor` branch that are fixed on main. In order to add the `deprecated` flag for `MsgSubmitMisbehavior` I fixed this (the linter was failing otherwise).

modules/core/keeper/msg_server.go
seantking: Done: see https://github.com/cosmos/ibc-go/pull/1188#discussion_r840666627

proto/ibc/core/channel/v1/tx.proto
damiannolan: Nice 👍 I fixed this on main in #1143 before but I think those changes have been pulled into `02-client-refactor` yet. 
I think we should probably merge the latest of main in, also includes the fixes for gh actions (link checker)

proto/ibc/core/client/v1/tx.proto
```diff
@@ -82,16 +82,17 @@ message MsgUpgradeClientResponse {}
 
 // MsgSubmitMisbehaviour defines an sdk.Msg type that submits Evidence for
 // light client misbehaviour.
+// Warning: DEPRECATED
```
damiannolan, line 85: Should we add something like: "Use MsgUpdateClient instead, this will be removed in a future release"

damiannolan (approved):

charleenfei (approved):

Issue #1190, opened 2022-03-29
07-tendermint should set own client/consensus states in UpdateState


colin-axner:
## Summary

The 07-tendermint `UpdateState` should set own client/consensus states. A check should be added in tests for this. Similar to #1144 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #1199, opened 2022-03-30
chore: 07-tendermint set client/consensus states in UpdateState

damiannolan:
## Description

- Set `ClientState` and `ConsensusState` in store within `07-tendermint` `UpdateState`
- Updating tests to assert against stored values in favour of return values

closes: #1190 

---

write a little note why.

### DIFF
--- a/modules/light-clients/07-tendermint/types/store.go
+++ b/modules/light-clients/07-tendermint/types/store.go
@@ -50,8 +50,8 @@ func setClientState(clientStore sdk.KVStore, cdc codec.BinaryCodec, clientState 
 	clientStore.Set(key, val)
 }
 
-// SetConsensusState stores the consensus state at the given height.
-func SetConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensusState *ConsensusState, height exported.Height) {
+// setConsensusState stores the consensus state at the given height.
+func setConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensusState *ConsensusState, height exported.Height) {
 	key := host.ConsensusStateKey(height)
 	val := clienttypes.MustMarshalConsensusState(cdc, consensusState)
 	clientStore.Set(key, val)
--- a/modules/light-clients/07-tendermint/types/update_test.go
+++ b/modules/light-clients/07-tendermint/types/update_test.go
@@ -415,10 +415,7 @@ func (suite *TendermintTestSuite) TestUpdateState() {
 				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().LT(clientMessage.GetHeight()))
 			},
 			func() {
-				bz := clientStore.Get(host.ClientStateKey())
-				updatedClientState := clienttypes.MustUnmarshalClientState(suite.chainA.App.AppCodec(), bz)
-
-				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().EQ(updatedClientState.GetLatestHeight())) // new update, updated client state should have changed
+				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().EQ(clientMessage.GetHeight())) // new update, updated client state should have changed
 			}, true,
 		},
 		{
@@ -486,10 +483,7 @@ func (suite *TendermintTestSuite) TestUpdateState() {
 				suite.Require().NoError(err)
 			},
 			func() {
-				bz := clientStore.Get(host.ClientStateKey())
-				updatedClientState := clienttypes.MustUnmarshalClientState(suite.chainA.App.AppCodec(), bz)
-
-				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().EQ(updatedClientState.GetLatestHeight())) // new update, updated client state should have changed
+				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().EQ(clientMessage.GetHeight())) // new update, updated client state should have changed
 
 				// ensure consensus state was pruned
 				_, found := path.EndpointA.Chain.GetConsensusState(path.EndpointA.ClientID, pruneHeight)
--- a/modules/light-clients/07-tendermint/types/update_test.go
+++ b/modules/light-clients/07-tendermint/types/update_test.go
@@ -398,10 +398,12 @@ func (suite *TendermintTestSuite) TestVerifyHeader() {
 
 func (suite *TendermintTestSuite) TestUpdateState() {
 	var (
-		path          *ibctesting.Path
-		clientMessage exported.ClientMessage
-		clientStore   sdk.KVStore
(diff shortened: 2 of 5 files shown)


modules/light-clients/07-tendermint/types/store.go
```diff
@@ -50,8 +50,8 @@ func setClientState(clientStore sdk.KVStore, cdc codec.BinaryCodec, clientState
 	clientStore.Set(key, val)
 }
 
-// SetConsensusState stores the consensus state at the given height.
-func SetConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensusState *ConsensusState, height exported.Height) {
+// setConsensusState stores the consensus state at the given height.
+func setConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensusState *ConsensusState, height exported.Height) {
```
damiannolan, line 54: Updated to make private / unexported

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -415,7 +415,10 @@ func (suite *TendermintTestSuite) TestUpdateState() {
 				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().LT(clientMessage.GetHeight()))
 			},
 			func() {
-				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().LT(updatedClientState.GetLatestHeight())) // new update, updated client state should have changed
+				bz := clientStore.Get(host.ClientStateKey())
+				updatedClientState := clienttypes.MustUnmarshalClientState(suite.chainA.App.AppCodec(), bz)
+
+				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().EQ(updatedClientState.GetLatestHeight())) // new update, updated client state should have changed
```
damiannolan, line 421: `path.EndpointA.GetClientState()` will now return latest client state as `UpdateState` sets it in store, so here we test for equality rather than less than

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -474,7 +486,10 @@ func (suite *TendermintTestSuite) TestUpdateState() {
 				suite.Require().NoError(err)
 			},
 			func() {
-				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().LT(updatedClientState.GetLatestHeight())) // new update, updated client state should have changed
+				bz := clientStore.Get(host.ClientStateKey())
+				updatedClientState := clienttypes.MustUnmarshalClientState(suite.chainA.App.AppCodec(), bz)
+
+				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().EQ(updatedClientState.GetLatestHeight())) // new update, updated client state should have changed
```
damiannolan, line 492: ditto

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -429,6 +432,9 @@ func (suite *TendermintTestSuite) TestUpdateState() {
 				suite.Require().True(path.EndpointA.GetClientState().GetLatestHeight().GT(clientMessage.GetHeight()))
 			},
 			func() {
+				bz := clientStore.Get(host.ClientStateKey())
+				updatedClientState := clienttypes.MustUnmarshalClientState(suite.chainA.App.AppCodec(), bz)
+
```
colin-axner, line 437: should this go in malleate? I'm not sure these checks perform any actual checks now since `GetClientState()` performs the same code

modules/light-clients/07-tendermint/types/update_test.go
colin-axner: maybe we should use `clientMessage.GetHeight()` instead of the `updatedClientState`?

modules/light-clients/07-tendermint/types/update_test.go
```diff
@@ -444,6 +450,12 @@ func (suite *TendermintTestSuite) TestUpdateState() {
 				suite.Require().Equal(path.EndpointA.GetClientState().GetLatestHeight(), clientMessage.GetHeight())
 			},
 			func() {
+				bz := clientStore.Get(host.ClientStateKey())
+				updatedClientState := clienttypes.MustUnmarshalClientState(suite.chainA.App.AppCodec(), bz)
+
+				bz = clientStore.Get(host.ConsensusStateKey(clientMessage.GetHeight()))
+				updatedConsensusState := clienttypes.MustUnmarshalConsensusState(suite.chainA.App.AppCodec(), bz)
```
colin-axner, line 457: ditto on moving into malleate

modules/light-clients/07-tendermint/types/update_test.go
colin-axner: ditto on using clientMessage.GetHeight

colin-axner (commented):
Excellent! Left some suggestions on improving the testing changes

modules/light-clients/07-tendermint/types/update_test.go
damiannolan: Right, yeah I get you! `GetClientState` is essentially just getting from the store and unmarshalling as well. So this is basically asserting true is true. 

Could you expand on what you're suggesting for moving to malleate? I'm not sure I follow, malleate will be called before `UpdateState` right, so the client state retrieved is not yet updated. Are you suggesting to retrieve the not yet updated state in malleate and then in expResult() assert that they are not equal?

modules/light-clients/07-tendermint/types/update_test.go
damiannolan: Nice yea, I agree, cleaner :)

Updated!

modules/light-clients/07-tendermint/types/update_test.go
colin-axner: >  Are you suggesting to retrieve the not yet updated state in malleate and then in expResult() assert that they are not equal?

These test cases are asserting the client state didn't get updated (since we updated to a previous height). So yes, I am suggesting we construct the expected client state in malleate and then ensure the client state doesn't change after calling `UpdateState`

colin-axner (approved):
ACK pending these [testing improvements](https://github.com/cosmos/ibc-go/pull/1199#discussion_r838471327). Wonderful that things are moving forward!

modules/light-clients/07-tendermint/types/update_test.go
damiannolan: Ah yes, sorry overlooked that when I wrote the comment. I've updated as you suggested to retrieve the "previous" states before update in malleate and check for equality in expResult! :))

modules/light-clients/07-tendermint/types/update_test.go
damiannolan: Done 👍

seantking (approved):

Issue #857, opened 2022-02-07
gRPC for Relayers to query which channels are incentivized


AdityaSripal:
Since channelID/portID does not include fee information, we need a rpc that tells relayers whether or not a given channel is incentivized

damiannolan:
I think we should possibly include two queries here. One to return all fee enabled channels and one to return whether or not a channel is fee enabled given its channel/port ID.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1219, opened 2022-04-06
feat: ics29 fee enabled channels queries

damiannolan:
## Description

- Adding queries for fee enabled channels
- Adding CLIs

closes: #857 

---

write a little note why.


damiannolan:
Tested using the environment outlined in the [QA doc](https://hackmd.io/0qHkPDIVSZWhhK8QdKLfCQ?view)

1. Deploy two chains using `simd`
2. Creates a new fee enabled channel using Hermes
3. Query the new endpoints via CLI

```bash
$ simd q ibc-fee channels --home ./data/test-1 --node tcp://localhost:16657
fee_enabled_channels:
- channel_id: channel-0
  port_id: transfer

$ simd q ibc-fee channel transfer channel-0 --home ./data/test-1 --node tcp://localhost:16657
fee_enabled: true
```

4. Create a second channel using Hermes

```bash
$  simd q ibc-fee channels --home ./data/test-1 --node tcp://localhost:16657 --height 100
fee_enabled_channels:
- channel_id: channel-0
  port_id: transfer

$ simd q ibc-fee channels --home ./data/test-1 --node tcp://localhost:16657
fee_enabled_channels:
- channel_id: channel-0
  port_id: transfer
- channel_id: channel-1
  port_id: transfer
```

modules/apps/29-fee/client/cli/query.go
```diff
@@ -189,3 +189,78 @@ func GetCmdCounterpartyAddress() *cobra.Command {
...
+		Args:    cobra.NoArgs,
+		Example: fmt.Sprintf("%s query ibc-fee channels", version.AppName),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			pageReq, err := client.ReadPageRequest(cmd.Flags())
+			if err != nil {
```
seantking, line 208: I think if we return an err here this means the pagination flags are not optional. Is that what we want?

seantking, line 208: I think if we return an err here this means the pagination flags are not optional. Is that what we want?

modules/apps/29-fee/keeper/grpc_query_test.go
```diff
@@ -493,3 +493,161 @@ func (suite *KeeperTestSuite) TestQueryCounterpartyAddress() {
...
+		},
+		{
+			"success, empty pagination",
+			func() {
+				req = &types.QueryFeeEnabledChannelsRequest{}
+			},
+			true,
+		},
+		{
+			"success, with multiple fee enabled channels",
```
crodriguezvega, line 521: Is the test description correct? I see below that only one channel is appended to `expFeeEnabledChannels`...

damiannolan: Below in the body of the test func, the first (default) expected channel is added to `expFeeEnabledChannels` on line 579, then `malleate()` is called after, and here we're just appending an additional expected channel.
Because `expFeeEnabledChannels` is defined at the top level scope of the test function we can mutate it in `malleate`

crodriguezvega: Ah, I see now, thanks. Should we make it explicit in the description that this test does not use pagination? Something like `success, no pagination with multiple fee enabled channels`?

damiannolan: Sure, can do, I'm going to close this PR and re-open with main as the target. I don't want to spoil the commit history if I change the base branch on this PR.

damiannolan: Just taking a second look at this now since I created a new branch for main. This test does use pagination. The default request that's built inside the body of the test function is used:

See line 581

```
req = &types.QueryFeeEnabledChannelsRequest{
	Pagination: &query.PageRequest{
		Limit:      5,
		CountTotal: false,
	},
	QueryHeight: 0,
}
```

Issue #1234, opened 2022-04-08
ics29 fee validation: return errors immediately


damiannolan:
## Summary

Return errors immediately as opposed to appending them as suggested in:
https://github.com/cosmos/ibc-go/pull/276#discussion_r844039714

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1235, opened 2022-04-08
chore: ics29 error returns on fee validation

damiannolan:
## Description

- Return errors immediately as opposed to appending them

closes: #1234 

---

write a little note why.

### DIFF
--- a/modules/apps/29-fee/types/fee.go
+++ b/modules/apps/29-fee/types/fee.go
@@ -1,8 +1,6 @@
 package types
 
 import (
-	"strings"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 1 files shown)


crodriguezvega:
Yes, I also agree that we should leave it as is. I think it's better UX when you get error messages for all the things that you might have wrong at once, instead of executing each time the command and getting only the next error message.

damiannolan:
Sounds good, in which case I'll close this PR and the issue!

colin-axner:
My only concern with the agreed upon decision is that this is not how the SDK functions. The SDK does not batch up all errors in a message execution. I believe `fee.Validate()` should get called on msg.ValidateBasic which occurs before a message gets broadcasted. It seems more appropriate to me to return one error at a time

Issue #862, opened 2022-02-07
If distribute fee fails, refund to original sender


AdityaSripal:
https://github.com/cosmos/ibc-go/blob/13f77de949ce84704db1c7307d31ba8c6e16842e/modules/apps/29-fee/keeper/escrow.go#L92

Should include a refund address argument which is original fee sender

If we are trying to refund to original sender and that fails, then we need to just no-op and continue.

damiannolan:
What are the concerns of the fee being left behind in the ics29 module account if this fails? 
For the same reasons a bank send fails to a forward/reverse relayer it may also fail to the refund account. Do you think this issue is still necessarily?

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1245, opened 2022-04-11
chore(ics29): attempt to refund fees on distribution failure

damiannolan:
## Description

- Attempt to refund fees on distribution failure

closes: #862 

---

write a little note why.


modules/apps/29-fee/keeper/escrow.go
```diff
@@ -85,31 +86,36 @@ func (k Keeper) DistributePacketFeesOnTimeout(ctx sdk.Context, timeoutRelayer sd
...
-	if err == nil {
-		// write the cache
-		writeFn()
-
-		// NOTE: The context returned by CacheContext() refers to a new EventManager, so it needs to explicitly set events to the original context.
-		ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
+	if err != nil && !bytes.Equal(receiver, refundAccAddress) {
+		err := k.bankKeeper.SendCoinsFromModuleToAccount(cacheCtx, types.ModuleName, refundAccAddress, fee)
+		if err != nil {
+			return
```
damiannolan, line 110: Is there anything we can do here in the case of both failing? Emit some events?

seantking: What would be the goal for emitting the event? Events are mostly just used by relayers I guess. 

I think it would be worth logging something to `INFO` or `DEBUG` logs. Although I haven't seen SDK modules utilizing logging much.

damiannolan: > What would be the goal for emitting the event?

Just some kind of signalling I guess, was just thinking aloud really. I think logging at `WARN` or `ERROR` here may be useful, but like you say, most sdk modules don't seem to do much logging. 
Feels a little wrong to just silently no-op, that's my only concern.

seantking (approved):

modules/apps/29-fee/keeper/escrow.go
seantking: Yeah, I know what you mean. I guess the problem is that even if we returned an error here we would simply dismiss it at the level above anyway, otherwise, we would err on `OnAcknowledgementPacket` which would throw away the ack for the ics20 transfer or w/e as well. 

I think we're limited in what we can do here so my vote would be some warn/err logs.

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -85,31 +86,36 @@ func (k Keeper) DistributePacketFeesOnTimeout(ctx sdk.Context, timeoutRelayer sd
...
 	cacheCtx, writeFn := ctx.CacheContext()
 
 	err := k.bankKeeper.SendCoinsFromModuleToAccount(cacheCtx, types.ModuleName, receiver, fee)
-	if err == nil {
-		// write the cache
-		writeFn()
-
-		// NOTE: The context returned by CacheContext() refers to a new EventManager, so it needs to explicitly set events to the original context.
-		ctx.EventManager().EmitEvents(cacheCtx.EventManager().Events())
+	if err != nil && !bytes.Equal(receiver, refundAccAddress) {
```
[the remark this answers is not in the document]
colin-axner: I think you need to separate these conditions as well. What happens if there is an err and refundAcc == receiver?

Maybe something like:
```go
if err != nil {
    // fee distribution failed, try to refund to original incentivizer
    if bytes.Equal(receiver, refundAccAddress) {
        return err
    }
    
    // attempt to distribute fee to refund address
}
```

modules/apps/29-fee/keeper/escrow.go
colin-axner: Seems like a bug if we cannot refund to the original sender? Maybe we should panic

modules/apps/29-fee/keeper/escrow.go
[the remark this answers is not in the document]
colin-axner: Could we add a test case for this case?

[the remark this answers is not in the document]
damiannolan: If there's an error and `refundAcc == receiver` then sending to the refundAcc has already failed, right? 
That's why I combined the conditions using the logical AND.
The issues writes:

> If we are trying to refund to original sender and that fails, then we need to just no-op and continue.

In this case, `writeFn()` is still called but there would be no state changes to commit (no-op).
I can adapt this tomorrow tho if we feel its better separated.

I'll look at adding a testcase 👍

modules/apps/29-fee/keeper/escrow.go
damiannolan: Hmm, maybe we should... originally @AdityaSripal had suggested to just no-op in the issue.

seantking: I guess the no-op is for the case that the other fees were distributed fine but this one errors. If we err or panic the other fees won't ever get distributed.

modules/apps/29-fee/keeper/escrow.go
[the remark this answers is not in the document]
damiannolan: Tests require some refactoring. In progress!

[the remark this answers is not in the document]
colin-axner: To be on the safe side I think it makes sense to ensure `writeFn()` doesn't get called. Otherwise we need to verify that no partial state transitions occurred in x/bank. It might assume state changes are discarded once the error is returned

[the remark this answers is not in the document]
damiannolan: Sure, I'll adapt the logic to represent that. I had similar concerns wrt `writeFn()` but wasn't 100%. Thanks :)

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -122,66 +122,94 @@ func (suite *KeeperTestSuite) TestEscrowPacketFee() {
...
+			},
+			func() {
+				// check if the refund acc has been refunded the timeoutFee & ackFee
+				expectedRefundAccBal := refundAccBal.Add(defaultTimeoutFee[0]).Add(defaultAckFee[0]).Add(defaultTimeoutFee[0]).Add(defaultAckFee[0])
+				balance := suite.chainA.GetSimApp().BankKeeper.GetBalance(suite.chainA.GetContext(), refundAcc, sdk.DefaultBondDenom)
+				suite.Require().Equal(expectedRefundAccBal, balance)
 			},
 		},
+		// {
+		// 	"invalid refund address: no-op, timeout fee remains in escrow",
```
damiannolan, line 202: I'll see if I can fix this test. Currently failing because we're relying on `EscrowPacketFee` and `malleate()` is called after this. If we call `malleate()` before escrowing the fees, then the `RefundAddress` (incentivizer) is the module account, which doesn't have balance.

Bit of a mouth full, but hope it makes sense.

damiannolan: The simplest solution is to mutate `packetFee.RefundAddress` in `malleate()` after the fees are initially escrowed and before `DistributePacketFees` is called 👍

colin-axner (approved):
Nice work! Thanks for following up on my suggested change

PR #1251, opened 2022-04-13
ics29: gracefully handle escrow out of balance edge case during fee distribution

colin-axner:
## Description

closes: #821

---

write a little note why.

### DIFF
--- a/modules/apps/29-fee/keeper/escrow.go
+++ b/modules/apps/29-fee/keeper/escrow.go
@@ -106,6 +106,7 @@ func (k Keeper) distributePacketFeeOnAcknowledgement(ctx sdk.Context, refundAddr
 
 }
 
+// DistributePacketsFeesOnTimeout pays all the timeout fees for a given packetID while refunding the acknowledgement & receive fees to the refund account.
 func (k Keeper) DistributePacketFeesOnTimeout(ctx sdk.Context, timeoutRelayer sdk.AccAddress, packetFees []types.PacketFee) {
 	// cache context before trying to distribute fees
 	// if the escrow account has insufficient balance then we want to avoid partially distributing fees
--- a/modules/apps/29-fee/keeper/escrow.go
+++ b/modules/apps/29-fee/keeper/escrow.go
@@ -50,62 +50,14 @@ func (k Keeper) EscrowPacketFee(ctx sdk.Context, packetID channeltypes.PacketId,
 }
 
 // DistributePacketFees pays the acknowledgement fee & receive fee for a given packetID while refunding the timeout fee to the refund account associated with the Fee.
-func (k Keeper) DistributePacketFees(ctx sdk.Context, forwardRelayer string, reverseRelayer sdk.AccAddress, feesInEscrow []types.PacketFee) {
+func (k Keeper) DistributePacketFees(ctx sdk.Context, forwardRelayer string, reverseRelayer sdk.AccAddress, packetFees []types.PacketFee, isTimeout bool) {
+	// cache context before trying to distribute fees
+	// if the escrow account has insufficient balance then we want to avoid partially distributing fees
+	cacheCtx, writeFn := ctx.CacheContext()
+
 	forwardAddr, _ := sdk.AccAddressFromBech32(forwardRelayer)
 
-	// cache context before trying to distribute fees
-	// if the escrow account has insufficient balance then we want to avoid partially distributing fees
-	cacheCtx, writeFn := ctx.CacheContext()
-
-	for _, packetFee := range feesInEscrow {
-		if !k.EscrowAccountHasBalance(cacheCtx, packetFee.Fee.Total()) {
-			// if the escrow account does not have sufficient funds then there must exist a severe bug
-			// the fee module should be locked until manual intervention fixes the issue
-			// a locked fee module will simply skip fee logic, all channels will temporarily function as
-			// fee disabled channels
-			// NOTE: we use the uncached context to lock the fee module so that the state changes from
-			// locking the fee module are persisted
-			k.lockFeeModule(ctx)
-
-			return
-		}
-
-		refundAddr, err := sdk.AccAddressFromBech32(packetFee.RefundAddress)
-		if err != nil {
-			panic(fmt.Sprintf("could not parse refundAcc %s to sdk.AccAddress", packetFee.RefundAddress))
-		}
-
-		// distribute fee to valid forward relayer address otherwise refund the fee
-		if !forwardAddr.Empty() && !k.bankKeeper.BlockedAddr(forwardAddr) {
-			// distribute fee for forward relaying
-			k.distributeFee(cacheCtx, forwardAddr, refundAddr, packetFee.Fee.RecvFee)
-		} else {
-			// refund onRecv fee as forward relayer is not valid address
-			k.distributeFee(cacheCtx, refundAddr, refundAddr, packetFee.Fee.RecvFee)
-		}
-
-		// distribute fee for reverse relaying
-		k.distributeFee(cacheCtx, reverseRelayer, refundAddr, packetFee.Fee.AckFee)
-
-		// refund timeout fee for unused timeout
-		k.distributeFee(cacheCtx, refundAddr, refundAddr, packetFee.Fee.TimeoutFee)
-	}
-
-	// write the cache
-	writeFn()
-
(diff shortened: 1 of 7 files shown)


modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -244,7 +261,7 @@ func (suite *KeeperTestSuite) TestDistributeFee() {
 			reverseRelayerBal = suite.chainA.GetSimApp().BankKeeper.GetBalance(suite.chainA.GetContext(), reverseRelayer, sdk.DefaultBondDenom)
 			refundAccBal = suite.chainA.GetSimApp().BankKeeper.GetBalance(suite.chainA.GetContext(), refundAcc, sdk.DefaultBondDenom)
 
-			suite.chainA.GetSimApp().IBCFeeKeeper.DistributePacketFees(suite.chainA.GetContext(), forwardRelayer, reverseRelayer, []types.PacketFee{packetFee, packetFee})
+			suite.chainA.GetSimApp().IBCFeeKeeper.DistributePacketFees(suite.chainA.GetContext(), forwardRelayer, reverseRelayer, packetFees, false)
```
colin-axner, line 264: The `TestDistributeTimeoutFee` isn't a table test yet, I think in a followup we should just expand this test `TestDistributeFee` to test both the `onAcknowledgement` and `onTimeout` case. The `expResult()` check would allow us to do this pretty easily

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -50,50 +50,76 @@ func (k Keeper) EscrowPacketFee(ctx sdk.Context, packetID channeltypes.PacketId,
 }
 
 // DistributePacketFees pays the acknowledgement fee & receive fee for a given packetID while refunding the timeout fee to the refund account associated with the Fee.
-func (k Keeper) DistributePacketFees(ctx sdk.Context, forwardRelayer string, reverseRelayer sdk.AccAddress, feesInEscrow []types.PacketFee) {
+func (k Keeper) DistributePacketFees(ctx sdk.Context, forwardRelayer string, reverseRelayer sdk.AccAddress, packetFees []types.PacketFee, isTimeout bool) {
```
seantking, line 53: The refactor here is great, I think it's a massive improvement splitting up the timeout and ack logic as you have. I can't help but feel like I'd prefer to have two separate functions though rather than passing a boolean. I feel like the fact that we're adding a boolean here could be a sign this [fn has too much responsibility](https://en.wikipedia.org/wiki/Single-responsibility_principle). 

What about the idea of having `DistributeAcknowledgementPacketFees` & `DistributeTimeoutPacketFees`?

I'm open to either way, but just thought I'd mention. Wdyt? :)

damiannolan: Yeah, I agree. I think it might be worth exploring the option of two separate functions in favour of the boolean arg, considering that the forward relayer is always an empty string for timeout scenarios also. 

I do appreciate that it would be a decent bit of duplicate code nonetheless, but I would lean towards having explicit APIs for each scenario (happy path/timeout)

colin-axner: My main concern is the sensitivity of performing the cacheCtx. It can be hard to verify that we pass in `cacheCtx` and `ctx` in the correct functions. Doing this redundantly in two separate functions seems like a recipe for one of the functions to get out of sync with the other and potentially lead to a bug. 

The actual distribution of fees on acknowledgement and fees on timeout are still separated. The logic this function is performing is wrapping packet fee distribution with an escrow out of balance check. The previously functionality not only distributed the packet fees for either ack or timeout but also wrapped it with an escrow out of balance check. 

For reference you can see how this file [differed before splitting](https://github.com/cosmos/ibc-go/blob/4d5a7474ca971e6aeb6b1dab55727b66353ed361/modules/apps/29-fee/keeper/escrow.go#L53-#L141). I actually find it harder to verify the [`distributeAcknowledgementPacketFees`](https://github.com/cosmos/ibc-go/blob/4d5a7474ca971e6aeb6b1dab55727b66353ed361/modules/apps/29-fee/keeper/escrow.go#L73-#L91) and [`distirbuteTimeoutPacketFees`](https://github.com/cosmos/ibc-go/blob/4d5a7474ca971e6aeb6b1dab55727b66353ed361/modules/apps/29-fee/keeper/escrow.go#L121-#L134) functionality in comparison to the proposed way [`distributePacketFeeOnAcknowledgement`](https://github.com/cosmos/ibc-go/blob/317aaa191ee92f64549694fad8bb0660272ccf14/modules/apps/29-fee/keeper/escrow.go#L95-#L109) and [`distirbutePacketFeeOnTimeout`](https://github.com/cosmos/ibc-go/blob/317aaa191ee92f64549694fad8bb0660272ccf14/modules/apps/29-fee/keeper/escrow.go#L114-#L122)

modules/apps/29-fee/keeper/escrow_test.go
```diff
@@ -244,59 +261,123 @@ func (suite *KeeperTestSuite) TestDistributeFee() {
...
+			suite.chainA.GetSimApp().IBCFeeKeeper.DistributePacketFeesOnAcknowledgement(suite.chainA.GetContext(), forwardRelayer, reverseRelayer, packetFees)
 
 			tc.expResult()
 		})
 	}
 }
 
-func (suite *KeeperTestSuite) TestDistributeTimeoutFee() {
-	suite.coordinator.Setup(suite.path) // setup channel
+func (suite *KeeperTestSuite) TestDistributePacketFeesOnTimeout() {
```
colin-axner, line 271: duplicated `TestDistributePacketFeesOnAcknowledgement` and adjusted the tests

modules/apps/29-fee/keeper/escrow.go
```diff
@@ -83,17 +106,38 @@ func (k Keeper) distributePacketFeeOnAcknowledgement(ctx sdk.Context, refundAddr
 
 }
 
-// DistributePacketsFeesOnTimeout pays all the timeout fees for a given packetID while refunding the acknowledgement & receive fees to the refund account.
-func (k Keeper) DistributePacketFeesOnTimeout(ctx sdk.Context, timeoutRelayer sdk.AccAddress, feesInEscrow []types.PacketFee) {
-	for _, feeInEscrow := range feesInEscrow {
+func (k Keeper) DistributePacketFeesOnTimeout(ctx sdk.Context, timeoutRelayer sdk.AccAddress, packetFees []types.PacketFee) {
```
damiannolan, line 109: Think we need a godoc for this!

modules/apps/29-fee/keeper/keeper_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	defaultReceiveFee = sdk.Coins{sdk.Coin{Denom: sdk.DefaultBondDenom, Amount: sdk.NewInt(100)}}
+	defaultRecvFee    = sdk.Coins{sdk.Coin{Denom: sdk.DefaultBondDenom, Amount: sdk.NewInt(100)}}
```
damiannolan, line 18: Nice, prefer this too! :)

damiannolan (approved):
Awesome work! Thanks for going to the trouble of exploring both approaches with this, much appreciated! While there's a bit of duplication with the code I think from an API perspective having no `forwardRelayer` arg in the timeout scenario looks much cleaner imo. 

❤️

modules/apps/29-fee/keeper/escrow.go
colin-axner: whoops!

seantking (approved):
Great work!

modules/apps/29-fee/keeper/escrow.go
crodriguezvega: I agree with @seantking and @damiannolan. I think having two separate functions will make the code more explicit and easier to read.

If that's not possible because of the concerns that @colin-axner has regarding the context, then a couple of other alternatives could be:
1. To have an enum value instead of a boolean as parameter. Then at least reading the function at call site would look like `..., OnTimeout)` or `..., OnAcknowledgement)`.
2. To pass the function that actually does the distribution instead of a boolean as parameter. So at call site you would have something like `..., distributePacketFeeOnAcknowledgement)` or `..., distributePacketFeeOnTimeout)`. But then you need to have both functions with the same signature, which is possible to work around, but might not lead to the most beautiful result...

Issue #722, opened 2022-01-13
Channel handshake ChanOpenInit improvements


colin-axner:
## Summary

Apply [spec changes](https://github.com/cosmos/ibc/pull/629). These changes have already been done for [ChanOpenTry](https://github.com/cosmos/ibc-go/issues/640). These changes should be broken into smaller pieces if multiple parts of core IBC need to be modified. Please see the workflow I did for #640 for inspiration. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
I think this is probably fine to do in one function since #640 help set the groundwork for these changes

Suggested ways of maybe breaking down the pr:
- [ ] modify OnChanOpenInit API to return version string, pass version into WriteChannelState
- [ ] ics20 - add a check if the version passed in is empty, if so return the ics20 version
- [ ] ics27 - no code should be changed, the version passed must be non empty since it indicates specific connection choices (as well as tx and encoding type)
- [ ] ics29 - if no version is specified it should not enable fees and pass onto underlying app
- [ ] add a field to MsgChannelOpenInitResponse to return the selected version (probably should be done for MsgChannelOpenTry as well)

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1283, opened 2022-04-21
refactor: channel handshake version improvements

seantking:
## Description

closes: #722 

---

write a little note why.


modules/core/keeper/msg_server.go
```diff
@@ -185,15 +185,17 @@ func (k Keeper) ChannelOpenInit(goCtx context.Context, msg *channeltypes.MsgChan
...
 	// Perform application logic callback
-	if err = cbs.OnChanOpenInit(ctx, msg.Channel.Ordering, msg.Channel.ConnectionHops, msg.PortId, channelID, cap, msg.Channel.Counterparty, msg.Channel.Version); err != nil {
+	version, err := cbs.OnChanOpenInit(ctx, msg.Channel.Ordering, msg.Channel.ConnectionHops, msg.PortId, channelID, cap, msg.Channel.Counterparty, msg.Channel.Version)
+	if err != nil {
 		return nil, sdkerrors.Wrap(err, "channel open init callback failed")
 	}
 
 	// Write channel into state
-	k.ChannelKeeper.WriteOpenInitChannel(ctx, msg.PortId, channelID, msg.Channel.Ordering, msg.Channel.ConnectionHops, msg.Channel.Counterparty, msg.Channel.Version)
+	k.ChannelKeeper.WriteOpenInitChannel(ctx, msg.PortId, channelID, msg.Channel.Ordering, msg.Channel.ConnectionHops, msg.Channel.Counterparty, version)
```
seantking, line 194: Changing this causes the fee tests to break. Looking into it :eyes:

modules/apps/29-fee/ibc_module.go
```diff
@@ -46,8 +46,14 @@ func (im IBCModule) OnChanOpenInit(
 			chanCap, counterparty, version)
 	}
 
+	if versionMetadata.FeeVersion == "" {
```
colin-axner, line 49: is this necessary?

seantking: Nope, indded it is not. I had this removed locally and forgot to push :)

modules/core/keeper/msg_server.go
seantking: Needed: https://github.com/cosmos/ibc-go/pull/1283/commits/0178c093a09265fa814491423cf3c9311e6ace1a

modules/apps/transfer/ibc_module.go
```diff
@@ -70,21 +71,25 @@ func (im IBCModule) OnChanOpenInit(
...
 	counterparty channeltypes.Counterparty,
 	version string,
-) error {
+) (string, error) {
 	if err := ValidateTransferChannelParams(ctx, im.keeper, order, portID, channelID); err != nil {
-		return err
+		return "", err
+	}
+
+	if strings.TrimSpace(version) == "" {
```
seantking, line 79: I can add a test case for this

modules/apps/27-interchain-accounts/controller/ibc_module.go
```diff
@@ -42,13 +42,13 @@ func (im IBCModule) OnChanOpenInit(
...
-		return types.ErrControllerSubModuleDisabled
+		return "", types.ErrControllerSubModuleDisabled
 	}
 
 	if err := im.keeper.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version); err != nil {
-		return err
+		return "", err
 	}
 
 	// call underlying app's OnChanOpenInit callback with the appVersion
```
colin-axner, line 54: we should be returning the version passed in, not the underlying application version

modules/apps/27-interchain-accounts/controller/ibc_module_test.go
```diff
@@ -197,7 +197,7 @@ func (suite *InterchainAccountsTestSuite) TestOnChanOpenInit() {
 			cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
 			suite.Require().True(ok)
 
-			err = cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
+			_, err = cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
```
colin-axner, line 200: we should be checking the returned version

modules/apps/29-fee/ibc_module.go
```diff
@@ -113,7 +123,7 @@ func (im IBCModule) OnChanOpenAck(
 	if im.keeper.IsFeeEnabled(ctx, portID, channelID) {
 		var versionMetadata types.Metadata
 		if err := types.ModuleCdc.UnmarshalJSON([]byte(counterpartyVersion), &versionMetadata); err != nil {
-			return sdkerrors.Wrap(types.ErrInvalidVersion, "failed to unmarshal ICS29 counterparty version metadata")
+			return sdkerrors.Wrapf(types.ErrInvalidVersion, "failed to unmarshal ICS29 counterparty version metadata: %s", counterpartyVersion)
```
colin-axner, line 126:
```suggestion
			return sdkerrors.Wrapf(err, "failed to unmarshal ICS29 counterparty version metadata: %s", counterpartyVersion)
```

modules/apps/transfer/ibc_module_test.go
```diff
@@ -85,7 +85,7 @@ func (suite *TransferTestSuite) TestOnChanOpenInit() {
 
 			tc.malleate() // explicitly change fields in channel and testChannel
 
-			err = cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
+			_, err = cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
```
colin-axner, line 88: we should be checking the returned version

modules/apps/27-interchain-accounts/controller/ibc_module.go
seantking: Are you suggesting changing the comment? 

We are passing the version the same as before.

modules/apps/27-interchain-accounts/controller/ibc_module_test.go
seantking: Yeah good point, thanks

modules/apps/27-interchain-accounts/controller/ibc_module.go
seantking: updated the comment anyway :+1:

modules/apps/29-fee/ibc_module.go
```diff
@@ -47,14 +47,24 @@ func (im IBCModule) OnChanOpenInit(
...
+	if err != nil {
+		return "", err
 	}
 
 	im.keeper.SetFeeEnabled(ctx, portID, channelID)
 
 	// call underlying app's OnChanOpenInit callback with the appVersion
-	return im.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID,
-		chanCap, counterparty, versionMetadata.AppVersion)
+	return string(versionBytes), nil
```
crodriguezvega, line 67: Why was the call to underlying app's `OnChanOpenInit` removed here?

CHANGELOG.md
```diff
@@ -43,6 +43,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### API Breaking
 
 * (transfer) [\#1250](https://github.com/cosmos/ibc-go/pull/1250) Deprecate `GetTransferAccount` since the `transfer` module account is never used.
+* (channel) [\#1283](https://github.com/cosmos/ibc-go/pull/1283) `OnChanOpenInit` now returns a version string in line with the latest [spec changes](https://github.com/cosmos/ibc/pull/629).
```
seantking, line 46: Open to improvements on the changelog message.

modules/apps/transfer/ibc_module_test.go
```diff
@@ -28,6 +28,11 @@ func (suite *TransferTestSuite) TestOnChanOpenInit() {
 		{
 			"success", func() {}, true,
 		},
+		{
+			"empty version string", func() {
+				channel.Version = ""
+			}, true,
+		},
```
crodriguezvega, line 35: Can you also add a test case when `channel.Version` is not empty but it's a string that does not match the expected `types.Version`?

crodriguezvega (commented):
Nice work, @seantking!

I just noticed the in ICA we don't use any `types.Version` like we have for transfer (`ics20-1`) or fee. Was there a reason not use one there?

modules/apps/27-interchain-accounts/controller/ibc_module.go
```diff
@@ -42,16 +42,16 @@ func (im IBCModule) OnChanOpenInit(
...
 
 	if err := im.keeper.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version); err != nil {
-		return err
+		return "", err
 	}
 
-	// call underlying app's OnChanOpenInit callback with the appVersion
+	// call underlying app's OnChanOpenInit callback with the passed in version
 	return im.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID,
 		chanCap, counterparty, version)
```
damiannolan, line 56: This gives an opportunity to the interchain accounts auth module to modify the version string, or perhaps mistakenly return the incorrect string, right? 

Is this something we want to support? i.e. providing the auth module the ability to mutate this? 
Currently the only expectation from this callback is the claiming of the channel capability.

The code below would complain about `appVersion` declared but not used...

```suggestion
        // call underlying app's OnChanOpenInit callback with the passed in version
        appVersion, err := im.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID,
		chanCap, counterparty, version)
	if err != nil {
	    return "", err
	}
	
	return version, nil
```

modules/apps/27-interchain-accounts/controller/ibc_module_test.go
```diff
@@ -197,11 +197,24 @@ func (suite *InterchainAccountsTestSuite) TestOnChanOpenInit() {
 			cbs, ok := suite.chainA.App.GetIBCKeeper().Router.GetRoute(module)
 			suite.Require().True(ok)
 
-			err = cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
+			version, err := cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
 				path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, chanCap, channel.Counterparty, channel.GetVersion(),
 			)
 
 			if tc.expPass {
+				expMetaData := icatypes.NewMetadata(
```
damiannolan, line 205: minor nit: casing 

```suggestion
				expMetadata := icatypes.NewMetadata(
```

modules/apps/29-fee/ibc_module.go
damiannolan: Its moved to above on line 53 so that the returned application version (for example: `ics20-1`) can be inserted into the `AppVersion` field of `types.Metadata`, and re-marshalled to a JSON encoded version string. Otherwise we would return directly from the app callback (writing `ics20-1` as the channel version into state), we would expect ics29 to be enabled but subsequent handshake callbacks on would fail as the channel version in state would not be the metadata object:

```json
{
    "AppVersion": "ics20-1",
    "FeeVersion": "ics29-1",
}
```

Does that make sense?

modules/apps/29-fee/ibc_module.go
```diff
@@ -47,14 +47,24 @@ func (im IBCModule) OnChanOpenInit(
...
+		return "", sdkerrors.Wrapf(types.ErrInvalidVersion, "expected %s, got %s", types.Version, versionMetadata.FeeVersion)
+	}
+
+	appVersion, err := im.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, versionMetadata.AppVersion)
+	if err != nil {
+		return "", err
+	}
+
+	versionMetadata.AppVersion = appVersion
+	versionBytes, err := types.ModuleCdc.MarshalJSON(&versionMetadata)
```
damiannolan, line 59: Super mega nit: could we add a line break between this so its consistent with the TRY code below, or alternatively remove the line break from the TRY handler?

modules/apps/29-fee/ibc_module_test.go
```diff
@@ -95,10 +95,27 @@ func (suite *FeeTestSuite) TestOnChanOpenInit() {
...
+					suite.Require().NoError(err)
+
+					suite.Require().Equal(version, string(versionBytes))
+				} else {
+					suite.Require().Equal(version, ibcmock.Version)
+				}
+
 				suite.Require().NoError(err, "unexpected error from version: %s", tc.version)
 			} else {
 				suite.Require().Error(err, "error not returned for version: %s", tc.version)
```
damiannolan, line 121: Could we also assert `version` is empty here too?

modules/apps/transfer/ibc_module_test.go
```diff
@@ -85,12 +90,13 @@ func (suite *TransferTestSuite) TestOnChanOpenInit() {
...
-			err = cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
+			version, err := cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
 				path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, chanCap, counterparty, channel.GetVersion(),
 			)
 
 			if tc.expPass {
 				suite.Require().NoError(err)
+				suite.Require().Equal(types.Version, version)
 			} else {
 				suite.Require().Error(err)
```
damiannolan, line 101: ditto on empty `version` assertion

damiannolan (commented):
Nice work! LGTM today, my only concern is the ICA auth module handling of the version string and if it should be given the opportunity to modify it, my initial thinking is no.
Other than that just some minor nits

modules/apps/transfer/ibc_module_test.go
seantking: We have this test case [already](https://github.com/cosmos/ibc-go/pull/1283/files/0db8c8672e36909b939101d685d19947ffc6f3d9#diff-84e1050545ac3bc948c89e7c6733f8130310338af194e367fa59890ce7796309R52), no?

crodriguezvega: oh, sorry. I missed it.

CHANGELOG.md
AdityaSripal, line 46: I would add that it may return a default version string if relayers pass in empty version

modules/apps/27-interchain-accounts/controller/ibc_module.go
seantking: Not really following your point with the code example. 

What would be the alternative here?

modules/apps/transfer/ibc_module.go
AdityaSripal, line 79: Shouldn't all of the applications/middleware have a similar code block?

modules/apps/29-fee/ibc_module.go
```diff
@@ -36,7 +36,7 @@ func (im IBCModule) OnChanOpenInit(
 	chanCap *capabilitytypes.Capability,
 	counterparty channeltypes.Counterparty,
 	version string,
-) error {
+) (string, error) {
 	var versionMetadata types.Metadata
 	if err := types.ModuleCdc.UnmarshalJSON([]byte(version), &versionMetadata); err != nil {
```
AdityaSripal, line 41: This will not enable fee by default. If provided string is empty then, get default version from underlying app and wrap with fee version.
If underlying app returns error with empty version then return error as well

modules/apps/27-interchain-accounts/controller/ibc_module.go
AdityaSripal: Agreed if ibc-auth is expected not to change the version at all. Then we should ignore it's return value. And if relayer passes in empty string, then just return our default version

AdityaSripal (changes requested):
Looks good so far. The default version logic you have in transfer should be extended to the other applications

modules/apps/29-fee/ibc_module.go
colin-axner: Do we want to enable fees by default? Otherwise there's no way to only use the default version of the base application

modules/apps/27-interchain-accounts/controller/ibc_module.go
damiannolan: Sorry if that was confusing, like @AdityaSripal says, we should ignore the `appVersion` field from the code example above by discarding it, as we don't want to depend on ica-auth to return the correct string(`version`) or risk modifying it:

```
_, err := im.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty version)
if err != nil {
    return "", err
}
```

> And if relayer passes in empty string, then just return our default version

But for ICA a relayer is never expected to initiate the channel handshake, it should be from some on-chain mechanism, correct?

modules/apps/29-fee/ibc_module.go
seantking: Yes, this was my assumption also. What is the benefit of enabling the fee middleware by default with an empty version passed in?

seantking:
> Looks good so far. The default version logic you have in transfer should be extended to the other applications

I left the default only for ics20. 

I think ics29 should not be enabled by default as @colin-axner suggests & relayers do not pass a version for ics-27. The controller for ics27 module always passes a metaData struct as a version to the `OnChanOpenInit` callback, it's not optional.

modules/apps/transfer/ibc_module.go
seantking: responded below

CHANGELOG.md
seantking: I left this out as I don't think the default is applicable to all modules.

modules/apps/27-interchain-accounts/controller/ibc_module.go
seantking: Updated this. Good catch :+1:

CHANGELOG.md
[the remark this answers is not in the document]
seantking: Added the first part but left out the version selection. I found that term confusing personally.

seantking:
> Nice work, @seantking!
> 
> I just noticed that in ICA we don't use any `types.Version` like we have for transfer (`ics20-1`) or fee. Was there a reason not use one there?

[Here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/keeper/account.go#L46) :)

modules/apps/27-interchain-accounts/controller/ibc_module.go
```diff
@@ -42,18 +42,21 @@ func (im IBCModule) OnChanOpenInit(
...
 
 	if err := im.keeper.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version); err != nil {
-		return err
+		return "", err
+	}
+
+	// call underlying app's OnChanOpenInit callback with the passed in version
+	if _, err := im.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version); err != nil {
+		return "", err
 	}
```
damiannolan, line 57: Could be useful to have a small explanation of why this is ignored in the comment.

damiannolan (approved):

modules/apps/29-fee/ibc_module.go
AdityaSripal: I do think fees should be enabled by default. My preference is that relayers get the default version of the stack if they don't explicitly specify otherwise. Some middleware will want to be activated by default  and some may want to be explicitly activated. For adoption purposes, I think it would be good to make incentivization the default rather than having to explicitly enable it. Because as I think the number of applications scale, most relayers will only be passing in empty strings and use non-empty strings only in very special cases (otherwise they would need to have context on every single port they relay for).

So a relayer that always defaults to what the application prefers will be opening incentivized channels.

Of course, if the relayer wants to explicitly disable fees then they can by passing in a version without the feeVersion. But we should design for a future where relayers are rarely passing in specific strings

CHANGELOG.md
AdityaSripal: Hence the word `may`, i think it could still be useful but maybe we want to leave that suggestion elsewhere (e.g. release notes or something)

modules/apps/29-fee/ibc_module.go
colin-axner: Sounds good to me. In favor of @AdityaSripal's approach. It should just be well documented that ics29 is an exceptional middleware application that we believe should be enabled by default, but that other applications (who might be copy/pasting our code) should give good consideration to whether their application should be enabled by default

seantking: Thanks for the explanation @AdityaSripal. Makes sense to me!

modules/apps/transfer/ibc_module_test.go
```diff
@@ -85,14 +90,16 @@ func (suite *TransferTestSuite) TestOnChanOpenInit() {
 
 			tc.malleate() // explicitly change fields in channel and testChannel
 
-			err = cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
+			version, err := cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
```
colin-axner, line 93: Change this to call the transfer IBCModule `OnChanOpenInit` directly. Currently this is getting the top level of the stack (ics29). It is better for us to just test ics20 in this unit test

We can either add a [`IBCTransferModule`](https://github.com/cosmos/ibc-go/blob/main/testing/simapp/app.go#L224) to simapp or call `transfer.NewIBCModule`

```go
ibcModule := transfer.NewIBCModule(suite.chainA.GetSimApp().TransferKeeper)
ibcModule.OnChanOpenInit(...)
```

seantking: This satisfies the success cases, but oddly fails on the test case for `capability already claimed`. Currently returning with an error:

`cannot claim nil capability: provided capability is nil`. 

Looking into it.

CHANGELOG.md
```diff
@@ -43,6 +43,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### API Breaking
 
 * (transfer) [\#1250](https://github.com/cosmos/ibc-go/pull/1250) Deprecate `GetTransferAccount` since the `transfer` module account is never used.
+* (channel) [\#1283](https://github.com/cosmos/ibc-go/pull/1283) The `OnChanOpenInit` application callback now returns a version string in line with the latest [spec changes](https://github.com/cosmos/ibc/pull/629).
```
colin-axner, line 46: can we add a note in the migration docs? See previous entry in v2 to v3

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
```diff
@@ -45,18 +45,21 @@ func (im IBCMiddleware) OnChanOpenInit(
...
+		return "", types.ErrControllerSubModuleDisabled
 	}
 
 	if err := im.keeper.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version); err != nil {
-		return err
+		return "", err
+	}
+
+	// call underlying app's OnChanOpenInit callback with the passed in version
+	if _, err := im.app.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version); err != nil {
```
colin-axner, line 58: should we add a comment why we discard the underlying application version?

modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -95,13 +100,30 @@ func (suite *FeeTestSuite) TestOnChanOpenInit() {
...
 			suite.Require().True(ok)
 
-			err = cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
+			version, err := cbs.OnChanOpenInit(suite.chainA.GetContext(), channel.Ordering, channel.GetConnectionHops(),
 				suite.path.EndpointA.ChannelConfig.PortID, suite.path.EndpointA.ChannelID, chanCap, counterparty, channel.Version)
 
 			if tc.expPass {
+				// check if the channel is fee enabled. If so version string should include metaData
+				isFeeEnabled := suite.chainA.GetSimApp().IBCFeeKeeper.IsFeeEnabled(suite.chainA.GetContext(), suite.path.EndpointA.ChannelConfig.PortID, suite.path.EndpointA.ChannelID)
+				if isFeeEnabled {
```
AdityaSripal, line 109: This is a strange way to test here because you're relying on the result of the callback execution to create your expected value. If the callback incorrectly creates a fee version and incorrectly sets fee enabled you won't catch it here. Similarly if the callback incorrectly doesn't set fee version and incorrectly doesn't set fee enabled you won't catch it.

I think we should add a boolean in the test cases that says `feeEnabled`. You then can add the expected value for each test case and then check if the feeEnabled flag is true and check if the version contains a fee version

colin-axner (approved):
Nice work! LGTM!

modules/core/05-port/types/module.go
```diff
@@ -11,10 +11,16 @@ import (
...
-	// OnChanOpenInit will verify that the relayer-chosen parameters are
-	// valid and perform any custom INIT logic.It may return an error if
-	// the chosen parameters are invalid in which case the handshake is aborted.
-	// OnChanOpenInit should return an error if the provided version is invalid.
+	// OnChanOpenInit will verify that the relayer-chosen parameters
+	// are valid and perform any custom INIT logic.
+	// It may return an error if the chosen parameters are invalid
+	// in which case the handshake is aborted.
+	// If the provided version string is non-empty, OnChanOpenInit should return
+	// the version string if valid or an error if the provided version is invalid.
```
AdityaSripal, line 19: Do we require that Init returns the exact same version if version is non-empty? I don't believe that is the case. Applications may want to modify the version string even if it is non-empty

AdityaSripal (commented):
Some testing and documentation nits :)

Would appreciate having them fixed, but doesn't need to happen in this PR

modules/apps/29-fee/ibc_middleware_test.go
seantking: Yup good catch. Thanks for the feedback :handshake:

modules/core/05-port/types/module.go
AdityaSripal: Yes I agree we should change the spec as well

AdityaSripal (approved):
Nice work! Thanks @seantking

Issue #641, opened 2021-12-16
Panic in core IBC Keeper if staking params are uninitialized 


colin-axner:
## Summary

If the staking keeper passed to the core IBC Keeper does not have initialized params, our keeper should panic. We need the UnbondingTime to complete connection handshakes

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
We need to add the check [in this function](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/keeper.go#L36).

catShaark:
> If the staking keeper passed to the core IBC Keeper does not have initialized params, our keeper should panic.

may be instead check if the passed in keepers is empty structs or not, cuz there's not a way to check in `KVStore` if the `stakingKeeper`'s params is set without the code looking weird ( I'm talking about using `sdk.Context` in `NewApp()` ) and also, If the `stakingKeeper`'s params is not set in the `KVStore`, there'll be panic anyway.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1284, opened 2022-04-21
add empty keepers checking in ibc NewKeeper

catShaark:
## Description

So I add the check to see if the keepers passed into `ibc` `NewKeeper()` is empty or not, and panic if any of those keepers is empty.

closes: #641

---

write a little note why.


modules/core/keeper/keeper.go
```diff
@@ -1,11 +1,16 @@
...
+	"fmt"
+	"reflect"
+
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	capabilitykeeper "github.com/cosmos/cosmos-sdk/x/capability/keeper"
 	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 
+	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
+	upgradekeeper "github.com/cosmos/cosmos-sdk/x/upgrade/keeper"
```
damiannolan, line 13: Thanks for taking the time to put together this PR.

This puts a hard dependency on the concrete implementations of the sdk staking and upgrade keepers. 
We generally use [expected keepers](https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/types/expected_keepers.go#L11-L26) to abstract these kinds of dependencies with a flexible interface.
For example, one may choose to use a custom staking module, that differs slightly from the standard sdk implementation but still honours the interfaces we depend on.

Unsure how others feel about this. cc @colin-axner @seantking @crodriguezvega @AdityaSripal

catShaark: Yeah I think you're right. I will try to make changes accordingly

catShaark: So I have remove the type assertion to sdk-defined keepers and use reflect.ValueOf(stakingKeeper).IsZero(), this should check the value of the interface to see if it's an empty struct or not. What do you think ?

modules/core/keeper/keeper.go
```diff
@@ -46,6 +49,11 @@ func NewKeeper(
 		paramSpace = paramSpace.WithKeyTable(keyTable)
 	}
 
+	err := checkEmptyKeepers(stakingKeeper, upgradeKeeper, scopedKeeper)
+	if err != nil {
+		panic(fmt.Errorf("cannot initialize ibc keeper: %w", err))
+	}
+
```
colin-axner, line 56:
```suggestion
	if err := checkEmptyKeepers(stakingKeeper, upgradeKeeper, scopedKeeper); err != nil {
		panic(fmt.Errorf("cannot initialize IBC keeper: %w", err))
	}

```

modules/core/keeper/keeper.go
```diff
@@ -76,3 +84,17 @@ func (k *Keeper) SetRouter(rtr *porttypes.Router) {
 	k.Router = rtr
 	k.Router.Seal()
 }
+
+func checkEmptyKeepers(stakingKeeper clienttypes.StakingKeeper, upgradeKeeper clienttypes.UpgradeKeeper,
```
colin-axner, line 88: I think it might be nicer not to abstract these into a function and just do the checks directly within `NewKeeper`. 

I agree with the approach of checking if the keeper is empty as opposed to if the params are initialized :+1:

colin-axner (commented):
Thanks for picking this one up! Do you think you could add some unit tests as well?

catShaark:
> Thanks for picking this one up! Do you think you could add some unit tests as well?

yess

crodriguezvega:
Thanks, @catShaark! We also need a changelog entry. :)

modules/core/keeper/keeper.go
```diff
@@ -46,6 +49,15 @@ func NewKeeper(
...
 	}
 
+	// panic if any of the keepers passed in is empty
+	if reflect.ValueOf(stakingKeeper).IsZero() {
+		panic(fmt.Errorf("cannot initialize ibc keeper: empty staking keeper"))
+	} else if reflect.ValueOf(upgradeKeeper).IsZero() {
+		panic(fmt.Errorf("cannot initialize ibc keeper: empty upgrade keeper"))
+	} else if reflect.DeepEqual(capabilitykeeper.ScopedKeeper{}, scopedKeeper) {
+		panic(fmt.Errorf("cannot initialize ibc keeper: empty scoped keeper"))
+	}
```
colin-axner, line 59: super nit (you don't need to apply it)
```suggestion
	if reflect.ValueOf(stakingKeeper).IsZero() {
		panic(fmt.Errorf("cannot initialize IBC keeper: empty staking keeper"))
	} 
	
	if reflect.ValueOf(upgradeKeeper).IsZero() {
		panic(fmt.Errorf("cannot initialize IBC keeper: empty upgrade keeper"))
	} 
	
	if reflect.DeepEqual(capabilitykeeper.ScopedKeeper{}, scopedKeeper) {
		panic(fmt.Errorf("cannot initialize IBC keeper: empty scoped keeper"))
	}
```

I find this easier to read

modules/core/keeper/keeper_test.go
```diff
@@ -0,0 +1,124 @@
...
+	dummyField string
+}
+
+func (d DummyStakingKeeper) GetHistoricalInfo(ctx sdk.Context, height int64) (stakingtypes.HistoricalInfo, bool) {
+	return stakingtypes.HistoricalInfo{}, true
+}
+
+func (d DummyStakingKeeper) UnbondingTime(ctx sdk.Context) time.Duration {
+	return 0
+}
```
colin-axner, line 67: Thanks for adding these! Could we use "Mock" instead of "Dummy"? I find "Dummy" to be non-neutral language

modules/core/keeper/keeper_test.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+type KeeperTestSuite struct {
+	suite.Suite
+
+	app           *simapp.SimApp
+	stakingKeeper clienttypes.StakingKeeper
+	upgradeKeeper clienttypes.UpgradeKeeper
+	scopedKeeper  capabilitykeeper.ScopedKeeper
+}
```
colin-axner, line 28: I think I'd prefer to use the existing `KeeperTestSuite` (as defined in msg_server_test). We can move the declaration existing `KeeperTestSuite` to this file

modules/core/keeper/msg_server_test.go
```diff
@@ -35,7 +33,7 @@ type KeeperTestSuite struct {
 }
 
 // SetupTest creates a coordinator with 2 test chains.
-func (suite *KeeperTestSuite) SetupTest() {
+func (suite *MsgServerTestSuite) SetupTest() {
```
colin-axner, line 36: I think I'm in favor of leaving as `KeeperTestSuite`

modules/core/keeper/keeper_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (suite *KeeperTestSuite) NewIBCKeeper() {
+	ibckeeper.NewKeeper(
+		suite.app.AppCodec(),
+		suite.app.GetKey(ibchost.StoreKey),
+		suite.app.GetSubspace(ibchost.ModuleName),
+		suite.stakingKeeper,
+		suite.upgradeKeeper,
+		suite.scopedKeeper,
+	)
```
colin-axner, line 49: Could we place this inside the `suite.Run` func for `TestNewKeeper`. Typically in our tests, when using table tests, we use local variable for each of the inputs which will change

So
```go
func TestNewKeeper() {
var (
    stakingKeeper
    upgradeKeeper
    scopedKeeper
)

// table tests
func() { // malleate
    stakingKeeper = emptyStakingKeeper
}
...

suite.Run() {
    // setup
    stakingKeeper = suite.chainA.GetSimApp().StakingKeeper // set passing defaults
    // etc
    
    if tc.expPass {
        require.NotPanics(
            keeper.NewKeeper(stakingKeeper, ...)
        )
    } else {
        require.Panics(
            keeper.NewKeeper(stakingKeeper, ...)
         )
    }
}
```

modules/core/keeper/keeper_test.go
catShaark: yess

modules/core/keeper/keeper_test.go
```diff
@@ -0,0 +1,138 @@
...
+		malleate func()
+		expPass  bool
+	}{
+		{"failure: empty staking keeper", func() {
+			emptyStakingKeeper := stakingkeeper.Keeper{}
+
+			stakingKeeper = emptyStakingKeeper
+
+		}, false},
+		{"failure: empty dummy staking keeper", func() {
```
colin-axner, line 88:
```suggestion
		{"failure: empty mock staking keeper", func() {
```

colin-axner (approved):
ACK, thank you @catShaark!

modules/core/keeper/keeper.go
damiannolan: Hey @catShaark, I also find this much easier to read and is probably considered more idiomatic Go, since the flow of execution will terminate with `panic()` I think a single conditional per keeper type is cleaner!
Personally I'm in favour of applying the suggested code snippet! :)

damiannolan (approved):
Thank you! Great work on these PRs! ❤️

modules/core/keeper/keeper_test.go
catShaark: ohh jeez, sorry I forgot this

modules/core/keeper/keeper.go
catShaark: yeah, it does look cleaner and I just applied the suggested change

Issue #101, opened 2021-03-31
DenomTrace gRPC should take in a IBC prefixed denom rather than a hash


colin-axner:
## Summary

The `DenomTrace` gRPC takes in just the hash as the request argument. This requires clients to parse out the `ibc` prefix. The hash is never used without the `ibc` prefix so we should parse out the `ibc` portion in the gRPC handler

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1312, opened 2022-04-29
chore: Change Query/DenomTrace gRPC

catShaark:
## Description

Change the `DenomTrace` grpc so that it takes the whole ibc denom instead of just the hash

closes: #101

---

write a little note why.


modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -22,17 +23,21 @@ func (q Keeper) DenomTrace(c context.Context, req *types.QueryDenomTraceRequest)
...
 
-	hash, err := types.ParseHexHash(req.Hash)
+	if !strings.HasPrefix(req.Denom, "ibc/") {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("invalid denom with no ibc prefix %s", req.Denom))
+	}
+
+	hash, err := types.ParseHexHash(req.Denom[4:])
 	if err != nil {
-		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("invalid denom trace hash %s, %s", req.Hash, err))
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("invalid denom trace hash %s, %s", req.Denom[4:], err))
```
crodriguezvega, line 32:
```suggestion
		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("invalid denom trace hash: %s, error: %s", req.Denom[4:], err))
```

crodriguezvega (commented):
Thanks, @catShaark. I left just a couple of nits. And we need a changelog entry. :) Since this changing the API, then I would say in the API breaking section.

modules/apps/transfer/keeper/grpc_query_test.go
```diff
@@ -21,10 +21,10 @@ func (suite *KeeperTestSuite) TestQueryDenomTrace() {
 		expPass  bool
 	}{
 		{
-			"invalid hex hash",
+			"invalid ibc denom",
```
seantking, line 24: nit: could we put the success case first? :=)

catShaark: yesss

seantking (approved):
Great work!

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -22,17 +23,21 @@ func (q Keeper) DenomTrace(c context.Context, req *types.QueryDenomTraceRequest)
 		return nil, status.Error(codes.InvalidArgument, "empty request")
 	}
 
-	hash, err := types.ParseHexHash(req.Hash)
+	if !strings.HasPrefix(req.Denom, "ibc/") {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("invalid denom with no ibc prefix: %s", req.Denom))
+	}
```
colin-axner, line 28: I think we should make this change backwards compatible. We could do so by checking if it has the ibc prefix and then remove the prefix before performing the query. I should have specified this on the original issue (but at the time there was a lot less dependencies on this gRPC

So essentially:
```go
if strings.HasPrefix(denom, "ibc/") {
    denom = denom[4:]
}

catShaark: So I'll make it like you said and leave the field name still be `denom`

crodriguezvega: That's a good suggestion from @colin-axner. If you make the change backwards compatible, then the changelog entry shouldn't be in the API breaking section, but in the improvements one instead.

catShaark: I'll apply the change accordingly

catShaark:
I'll close this PR and make a new one

Issue #1322, opened 2022-05-01
Check channel ID exists in `RegisterCounterpartyAddress`


crodriguezvega:
Check if [channel ID coming in `MsgRegisterCounterpartyAddres`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/29-fee/types/tx.pb.go#L39) exists inside [RegisterCounterpartyAddress](https://github.com/cosmos/ibc-go/blob/main/modules/apps/29-fee/keeper/msg_server.go#L17).

colin-axner:
The check is to ensure we don't allow a relayer to register an address for a channel that does not exist

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1339, opened 2022-05-04
refactor: adding check if channel exists before register counterparty

seantking:
## Description

In order to add this check we needed access to the PortID. I have added portID to the Msg + updated CLI etc. Wdyt?

closes: #1322 

---

write a little note why.


modules/apps/29-fee/types/msgs.go
```diff
@@ -36,6 +37,11 @@ func (msg MsgRegisterCounterpartyAddress) ValidateBasic() error {
 		return ErrCounterpartyAddressEmpty
 	}
 
+	// validate portId
+	if err := host.PortIdentifierValidator(msg.PortId); err != nil {
```
colin-axner, line 41: could we add a test case for this?

modules/apps/29-fee/keeper/msg_server_test.go
```diff
@@ -29,16 +30,25 @@ func (suite *KeeperTestSuite) TestRegisterCounterpartyAddress() {
...
 		sender = suite.chainA.SenderAccount.GetAddress().String()
 		counterparty = suite.chainB.SenderAccount.GetAddress().String()
+		channelID = suite.path.EndpointA.ChannelID
+
 		tc.malleate()
-		msg := types.NewMsgRegisterCounterpartyAddress(sender, counterparty, ibctesting.FirstChannelID)
+
+		msg := types.NewMsgRegisterCounterpartyAddress(sender, counterparty, suite.path.EndpointA.ChannelConfig.PortID, channelID)
 
 		_, err := suite.chainA.SendMsgs(msg)
```
seantking, line 53: For some reason, the test suite is exploding here and returning the error message we expect (channel does not exist) but it's panicking and exiting rather than being handled.

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -17,6 +17,12 @@ var _ types.MsgServer = Keeper{}
 func (k Keeper) RegisterCounterpartyAddress(goCtx context.Context, msg *types.MsgRegisterCounterpartyAddress) (*types.MsgRegisterCounterpartyAddressResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	// only register counterparty address if the channel exists
+	_, found := k.channelKeeper.GetChannel(ctx, msg.PortId, msg.ChannelId)
```
colin-axner, line 21: can we add a check to see if the channel is fee enabled as well?

modules/apps/29-fee/keeper/msg_server_test.go
colin-axner: `SendMsgs` assumes the message to pass and errors [out otherwise](https://github.com/cosmos/ibc-go/blob/bd086506f9256c4d5776bc38d2e930b523eb5125/testing/simapp/test_helpers.go#L261). We could another function to the testing package to send a message expected to fail

modules/apps/29-fee/keeper/msg_server.go
seantking: Yup I can do that. I was going to mention this also. I feel like we discussed this and said not to add it for some reason but I can't remember why (can't think of any reason why not also)

modules/apps/29-fee/keeper/msg_server_test.go
seantking: Gotcha. I'll just call the function directly similar to the other tests

modules/apps/29-fee/types/msgs.go
seantking: done

modules/apps/29-fee/keeper/msg_server.go
seantking: added

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### Improvements
 
+* (modules/app/29-fee) [\#1339](https://github.com/cosmos/ibc-go/pull/1339) The `RegisterCounterpartyAddress` gRPC endpoint now checks if the channel exists before registering.
```
colin-axner, line 52: Changelog entry isn't super necessary since we haven't release ics29, but guess it doesn't hurt

colin-axner (approved):
Wahoo! Very clean work! :)

AdityaSripal (approved):

modules/apps/29-fee/client/cli/tx.go
```diff
@@ -101,18 +101,18 @@ func NewPayPacketFeeAsyncTxCmd() *cobra.Command {
 // NewRegisterCounterpartyAddress returns the command to create a MsgRegisterCounterpartyAddress
 func NewRegisterCounterpartyAddress() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "register-counterparty [address] [counterparty-address] [channel-id]",
+		Use:     "register-counterparty [address] [counterparty-address] [channel-id] [port-id]",
```
crodriguezvega, line 104: It would be more consistent with the rest of CLIs if the order of the arguments is reversed, so `[port-id] [channel-id]`.

modules/apps/29-fee/client/cli/tx.go
```diff
@@ -101,18 +101,18 @@ func NewPayPacketFeeAsyncTxCmd() *cobra.Command {
...
+		Example: fmt.Sprintf("%s tx ibc-fee register-counterparty cosmos1rsp837a4kvtgp2m4uqzdge0zzu6efqgucm0qdh osmo1v5y0tz01llxzf4c2afml8s3awue0ymju22wxx2 channel-0 transfer", version.AppName),
+		Args:    cobra.ExactArgs(4),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			clientCtx, err := client.GetClientTxContext(cmd)
 			if err != nil {
 				return err
 			}
 
-			msg := types.NewMsgRegisterCounterpartyAddress(args[0], args[1], args[2])
+			msg := types.NewMsgRegisterCounterpartyAddress(args[0], args[1], args[2], args[3])
```
crodriguezvega, line 115: And I think this is wrong because `args[3]` is the `port-id`, but it should pass the `channel-id` as the last argument... But if we revert the order or the CLI parameters, then this will be fine. :)

crodriguezvega (commented):
Maybe we can address the reversal of parameters in a separate PR?

modules/apps/29-fee/client/cli/tx.go
seantking: Good catch.

Issue #1318, opened 2022-05-01
Return error when trying to incentivize an already relayed packet


crodriguezvega:
To prevent that a packet that has already been relayed can still be incentivized (and therefore fees end up being locked up forever), on [`PayPacketFeeAsync`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/29-fee/keeper/msg_server.go#L62), if the `NextSequenceSend` is greater than the packet sequence that is intended to be incentivized AND the packet commitment does not exist (since packet commitments are deleted when packets are [acknowledged](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/packet.go#L493) or [timed out](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/timeout.go#L154)), return an error.

colin-axner:
```go
if packet is not sent {
    return err
}

if packet acknowledged || timed out {
    return err
}
```

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1347, opened 2022-05-11
fix: disallow incentivizing packets which have not been sent or have been already relayed

colin-axner:
## Description

closes: #1318 

---

write a little note why.

### DIFF
--- a/modules/apps/29-fee/keeper/msg_server.go
+++ b/modules/apps/29-fee/keeper/msg_server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 
 	"github.com/cosmos/ibc-go/v3/modules/apps/29-fee/types"
 	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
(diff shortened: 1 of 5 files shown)


modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -71,6 +73,21 @@ func (k Keeper) PayPacketFeeAsync(goCtx context.Context, msg *types.MsgPayPacket
 		return nil, types.ErrFeeModuleLocked
 	}
 
+	nextSeqSend, found := k.GetNextSequenceSend(ctx, msg.PacketId.PortId, msg.PacketId.ChannelId)
+	if !found {
+		return nil, sdkerrors.Wrapf(channeltypes.ErrSequenceSendNotFound, "channel does not exist, port: %s, channel: %s", msg.PacketId.PortId, msg.PacketId.ChannelId)
```
colin-axner, line 78:
```suggestion
		return nil, sdkerrors.Wrapf(channeltypes.ErrSequenceSendNotFound, "channel does not exist, portID: %s, channelID: %s", msg.PacketId.PortId, msg.PacketId.ChannelId)
```

modules/apps/29-fee/keeper/msg_server.go
```diff
@@ -71,6 +73,21 @@ func (k Keeper) PayPacketFeeAsync(goCtx context.Context, msg *types.MsgPayPacket
 		return nil, types.ErrFeeModuleLocked
 	}
 
+	nextSeqSend, found := k.GetNextSequenceSend(ctx, msg.PacketId.PortId, msg.PacketId.ChannelId)
+	if !found {
+		return nil, sdkerrors.Wrapf(channeltypes.ErrSequenceSendNotFound, "channel does not exist, portID: %s, channelID: %s", msg.PacketId.PortId, msg.PacketId.ChannelId)
+	}
+
+	// only allow incentivizing of packets which have been sent
```
seantking, line 81: Should we also check if the channel is in an `OPEN` state?

seantking (approved):

modules/apps/29-fee/keeper/msg_server.go
colin-axner: This is a great question. The answer isn't immediately obvious. I would initially think it is a question of supporting optimistic sends, but we also need to consider the case where a channel closes. 

If a channel closes after a packet is sent, it is possible a user may want to incentivize a timeout relayer to get their packet timed out. I think that is a strong enough reason to allow incentivization on any channel state (so long as the channel exists)

AdityaSripal (approved):
Nice work!

PR #1401, opened 2022-05-20
update getStakingKeeper to use expected_keeper interface

okwme:
## Description

Interchain Security is using the IBC go test package `ibctesting.TestingApp` that requires a method `GetStakingKeeper`. Currently this is returning the struct literal of the staking keeper. If it instead returned the expected_interface from the ibc client types package it would be possible to substitute the stakingKeeper for the consumerKeeper (replacement for staking on consumer chains) and keep using the test framework.

closes: #XXXX

---

write a little note why.


crodriguezvega (commented):
Thanks for the PR, @okwme!

However, I am wondering if this needs to be merged to ibc-go right now or it should be merged later on as part of the other changes of the testing package that are happening in [this fork](https://github.com/informalsystems/ibc-go/tree/interchain-security-2) to support testing of interchain security... 

Do you really need this change in ibc-go right now or can you better make this change in the interchain security fork instead?

okwme:
We can use this branch for now.
In fact it's still being worked out whether this solution will suffice or if we need to use a different testing framework since the IBC testing package is really expecting a blockchain that has a staking module (and interchain security does not).

It may be worth having a session to discuss IBC testing in case interchain security chains should be accomodated within the IBC testing framework or not. In the short term we can work from a fork.

Issue #1446, opened 2022-05-27
bug(transfer): Cannot send namespaced IBC denom using slashes


fedekunze:
## Summary of Bug

#894 wasn't fully resolved as `ParseDenomTrace` still assumes no slashes on the raw IBC denomination. Sending an `erc20/` prefixed denomination from Evmos

```bash
evmosd tx ibc-transfer transfer transfer channel-7 osmo1anwz9u6nrtq20rumr0hhmgx48j29wakt9wle3f   "1000000000000000000erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA" \
--fees 80atevmos --from hot2 \
--node="https://tendermint.bd.evmos.dev:26657/" --chain-id evmos_9000-4 -b block
```

results in

```bash
raw_log: 'failed to execute message; message index: 0: 87AAE606C6372DE2C90C3D88B29CE730D9282D95F3238D2CD71D57B43576A3BD:
  denomination trace not found'
```

The bug is in this line: [transfer/types/trace.go#L36](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/trace.go#L36)

## Version

`main` / `v3` (latest)

## Steps to Reproduce

- Send the transaction from Evmos testnet as defined above, or
- Add a unit test on `TestParceDenomTrace` using `erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA` or `gamm/pool/1` as base denom

```go
--- FAIL: TestParseDenomTrace (0.00s)
    /Users/federico/ibc-go/modules/apps/transfer/types/trace_test.go:26: 
        	Error Trace:	trace_test.go:26
        	Error:      	Not equal: 
        	            	expected: types.DenomTrace{Path:"transfer/channelToA", BaseDenom:"erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA"}
        	            	actual  : types.DenomTrace{Path:"transfer/channelToA/erc20", BaseDenom:"0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,4 @@
        	            	 (types.DenomTrace) {
        	            	- Path: (string) (len=19) "transfer/channelToA",
[...]

PR #1451, opened 2022-05-27
fix to correctly parse denoms with slashes in the base denom

crodriguezvega:
## Description

closes: #1446 

---

write a little note why.

### DIFF
--- a/modules/apps/transfer/types/trace.go
+++ b/modules/apps/transfer/types/trace.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -12,6 +13,7 @@ import (
 	tmbytes "github.com/tendermint/tendermint/libs/bytes"
 	tmtypes "github.com/tendermint/tendermint/types"
 
+	channeltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
 	host "github.com/cosmos/ibc-go/v3/modules/core/24-host"
 )
 
@@ -21,6 +23,9 @@ import (
 // Examples:
 //
 // 	- "portidone/channelidone/uatom" => DenomTrace{Path: "portidone/channelidone", BaseDenom: "uatom"}
+//  - "portidone/channelidone/portidtwo/channelidtwo/uatom" => DenomTrace{Path: "portidone/channelidone/portidtwo/channelidtwo", BaseDenom: "uatom"}
+//  - "portidone/channelidone/gamm/pool/1" => DenomTrace{Path: "portidone/channelidone", BaseDenom: "gamm/pool/1"}
+//  - "gamm/pool/1" => DenomTrace{Path: "", BaseDenom: "gamm/pool/1"}
 // 	- "uatom" => DenomTrace{Path: "", BaseDenom: "uatom"}
 func ParseDenomTrace(rawDenom string) DenomTrace {
 	denomSplit := strings.Split(rawDenom, "/")
@@ -32,9 +37,24 @@ func ParseDenomTrace(rawDenom string) DenomTrace {
 		}
 	}
 
+	j := 1
+	path := []string{}
+	baseDenom := []string{}
+	length := len(denomSplit)
+	r, _ := regexp.Compile(fmt.Sprintf("%s[0-9]+", channeltypes.ChannelPrefix))
+	for i := 0; i < length; i = i + j {
+		if denomSplit[i] == PortID && r.MatchString(denomSplit[i+1]) {
+			path = append(path, denomSplit[i], denomSplit[i+1])
+			j = 2
+		} else {
+			baseDenom = append(baseDenom, denomSplit[i])
+			j = 1
+		}
+	}
+
 	return DenomTrace{
-		Path:      strings.Join(denomSplit[:len(denomSplit)-1], "/"),
-		BaseDenom: denomSplit[len(denomSplit)-1],
+		Path:      strings.Join(path, "/"),
+		BaseDenom: strings.Join(baseDenom, "/"),
 	}
 }
 
--- a/modules/apps/transfer/types/trace_test.go
+++ b/modules/apps/transfer/types/trace_test.go
@@ -14,10 +14,19 @@ func TestParseDenomTrace(t *testing.T) {
 	}{
 		{"empty denom", "", DenomTrace{}},
 		{"base denom", "uatom", DenomTrace{BaseDenom: "uatom"}},
-		{"trace info", "transfer/channelToA/uatom", DenomTrace{BaseDenom: "uatom", Path: "transfer/channelToA"}},
-		{"incomplete path", "transfer/uatom", DenomTrace{BaseDenom: "uatom", Path: "transfer"}},
-		{"invalid path (1)", "transfer//uatom", DenomTrace{BaseDenom: "uatom", Path: "transfer/"}},
-		{"invalid path (2)", "transfer/channelToA/uatom/", DenomTrace{BaseDenom: "", Path: "transfer/channelToA/uatom"}},
+		{"base denom ending with '/'", "uatom/", DenomTrace{BaseDenom: "uatom/"}},
+		{"base denom with single '/'s", "gamm/pool/1", DenomTrace{BaseDenom: "gamm/pool/1"}},
+		{"base denom with double '/'s", "gamm//pool//1", DenomTrace{BaseDenom: "gamm//pool//1"}},
+		{"trace info", "transfer/channel-0/uatom", DenomTrace{BaseDenom: "uatom", Path: "transfer/channel-0"}},
+		{"trace info with base denom ending in '/'", "transfer/channel-0/uatom/", DenomTrace{BaseDenom: "uatom/", Path: "transfer/channel-0"}},
+		{"trace info with single '/' in base denom", "transfer/channel-0/erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA", DenomTrace{BaseDenom: "erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA", Path: "transfer/channel-0"}},
+		{"trace info with multiple '/'s in base denom", "transfer/channel-0/gamm/pool/1", DenomTrace{BaseDenom: "gamm/pool/1", Path: "transfer/channel-0"}},
+		{"trace info with multiple double '/'s in base denom", "transfer/channel-0/gamm//pool//1", DenomTrace{BaseDenom: "gamm//pool//1", Path: "transfer/channel-0"}},
+		{"trace info with multiple port/channel pairs", "transfer/channel-0/transfer/channel-1/uatom", DenomTrace{BaseDenom: "uatom", Path: "transfer/channel-0/transfer/channel-1"}},
+		{"incomplete path", "transfer/uatom", DenomTrace{BaseDenom: "transfer/uatom"}},
+		{"invalid path (1)", "transfer//uatom", DenomTrace{BaseDenom: "transfer//uatom"}},
+		{"invalid path (2)", "transfer/channelToA/uatom", DenomTrace{BaseDenom: "transfer/channelToA/uatom"}},
+		{"invalid path (3)", "channel-0/transfer/uatom", DenomTrace{BaseDenom: "channel-0/transfer/uatom"}},
 	}
 
 	for _, tc := range testCases {
@@ -131,6 +140,7 @@ func TestValidateIBCDenom(t *testing.T) {
 	}{
 		{"denom with trace hash", "ibc/7F1D3FCF4AE79E1554D670D1AD949A9BA4E4A3C76C63093E17E446A46061A7A2", false},
 		{"base denom", "uatom", false},
+		{"base denom ending with '/'", "uatom/", false},
 		{"base denom with single '/'s", "gamm/pool/1", false},
 		{"base denom with double '/'s", "gamm//pool//1", false},
 		{"non-ibc prefix with hash", "notibc/7F1D3FCF4AE79E1554D670D1AD949A9BA4E4A3C76C63093E17E446A46061A7A2", false},


modules/apps/transfer/types/trace.go
```diff
@@ -32,9 +37,25 @@ func ParseDenomTrace(rawDenom string) DenomTrace {
...
 
+	path := []string{}
+	baseDenom := []string{}
+	length := len(denomSplit)
+	r, err := regexp.Compile(fmt.Sprintf("%s[0-9]+", channeltypes.ChannelPrefix))
+	if err != nil {
+		panic("could not compile regexp 'channel-[0-9]+'")
+	}
+	for i := 0; i < length; i = i + 2 {
+		if i < length-1 && denomSplit[i] == PortID && r.MatchString(denomSplit[i+1]) {
```
crodriguezvega, line 48: Checking that both the port ID is `transfer` and the channel ID matches the regex `channel-[0-9]+` might be too restrictive, right? Because the specs don't force the implementations to follow these two patterns, right? 
If I remember correctly the transfer module of Agoric actually uses a custom port ID (at least it did back in September). And other implementations of IBC might also decide to use a different pattern for the channel identifiers. And don't get me started if a custom transfer module decides to use slashes in the port ID...
So it seems like it's a bit tricky to parse the denom and be able to cover all these possible scenarios.

danburck (approved):
Nice, LGTM!

modules/apps/transfer/types/trace.go
AdityaSripal: We should check that the portID is transfer. Since this will always pass through the transfer module and even if tokens came from a previous app, we care about the chain of `transfer` module hops. But we do not enforce channel-id naming.

AdityaSripal: So keep transfer and remove regex for channelIDs

AdityaSripal:
Let's manually test the following:

- Sending slashed coins works
- Receiving slashed coins works, verify trace and base_denom is as expected
- (ideally but maybe not super necessary) receive to a third chain verify trace and base_denom is as expected
- Unwind token path completely and check that we get back the original token

We should also test a packet refund, I think the second chain will be most relevant here

fedekunze (changes requested):
Can we add the 2 denom examples (`gamm/pool/1` and `erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA`) to `TestValidatePrefixedDenom` and `TestDenomTrace_Validate` just to make sure it won't fail elsewhere?

modules/apps/transfer/types/trace.go
```diff
@@ -32,9 +37,25 @@ func ParseDenomTrace(rawDenom string) DenomTrace {
 		}
 	}
 
+	path := []string{}
+	baseDenom := []string{}
+	length := len(denomSplit)
+	r, err := regexp.Compile(fmt.Sprintf("%s[0-9]+", channeltypes.ChannelPrefix))
+	if err != nil {
+		panic("could not compile regexp 'channel-[0-9]+'")
```
chatton, line 45: My understanding is that we recover from the panics higher up in the framework, but if we can now fail parsing the string, maybe it makes sense to add an `error` as a return value (it looks like the calling methods already return an error)

If you would prefer to leave the panic in, we could use the regexp.MustCompile option which already panics.

crodriguezvega: Thanks for the suggestion, @chatton! Since Aditya recommends to remove the regex for channel identifier, then we will not need this anymore. :)

modules/apps/transfer/types/trace.go
crodriguezvega: Done.

crodriguezvega:
I've been doing some manual testing, so I just want to report back the results so far... I have tested 2 scenarios so far:

### Scenario 1

Setup of 2 chains:

- `chain1` and `chain2` are both using the simd from [the branch of this PR](https://github.com/cosmos/ibc-go/tree/carlos/1446-bugtransfer-cannot-send-namespaced-ibc-denom-using-slashes).

I setup the transfer channel between both chains and I try to send `100erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA` from an account on `chain1` to an account on `chain2`. The transfer succeeds and `denom-trace` and `denom-hash` CLI commands return the expected output on `chain2`:

```
> simd q ibc-transfer denom-trace C1D5D3AE1B5C9C9778DF7E8972352937B52C8E97C66505FF6639ED12BD9E8084
denom_trace:
  base_denom: erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA
  path: transfer/channel-0
```
```
> simd q ibc-transfer denom-hash transfer/channel-0/erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA
hash: C1D5D3AE1B5C9C9778DF7E8972352937B52C8E97C66505FF6639ED12BD9E8084
```

### Scenario 2

Setup of 2 chains:

- `chain1` is using the simd from [the branch of this PR](https://github.com/cosmos/ibc-go/tree/carlos/1446-bugtransfer-cannot-send-namespaced-ibc-denom-using-slashes).
- `chain2` is using the simd from main. 

I setup the transfer channel between both chains and I try to send `100erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA` from an account on `chain1` to an account on `chain2`. The transfer fails:

```json
{
  "type":"fungible_token_packet",
  "attributes":[
     {
        "key":"module",
        "value":"transfer"
     },
     {
        "key":"sender",
        "value":"cosmos1sc3r6elp0vh8qftt55kr9qawrwe6xq6qsl74xx"
     },
     {
        "key":"receiver",
        "value":"cosmos1qufuxgutlt0skpccyxlzmwtmvge0ccfrl96l80"
     },
     {
        "key":"denom",
        "value":"erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA"
     },
     {
        "key":"amount",
        "value":"100"
     },
     {
        "key":"acknowledgement",
        "value":"error:\"ABCI code: 1: error handling packet on destination chain: see events for details\" "
     },
     {
        "key":"error",
        "value":"ABCI code: 1: error handling packet on destination chain: see events for details"
     }
  ]
}
```

The error message is actually not very useful and I think there is nothing else in the events to give more information about what went wrong. But I suspect that the problem is the following (since `chain2` is running the code from main without this fix):

In the `OnRecvPacket` callback there is [a validation of the data in the `FungibleTokenPacketData`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L190). The last step of this validation [calls `ValidatePrefixedDenom`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/packet.go#L54) with the denom `erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA`. Inside `ValidatePrefixedDenom` the last step is [to validate the trace identifiers](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/trace.go#L159-L160). The input to `validateTraceIdentifiers` is going to be a string slice with only the element `erc20`. And then inside `validateTraceIdentifiers` I think an error is returned because [len(identifiers)%2 != 0](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/trace.go#L74-L76).

So I am afraid that for IBC transfers of coins with slashed denoms to succeed we are going to need to have this fix deployed as well in all the counterparty chains.

modules/apps/transfer/types/trace.go
```diff
@@ -70,6 +74,22 @@ func (dt DenomTrace) GetFullDenomPath() string {
 	return dt.GetPrefix() + dt.BaseDenom
 }
 
+func extractPathAndBaseDenomFromDenomTrace(denomTraceItems []string) (string, string) {
```
fedekunze, line 77: would be nice to add a high-level overview of the procedure used to extract the path

fedekunze (approved):
LGTM, pending changelog entry

modules/apps/transfer/types/trace.go
```diff
@@ -20,7 +20,10 @@ import (
 //
 // Examples:
 //
-// 	- "portidone/channelidone/uatom" => DenomTrace{Path: "portidone/channelidone", BaseDenom: "uatom"}
+// 	- "transfer/channelidone/uatom" => DenomTrace{Path: "transfer/channelidone", BaseDenom: "uatom"}
+//  - "transfer/channelidone/transfer/channelidtwo/uatom" => DenomTrace{Path: "transfer/channelidone/transfer/channelidtwo", BaseDenom: "uatom"}
+//  - "transfer/channelidone/gamm/pool/1" => DenomTrace{Path: "transfer/channelidone", BaseDenom: "gamm/pool/1"}
+//  - "gamm/pool/1" => DenomTrace{Path: "", BaseDenom: "gamm/pool/1"}
 // 	- "uatom" => DenomTrace{Path: "", BaseDenom: "uatom"}
```
damiannolan, line 27: nit: aligning

```suggestion
// - "transfer/channelidone/uatom" => DenomTrace{Path: "transfer/channelidone", BaseDenom: "uatom"}
// - "transfer/channelidone/transfer/channelidtwo/uatom" => DenomTrace{Path: "transfer/channelidone/transfer/channelidtwo", BaseDenom: "uatom"}
// - "transfer/channelidone/gamm/pool/1" => DenomTrace{Path: "transfer/channelidone", BaseDenom: "gamm/pool/1"}
// - "gamm/pool/1" => DenomTrace{Path: "", BaseDenom: "gamm/pool/1"}
// - "uatom" => DenomTrace{Path: "", BaseDenom: "uatom"}
```

modules/apps/transfer/types/trace.go
```diff
@@ -70,6 +74,22 @@ func (dt DenomTrace) GetFullDenomPath() string {
 	return dt.GetPrefix() + dt.BaseDenom
 }
 
+func extractPathAndBaseDenomFromDenomTrace(denomTraceItems []string) (string, string) {
+	path := []string{}
+	baseDenom := []string{}
```
damiannolan, line 79: more nits on declare vs assign, personally I prefer declare but its minor anyways, maybe we could benefit from a style guide at some point! 

```suggestion
	var path []string
	var baseDenom []string
```

modules/apps/transfer/types/trace.go
```diff
@@ -156,7 +178,13 @@ func ValidatePrefixedDenom(denom string) error {
 		return sdkerrors.Wrap(ErrInvalidDenomForTransfer, "base denomination cannot be blank")
 	}
 
-	identifiers := denomSplit[:len(denomSplit)-1]
+	path, _ := extractPathAndBaseDenomFromDenomTrace(denomSplit)
```
damiannolan, line 181: How come the base denom arg is ignored? Should we add a comment? Or is the second return value needed?

damiannolan (approved):
Looks good to go, I just left some nits and a quick question! Thanks 🙏

modules/apps/transfer/types/trace.go
```diff
@@ -70,6 +74,24 @@ func (dt DenomTrace) GetFullDenomPath() string {
 	return dt.GetPrefix() + dt.BaseDenom
 }
 
+// extractPathAndBaseDenomFromDenomTrace returns the trace path and the base denom from
+// the elements that constitute the complete denom trace.
+func extractPathAndBaseDenomFromDenomTrace(denomTraceItems []string) (string, string) {
```
AdityaSripal, line 79: nit: This can be named better since denom trace is just the path in the rest of the repo

Perhaps: `extractPathAndBaseFromFullDenom`

modules/apps/transfer/types/trace_test.go
```diff
@@ -14,10 +14,23 @@ func TestParseDenomTrace(t *testing.T) {
 	}{
 		{"empty denom", "", DenomTrace{}},
 		{"base denom", "uatom", DenomTrace{BaseDenom: "uatom"}},
+		{"base denom ending with '/'", "uatom/", DenomTrace{BaseDenom: "uatom/"}},
```
AdityaSripal, line 17: excellent testcases!

AdityaSripal (approved):

AdityaSripal:
> In the OnRecvPacket callback there is [a validation of the data in the FungibleTokenPacketData](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L190). The last step of this validation [calls ValidatePrefixedDenom](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/packet.go#L54) with the denom erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA. Inside ValidatePrefixedDenom the last step is [to validate the trace identifiers](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/trace.go#L159-L160). The input to validateTraceIdentifiers is going to be a string slice with only the element erc20. And then inside validateTraceIdentifiers I think an error is returned because [len(identifiers)%2 != 0](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/trace.go#L74-L76).

Yes in order for this fix to work correctly both sides must have the updated code.

Otherwise we enter into different scenarios depending on the denomination passed in:

1. If we have an even split like `erc20/0x85bcBCd7e79Ec36f4fBBDc54F90C643d921151AA`

Then we will fail as described by @crodriguezvega at packet data validation. The receiving chain will write an error ack, and then the sending chain will simply refund to the sender. This is not a bad outcome.

2. If we have an odd split like `gamm/pool/1`

Then we will pass packet validation but the token will be issued on the receiving chain with BaseDenom: `1` and Trace: `gamm/pool` which is incorrect. The internal representation of the token would then be `ibc/hash{gamm/pool/1}`

This is not ideal for the receiving chain since it may mess with client UX but it isn't catastrophic.

When the chains get moved back to the original sender, since it is updated it will recognize that `gamm/pool/1` is the base denom and appropriately unescrow funds.

When the funds go to other receiving chains, if they have the updated code then they will generate the trace and base denom correctly. If the other receiving chains don't have the update, they will also include `gamm/pool` to the end of the trace and have `1` in the base denom.

We should decide how to deal with the second scenario.

One option is to simply allow it and have some chains temporarily have the wrong trace information on their chains. This will only affect internal representations and users of that chain by giving them poor UX.

We can try to prevent slashed denoms from being received on chains that do not have this update. There is no great way to do this apart from creating middleware filters on updated sending chain that explicitly permission approved receiver chains that are known to have the update. This middleware can be added to until it is no longer necessary and can be removed. This would be an annoying mechanism to deal with, but is the cleanest one I can think of if we want to be very strict and avoid the problem all together.

fedekunze:
> We should decide how to deal with the second scenario.

@AdityaSripal IMO there should be an upgrade migration logic so that the denom hash is updated everywhere

AdityaSripal:
> IMO there should be an upgrade migration logic so that the denom hash is updated everywhere

Not sure I understand the proposal. Is the proposal that we allow slashed denoms to enter into all chains as is, and then once the upgrade happens we run migration logic to fix the traces on chains that did not upgrade in time?

modules/apps/transfer/types/trace.go
crodriguezvega: In this case the base denom is not needed at all because the only validation that happens here is for the trace path information.

crodriguezvega:
Running `chain1` with the fix and `chain2` without it, I tested as well with denom `gamm/pool/1` but the transfer fails on the receiving chain [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/trace.go#L83) because `pool` has [fewer than 8 characters](https://github.com/cosmos/ibc-go/blob/main/modules/core/24-host/validate.go#L79)...

So I tried with denom `gamm/poolpoolpool/1` to able to pass that validation and the transfer succeeded and on the receiving chain I can query for the denom trace and get the following:

```
denom_traces:
- base_denom: "1"
  path: transfer/channel-0/gamm/poolpoolpool
```  

I can transfer the tokens back to the sending chain and the denom is correctly unwinded.

AdityaSripal:
Great, thanks @crodriguezvega! So we can confirm the different behavior.

Since it gets correctly unwinded at the source and the trace information is purely local this isn't a critical issue.

I would say that we should release, and as Fede says, we should have a migration script that will update the trace information on receiver chains once they upgrade. Perhaps we can have that at some point after with one of the upcoming releases since it is more of a UX fix.

Note that in the interim we will have to live with the wrong trace information living on unupgraded receiver chains when they receive slashed denoms.

It looks to me like @fedekunze usecase of: `erc20/{coin}` will not be receivable anyway until the counterparty also upgrades so it is not as relevant an issue in evmos case

modules/apps/transfer/types/trace.go
```diff
@@ -70,6 +74,24 @@ func (dt DenomTrace) GetFullDenomPath() string {
...
+// extractPathAndBaseFromFullDenom returns the trace path and the base denom from
+// the elements that constitute the complete denom.
+func extractPathAndBaseFromFullDenom(fullDenomItems []string) (string, string) {
+	var path []string
+	var baseDenom []string
+	length := len(fullDenomItems)
+	for i := 0; i < length; i = i + 2 {
+		if i < length-1 && length > 2 && fullDenomItems[i] == PortID {
+			path = append(path, fullDenomItems[i], fullDenomItems[i+1])
+		} else {
```
colin-axner, line 86: Nice work @crodriguezvega! It should be noted this introduces undocumented changes. Specifically the usage of `PortID`. ICS20 does not specify the PortID that should be used for the transfer channel. See [specification callbacks](https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer#channel-lifecycle-management). This is also true in our [code](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go#L57). This code now assumes the PortID to be `transfer`. This should be enforced (in channel handshakes) if this the intended behaviour

I'm unaware of any chains which have changed the portID, but last I checked, Agoric used a modified portID in testing

AdityaSripal: Nice catch @colin-axner ! Unfortunately without a standardized port, it's impossible to correctly differentiate between what is trace info and what is the base denom. Since the trace is largely internal information for a chain and will not affect correctness, I think we can live with the fact that non-standard ports and their subsequent traces will be part of the base denom.

Reiterating for future viewers that this will not affect correctness of the protocol. But when querying trace information, any nonstandard trace will be in the base denom

PR #1518, opened 2022-06-09
Add slash migration guide

AdityaSripal:
## Description

Adds a migration guide for chain developers that are upgrading from a version that does not support slashed denominations to a version that does. Only necessary if the chain has malformed trace information in its state, but recommended to run for all upgrading chains regardless.

closes: #XXXX

---

write a little note why.

### DIFF
--- a/docs/migrations/support-slashed-denoms.md
+++ b/docs/migrations/support-slashed-denoms.md
@@ -1,4 +1,4 @@
-# Migrating from Unsupported Base Denom Slashes to Supporting Slashed BaseDenoms
+# Migrating from not supporing base denoms with slashes to supporting base denoms with slashes
 
 This document is intended to highlight significant changes which may require more information than presented in the CHANGELOG.
 Any changes that must be done by a user of ibc-go should be documented here.
@@ -9,24 +9,19 @@ There are four sections based on the four potential user groups of this document
 - Relayers
 - IBC Light Clients
 
-**Note:** ibc-go supports golang semantic versioning and therefore all imports must be updated to bump the version number on major releases.
-```go
-github.com/cosmos/ibc-go/v3 -> github.com/cosmos/ibc-go/v4
-```
+This document is necessary when chains are upgrading from a version that does not support base denoms with slashes (e.g. v3.0.0) to a version that does (e.g. v3.1.0). All versions of ibc-go smaller than v1.5.0 for the v1.x release line, v2.3.0 for the v2.x release line, and v3.1.0 for the v3.x release line do *NOT** support IBC token transfers of coins whose base denoms contain slashes. Therefore the in-place of genesis migration described in this document are required when upgrading.
 
-This document is necessary when chains are upgrading from a version that does not support slashed base denoms (e.g. v3.0.0) to a version that does (e.g. v3.1.0).
+If a chain receives coins of a base denom with slashes before it upgrades to supporting it, the receive may pass however the trace information will be incorrect.
 
-If a chain receives a slashed denom before it upgrades to supporting it, the receive may pass however the trace information will be incorrect.
+E.g. If a base denom of `testcoin/testcoin/testcoin` is sent to a chain that does not support slashes in the base denom, the receive will be successful. However, the trace information stored on the receiving chain will be: `Trace: "transfer/{channel-id}/testcoin/testcoin", BaseDenom: "testcoin"`.
 
-E.g. If a base denom of `testcoin/testcoin/testcoin` is sent to a chain that does not support slashes in the base denom; the receive will be successful. However, the trace information stored on the receiving chain will be: `Trace: "testcoin/testcoin", BaseDenom: "testcoin"`
+This incorrect trace information must be corrected when the chain does upgrade to fully supporting denominations with slashes.
 
-This incorrect trace information must be corrected when the chain does upgrade to fully supporting slashed denominations.
-
-To do so, chain binaries should include a migration script that will run when the chain upgrades from not supporting slashed base denominations to supporting slashed base denominations.
+To do so, chain binaries should include a migration script that will run when the chain upgrades from not supporting base denominations with slashes to supporting base denominations with slashes.
 
 ## Chains
 
-### Transfer
+### ICS20 - Transfer
 
 The transfer module will now support slashes in base denoms, so we must iterate over current traces to check if any of them are incorrectly formed and correct the trace information.
 
@@ -34,33 +29,36 @@ The transfer module will now support slashes in base denoms, so we must iterate 
 
 ```go
 // Here the upgrade name is the upgrade name set by the chain
-app.UpgradeKeeper.SetUpgradeHandler("supportSlashingDenomUpgrade",
+app.UpgradeKeeper.SetUpgradeHandler("supportSlashedDenomsUpgrade",
     func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
         // list of traces that must replace the old traces in store
-        var newTraces []transfertypes.DenomTrace
+        var newTraces []ibctransfertypes.DenomTrace
+        app.TransferKeeper.IterateDenomTraces(ctx,
+            func(dt ibctransfertypes.DenomTrace) bool {
+                // check if the new way of splitting FullDenom
+                // into Trace and BaseDenom passes validation and
+                // is the same as the current DenomTrace.
+                // If it isn't then store the new DenomTrace in the list of new traces.
+                newTrace := ibctransfertypes.ParseDenomTrace(dt.GetFullDenomPath())
+                if err := newTrace.Validate(); err == nil && !reflect.DeepEqual(newTrace, dt) {
+                    newTraces = append(newTraces, newTrace)
+                }
 
-        transferKeeper.IterateDenomTraces(ctx,
-        func(dt transfertypes.DenomTrace) bool {
-            // check if the new way of splitting FullDenom
-            // into Trace and BaseDenom is the same as the current
-            // DenomTrace.
-            // If it isn't then store the new DenomTrace in the list of new traces.
-            newTrace := transfertypes.ParseDenomTrace(dt.GetFullDenomPath())
-
-            if !reflect.DeepEqual(newTrace, dt) {
-                append(newTraces, newTrace)
-            }
-        })
+                return false
+            })
 
         // replace the outdated traces with the new trace information
         for _, nt := range newTraces {
-            transferKeeper.SetDenomTrace(ctx, nt)
+            app.TransferKeeper.SetDenomTrace(ctx, nt)
         }
-    }
-)
+
+        return app.mm.RunMigrations(ctx, app.configurator, fromVM)
+    })
 ```
 
-This is only necessary if there are DenomTraces in the store with incorrect trace information from previously received coins that had a slash in the base denom. However, it is recommended that any chain upgrading to support slashed denominations runs this code for safety.
+This is only necessary if there are denom traces in the store with incorrect trace information from previously received coins that had a slash in the base denom. However, it is recommended that any chain upgrading to support base denominations with slashes runs this code for safety.
+
+For a more detailed sample, please check out the code changes in [this pull request](https://github.com/cosmos/ibc-go/pull/1527).
 
 ### Genesis Migration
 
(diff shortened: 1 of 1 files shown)


docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,113 @@
...
+## Chains
+
+### Transfer
+
+The transfer module will now support slashes in base denoms, so we must iterate over current traces to check if any of them are incorrectly formed and correct the trace information.
+
+### Upgrade Propsoal
+
+```go
+app.UpgradeKeeper.SetUpgradeHandler("v3.1.0",
```
fedekunze, line 36: I think it's important to mention that this version is the one from the chain that is importing IBC-go

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,113 @@
...
+
+To do so, chain binaries should include a migration script that will run when the chain upgrades from not supporting slashed base denominations to supporting slashed base denominations.
+
+## Chains
+
+### Transfer
+
+The transfer module will now support slashes in base denoms, so we must iterate over current traces to check if any of them are incorrectly formed and correct the trace information.
+
+### Upgrade Propsoal
```
fedekunze, line 33:
```suggestion
### Upgrade Proposal
```

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,113 @@
...
+if upgradeInfo.Name == "supportSlashingDenomUpgrade" && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height)  {
+    storeUpgrades := store.StoreUpgrades{
+        Added: []string{transfertypes.StoreKey}
+    }
+
+    app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
+}
+```
+
+This ensures that the transfer module's stores are added to the multistore before the migrations begin.
```
fedekunze, line 79: isn't the `Added` meant for new modules only? cc @alexanderbez

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,113 @@
...
+- IBC Apps
+- Relayers
+- IBC Light Clients
+
+**Note:** ibc-go supports golang semantic versioning and therefore all imports must be updated to bump the version number on major releases.
+```go
+github.com/cosmos/ibc-go/v3 -> github.com/cosmos/ibc-go/v4
+```
+
+This document is necessary when chains are upgrading from a version that does not support slashed base denoms (e.g. v3.0.0) to a version that does (e.g. v3.1.0).
```
fedekunze, line 17: q: shouldn't this be v4.0.0 if you are bumping the go version?

docs/migrations/support-slashed-denoms.md
AdityaSripal: Ahh ok, yes this might be the case. I copied from the ICA example. Please let me know if it incorrect @alexanderbez and I will remove

docs/migrations/support-slashed-denoms.md
AdityaSripal: I believe the plan is to do a minor release and backport to v1 and v2. cc: @crodriguezvega

AdityaSripal: https://github.com/cosmos/ibc-go/blob/main/RELEASES.md

docs/migrations/support-slashed-denoms.md
AdityaSripal: Reading through the docs more carefully. You are indeed correct.

docs/migrations/support-slashed-denoms.md
crodriguezvega: Yes, the idea is to release this fix in v3.1.0, but also as a minor backport in v2.3.0 and v1.5.0, so that chains not yet on v3 can still benefit from the fix. I think this is the fastest way to get this fix out. 

Alternatively, we could release this fix only with v4.0.0. That would probably make it easier for us to communicate this fix, because we can add this migration guide to the v3-to-v4 migration guide, and all chains migrating to v4.0.0 will need to run this migration. But v4.0.0 needs to come out with fee middleware and it could still take easily another 3 weeks before we cut the final release. So the viability of this option depends on how soon Evmos needs to have this fix.

Having a quick look at mintscan and it looks like most of the chains are still on v2, so if we release this only on v4 and those chains would like to benefit from the fix, then we are forcing them to run the migrations to upgrade to v3 and v4, which might be a bit too much to ask?

Does Evmos expect their slashed-denom tokens to be transferred to a particular set of chains or they could just go to any chain? Because if most likely the receiving chains of those tokens are going to be a handful of chains, maybe we could coordinate with an upgrade to v4?

@fedekunze @AdityaSripal what of these two options you think it's best?

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,97 @@
...
+
+If a chain receives a slashed denom before it upgrades to supporting it, the receive may pass however the trace information will be incorrect.
+
+E.g. If a base denom of `testcoin/testcoin/testcoin` is sent to a chain that does not support slashes in the base denom; the receive will be successful. However, the trace information stored on the receiving chain will be: `Trace: "testcoin/testcoin", BaseDenom: "testcoin"`
```
crodriguezvega, line 21: The trace information stored will actually be `transfer/{channel-id}/testcoin/testcoin`.

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,97 @@
...
+
+There are four sections based on the four potential user groups of this document:
+- Chains
+- IBC Apps
+- Relayers
+- IBC Light Clients
+
+**Note:** ibc-go supports golang semantic versioning and therefore all imports must be updated to bump the version number on major releases.
+```go
+github.com/cosmos/ibc-go/v3 -> github.com/cosmos/ibc-go/v4
```
crodriguezvega, line 14: Maybe we should remove this part to avoid confusion...

crodriguezvega (commented):
Thank so much, @AdityaSripal, for writing this migration guide!

I can do some testing of these migrations in the next couple of days.

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,97 @@
...
+            subsituteTraces[i] = newTrace
+        }
+
+        transferGenState.Traces = substituteTraces
+
+        // delete old genesis state
+		delete(appState, transfertypes.ModuleName)
+
+        // set new ibc transfer genesis state
+		appState[transfertypes.ModuleName] = clientCtx.JSONCodec.MustMarshalJSON(transferGenState)
```
damiannolan, line 92: nit: indentation

```suggestion
        // delete old genesis state
	delete(appState, transfertypes.ModuleName)

        // set new ibc transfer genesis state
	appState[transfertypes.ModuleName] = clientCtx.JSONCodec.MustMarshalJSON(transferGenState)
```

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,97 @@
...
+
+            if !reflect.DeepEqual(newTrace, dt) {
+                append(newTraces, newTrace)
+            }
+        })
+
+        // replace the outdated traces with the new trace information
+        for _, nt := range newTraces {
+            transferKeeper.SetDenomTrace(ctx, nt)
+        }
```
damiannolan, line 58: Is it worth adding a call to [`Validate`](https://github.com/cosmos/ibc-go/blob/carlos/1446-bugtransfer-cannot-send-namespaced-ibc-denom-using-slashes/modules/apps/transfer/types/trace.go#L113) on new traces?

docs/migrations/support-slashed-denoms.md
damiannolan: Maybe a asterisk note with special case for this release. I think this is part of the migrations template right?

damiannolan (approved):
Nice! Guide looks pretty much good to me, left a nit or two, pretty minor!

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Transfer
+
+The transfer module will now support slashes in base denoms, so we must iterate over current traces to check if any of them are incorrectly formed and correct the trace information.
+
+### Upgrade Proposal
+
+```go
+// Here the upgrade name is the upgrade name set by the chain
+app.UpgradeKeeper.SetUpgradeHandler("supportSlashingDenomUpgrade",
```
crodriguezvega, line 37: The code needed some adjustments to make it compile in the ibc-go simapp:

```golang
app.UpgradeKeeper.SetUpgradeHandler("supportSlashingDenomUpgrade",
	func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
		// list of traces that must replace the old traces in store
		var newTraces []ibctransfertypes.DenomTrace
		app.TransferKeeper.IterateDenomTraces(ctx,
			func(dt ibctransfertypes.DenomTrace) bool {
				// check if the new way of splitting FullDenom
				// into Trace and BaseDenom is the same as the current
				// DenomTrace.
				// If it isn't then store the new DenomTrace in the list of new traces.
				newTrace := ibctransfertypes.ParseDenomTrace(dt.GetFullDenomPath())
				if !reflect.DeepEqual(newTrace, dt) {
					newTraces = append(newTraces, newTrace)
				}
				return false
			})

		// replace the outdated traces with the new trace information
		for _, nt := range newTraces {
			app.TransferKeeper.SetDenomTrace(ctx, nt)
		}

		return app.mm.RunMigrations(ctx, app.configurator, fromVM)
	})
```

docs/migrations/support-slashed-denoms.md
danburck: @crodriguezvega The idea is that slashed-denom tokens are supported on any cosmos chain and that there is no specific set of chains. 

As this functionality is blocking the adoption of our `x/erc20` module, it would be great to go for the first option and release it in v3.1.0, is it is the faster option and requires less coordination.

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,97 @@
...
+This is only necessary if there are DenomTraces in the store with incorrect trace information from previously received coins that had a slash in the base denom. However, it is recommended that any chain upgrading to support slashed denominations runs this code for safety.
+
+### Genesis Migration
+
+If the chain chooses to add support for slashes in base denoms via genesis export, then the trace information must be corrected during genesis migration.
+
+The migration code required may look like:
+
+```go
+func MigrateGenesis(appState genutiltypes.AppMap, clientCtx client.Context, genDoc tmtypes.GenesisDoc) (genutiltypes.AppMap, error) {
```
crodriguezvega, line 72: This code also needed some adjustments to compile:

```go
func MigrateGenesis(appState genutiltypes.AppMap, clientCtx client.Context, genDoc *tmtypes.GenesisDoc) (genutiltypes.AppMap, error) {
	if appState[ibctransfertypes.ModuleName] != nil {
		transferGenState := &ibctransfertypes.GenesisState{}
		clientCtx.Codec.MustUnmarshalJSON(appState[ibctransfertypes.ModuleName], transferGenState)

		substituteTraces := make([]ibctransfertypes.DenomTrace, len(transferGenState.DenomTraces))
		for i, dt := range transferGenState.DenomTraces {
			// replace all previous traces with the latest trace
			// note most traces will have same value
			newTrace := ibctransfertypes.ParseDenomTrace(dt.GetFullDenomPath())

			substituteTraces[i] = newTrace
		}

		transferGenState.DenomTraces = substituteTraces

		// delete old genesis state
		delete(appState, ibctransfertypes.ModuleName)

		// set new ibc transfer genesis state
		appState[ibctransfertypes.ModuleName] = clientCtx.Codec.MustMarshalJSON(transferGenState)
	}

	return appState, nil
}
```

crodriguezvega: I haven't been able to test this migration because the [app state export panics](https://github.com/cosmos/cosmos-sdk/issues/12227).

docs/migrations/support-slashed-denoms.md
crodriguezvega: I tested the upgrade handler locally and it works. 

Before migration:
```
denom_traces:
- base_denom: "1"
  path: transfer/channel-0/gamm/poolpoolpool
- base_denom: "1"
  path: transfer/channel-0/transfer/channel-1/gamm/testtesttest
pagination:
  next_key: null
  total: "0"
```

After migration:
```
denom_traces:
- base_denom: gamm/poolpoolpool/1
  path: transfer/channel-0
- base_denom: gamm/testtesttest/1
  path: transfer/channel-0/transfer/channel-1
pagination:
  next_key: null
  total: "0"  
```

docs/migrations/support-slashed-denoms.md
crodriguezvega: Thanks for the feedback, @danburck! Yes, I think that it's the option that makes most sense. I will try to release is asap!

docs/migrations/support-slashed-denoms.md
crodriguezvega: I did a trick (I skipped the app state export, but instead I manually modified the genesis.json file to add malformed denom traces) and I was able to test this migration functionality. It also works as expected.

docs/migrations/support-slashed-denoms.md
crodriguezvega: Maybe we could add it in `if !reflect.DeepEqual(newTrace, dt)` as an extra condition. But if the validation fails, should we keep then  the existing denom trace?

crodriguezvega: Yeah, I think we can add the `Validate` on that `if` condition and if it doesn't pass, then we just leave that denom trace as it is...

docs/migrations/support-slashed-denoms.md
```diff
@@ -0,0 +1,101 @@
...
+        // list of traces that must replace the old traces in store
+        var newTraces []ibctransfertypes.DenomTrace
+        app.TransferKeeper.IterateDenomTraces(ctx,
+            func(dt ibctransfertypes.DenomTrace) bool {
+                // check if the new way of splitting FullDenom
+                // into Trace and BaseDenom passes validation and
+                // is the same as the current DenomTrace.
+                // If it isn't then store the new DenomTrace in the list of new traces.
+                newTrace := ibctransfertypes.ParseDenomTrace(dt.GetFullDenomPath())
+                if err := newTrace.Validate(); err == nil && !reflect.DeepEqual(newTrace, dt) {
```
damiannolan, line 43: Quick question, with the additionof the `reflect.DeepEqual` check, doesn't this mean that non-malformed traces that were previously in the store will be essentially discard here? 
Do we need the `reflect.DeepEqual` check?

This logic seems to differ from the Genesis Migration below.

damiannolan: Ah apologies, we're not setting the whole list and overriding, its looping below and setting individually

PR #1550, opened 2022-06-16
chore: allow ics27 to select and return default JSON encoded metadata

damiannolan:
## Description

Allow interchain accounts to accept an empty channel version and select a default.

- Adding `version` return to `OnChanOpenInit` cb handler in controller
- Updating test cases

closes: #1551 

---

write a little note why.


modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -58,15 +67,15 @@ func (k Keeper) OnChanOpenInit(
...
 		}
 
 		if !icatypes.IsPreviousMetadataEqual(channel.Version, metadata) {
-			return sdkerrors.Wrap(icatypes.ErrInvalidVersion, "previous active channel metadata does not match provided version")
+			return "", sdkerrors.Wrap(icatypes.ErrInvalidVersion, "previous active channel metadata does not match provided version")
 		}
 	}
 
-	return nil
+	return string(icatypes.ModuleCdc.MustMarshalJSON(&metadata)), nil
```
chatton, line 78: just to make sure I understand, if this panics this is handled at the sdk level right?

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -28,26 +28,35 @@ func (k Keeper) OnChanOpenInit(
...
 	var metadata icatypes.Metadata
-	if err := icatypes.ModuleCdc.UnmarshalJSON([]byte(version), &metadata); err != nil {
-		return sdkerrors.Wrapf(icatypes.ErrUnknownDataType, "cannot unmarshal ICS-27 interchain accounts metadata")
+	if strings.TrimSpace(version) == "" {
+		connection, err := k.channelKeeper.GetConnection(ctx, connectionHops[0])
+		if err != nil {
+			return "", err
+		}
+
+		metadata = icatypes.NewDefaultMetadata(connectionHops[0], connection.GetCounterparty().GetConnectionID())
```
chatton, line 51: I assume it will not be possible to have a connectionHops that is empty at this stage right?

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -28,26 +28,35 @@ func (k Keeper) OnChanOpenInit(
 	chanCap *capabilitytypes.Capability,
 	counterparty channeltypes.Counterparty,
 	version string,
-) error {
+) (string, error) {
```
chatton, line 31: what is the implication of changing an `OnChanOpenInit` signature like this? Are we free to do this without causing any issues?

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
```diff
@@ -50,7 +50,8 @@ func (im IBCMiddleware) OnChanOpenInit(
 		return "", types.ErrControllerSubModuleDisabled
 	}
 
-	if err := im.keeper.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version); err != nil {
+	version, err := im.keeper.OnChanOpenInit(ctx, order, connectionHops, portID, channelID, chanCap, counterparty, version)
```
chatton, line 53: nice readability improvement 🥇

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -28,26 +28,35 @@ func (k Keeper) OnChanOpenInit(
...
-		return sdkerrors.Wrapf(icatypes.ErrUnknownDataType, "cannot unmarshal ICS-27 interchain accounts metadata")
+	if strings.TrimSpace(version) == "" {
+		connection, err := k.channelKeeper.GetConnection(ctx, connectionHops[0])
+		if err != nil {
+			return "", err
+		}
+
+		metadata = icatypes.NewDefaultMetadata(connectionHops[0], connection.GetCounterparty().GetConnectionID())
+	} else {
+		if err := icatypes.ModuleCdc.UnmarshalJSON([]byte(version), &metadata); err != nil {
```
chatton, line 53: [nit] rather than having nested if/else clauses it might be cleaner to have a helper function to return the metadata (using early returns)

chatton (commented):
Looks great! Just had a few questions/comments 👍

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
damiannolan: Yep, its caught by the grpc recovery interceptor (essentially a middleware handler in grpc land)

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
damiannolan: Yep, correct! 

https://github.com/cosmos/ibc-go/blob/31e74552a73817bb12ceb89bdf67cce219f67ca6/modules/core/04-channel/keeper/handshake.go#L30-L34

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
damiannolan: I took a look at this and feel like there's not a massive difference making a helper function. You'd either need to return a new metadata type or take a pointer and a bunch of other args too... I'm pretty indifferent about it

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
seantking: to add to that: in general, we would also update the ibc spec for changes to these fn sigs. But like @damiannolan said the calling func already has this signature.

seantking (approved):
LGTM! Nice work. 

Should we test this out manually as well?

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -28,26 +28,35 @@ func (k Keeper) OnChanOpenInit(
...
-		return sdkerrors.Wrapf(icatypes.ErrInvalidHostPort, "expected %s, got %s", icatypes.PortID, counterparty.PortId)
+		return "", sdkerrors.Wrapf(icatypes.ErrInvalidHostPort, "expected %s, got %s", icatypes.PortID, counterparty.PortId)
 	}
 
 	var metadata icatypes.Metadata
-	if err := icatypes.ModuleCdc.UnmarshalJSON([]byte(version), &metadata); err != nil {
-		return sdkerrors.Wrapf(icatypes.ErrUnknownDataType, "cannot unmarshal ICS-27 interchain accounts metadata")
+	if strings.TrimSpace(version) == "" {
+		connection, err := k.channelKeeper.GetConnection(ctx, connectionHops[0])
+		if err != nil {
```
charleenfei, line 47: just a super tiny nit but you have the err statement in the second else clause in the  `; err != nil` setup, would be nice if they were both like this or both the other way just as a tiny remark on style

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -28,26 +28,35 @@ func (k Keeper) OnChanOpenInit(
...
 	if counterparty.PortId != icatypes.PortID {
-		return sdkerrors.Wrapf(icatypes.ErrInvalidHostPort, "expected %s, got %s", icatypes.PortID, counterparty.PortId)
+		return "", sdkerrors.Wrapf(icatypes.ErrInvalidHostPort, "expected %s, got %s", icatypes.PortID, counterparty.PortId)
 	}
 
 	var metadata icatypes.Metadata
-	if err := icatypes.ModuleCdc.UnmarshalJSON([]byte(version), &metadata); err != nil {
-		return sdkerrors.Wrapf(icatypes.ErrUnknownDataType, "cannot unmarshal ICS-27 interchain accounts metadata")
+	if strings.TrimSpace(version) == "" {
+		connection, err := k.channelKeeper.GetConnection(ctx, connectionHops[0])
```
charleenfei, line 46: why is connectionHops hardcoded to the first value of the array?

seantking: The first value is the source chain connection end

damiannolan: ConnectionHops is an ordered list of connection ids, so using `connectionHops[0]` here takes the first connection hop through which packets travel. There's a bunch of defensive checks in ibc core ensuring that this is always of length one at the moment. So on source chain `connectionHops` will contain the connection ID of self, in this instance its the controller chain. And likewise, the counterparty handlers would have a `connectionHops` array containing the connection of themself (host). Obviously in future with multi hop the semantics will change and ibc core will have to adapt for it. But due to the ordering, we can safely assume that the 0 index element is self.

Channel spec says:

> The connectionHops stores the list of connection identifiers, in order, along which packets sent on this channel will travel. At the moment this list must be of length 1. In the future multi-hop channels may be supported.

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
damiannolan: But we need to use the `connection` below outside the scope of the returned error.
Generally we use the below for function return sigs with one single `error` return:
```go
if err := invokeFunc(); err != nil {
    // handle err
}
```

If we use the following then we can't use the connection below, outside the scope of the conditional:
```go
if connection, err := k.channelKeeper.GetConnection(ctx, connectionHops[0]); err != nil {
   // connection and err are only available at this level of scope
}
```

chatton: you should technically be able to if you put the usage in an else clause, but I personally prefer the way you have it now rather than adding additional `else` clauses!

crodriguezvega (approved):
Awesome work, @damiannolan!

I just left a suggestion to potentially remove some duplicate code...

Issue #1648, opened 2022-07-04
Add Baseline E2E Test Suite type (E2E #3)


chatton:

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1650, opened 2022-07-04
Add E2ETestSuite Type (E2E #3)

chatton:
## Description

~Should be reviewed after #1646  is merged~

This PR adds the `E2ETestSuite` type. This type has helper functions that interact with the `ibctest` libraries. All of our test suites should be composed of this type.

closes: #1648

---

write a little note why.


.github/scripts/determine_simd_tag.go
```diff
@@ -3,32 +3,24 @@ package main
 import (
```
chatton, line 3: I had misunderstood the behaviour of the ref/pr values in GithubActions. I simplified this script to simply accept a PR number as a string, or use "main" if we are not in a PR.

e2e/fee_middleware_test.go
```diff
@@ -1,29 +1,37 @@
...
 func TestFeeMiddlewareTestSuite(t *testing.T) {
 	suite.Run(t, new(FeeMiddlewareTestSuite))
 }
 
 type FeeMiddlewareTestSuite struct {
-	suite.Suite
+	testsuite.E2ETestSuite
 }
 
 func (s *FeeMiddlewareTestSuite) TestPlaceholder() {
```
chatton, line 21: the place holder test now creates chains in a container and starts a relayer (no assertions are made yet!)

chatton:
@charleenfei @colin-axner 

I plan on applying the changes we discussed in the call today in a followup PR.

Dockerfile
```diff
@@ -20,4 +20,5 @@ FROM ubuntu:20.04
 
 COPY --from=builder /go/build/simd /bin/simd
 
-ENTRYPOINT ["simd"]
+# TODO(chatton): uncomment once https://github.com/strangelove-ventures/ibctest/issues/183 is resolved.
```
colin-axner, line 23: has this been resolved? I see the linked issue is closed

e2e/fee_middleware_test.go
```diff
@@ -1,29 +1,37 @@
...
+}
 
-	s.T().Logf("Placeholder test")
-	s.Require().True(true)
+// feeMiddlewareChannelOptions configures both of the chains to have fee middleware enabled.
+func feeMiddlewareChannelOptions() func(options *ibc.CreateChannelOptions) {
+	return func(opts *ibc.CreateChannelOptions) {
+		opts.Version = "{\"fee_version\":\"ics29-1\",\"app_version\":\"ics20-1\"}"
+		opts.DestPortName = "transfer"
+		opts.SourcePortName = "transfer"
```
colin-axner, line 35: nit: rename to avoid `dest`/`source` naming. Maybe in a followup issue?

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/stretchr/testify/suite"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
+
+	"github.com/cosmos/ibc-go/v4/e2e/testconfig"
+)
+
+const (
+	SourceRelayerName      = "srcRly"
+	DestinationRelayerName = "dstRly"
```
colin-axner, line 26: add godoc?

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+	s.chainPairs[s.T().Name()] = cp
+
+	return cp.srcChain, cp.dstChain
+}
+
+// CreateChainsRelayerAndChannel create two chains, a relayer, establishes a connection and creates a channel
+// using the given channel options. The relayer returned by this function has not yet started. It should be started
+// with E2ETestSuite.StartRelayer if needed.
+// This should be called at the start of every test, unless fine grained control is required.
+func (s *E2ETestSuite) CreateChainsRelayerAndChannel(ctx context.Context, channelOpts ...func(*ibc.CreateChannelOptions)) ibc.Relayer {
```
colin-axner, line 85: I'd rename to `Setup`. I also suggest making this the first function after type declarations

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+	chainPairs     map[string]chainPair
+	logger         *zap.Logger
+	Client         *dockerclient.Client
+	network        string
+	startRelayerFn func(relayer ibc.Relayer)
+}
+
+// chainPair is a pairing of two chains which will be used in a test.
+type chainPair struct {
+	srcChain, dstChain *cosmos.CosmosChain
```
colin-axner, line 41: I presume there is a difficulty with making this ibc.Chain?

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+// ChainOptionConfiguration enables arbitrary configuration of ChainOptions.
+type ChainOptionConfiguration func(options *ChainOptions)
+
+// GetChains returns a src and dst chain that can be used in a test. The pair returned
+// is unique to the current test being run. Note: this function does not create containers.
+func (s *E2ETestSuite) GetChains(chainOpts ...ChainOptionConfiguration) (*cosmos.CosmosChain, *cosmos.CosmosChain) {
+	if s.chainPairs == nil {
+		s.chainPairs = map[string]chainPair{}
+	}
+	cp, ok := s.chainPairs[s.T().Name()]
```
colin-axner, line 61: I have a mild preference for avoiding one/two letter variable naming unless it is contained within a for loop. I think I'd prefer to just use `chainPair`

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+	srcChain, dstChain := s.GetChains()
+	srcRelayWallet, ok := relayer.GetWallet(srcChain.Config().ChainID)
+	if !ok {
+		return ibc.RelayerWallet{}, ibc.RelayerWallet{}, fmt.Errorf("unable to find source chain relayer wallet")
+	}
+	dstRelayWallet, ok := relayer.GetWallet(dstChain.Config().ChainID)
+	if !ok {
+		return ibc.RelayerWallet{}, ibc.RelayerWallet{}, fmt.Errorf("unable to find destination chain relayer wallet")
+	}
+	return srcRelayWallet, dstRelayWallet, nil
```
colin-axner, line 148:
```suggestion
func (s *E2ETestSuite) GetRelayerWallets(relayer ibc.Relayer) (ibc.RelayerWallet, ibc.RelayerWallet, error) {
	srcChain, dstChain := s.GetChains()
	srcRelayWallet, ok := relayer.GetWallet(srcChain.Config().ChainID)
	if !ok {
		return ibc.RelayerWallet{}, ibc.RelayerWallet{}, fmt.Errorf("unable to find source chain relayer wallet")
	}
	
	dstRelayWallet, ok := relayer.GetWallet(dstChain.Config().ChainID)
	if !ok {
		return ibc.RelayerWallet{}, ibc.RelayerWallet{}, fmt.Errorf("unable to find destination chain relayer wallet")
	}
	
	return srcRelayWallet, dstRelayWallet, nil
```

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/cosmos/ibc-go/v4/e2e/testconfig"
+)
+
+const (
+	SourceRelayerName      = "srcRly"
+	DestinationRelayerName = "dstRly"
+)
+
+// E2ETestSuite has methods and functionality which can be shared among all test suites.
+type E2ETestSuite struct {
```
colin-axner, line 30: so the idea is that this suite is used for a number of test cases? Each test case has a chain pair (thus keeping them logically isolated). Each pair is mapped via the test name?

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+	SrcChainConfig *ibc.ChainConfig
+	DstChainConfig *ibc.ChainConfig
+}
+
+// ChainOptionConfiguration enables arbitrary configuration of ChainOptions.
+type ChainOptionConfiguration func(options *ChainOptions)
+
+// GetChains returns a src and dst chain that can be used in a test. The pair returned
+// is unique to the current test being run. Note: this function does not create containers.
+func (s *E2ETestSuite) GetChains(chainOpts ...ChainOptionConfiguration) (*cosmos.CosmosChain, *cosmos.CosmosChain) {
```
colin-axner, line 57: I have a mild preference for not initializing the chains within this function. I think I'd prefer a function which initializes the chains and then this function would return the chains (or error/panic if they chains haven't been created yet)

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return nil
+}
+
+// StartRelayer starts the given relayer.
+func (s *E2ETestSuite) StartRelayer(relayer ibc.Relayer) {
+	if s.startRelayerFn == nil {
+		panic("cannot start relayer before it is created!")
+	}
+	s.startRelayerFn(relayer)
```
colin-axner, line 175:
```suggestion
func (s *E2ETestSuite) StartRelayer(relayer ibc.Relayer) {
	if s.startRelayerFn == nil {
		panic("cannot start relayer before it is created!")
	}
	
	s.startRelayerFn(relayer)
```

I like adding extra spacing in these cases, just to separate the code based on error (in this case a nil startRelayerFn) and the code based on a successful executrion (starting the relayer)

colin-axner (commented):
Excellent work @chatton! Reviewed mostly everything (still have the second half of testsuite.go). Left a lot of suggestions. Always open to opposing viewpoints on these suggestions. I realize some of the comments I suggested might require changes to ibctest, but I'd prefer to open issues and request the changes in these cases instead of leaving them as is

Will continue my review when I get a chance. If you feel any of my requested changes require a non-trivial amount of work, please feel free to open an issue for them. I find it easier to make progress in such situations

e2e/fee_middleware_test.go
chatton: This one is trickier as it is part of the ibctest library.

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
+package testsuite
```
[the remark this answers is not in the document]
chatton: I was planning on doing these renamings in a followup but I think it's actually better to do them here, I'll make these changes 👍

e2e/testsuite/testsuite.go
chatton: I tend to use short variable names when the scope is limited (loops as you mentioned or small functions). I don't feel very strongly about it, so I'm happy to rename to `chainPair` in this case.

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,286 @@
...
+	suite.Suite
+	chainPairs     map[string]chainPair
+	logger         *zap.Logger
+	Client         *dockerclient.Client
+	network        string
+	startRelayerFn func(relayer ibc.Relayer)
+}
+
+// chainPair is a pairing of two chains which will be used in a test.
+type chainPair struct {
```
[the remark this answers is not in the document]
chatton: I'm happy to rename this type to be called `path` if it's more consistent with what we already have.

e2e/testsuite/testsuite.go
chatton: That's right, in any given test we are expected to running a single Suite::Test function in isolation. Each will have its own pair of chains.

e2e/testsuite/testsuite.go
chatton: Ah maybe this wasn't clear but the `createCosmosChains` function does not create an containers, it simply creates the structs. The chains themselves are created in `CreateChainsRelayerAndChannel`. The naming here isn't quite clear I think.

e2e/testsuite/testsuite.go
chatton: In certain places we depend on the concrete type `*cosmos.CosmosChain`. Not in this PR but in the follow up tests. I'll make a note of this to verify that this can't just be `ibc.Chain`.

e2e/fee_middleware_test.go
colin-axner: Is there already an issue created? Would you like me to open one?

e2e/testsuite/testsuite.go
damiannolan: Agree with both, I usually prefer descriptive/informative names. For example `rly` over `r` says much more to me. 
But in general it depends on the scope and lifespan of the variable, for small loops and functions I feel fine about one or two letter namings. In my opinion its usually fine if the type/variable is inferred from the function name that was invoked too. The exception is func receivers, I usually prefer short-hand for them.

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+}
+
+// SetupChainsRelayerAndChannel create two chains, a relayer, establishes a connection and creates a channel
+// using the given channel options. The relayer returned by this function has not yet started. It should be started
+// with E2ETestSuite.StartRelayer if needed.
+// This should be called at the start of every test, unless fine grained control is required.
+func (s *E2ETestSuite) SetupChainsRelayerAndChannel(ctx context.Context, channelOpts ...func(*ibc.CreateChannelOptions)) ibc.Relayer {
+	chainA, chainB := s.GetChains()
+	home, err := ioutil.TempDir("", "")
```
damiannolan, line 58: Should we be using empty strings here? Is this the chain home directory?

e2e/testsuite/testsuite.go
```diff
@@ -0,0 +1,242 @@
...
+
+	path, ok := s.paths[s.T().Name()]
+	if ok {
+		return path.chainA, path.chainB
+	}
+
+	chainOptions := testconfig.DefaultChainOptions()
+	for _, opt := range chainOpts {
+		opt(&chainOptions)
+	}
```
damiannolan, line 123: I'm a big fan of the functional opts pattern 👍

e2e/testsuite/testsuite.go
chatton: with empty strings we will just have a random string generated for the home directory. After this we don't need to interact with the home dir directly.

damiannolan (approved):
Awesome job on this, excited to see the e2e suite come together! Super clean work 🚀

e2e/testsuite/testsuite.go
colin-axner: oh whoops, I meant to tag you on this [comment](https://github.com/cosmos/ibc-go/pull/1650#discussion_r919963422) @damiannolan. Agree with your response though

colin-axner (approved):
Fantastic work! Thanks for pushing all of this forward!

e2e/fee_middleware_test.go
colin-axner: Yea you are right. Do you think it is too much work to request changes for? 

In the past I've spent a fair amount of time debugging the difference between source/destination, that it seems worthwhile to me to consider changing naming conventions if a library is going to have increased usage. Happy to let the maintainers decide if they want to do the changes, but the "source" and "destination" terminology doesn't make much sense when packet flow is bidirectional

chatton: looking at the ibctest code I see the following

```golang
func (commander) CreateChannel(pathName string, opts ibc.CreateChannelOptions, homeDir string) []string {
	return []string{
		"rly", "tx", "channel", pathName,
		"--src-port", opts.SourcePortName,
		"--dst-port", opts.DestPortName,
		"--order", opts.Order.String(),
		"--version", opts.Version,
		"--home", homeDir,
	}
}
```

It looks like the field names are mappings to the go relayer arguments for channel creation so in this case I think the name makes sense.

I definitely agree that overall we shouldn't use `src/dst` for our chain names.

Issue #1542, opened 2022-06-14
Make enabling slashes on transfers a one-liner


ValarDragon:
## Summary

Lets make the upgrade handler code for chains to upgrade to support upon switching to IBC v3.1.0 a one-liner, that they import from the app-transfer module.

## Problem Definition /  Proposal

The code for the upgrade is currently including the following in your upgrade struct:
```golang
        var newTraces []ibctransfertypes.DenomTrace
        app.TransferKeeper.IterateDenomTraces(ctx,
            func(dt ibctransfertypes.DenomTrace) bool {
                // check if the new way of splitting FullDenom
                // into Trace and BaseDenom passes validation and
                // is the same as the current DenomTrace.
                // If it isn't then store the new DenomTrace in the list of new traces.
                newTrace := ibctransfertypes.ParseDenomTrace(dt.GetFullDenomPath())
                if err := newTrace.Validate(); err == nil && !equalTraces(newTrace, dt) {
                    newTraces = append(newTraces, newTrace)
                }

                return false
            })

        // replace the outdated traces with the new trace information
        for _, nt := range newTraces {
            app.TransferKeeper.SetDenomTrace(ctx, nt)
        }
```

Seems to me that this would be better being:
```golang
ibctransfertypes.UpgradeToSupportSlashes(ctx, app.TransferKeeper)
```

for chains to all adopt this more quickly. This would save several lines from the upgrade. Or you could make this use migrations on the TransferKeeper directly. This would be preferrred, since then the upgrade would just be a go.mod update!

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

ValarDragon:
Actually is it viable to make a new patch or minor release, and just bump the transfer consensus version to 2.0 https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/module.go, and setup a migration handler? That way chains can update the import in the go.mod, and the rest will 'justwork.TM'  :)

You'd do this by bumping consensus version to 2, and setting up a migration handler:  https://github.com/cosmos/cosmos-sdk/blob/main/x/bank/module.go#L109-L112

crodriguezvega:
Thanks for the idea, @ValarDragon! This would be indeed a good improvement for the migrations!

PR #1680, opened 2022-07-08
chore: denom traces migration handler

charleenfei:
## Description

 closes: #1542 
---

write a little note why.


modules/apps/transfer/keeper/migrations.go
```diff
@@ -0,0 +1,52 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/ibc-go/v4/modules/apps/transfer/types"
```
crodriguezvega, line 5: Please separate imports :)

modules/apps/transfer/keeper/migrations.go
```diff
@@ -0,0 +1,52 @@
...
+			}
+			iterErr = err
+			return false
+		})
+
+	// replace the outdated traces with the new trace information
+	for _, nt := range newTraces {
+		m.keeper.SetDenomTrace(ctx, nt)
+	}
+	return iterErr
```
crodriguezvega, line 47: If a non-`nil` error is returned here, will the changes that may have happened for some denom traces be rolled back?

modules/apps/transfer/keeper/migrations.go
```diff
@@ -0,0 +1,52 @@
...
+			newTrace := types.ParseDenomTrace(dt.GetFullDenomPath())
+			err := newTrace.Validate()
+			if err != nil {
+				iterErr = err
+				return true
+			}
+			if !equalTraces(newTrace, dt) {
+				newTraces = append(newTraces, newTrace)
+			}
+			iterErr = err
```
crodriguezvega, line 39: Nit: I guess it doesn't hurt to have it, but I think that you could remove this assignment, since if all denom traces validate successfully `iterErr` will be `nil` because it was never initialized, right?

crodriguezvega (commented):
Nice work, @charleenfei!

Should we add changelog entry and also an explanation in the v3-to-v4 migration docs?

docs/migrations/v3-to-v4.md
```diff
@@ -18,7 +18,27 @@ No genesis or in-place migrations required when upgrading from v1 or v2 of ibc-g
...
+
+This incorrect trace information must be corrected when the chain does upgrade to fully supporting denominations with slashes.
+
+If the chain will migrate to supporting base denoms with slashes, it must set the appropriate params during the execution of the upgrade handler in `app.go`: 
+```go
+app.UpgradeKeeper.SetUpgradeHandler("MigrateTraces",
+    func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+        // bump the consensus version of the IBC transfer module 
+        fromVM[ibctransfertypes.ModuleName] = ibctransfermodule.ConsensusVersion()
+
```
ValarDragon, line 37: we shouldn't have to do this right? Because the fromVm version would be the old one, and the live code one is `ibctransfermodule.ConsensusVersion()`, hence the `app.mm.RunMigrations` will detect the difference, update the stored version manager, and run the migration right

ValarDragon (approved):
This is awesome!!

docs/migrations/v3-to-v4.md
charleenfei: aha, got it. i wasnt aware that `app.mm.RunMigrations` would make this redundant, and wanted to make the version bumping explicit. but I can remove this line!

docs/migrations/v3-to-v4.md
```diff
@@ -18,7 +18,25 @@ No genesis or in-place migrations required when upgrading from v1 or v2 of ibc-g
 
 ## Chains
 
-- No relevant changes were made in this release.
+# Migrating from not supporing base denoms with slashes to supporting base denoms with slashes
+
+This is necessary when chains are upgrading from a version that does not support base denoms with slashes (e.g. v3.0.0) to a version that does (e.g. v3.1.0). All versions of ibc-go smaller than v1.5.0 for the v1.x release line, v2.3.0 for the v2.x release line, and v3.1.0 for the v3.x release line do *NOT** support IBC token transfers of coins whose base denoms contain slashes. Therefore the in-place of genesis migration described in this document are required when upgrading.
```
crodriguezvega, line 23: I am thinking that maybe we can reduce the text here and just make a reference to the migration that was documented in v3.1.0, v2.3.0 and v1.5.0 and say that this is an improved, simplified way to run the same migration. What about something like this?

_As part of [v1.5.0](https://github.com/cosmos/ibc-go/releases/tag/v1.5.0), [v2.3.0](https://github.com/cosmos/ibc-go/releases/tag/v2.3.0) and [v3.1.0](https://github.com/cosmos/ibc-go/releases/tag/v3.1.0) some [migration handler code sample was documented](https://github.com/cosmos/ibc-go/blob/main/docs/migrations/support-denoms-with-slashes.md#upgrade-proposal) that needs to run in order to correct the trace information of coins transferred using ICS20 whose base denom contains slashes.
Based on feedback from the community we add now an improved solution to run the same migration that does not require copying a large piece of code over from the migration document, but instead requires only adding a one-line upgrade handler._

What do you think? Do you agree with writing something like this? Please change whatever you see fit.

crodriguezvega (commented):
The is a great improvement, @charleenfei!

Question: should we also update the [existing sample in the separate migration doc](https://github.com/cosmos/ibc-go/blob/main/docs/migrations/support-denoms-with-slashes.md#upgrade-proposal) so that we can backport this change to v2 and v3? Then, for example, we could release v3.2.0 and v2.4.0 and chains would benefit from using the simpler upgrade handler. What do you think?

modules/apps/transfer/keeper/migrations.go
```diff
@@ -0,0 +1,52 @@
...
+	keeper Keeper
+}
+
+// NewMigrator returns a new Migrator.
+func NewMigrator(keeper Keeper) Migrator {
+	return Migrator{keeper: keeper}
+}
+
+// This migration will update the denom
+func (m Migrator) MigrateTraces(ctx sdk.Context) error {
```
damiannolan, line 20: nit: godocs should start with function name. e.g. "MigrateTraces will update the denom traces stored in state"

modules/apps/transfer/keeper/migrations.go
```diff
@@ -0,0 +1,52 @@
...
+	m.keeper.IterateDenomTraces(ctx,
+		func(dt types.DenomTrace) (stop bool) {
+			// check if the new way of splitting FullDenom
+			// into Trace and BaseDenom passes validation and
+			// is the same as the current DenomTrace.
+			// If it isn't then store the new DenomTrace in the list of new traces.
+			newTrace := types.ParseDenomTrace(dt.GetFullDenomPath())
+			err := newTrace.Validate()
+			if err != nil {
+				iterErr = err
```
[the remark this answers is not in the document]
damiannolan: Maybe it should just panic and then have the chain devs export a snapshot and manually address the situation? What do you think?

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -0,0 +1,108 @@
+package keeper_test
+
+import (
+	"fmt"
+
+	transferkeeper "github.com/cosmos/ibc-go/v4/modules/apps/transfer/keeper"
+	transfertypes "github.com/cosmos/ibc-go/v4/modules/apps/transfer/types"
+)
+
+func (suite *KeeperTestSuite) TestMigrator_MigrateTraces() {
```
damiannolan, line 10: nit: I think you can remove the underscore in the test func name. This is more like a java-esque style naming convention for tests. I think `TestMigratorMigrateTraces` is good and follows Go conventions

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		suite.Run(fmt.Sprintf("case %s", tc.msg), func() {
+			suite.SetupTest() // reset
+			tc.malleate()     // explicitly set up denom traces
+			migrator := transferkeeper.NewMigrator(suite.chainA.GetSimApp().TransferKeeper)
+			err := migrator.MigrateTraces(suite.chainA.GetContext())
+			suite.Require().NoError(err)
```
damiannolan, line 102: nit: I think some kind of logical spacing is nice for readability, but not required! :) 

```suggestion
			suite.SetupTest() // reset
			
			tc.malleate() // explicitly set up denom traces
			
			migrator := transferkeeper.NewMigrator(suite.chainA.GetSimApp().TransferKeeper)
			err := migrator.MigrateTraces(suite.chainA.GetContext())
			suite.Require().NoError(err)
```

damiannolan (commented):
Awesome work, looks more or less good to go, just left a few nits and suggestion re. malformed traces

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -0,0 +1,108 @@
...
+						BaseDenom: "customport/channel-7/uatom", Path: "transfer/channel-0/transfer/channel-1",
+					})
+			},
+			transfertypes.Traces{
+				{
+					BaseDenom: "customport/channel-7/uatom", Path: "transfer/channel-0/transfer/channel-1",
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
charleenfei: see below comment

modules/apps/transfer/keeper/migrations.go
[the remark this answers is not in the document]
damiannolan: In which case, couldn't we return the error immediately? Instead of continuing to do unnecessary work... I'm assuming that the upgrade handler and `RunMigrations` will [likely bubble up the error](https://github.com/cosmos/cosmos-sdk/blob/main/types/module/module.go#L444-L447) and revert any changes so I think it seems unnecessary to continue parsing traces if we encounter an error for some unforeseen reason!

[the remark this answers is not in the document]
charleenfei: we do stop the parsing and then return the error in the current code, is there another way to return the error that you think would make more sense?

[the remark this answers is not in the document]
damiannolan: Ah sorry, I missed the `return true` underneath here due to the thread of comments!

damiannolan (approved):
Great work! 🚀

docs/migrations/v3-to-v4.md
```diff
@@ -18,7 +18,27 @@ No genesis or in-place migrations required when upgrading from v1 or v2 of ibc-g
...
+As part of [v1.5.0](https://github.com/cosmos/ibc-go/releases/tag/v1.5.0), [v2.3.0](https://github.com/cosmos/ibc-go/releases/tag/v2.3.0) and [v3.1.0](https://github.com/cosmos/ibc-go/releases/tag/v3.1.0) some [migration handler code sample was documented](https://github.com/cosmos/ibc-go/blob/main/docs/migrations/support-denoms-with-slashes.md#upgrade-proposal) that needs to run in order to correct the trace information of coins transferred using ICS20 whose base denom contains slashes.
+
+Based on feedback from the community we add now an improved solution to run the same migration that does not require copying a large piece of code over from the migration document, but instead requires only adding a one-line upgrade handler.
+
+If the chain will migrate to supporting base denoms with slashes, it must set the appropriate params during the execution of the upgrade handler in `app.go`: 
+```go
+app.UpgradeKeeper.SetUpgradeHandler("MigrateTraces",
+    func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+        // transfer module consensus version has been bumped to 2
+        return app.mm.RunMigrations(ctx, app.configurator, fromVM)
```
chatton, line 32: Where/how is `RunMigrations` defined? Does this get run at the sdk level as a callback?

This one liner is very nice compared to previous implementation btw, nice job!

modules/apps/transfer/keeper/migrations.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/cosmos/ibc-go/v4/modules/apps/transfer/types"
+)
+
+// Migrator is a struct for handling in-place store migrations.
+type Migrator struct {
+	keeper Keeper
+}
+
+// NewMigrator returns a new Migrator.
+func NewMigrator(keeper Keeper) Migrator {
```
chatton, line 15: is the intention that all transfer related migrations will happen here?

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -0,0 +1,110 @@
+package keeper_test
+
+import (
+	"fmt"
+
+	transferkeeper "github.com/cosmos/ibc-go/v4/modules/apps/transfer/keeper"
+	transfertypes "github.com/cosmos/ibc-go/v4/modules/apps/transfer/types"
+)
+
+func (suite *KeeperTestSuite) TestMigratorMigrateTraces() {
```
chatton, line 10: 🥇

chatton (approved):
This looks great, amazing test coverage! I just had a few question about how the migrations actually run.

docs/migrations/v3-to-v4.md
colin-axner: see [docs](https://docs.cosmos.network/main/core/upgrade.html). Core IBC has also had its "module version" [incremented](https://github.com/cosmos/ibc-go/blob/9776639b1a13811191c875c73e185efee577468f/modules/core/module.go#L140) previously

modules/apps/transfer/types/trace.go
```diff
@@ -71,6 +71,9 @@ func (dt DenomTrace) GetFullDenomPath() string {
 	if dt.Path == "" {
 		return dt.BaseDenom
 	}
+	if dt.BaseDenom == "" {
+		return dt.Path
+	}
```
colin-axner, line 76: why was this added? I think in this situation, there should be a panic

modules/apps/transfer/keeper/migrations_test.go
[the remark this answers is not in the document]
colin-axner: I agree this situation is unlikely. It can still be useful to add a test case to ensure there is an if statement checking that iterErr is non nil (I have seen those missing if statements go unnoticed before)

modules/apps/transfer/types/trace.go
charleenfei: i added it bc in the case that the incorrectly parsed BaseDenom is an empty string for whatever reason, the `GetPrefix()` method will actually append an additional `/` onto the end and the BaseDenom will end up looking like ie `erc/0x00000000000etc/` at the conclusion of this logic. I can also set a panic, which do you think makes more sense?

modules/apps/transfer/keeper/migrations.go
charleenfei: yes -- the cosmos sdk has all migrations in a [separate directory](https://github.com/cosmos/cosmos-sdk/tree/main/x/bank/migrations) but imo that might be overkill for us. wdyt?

modules/apps/transfer/types/trace.go
colin-axner: I think we can cleanup a lot of this code when we look into refactoring transfer and improving upon it. For now, I think it'd be best to remove this check and ensure during migrations that the previously set denom trace passes the `Validate()` call. If it doesn't, we should probably panic

charleenfei: but i dont think some of the previous set denom traces would pass the Validate() call... ie something like ```transfertypes.DenomTrace{
BaseDenom: "1", Path: "transfer/channel-0/gamm/pool",
})```

wouldnt this make the point of the migration code redundant if we are trying to `Validate` before setting the new trace?

modules/apps/transfer/keeper/migrations.go
chatton: I think for us it makes sense to have them here, we can worry about additional directories if we end up with a large number of migrations.

modules/apps/transfer/types/trace.go
colin-axner: previous invalid denom traces aren't possible in the current code as they would fail on packet [Validation](https://github.com/cosmos/ibc-go/blob/aaccbd193f19f97f8068278141044bf324e6425c/modules/apps/transfer/types/packet.go#L54). The purpose of the migration code is to fix the stored trace info, not validate the denomination. denomTrace.Validate() should pass for all existing denominations

The specific example, "gamm/pool/1", should have failed due to port/channel identifier validation (insufficient length) when being received or sent in a packet

charleenfei: got it, that makes sense!

AdityaSripal (approved):
Nice work!

PR #1701, opened 2022-07-14
Separate go mod for e2e (E2E #4)

chatton:
## Description

This PR removes `ibctest` as a dependency of the `ibc-go` module and creates a seperate go.mod file for our e2e tests.

The E2E specific Makefile commands have been extracted and moved into a separate Makefile under the `e2e` directory.

---

write a little note why.


e2e/Makefile
```diff
@@ -0,0 +1,11 @@
...
+TEST_CONTAINERS=$(shell docker ps --filter "label=ibc-test" -a -q)
+
+cleanup-ibc-test-containers:
+	for id in $(TEST_CONTAINERS) ; do \
+		$(DOCKER) stop $$id ; \
+		$(DOCKER) rm $$id ; \
+	done
+
+e2e-test: cleanup-ibc-test-containers
+	@go test -v ./ --run $(suite) -testify.m ^$(test)$$
```
damiannolan, line 11: I'm not exactly a Makefile guru, but should these `.PHONY` targets as they're not creating an output file?

chatton: you are absolutely correct, thanks for spotting that!

e2e/go.mod
```diff
@@ -0,0 +1,142 @@
+module e2e
```
colin-axner, line 1: Is this common practice? I would think the import should be "github.com/cosmos/ibc-go/e2e"? What would the differences be between using the different module paths?

chatton: I was initially thinking of to do something like this, but opted to remove the `ibc-go` prefix to make it clear it is a separate entity. If you think it's better to have the fully qualified name we can always rename it.

Issue #1395, opened 2022-05-18
Simplify SendPacket interface


colin-axner:
## Summary

Apply changes from [cosmos/ibc#731](https://github.com/cosmos/ibc/pull/731) and it's followup prs

low priority, primarily an application developer UX improvement

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1703, opened 2022-07-14
Simplify SendPacket API

AdityaSripal:
## Description

closes: #1395 

---

write a little note why.


damiannolan (approved):
Nice!! Looks super clean, love how much simpler the `SendTx` code in ICA is. 
Left one question about telemetry, not sure how critical it really is tho.

🚀

modules/apps/transfer/keeper/relay.go
```diff
@@ -100,10 +82,7 @@ func (k Keeper) SendTransfer(
 		}
 	}
 
-	labels := []metrics.Label{
-		telemetry.NewLabel(coretypes.LabelDestinationPort, destinationPort),
-		telemetry.NewLabel(coretypes.LabelDestinationChannel, destinationChannel),
-	}
```
chatton, line 106: [q] where did these values end up previously? Is it safe to remove them?

modules/core/04-channel/keeper/packet.go
```diff
@@ -40,22 +40,24 @@ func (k Keeper) SendPacket(
...
 	}
 
-	if !k.scopedKeeper.AuthenticateCapability(ctx, channelCap, host.ChannelCapabilityPath(packet.GetSourcePort(), packet.GetSourceChannel())) {
-		return sdkerrors.Wrapf(types.ErrChannelCapabilityNotFound, "caller does not own capability for channel, port ID (%s) channel ID (%s)", packet.GetSourcePort(), packet.GetSourceChannel())
+	if !k.scopedKeeper.AuthenticateCapability(ctx, channelCap, host.ChannelCapabilityPath(srcPort, srcChannel)) {
+		return sdkerrors.Wrapf(types.ErrChannelCapabilityNotFound, "caller does not own capability for channel, port ID (%s) channel ID (%s)", srcPort, srcChannel)
 	}
 
-	if packet.GetDestPort() != channel.Counterparty.PortId {
+	sequence, found := k.GetNextSequenceSend(ctx, srcPort, srcChannel)
```
chatton, line 47: [q] Could you help me understand why this check is a substitute for the previous `packet.GetDestPort() != channel.Counterparty.PortId` and `packet.GetDestChannel() != channel.Counterparty.ChannelId`? It's not immediately clear to me looking into `GetNextSequenceSend`

chatton (commented):
Looks great! Nice cleanup. I just had a few small questions

crodriguezvega (commented):
Thanks, @AdityaSripal! I think we need to add also some documentation in the migration docs for app developers, right? And we should decide in what version this should be released...

modules/core/04-channel/keeper/packet.go
AdityaSripal: @crodriguezvega is correct. Yea the diffs on this PR look confusing in the merged view. But these checks have been removed completely since we are directly grabbing the destination info from the channelEnd

Before we were requiring callers to pass this in and checking it is the exact value we expect. But if we already have the value, then we could just plug it in ourselves. That is the point of this PR.

modules/apps/transfer/keeper/relay.go
```diff
@@ -100,10 +82,7 @@ func (k Keeper) SendTransfer(
...
+	labels := []metrics.Label{}
```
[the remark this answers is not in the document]
AdityaSripal: Yes. cc: @colin-axner

modules/core/04-channel/keeper/packet.go
chatton: Ah I see, thanks for the clarification 👍

modules/core/04-channel/keeper/packet.go
```diff
@@ -22,15 +22,15 @@ import (
 func (k Keeper) SendPacket(
 	ctx sdk.Context,
 	channelCap *capabilitytypes.Capability,
-	packet exported.PacketI,
+	srcPort string,
+	srcChannel string,
+	timeoutHeight clienttypes.Height,
+	timeoutTimestamp uint64,
+	data []byte,
 ) error {
```
colin-axner, line 30: The sequence should be returned

modules/core/04-channel/keeper/packet.go
```diff
@@ -22,15 +22,15 @@ import (
 func (k Keeper) SendPacket(
 	ctx sdk.Context,
 	channelCap *capabilitytypes.Capability,
-	packet exported.PacketI,
+	srcPort string,
+	srcChannel string,
```
colin-axner, line 26: I'm not sure how often we use `src`. Maybe we could just use `source`?

modules/core/04-channel/keeper/packet.go
asalzmann: +1 please add this. we're currently doing this on a fork, but it would be easier for us to maintain if this was the default behavior https://github.com/Stride-Labs/ibc-go/pull/1/files

colin-axner: Thanks for the feedback @asalzmann! I've noted it on #1969. I'll see if we can get someone to pick it up relatively soon

modules/apps/27-interchain-accounts/controller/keeper/relay.go
```diff
@@ -21,57 +21,25 @@ func (k Keeper) SendTx(ctx sdk.Context, chanCap *capabilitytypes.Capability, con
...
-		timeoutTimestamp,
-	)
-
-	if err := k.ics4Wrapper.SendPacket(ctx, chanCap, packet); err != nil {
+	if err := k.ics4Wrapper.SendPacket(ctx, chanCap, portID, activeChannelID, clienttypes.ZeroHeight(), timeoutTimestamp, icaPacketData.GetBytes()); err != nil {
 		return 0, err
 	}
 
-	return packet.Sequence, nil
+	return sequence, nil
```
colin-axner, line 42: sequence should be returned by SendPacket so you can remove the get next sequence call

modules/apps/27-interchain-accounts/controller/keeper/relay.go
```diff
@@ -21,57 +21,25 @@ func (k Keeper) SendTx(ctx sdk.Context, chanCap *capabilitytypes.Capability, con
 		return 0, sdkerrors.Wrapf(icatypes.ErrActiveChannelNotFound, "failed to retrieve active channel on connection %s for port %s", connectionID, portID)
 	}
 
-	sourceChannelEnd, found := k.channelKeeper.GetChannel(ctx, portID, activeChannelID)
-	if !found {
-		return 0, sdkerrors.Wrap(channeltypes.ErrChannelNotFound, activeChannelID)
-	}
```
colin-axner, line 27: can you double check the test cases here, there might be some now that are unnecessary/illogical

modules/apps/27-interchain-accounts/types/expected_keepers.go
```diff
@@ -20,7 +21,7 @@ type AccountKeeper interface {
 
 // ICS4Wrapper defines the expected ICS4Wrapper for middleware
 type ICS4Wrapper interface {
-	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, packet ibcexported.PacketI) error
+	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, srcPort, srcChannel string, timeoutHeight clienttypes.Height, timeoutTimestamp uint64, data []byte) error
```
colin-axner, line 24: I think I'd prefer to just use `sourcePort` and `sourceChannel`

modules/apps/transfer/keeper/relay.go
```diff
@@ -62,24 +62,6 @@ func (k Keeper) SendTransfer(
...
-	destinationChannel := sourceChannelEnd.GetCounterparty().GetChannelID()
-
-	// get the next sequence
-	sequence, found := k.channelKeeper.GetNextSequenceSend(ctx, sourcePort, sourceChannel)
-	if !found {
-		return sdkerrors.Wrapf(
-			channeltypes.ErrSequenceSendNotFound,
-			"source port: %s, source channel: %s", sourcePort, sourceChannel,
-		)
-	}
```
colin-axner, line 80: same here for checking unnecessary test cases

modules/apps/transfer/keeper/relay.go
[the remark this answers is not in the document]
colin-axner: I think it is okay, so long as there is telemetry within core IBC showing where packets are being sent from/to, then the telemetry could filter on port ID to get the same info. This might break external UX though, so might make more sense to do in a followup

modules/core/04-channel/keeper/packet.go
```diff
@@ -105,34 +106,19 @@ func (k Keeper) SendPacket(
...
-	}
-
 	commitment := types.CommitPacket(k.cdc, packet)
 
-	nextSequenceSend++
-	k.SetNextSequenceSend(ctx, packet.GetSourcePort(), packet.GetSourceChannel(), nextSequenceSend)
-	k.SetPacketCommitment(ctx, packet.GetSourcePort(), packet.GetSourceChannel(), packet.GetSequence(), commitment)
+	sequence++
+	k.SetNextSequenceSend(ctx, srcPort, srcChannel, sequence)
+	k.SetPacketCommitment(ctx, srcPort, srcChannel, packet.GetSequence(), commitment)
```
colin-axner, line 113: Hmm, I'd be careful here if we are also trying to return the sequence in the API, you don't want to make it so the nextSequenceSend is returned

Maybe it makes more sense to do:
```go
k.SetNextSequenceSend(ctx, srcPort, srcChannel, sequence+1)
```

colin-axner: Technically you could be fine by ensuring you return packet.GetSequence(), but I think I'd actually prefer not to increment the sequence variable. 

You are using the variable for two different purposes:
- to get the sequence of the packet to send
- to set the next packet sequence to send

I'd rather not just the behaviour unexplicitly partway through the code

modules/core/05-port/types/module.go
```diff
@@ -110,7 +111,11 @@ type ICS4Wrapper interface {
 	SendPacket(
 		ctx sdk.Context,
 		chanCap *capabilitytypes.Capability,
-		packet exported.PacketI,
+		srcPort string,
+		srcChannel string,
```
colin-axner, line 115: My preference is `sourcePort` `sourceChannel` since there isn't a reason not to be explicit

testing/endpoint.go
```diff
@@ -375,7 +375,9 @@ func (endpoint *Endpoint) SendPacket(packet exported.PacketI) error {
 	channelCap := endpoint.Chain.GetChannelCapability(packet.GetSourcePort(), packet.GetSourceChannel())
 
 	// no need to send message, acting as a module
-	err := endpoint.Chain.App.GetIBCKeeper().ChannelKeeper.SendPacket(endpoint.Chain.GetContext(), channelCap, packet)
+	// TODO: Change Endpoint SendPacket to take in arguments directly
```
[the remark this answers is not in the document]
colin-axner: Can you open an issue to track this work?

modules/core/04-channel/keeper/packet.go
colin-axner: When this change is made, a test should be added to ensure the correct sequence is returned (and not the next sequence send value)

modules/core/04-channel/keeper/packet.go
colin-axner: Can you double check that we verify these values are correctly set in the tests?

colin-axner (commented):
Concept ACK, see suggested feedback. Seems like a straightforward change. Nice improvement! :mechanical_arm:

colin-axner:
Marking as draft just so it is easier to know when suggestions and merge conflicts have been addressed and the pr is ready for a full review again

modules/core/04-channel/keeper/packet.go
AdityaSripal: Done. I ended up using packet.GetSequence() since that is what gets used in the rest of the lines, but I did switch to just setting sequence+1. So future changes don't affect anything.

I also added tests, so it would be caught anyway

modules/apps/transfer/keeper/relay.go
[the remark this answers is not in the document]
AdityaSripal: I don't think telemetry is used on external UX? Also I just think it makes sense to keep it clean here, rather than grabbing channel just so that we can emit the same telemetry.

Afaiu, breaking telemetry does not count as an api breaking or state machine breaking change. So i don't want to be so restrictive and conservative in this case.

I'm in favor of just removing it. We can add it in the changelog so that people who need this information are informed of it beforehand

testing/endpoint.go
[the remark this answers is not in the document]
AdityaSripal: https://github.com/cosmos/ibc-go/issues/2452

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -25,7 +26,7 @@ type Keeper struct {
 	cdc        codec.BinaryCodec
 	paramSpace paramtypes.Subspace
 
-	ics4Wrapper   icatypes.ICS4Wrapper
+	ics4Wrapper   porttypes.ICS4Wrapper
```
AdityaSripal, line 29: No reason for us to duplicate ICS4Wrapper in the expected keepers. the interface already exists in port types. This also prevents accidentally missing a function. Done for all apps

modules/apps/27-interchain-accounts/types/expected_keepers.go
```diff
@@ -18,15 +18,9 @@ type AccountKeeper interface {
...
-// ICS4Wrapper defines the expected ICS4Wrapper for middleware
-type ICS4Wrapper interface {
-	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, packet ibcexported.PacketI) error
-	GetAppVersion(ctx sdk.Context, portID, channelID string) (string, bool)
-}
-
 // ChannelKeeper defines the expected IBC channel keeper
 type ChannelKeeper interface {
-	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
+	GetChannel(ctx sdk.Context, sourcePort, srcChan string) (channel channeltypes.Channel, found bool)
```
AdityaSripal, line 23:
```suggestion
	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
```

modules/apps/29-fee/types/expected_keepers.go
```diff
@@ -15,16 +14,9 @@ type AccountKeeper interface {
...
-type ICS4Wrapper interface {
-	WriteAcknowledgement(ctx sdk.Context, chanCap *capabilitytypes.Capability, packet ibcexported.PacketI, acknowledgement ibcexported.Acknowledgement) error
-	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, packet ibcexported.PacketI) error
-	GetAppVersion(ctx sdk.Context, portID, channelID string) (string, bool)
-}
-
 // ChannelKeeper defines the expected IBC channel keeper
 type ChannelKeeper interface {
-	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
+	GetChannel(ctx sdk.Context, sourcePort, srcChan string) (channel channeltypes.Channel, found bool)
```
AdityaSripal, line 19:
```suggestion
	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
```

modules/apps/transfer/types/expected_keepers.go
```diff
@@ -26,14 +26,9 @@ type BankKeeper interface {
...
 
-// ICS4Wrapper defines the expected ICS4Wrapper for middleware
-type ICS4Wrapper interface {
-	SendPacket(ctx sdk.Context, channelCap *capabilitytypes.Capability, packet ibcexported.PacketI) error
-}
-
 // ChannelKeeper defines the expected IBC channel keeper
 type ChannelKeeper interface {
-	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
+	GetChannel(ctx sdk.Context, sourcePort, srcChan string) (channel channeltypes.Channel, found bool)
```
AdityaSripal, line 31:
```suggestion
	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
```

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -20,7 +20,7 @@ import (
 var (
 	disabledTimeoutTimestamp = uint64(0)
 	disabledTimeoutHeight    = clienttypes.ZeroHeight()
-	timeoutHeight            = clienttypes.NewHeight(1, 100)
+	defaultTimeoutHeight     = clienttypes.NewHeight(1, 100)
```
AdityaSripal, line 23: I had to change this back since it was conflicting with the local var in `TestSendPacket`. I also just find it makes more sense, since there are some non default values I put in the test cases.

crodriguezvega (commented):
Thanks, @AdityaSripal!

We need to add these changes to the migration docs, but maybe that can be done in @damiannolan's [PR](https://github.com/cosmos/ibc-go/pull/2417).

We should also update all the documentation where `SendPacket` is mentioned; there are a few places...

colin-axner (changes requested):
Needs migration docs. My main remaining concern is around telemetry. I wouldn't want to accidentally make a disruptive change, so we should ensure we have given sufficient consideration to the matter

docs/migrations/v5-to-v6.md
```diff
@@ -0,0 +1,41 @@
...
+
+## IBC Apps
+
+### ICS04 - SendPacket API change
+
+The `SendPacket` API has been simplified:
+
+```diff
+// SendPacket is called by a module in order to send an IBC packet on a channel
+@@ -22,122 +22,98 @@ import (
```
damiannolan, line 25: I think you can remove this line

```suggestion
```

modules/core/04-channel/keeper/packet.go
```diff
@@ -22,113 +22,98 @@ import (
 func (k Keeper) SendPacket(
```
damiannolan, line 22: Should we update the godoc to mention that the sequence is now returned?

modules/core/05-port/types/module.go
```diff
@@ -110,8 +111,12 @@ type ICS4Wrapper interface {
...
 		ctx sdk.Context,
 		chanCap *capabilitytypes.Capability,
-		packet exported.PacketI,
-	) error
+		sourcePort string,
+		sourceChannel string,
+		timeoutHeight clienttypes.Height,
+		timeoutTimestamp uint64,
+		data []byte,
+	) (sequence uint64, err error)
```
damiannolan, line 119: Should we leave these as unnamed return args as they are in other methods such as `GetAppVersion`?

damiannolan (approved):
LGTM! Nice work 🚀

modules/core/05-port/types/module.go
colin-axner: It is good as is in my opinion. OnChanOpenInit has named returns. GetAppVersion probably should have had a named return

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
```diff
@@ -227,8 +228,12 @@ func (im IBCMiddleware) OnTimeoutPacket(
...
 	ctx sdk.Context,
 	chanCap *capabilitytypes.Capability,
-	packet ibcexported.PacketI,
-) error {
+	sourcePort string,
+	sourceChannel string,
+	timeoutHeight clienttypes.Height,
+	timeoutTimestamp uint64,
+	data []byte,
+) (uint64, error) {
```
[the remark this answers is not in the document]
colin-axner: updated

colin-axner:
@crodriguezvega Are there any changes left you think we should make? With regards to your current suggestions, I think it probably makes sense to continue using the packet variables so as to ensure we are referencing the correct values

Otherwise I think this pr is ready to be merged

docs/ibc/apps.md
```diff
@@ -238,10 +238,21 @@ DecodePacketData(encoded []byte) (CustomPacketData) {
 Then a module must encode its packet data before sending it through IBC.
```
crodriguezvega, line 238: I realized that this file is actually not used in the docs website, so could probably be removed...

modules/core/05-port/types/module.go
damiannolan: Maybe we can make them all consistent in housekeeping followup.

crodriguezvega:
> @crodriguezvega Are there any changes left you think we should make?

Yes, I just updated the documentation. I would appreciate a review to check it all looks ok. After that, I agree we can merge it.

docs/ibc/apps.md
colin-axner: I see, looks like we took this info and split it up into separate files? Lets open an issue whose task is to see if any info in here should be moved before deleting the file

colin-axner: #2480

colin-axner (approved):
Thanks @crodriguezvega for updating docs! I re-reviewed all changes

Issue #1398, opened 2022-05-19
Allow governance upgrades to change the TrustingPeriod of an IBC client


charleenfei:
## Summary

currently, changing the trusting period will cause a governance upgrade of a client to fail as we assume that the security properties should remain the same across client upgrades. however, there are situations where a trusting period has been incorrectly misconfigured for a canonical channel, governance should be allowed to upgrade this parameter (see: EVMOS)

## Proposal

Null trusting period in [IsMatchingClientState](https://github.com/cosmos/ibc-go/blob/2ae4f10143e233ab69d408c12fe48a9ff58f24c7/modules/light-clients/07-tendermint/types/proposal_handle.go#L96)

[Set new trusting period](https://github.com/cosmos/ibc-go/blob/2ae4f10143e233ab69d408c12fe48a9ff58f24c7/modules/light-clients/07-tendermint/types/proposal_handle.go#L86)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1713, opened 2022-07-15
feat: allow governance to update the TrustingPeriod of the 07-tendermint light client

charleenfei:
## Description

closes: #1398 

---

write a little note why.


docs/architecture/adr-026-ibc-client-recovery-mechanisms.md
```diff
@@ -51,6 +52,9 @@ We elect not to deal with chains which have actually halted, which is necessaril
 
     Previously, `AllowUpdateAfterExpiry` and `AllowUpdateAfterMisbehaviour` were used to signal the recovery options for an expired or frozen client, and governance proposals were not allowed to overwrite the client if these parameters were set to false. However, this has now been deprecated because a code migration can overwrite the client and consensus states regardless of the value of these parameters. If governance would vote to overwrite a client or consensus state, it is likely that governance would also be willing to perform a code migration to do the same.
 
+    In addition, `TrustingPeriod` was initally not allowed to be updated by a client upgrade proposal. However, due to the number of situations experienced in production where the `TrustingPeriod` of a client should be allowed to be updated because of ie: initial misconfiguration for a canonical channel, governance should be allowed to update this client parameter.
```
crodriguezvega, line 55:
```suggestion
    In addition, `TrustingPeriod` was initally not allowed to be updated by a client upgrade proposal. However, due to the number of situations experienced in production where the `TrustingPeriod` of a client should be allowed to be updated because of, for example, initial misconfiguration for a canonical channel, governance should be allowed to update this client parameter. 
```

docs/architecture/adr-026-ibc-client-recovery-mechanisms.md
```diff
@@ -51,6 +52,9 @@ We elect not to deal with chains which have actually halted, which is necessaril
...
 
+    In addition, `TrustingPeriod` was initally not allowed to be updated by a client upgrade proposal. However, due to the number of situations experienced in production where the `TrustingPeriod` of a client should be allowed to be updated because of ie: initial misconfiguration for a canonical channel, governance should be allowed to update this client parameter. 
+    
+    Note that this should NOT be lightly updated, as there may be a gap in time between when misbehaviour has occured and when the evidence of misbehaviour is submitted. For example, if the `UnbondingPeriod` is 2 weeks and the `TrustingPeriod` has also been set to two weeks, a validator could wait until right before `UnbondingPeriod` finishes, submit false information, then unbond and exit without being slashed for misbehaviour. Therefore, we recommend that the trusting period for the 07-tendermint client be set at a minimum to 2/3 of the `UnbondingPeriod`.
```
crodriguezvega, line 57:
```suggestion
    Note that this should NOT be lightly updated, as there may be a gap in time between when misbehaviour has occured and when the evidence of misbehaviour is submitted. For example, if the `UnbondingPeriod` is two weeks and the `TrustingPeriod` has also been set to two weeks, a validator could wait until right before `UnbondingPeriod` finishes, submit false information, then unbond and exit without being slashed for misbehaviour. Therefore, we recommend that the trusting period for the 07-tendermint client be set at a minimum to 2/3 of the `UnbondingPeriod`.
```

crodriguezvega (commented):
Thanks, @charleenfei!

I think we need to update this text in [`docs/ibc/upgrades/developer-guide.md`](https://github.com/cosmos/ibc-go/blob/main/docs/ibc/upgrades/developer-guide.md), right? 

_... while ensuring that the relayer submitting the UpgradeClientMsg cannot alter the client-chosen parameters that the users are relying upon (examples in Tendermint client include TrustingPeriod, TrustLevel, MaxClockDrift, etc)._

docs/architecture/adr-026-ibc-client-recovery-mechanisms.md
charleenfei: i will leave it `2` for readability :)

docs/architecture/adr-026-ibc-client-recovery-mechanisms.md
charleenfei: i will leave ie: for brevity

crodriguezvega: Although [Google's developer documentation style guide](https://developers.google.com/style/word-list) recommends not to use the abbreviation (_Instead, use phrases like for example or such as. Too many people mix up e.g. and i.e._), then I think it would be better to use `e.g.`, since you are giving an example (instead of clarifying the previous statement), right?

docs/architecture/adr-026-ibc-client-recovery-mechanisms.md
crodriguezvega: Sorry to bring up [Google's developer documentation style guide](https://developers.google.com/style/numbers) again, but they recommend spelling out _numbers from zero through nine_. If you still prefer to leave it as `2`, then I would suggest to make it also a `2` a bit later in the same sentence, for consistency. :)

damiannolan (approved):
🚀

AdityaSripal (approved):
Thanks for the fix @charleenfei ! This works for the immediate issue at hand, but actually i think governance should be able to change all parameters of the client.

Governance basically has sudo privilege on the state machine. so it should be able to do what it wants. It can do what it wants anyway by just explicitly voting to change the state, so we should make things easy for them when we anticipate changes like this.

Though that fix can go in a future PR.

docs/architecture/adr-026-ibc-client-recovery-mechanisms.md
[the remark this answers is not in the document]
charleenfei: makes sense! i'll take it out

colin-axner (commented):
Post merge ACK

PR #1687, opened 2022-07-11
refactor: solomachine generic verification methods and signbytes simplification

damiannolan:
## Description

- Adding new `SignBytes` message type ( note `V2` suffix will be dropped when PRs are upstreamed to this branch)
- Adding generic `VerifyMembership` implementation
- Updating tests

needs: 
- https://github.com/cosmos/ibc-go/pull/1715 
- https://github.com/cosmos/ibc-go/pull/1716
- https://github.com/cosmos/ibc-go/pull/1720

closes: #1292 #1295 #1141 

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #1715, opened 2022-07-15
refactor: solomachine misbehaviour checking

damiannolan:
## Description

- Updates `SignatureAndData` type temporarily to `SignatureAndDataV2` 
- Refactors misbehaviour checking code

NOTE: branch is targeted for https://github.com/cosmos/ibc-go/pull/1687

closes: #XXXX

---

write a little note why.


modules/light-clients/06-solomachine/misbehaviour.go
```diff
@@ -71,6 +71,24 @@ func (sd SignatureAndData) ValidateBasic() error {
 	return nil
 }
 
+// ValidateBasic ensures that the signature and data fields are non-empty.
+func (sd SignatureAndDataV2) ValidateBasic() error {
+	if len(sd.Signature) == 0 {
```
crodriguezvega, line 76: This is a nit, but maybe UX could be slightly improved if we group errors, in case more than one thing is empty. For example, if both signature and data are empty, then you can return one error saying that `signature and data cannot be empty`.

charleenfei: i kind of like having the separate signatures rather than arbitrarily grouping errors (sometimes two? sometimes 3?), this way we can keep it consistent across the codebase...

charleenfei (approved):

modules/light-clients/06-solomachine/misbehaviour.go
damiannolan: I think this point has come up a few times before. I tend to agree with the general behaviour of the sdk of "fail fast" where the error is returned immediately rather than grouping. I think it makes for more idiomatic go code. I do see the reasoning for validating the entire message and returning a group of error responses... it is in line with more traditional rest api design in a way. But all things considered I'd prefer to keep the code consistent and stay aligned with the sdk pattern.

If we were to change the behaviour here it should be changed across our entire codebase. And still, it would differ from standard sdk modules then like `x/auth`,`x/bank`, `x/gov` ..etc

colin-axner (approved):
Thank you! :raised_hands: Changes look great! Very clean and easy to review :)

Issue #1430, opened 2022-05-24
Question about multiple OpenTry messages 


ancazamfir:
## Summary of Bug
Testing connection handshake with multiple relayers (https://github.com/informalsystems/ibc-rs/issues/2168), i see that multiple `MsgConnectionOpenTry` messages with same counterparty are accepted and they generate multiple connection ends in tryOpen state. Only for one of them will the handshake be succesful of course but was wondering why the `ConnOpenTry` handler does not check if the counterparty is already bound to an existing connection.
I remember asking this a while ago but cannot remember the answer :)

To fix this, hermes has to perform some ID "fixup". It is true that even if the `ConnOpenTry` handler would reject subsequent messages the relayer would need to figure out which ID to switch to (probably by querying clientConnections).

Note: the same is true for `ChanOpenTry` handler.

## Version
master

## Steps to Reproduce
```
hermes create client ibc-0 ibc-1
hermes create client ibc-1 ibc-0
hermes tx raw conn-init ibc-0 ibc-1 07-tendermint-0 07-tendermint-0
hermes tx raw conn-try ibc-1 ibc-0 07-tendermint-0 07-tendermint-0 -s connection-0
hermes tx raw conn-try ibc-1 ibc-0 07-tendermint-0 07-tendermint-0 -s connection-0
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

AdityaSripal:
Hmm, yes this seems like a thing we should be able to check on TRY. We would have to probably store some additional state, since on the TRY step, we wouldn't be able to efficiently search through our own existing connectionIDs to see if they have the same counterparty.

So we would have to store a mapping from the proposed counterparty to the ID that we have generated for a handshake in progress on TRY. This would then get deleted on CONFIRM if the first handshake fails in order to allow for a subsequent handshake against the same counterparty.

Similar logic would need to be applied on connection/channel handshakes

cc: @damiannolan @seantking

crodriguezvega:
>So we would have to store a mapping from the proposed counterparty to the ID that we have generated for a handshake in progress on TRY

[...]

PR #1723, opened 2022-07-18
ICS3: Remove ability for multiple TRY msgs to succeed for single INIT msg

AdityaSripal:
## Description

- [x] bump to main
- [ ] add genesis logic
- [ ] e2e
- [ ] needs queries? 

closes: #1430 for connections. A separate PR will do the same logic for channels once the approach here is approved

---

write a little note why.


modules/core/03-connection/keeper/handshake.go
```diff
@@ -242,6 +247,10 @@ func (k Keeper) ConnOpenAck(
 	connection.Counterparty.ConnectionId = counterpartyConnectionID
 	k.SetConnection(ctx, connectionID, connection)
 
+	// delete generatedConnectionID mapping now that handshake attempt is successful
+	// we no longer need to store it for redundancy protection
+	k.DeleteGeneratedConnectionID(ctx, connection.ClientId, counterpartyConnectionID)
```
AdityaSripal, line 252: Not sure if I should be deleting in ACK and CONFIRM. No longer strictly necessary for the state machine but may be useful as a query even after handshake is complete

modules/core/03-connection/keeper/handshake_test.go
```diff
@@ -535,6 +553,11 @@ func (suite *KeeperTestSuite) TestConnOpenAck() {
 
 			if tc.expPass {
 				suite.Require().NoError(err)
+
+				generatedID := suite.chainA.App.GetIBCKeeper().ConnectionKeeper.GetGeneratedConnectionID(
+					suite.chainA.GetContext(), path.EndpointA.ClientID, path.EndpointB.ConnectionID,
+				)
+				suite.Require().Equal("", generatedID, "generatedConnectionID mapping not deleted after handshake complete")
```
AdityaSripal, line 560: Should I be using empty here?

modules/core/03-connection/types/keys.go
```diff
@@ -60,3 +63,7 @@ func ParseConnectionSequence(connectionID string) (uint64, error) {
 
 	return sequence, nil
 }
+
+func GeneratedConnectionIDKey(clientID, counterpartyConnectionID string) string {
```
AdityaSripal, line 67: Open to suggestions on a better name

chatton: how about `ExistingConnectionIDKey` or even `ConnectionIDKey`, does the fact that it was generated in some way matter in this context?

AdityaSripal: I like `ExistingConnectionIDKey` since its more descriptive of the problem this is solving and `ConnectionIDKey` can be confused with the ConnectionKey that stores the ConnectionEnd.

Will wait for quorum before changing

modules/core/03-connection/keeper/handshake.go
damiannolan: Shouldn't we only need to delete in the `ConnOpenConfirm` handler?

modules/core/03-connection/types/keys.go
damiannolan: +1 for `ExistingConnectionIDKey`

modules/core/03-connection/keeper/handshake.go
AdityaSripal: Well, theoretically we could have a crossing hello; where ACK gets called on both ends

damiannolan: In this case does CONFIRM not get called on both ends too?

modules/core/03-connection/types/keys.go
AdityaSripal: Done

modules/core/03-connection/keeper/handshake.go
AdityaSripal: nope because both sides go to OPEN

modules/core/03-connection/keeper/keeper.go
```diff
@@ -141,6 +141,26 @@ func (k Keeper) SetNextConnectionSequence(ctx sdk.Context, sequence uint64) {
...
+func (k Keeper) GetExistingConnectionID(ctx sdk.Context, clientID, counterpartyConnectionID string) string {
+	store := ctx.KVStore(k.storeKey)
+	return string(store.Get([]byte(types.ExistingConnectionIDKey(clientID, counterpartyConnectionID))))
+}
+
+// DeleteExistingConnectionID removes the ExistingConnectionID mapping from the store
+func (k Keeper) DeleteExistingConnectionID(ctx sdk.Context, clientID, counterpartyConnectionID string) {
+	store := ctx.KVStore(k.storeKey)
+	store.Delete([]byte(types.ExistingConnectionIDKey(clientID, counterpartyConnectionID)))
+}
```
colin-axner, line 162: requires genesis export/import logic

colin-axner:
Closing pr for now (to reduce noise) as we await for more information on the upstream issue. We can reopen when necessary

PR #1742, opened 2022-07-20
modules/core/23-commitment/types: Add godoc to `VerifyMembership` regarding path

renaynay:
This PR adds an additional godoc to the `VerifyMembership` method on `MerkleProof` to clarify how `path` is expected.

It also includes an additional typo found in godoc of Proof interface (happy to extract in a separate PR, but figured it was fine).


colin-axner (commented):
Thanks @renaynay! Would like to get @AdityaSripal to review this. The documentation makes sense to me, but I can't remember if this is a requirement of ibc-go usage of this function or ics23

AdityaSripal (approved):
Thanks @renaynay ! Yes this happens to be true of sdk chains at the moment. So I think it makes sense to document this at the ibc-go layer

Issue #1685, opened 2022-07-11
E2E: Fee middleware single sender timeout 


chatton:
## Summary

Write an e2e test which verifies the following

> When packet is incentivized from multiple senders AND token transfer times out, then timeout fees are paid.

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1751, opened 2022-07-21
E2E Test: TestMsgPayPacketFeeSingleSenderTimesOut

chatton:
## Description

closes: #1685

---

write a little note why.


e2e/fee_middleware_test.go
```diff
@@ -79,7 +80,7 @@ func (s *FeeMiddlewareTestSuite) QueryIncentivizedPacketsForChannel(
 	return res.IncentivizedPackets, err
 }
 
-func (s *FeeMiddlewareTestSuite) TestMsgPayPacketFeeAsyncSingleSender() {
+func (s *FeeMiddlewareTestSuite) TestMsgPayPacketFee_AsyncSingleSender_Succeeds() {
```
chatton, line 83: added underscores to make the names a little more readable.

e2e/fee_middleware_test.go
```diff
@@ -309,6 +310,130 @@ func (s *FeeMiddlewareTestSuite) TestMultiMsg_MsgPayPacketFeeSingleSender() {
...
+	})
+
+	walletAmount := ibc.WalletAmount{
+		Address: chainBWallet.Bech32Address(chainB.Config().Bech32Prefix), // destination address
+		Denom:   chainA.Config().Denom,
+		Amount:  testvalues.IBCTransferAmount,
+	}
+
+	t.Run("Send IBC transfer", func(t *testing.T) {
+		var err error
```
charleenfei, line 362: nit: `chainATx, err :=` below

e2e/testvalues/values.go
```diff
@@ -7,8 +7,9 @@ import (
 )
 
 const (
-	StartingTokenAmount int64 = 10_000_000
-	IBCTransferAmount   int64 = 10_000
+	StartingTokenAmount int64  = 10_000_000
+	IBCTransferAmount   int64  = 10_000
+	ImmediatelyTimeout  uint64 = 1
```
charleenfei, line 12: not sure if this makes sense in the const file as we wouldnt necessarily change this value (changing the value would change the function) -- maybe you could just add a comment in the code above the line this value is used to explain that it means immediate timeout?

e2e/fee_middleware_test.go
```diff
@@ -309,6 +310,130 @@ func (s *FeeMiddlewareTestSuite) TestMultiMsg_MsgPayPacketFeeSingleSender() {
...
+		chainATx, err = chainA.SendIBCTransfer(ctx, channelA.ChannelID, chainAWallet.KeyName, walletAmount, &ibc.IBCTimeout{
+			NanoSeconds: testvalues.ImmediatelyTimeout,
```
charleenfei, line 364: see comment in the const file

e2e/fee_middleware_test.go
```diff
@@ -309,6 +310,130 @@ func (s *FeeMiddlewareTestSuite) TestMultiMsg_MsgPayPacketFeeSingleSender() {
...
+		s.Require().NoError(err)
+
+		expected := testvalues.StartingTokenAmount - walletAmount.Amount
+		s.Require().Equal(expected, actualBalance)
+	})
+
+	t.Run("pay packet fee", func(t *testing.T) {
+
+		packetId := channeltypes.NewPacketId(channelA.PortID, channelA.ChannelID, 1)
+		packetFee := feetypes.NewPacketFee(testFee, chainAWallet.Bech32Address(chainA.Config().Bech32Prefix), nil)
```
charleenfei, line 382: chainAWallet.Address?

e2e/fee_middleware_test.go
```diff
@@ -309,6 +310,130 @@ func (s *FeeMiddlewareTestSuite) TestMultiMsg_MsgPayPacketFeeSingleSender() {
...
+	t.Run("relayer wallets recovered", func(t *testing.T) {
+		s.Require().NoError(s.RecoverRelayerWallets(ctx, relayer))
+	})
+
+	chainARelayerWallet, chainBRelayerWallet, err := s.GetRelayerWallets(relayer)
+	t.Run("relayer wallets fetched", func(t *testing.T) {
+		s.Require().NoError(err)
+	})
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 1, chainA, chainB), "failed to wait for blocks")
```
charleenfei, line 339: why waiting for blocks here?

e2e/fee_middleware_test.go
```diff
@@ -309,6 +310,130 @@ func (s *FeeMiddlewareTestSuite) TestMultiMsg_MsgPayPacketFeeSingleSender() {
...
+		s.AssertValidTxResponse(resp)
+	})
+
+	t.Run("verify counter party payee", func(t *testing.T) {
+		address, err := s.QueryCounterPartyPayee(ctx, chainB, chainBRelayerWallet.Address, channelA.Counterparty.ChannelID)
+		s.Require().NoError(err)
+		s.Require().Equal(chainARelayerWallet.Address, address)
+	})
+
+	walletAmount := ibc.WalletAmount{
```
charleenfei, line 355: nit: maybe more explicit naming here, it's actually a chainBWalletAmount --

also, maybe i am misreading this but can't you set the address beneath this just by using chainBWallet.Address?

e2e/fee_middleware_test.go
```diff
@@ -309,6 +310,130 @@ func (s *FeeMiddlewareTestSuite) TestMultiMsg_MsgPayPacketFeeSingleSender() {
...
+		Address: chainBWallet.Bech32Address(chainB.Config().Bech32Prefix), // destination address
+		Denom:   chainA.Config().Denom,
+		Amount:  testvalues.IBCTransferAmount,
+	}
+
+	t.Run("Send IBC transfer", func(t *testing.T) {
+		var err error
+		chainATx, err = chainA.SendIBCTransfer(ctx, channelA.ChannelID, chainAWallet.KeyName, walletAmount, &ibc.IBCTimeout{
```
charleenfei, line 363: walletAmount gets a bit confusing here bc it was set as a chainB wallet amount but now youre sending a chainA tx?

e2e/fee_middleware_test.go
chatton: good point this probably is not required.

e2e/fee_middleware_test.go
chatton: `chainAWallet.Address` does not contain the prefix which is required in the packet fee message.

charleenfei: got it, i left the issue in the strangelove repo :)

charleenfei (approved):
nice work! lgtm!

e2e/testvalues/values.go
```diff
@@ -9,6 +9,9 @@ import (
 const (
 	StartingTokenAmount int64 = 10_000_000
 	IBCTransferAmount   int64 = 10_000
+	// ImmediatelyTimeout can be used as the value for timeout in nano seconds when we want
```
charleenfei, line 12: nit: this might be redundant bc it will always be 1, and usually we should use these const values as something we might want to change update in the long run? maybe another way could be to set it as one and then add this comment you have describing the const in the code above where you use "1" as the TimeoutValue? not a blocker tho imo

chatton: actually after thinking some more, maybe the clearest solution is to simply make the constant return

```golang
&ibc.IBCTimeout{
  NanoSeconds: 1,
}
```

This way the result can be passed directly into the `SendIBCTransfer` function. WDYT?

chatton: The above would look like

```golang
chainATx, err = chainA.SendIBCTransfer(ctx, channelA.ChannelID, chainAWallet.KeyName, chainBWalletAmount, testvalues.ImmediatelyTimeout())
```

charleenfei: yes! this lgtm :)

PR #1768, opened 2022-07-22
chore: remove refs to AllowUpdateToExpired/Frozen client booleans

### LINKED ISSUE #1237: Remove AllowUpdateAfterExpiry and AllowUpdateAfterMisbehaviour references
testing/config.go fields can be removed -- this is an API breaking change

charleenfei:
## Description

closes #1237

write a little note why.


charleenfei:
question  -- should the proto fields also be removed? currently it's API breaking (still need to finish updating the CHANGELOG) but updated the protos would make it also state machine breaking (right?) should we just do this in one go?

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * (transfer)[\#1565](https://github.com/cosmos/ibc-go/pull/1565) Removing `NewErrorAcknowledgement` in favour of `channeltypes.NewErrorAcknowledgement`.
 * (channel)[\#1565](https://github.com/cosmos/ibc-go/pull/1565) Updating `NewErrorAcknowledgement` to accept an error instead of a string and removing the possibility of non-deterministic writes to application state.
 * (core/04-channel)[\#1636](https://github.com/cosmos/ibc-go/pull/1636) Removing `SplitChannelVersion` and `MergeChannelVersions` functions since they are not used.
+* (light-clients/tendermint)[\#1768](https://github.com/cosmos/ibc-go/pull/1768/files) Removed `AllowUpdateAfter...` booleans as they are deprecated (see ADR026)
```
chatton, line 57: [nit] maybe we should explicitly list the booleans to make it more clear what exactly was removed?

chatton (approved):
LGTM! Just one small comment on the changelog entry 🥇

modules/light-clients/07-tendermint/types/upgrade.go
```diff
@@ -101,7 +101,6 @@ func (cs ClientState) VerifyUpgradeAndUpdateState(
 	newClientState := NewClientState(
 		tmUpgradeClient.ChainId, cs.TrustLevel, cs.TrustingPeriod, tmUpgradeClient.UnbondingPeriod,
 		cs.MaxClockDrift, tmUpgradeClient.LatestHeight, tmUpgradeClient.ProofSpecs, tmUpgradeClient.UpgradePath,
-		cs.AllowUpdateAfterExpiry, cs.AllowUpdateAfterMisbehaviour,
```
colin-axner, line 104: I guess there's a sideaffect here. These parameters will be set to false, false upon upgrades (even if the previous values were true, true). Shouldn't cause an issue, but is worth being aware of

colin-axner (approved):
LGTM, nice work!

colin-axner:
It probably makes sense to include this in v6 instead of v5 (ie wait for the release branch to be created), just in case there's any issues for users with the changing of `NewClientState` API (thinking of backwards compatibility to older versions)

CHANGELOG.md
charleenfei: makes sense, i will update!

modules/light-clients/07-tendermint/types/upgrade.go
charleenfei: i can add an inline doc for this :)

charleenfei:
> It probably makes sense to include this in v6 instead of v5 (ie wait for the release branch to be created), just in case there's any issues for users with the changing of `NewClientState` API (thinking of backwards compatibility to older versions)

cc @crodriguezvega

modules/light-clients/07-tendermint/types/upgrade.go
charleenfei: added a note in upgrade docs about this

PR #1772, opened 2022-07-25
11-beefy light client implementation

seunlanlege:


seunlanlege:
yeah we definitely fucked the history of this one up.

Issue #1602, opened 2022-06-29
Add `UpgradeErrorPath` to `24-host`


damiannolan:
## Summary

> The upgrade error path is a public path that can signal an error of the upgrade to the counterparty for the given upgrade attempt. It does not store anything in the successful case, but it will store the ErrorReceipt in the case that a chain does not accept the proposed upgrade.

Add `UpgradeErrorPath` to `24-host` with associated key func.

- `UpgradeErrorPath`
- `UpgradeErrorKey`

Ref: [24-host/keys.go](https://github.com/cosmos/ibc-go/blob/main/modules/core/24-host/keys.go)

```typescript
function errorPath(portIdentifier: Identifier, channelIdentifier: Identifier): Path {
    return "channelUpgrade/ports/{portIdentifier}/channels/{channelIdentifier}/upgradeError"
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
Review spec for any updates to key paths

PR #1822, opened 2022-07-28
chore: add UpgradeErrorPath to 24-host

charleenfei:
## Description

closes: #1602 

---

write a little note why.

### DIFF
--- a/modules/core/24-host/channel_keys.go
+++ b/modules/core/24-host/channel_keys.go
@@ -9,7 +9,7 @@ const (
 	KeyChannelRestorePrefix = "restore"
 	KeyUpgradeTimeoutPrefix = "upgradeTimeout"
 
-	KeyUpgradeError = "upgradeError"
+	KeyUpgradeErrorPrefix = "upgradeError"
 )
 
 // ICS04
@@ -41,6 +41,16 @@ func ChannelUpgradeTimeoutKey(portID, channelID string) []byte {
 	return []byte(ChannelUpgradeTimeoutPath(portID, channelID))
 }
 
+// ChannelUpgradeErrorPath defines the path under which the ErrorReceipt is stored in the case that a chain does not accept the proposed upgrade
+func ChannelUpgradeErrorPath(portID, channelID string) string {
+	return fmt.Sprintf("%s/%s/%s", KeyChannelUpgradePrefix, KeyUpgradeErrorPrefix, channelPath(portID, channelID))
+}
+
+// ChannelUpgradeErrorKey returns the store key for a particular channelEnd used to stor the ErrorReceipt in the case that a chain does not accept the proposed upgrade
+func ChannelUpgradeErrorKey(portID, channelID string) []byte {
+	return []byte(ChannelUpgradeErrorPath(portID, channelID))
+}
+
 // ChannelRestorePath defines the path under which channel ends are stored for restoration in the event of upgrade handshake failure
 func ChannelRestorePath(portID, channelID string) string {
 	return fmt.Sprintf("%s/%s/%s", KeyChannelUpgradePrefix, KeyChannelRestorePrefix, channelPath(portID, channelID))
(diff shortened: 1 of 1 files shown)


modules/core/24-host/channel_keys.go
```diff
@@ -52,3 +54,8 @@ func ChannelRestoreKey(portID, channelID string) []byte {
 func channelPath(portID, channelID string) string {
 	return fmt.Sprintf("%s/%s/%s/%s", KeyPortPrefix, portID, KeyChannelPrefix, channelID)
 }
+
+// ErrorPath stores the ErrorReceipt in the case that a chain does not accept the proposed upgrade
+func ErrorPath(portID, channelID string) string {
+	return fmt.Sprintf("%s/%s/%s/%s/%s/%s", KeyChannelUpgradePrefix, KeyPortPrefix, portID, KeyChannelPrefix, channelID, KeyUpgradeError)
+}
```
chatton, line 61: do we need to also add an `ErrorKey`, or maybe a more explicit `ChannelUpgradeErrorKey`

modules/core/24-host/channel_keys.go
```diff
@@ -8,6 +8,8 @@ const (
 	KeyChannelUpgradePrefix = "channelUpgrades"
 	KeyChannelRestorePrefix = "restore"
 	KeyUpgradeTimeoutPrefix = "upgradeTimeout"
+
+	KeyUpgradeError = "upgradeError"
```
damiannolan, line 12:
```suggestion
	KeyUpgradeErrorPrefix = "upgradeError"
```

modules/core/24-host/channel_keys.go
```diff
@@ -52,3 +54,8 @@ func ChannelRestoreKey(portID, channelID string) []byte {
 func channelPath(portID, channelID string) string {
 	return fmt.Sprintf("%s/%s/%s/%s", KeyPortPrefix, portID, KeyChannelPrefix, channelID)
 }
+
+// ErrorPath stores the ErrorReceipt in the case that a chain does not accept the proposed upgrade
+func ErrorPath(portID, channelID string) string {
+	return fmt.Sprintf("%s/%s/%s/%s/%s/%s", KeyChannelUpgradePrefix, KeyPortPrefix, portID, KeyChannelPrefix, channelID, KeyUpgradeError)
```
damiannolan, line 60: I think we can simplify this like below:

```suggestion
	return fmt.Sprintf("%s/%s/%s", KeyChannelUpgradePrefix, KeyUpgradeErrorPrefix, channelPath(portID, channelID))
```

modules/core/24-host/channel_keys.go
damiannolan: I think @chatton is referring to the key function which returns the path as a `[]byte`

modules/core/24-host/channel_keys.go
```diff
@@ -52,3 +54,8 @@ func ChannelRestoreKey(portID, channelID string) []byte {
 func channelPath(portID, channelID string) string {
 	return fmt.Sprintf("%s/%s/%s/%s", KeyPortPrefix, portID, KeyChannelPrefix, channelID)
 }
+
+// ErrorPath stores the ErrorReceipt in the case that a chain does not accept the proposed upgrade
+func ErrorPath(portID, channelID string) string {
```
damiannolan, line 59: Could we rename to `ChannelUpgradeErrorPath`, I think it would be best to be explicit. Also another mega nit but could we add the key and path funcs above the `channelPath` private func such that all exported funcs read first in the file

modules/core/24-host/channel_keys.go
chatton: I was referring to having a `ErrorPath -> string` and `ErrorKey -> []byte`

modules/core/24-host/channel_keys.go
charleenfei: why is this called a Prefix if it's not really prefixing anything?

modules/core/24-host/channel_keys.go
charleenfei: good call! i'll update

chatton: @damiannolan this change isn't equivalent to the current string, but I believe this new format is correct. The existing string has `upgradeError` at the end, when we actually want it as `channelUpgrades/upgradeError/...`

modules/core/24-host/channel_keys.go
charleenfei: got it, that makes sense now 🙈 . i'll add that

chatton (approved):
LGTM!

AdityaSripal (approved):

Issue #1622, opened 2022-06-29
Implement `sdk.Msg` interface for channel upgrade handshake messages


damiannolan:
## Summary

Implement the `sdk.Msg` interface for the `04-channel` upgrade handshake messages.
Add the `sdk.Msg` methods to the `04-channel` [msgs.go](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/msgs.go) file.

- [x] `MsgChannelUpgradeInit` (@charleenfei)
  - `ValidateBasic`
  - `GetSigners`
- [x] `MsgChannelUpgradeTry` (@charleenfei)
  - `ValidateBasic`
  - `GetSigners`
- [x] `MsgChannelUpgradeAck` (@charleenfei)
  - `ValidateBasic`
  - `GetSigners`
- [x] `MsgChannelUpgradeConfirm`  (@chatton)
  - `ValidateBasic`
  - `GetSigners`
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
Preferably split the work into 4 PRs, one per message type

chatton:
I will pick up `MsgChannelUpgradeConfirm`

PR #1836, opened 2022-07-31
chore: adding MsgChannelUpgradeTry fields and sdk.Msg interface

charleenfei:
## Description

addresses #1622 

- Adding fields to MsgChannelUpgradeTry
- Implement ValidateBasic and GetSigners methods for sdk.Msg interface
- Adding tests

---

write a little note why.


modules/core/04-channel/types/msgs.go
```diff
@@ -517,18 +517,97 @@ var _ sdk.Msg = &MsgChannelUpgradeTry{}
...
+
+	if msg.ProposedUpgradeChannel.State != TRYUPGRADE {
+		return sdkerrors.Wrapf(ErrInvalidChannelState, "expected: %s, got: %s", TRYUPGRADE, msg.CounterpartyChannel.State)
+	}
+
+	if msg.TimeoutHeight.IsZero() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidHeight, "timeout height must be non-zero")
+	}
+
+	if msg.TimeoutTimestamp == 0 {
```
charleenfei, line 575: the spec has both timeout timestamp and timeout height, but do both of these need to have a value or just one?

modules/core/04-channel/types/msgs.go
```diff
@@ -517,18 +517,97 @@ var _ sdk.Msg = &MsgChannelUpgradeTry{}
...
+
+	if !IsValidChannelID(msg.ChannelId) {
+		return ErrInvalidChannelIdentifier
+	}
+
+	if msg.CounterpartyChannel.State != INITUPGRADE {
+		return sdkerrors.Wrapf(ErrInvalidChannelState, "expected: %s, got: %s", INITUPGRADE, msg.CounterpartyChannel.State)
+	}
+
+	if msg.CounterpartySequence == 0 {
```
charleenfei, line 563: our sequences are not zero-indexed right?

modules/core/04-channel/types/msgs.go
crodriguezvega: I see this part of pseudocode in the spec in the `chanUpgradeTry` function:

```golang
// either timeout height or timestamp must be non-zero
// if the upgrade feature is implemented on the TRY chain, then a relayer may submit a TRY transaction after the timeout.
// this will restore the channel on the executing chain and allow counterparty to use the ChannelUpgradeCancelMsg to restore their channel.
if timeoutHeight == 0 && timeoutTimestamp == 0 {
    restoreChannel()
    return
}
```

which would indicate that both can be zero. 

Would the relayer submit both zero timeouts in a try message after a timeout was processed on the side that initialized the upgrade to restore the original channel on the counterparty? @AdityaSripal

crodriguezvega: Although on `chanOpenTry` there's a verification that the init side has stored the upgrade timeout:

```golang
abortTransactionUnless(verifyChannelUpgradeTimeout(connection, proofHeight, proofUpgradeTimeout, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, upgradeTimeout)) 
```

And on `chanOpenInit` there's a condition to abort unless one of the timeouts is zero:

```
// either timeout height or timestamp must be non-zero
abortTransactionUnless(counterpartyTimeoutHeight != 0 || counterpartyTimeoutTimestamp != 0)
```

So how would it be possible to reach this check on `chanUpgradeTry` then?

```golang
// either timeout height or timestamp must be non-zero
// if the upgrade feature is implemented on the TRY chain, then a relayer may submit a TRY transaction after the timeout.
// this will restore the channel on the executing chain and allow counterparty to use the ChannelUpgradeCancelMsg to restore their channel.
if timeoutHeight == 0 && timeoutTimestamp == 0 {
    restoreChannel()
    return
}
```

modules/core/04-channel/types/msgs.go
crodriguezvega: Yeah, one of the items from the kick-off call was to initialize the sequence to 1 in `chanUpgradeInit` if there was no upgrade sequence stored before; and if there was a sequence stored, then increment it.

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -276,6 +276,23 @@ message MsgChannelUpgradeInitResponse {}
 message MsgChannelUpgradeTry {
   option (gogoproto.equal)           = false;
   option (gogoproto.goproto_getters) = false;
+
+  string  port_id              = 1 [(gogoproto.moretags) = "yaml:\"port_id\""];
```
seantking, line 280: same nit here for the proto file formatting

modules/core/04-channel/types/msgs.go
```diff
@@ -517,18 +517,97 @@ var _ sdk.Msg = &MsgChannelUpgradeTry{}
...
+		return sdkerrors.Wrapf(ErrInvalidChannelState, "expected: %s, got: %s", TRYUPGRADE, msg.CounterpartyChannel.State)
+	}
+
+	if msg.TimeoutHeight.IsZero() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidHeight, "timeout height must be non-zero")
+	}
+
+	if msg.TimeoutTimestamp == 0 {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidHeight, "timeout timestamp must be non-zero")
+	}
```
damiannolan, line 577: I think these checks should grouped with a logical AND:

```go
if msg.TimeoutHeight.IsZero() && msg.TimeoutTimestamp == 0 {
	return sdkerrors.Wrap(sdkerrors.ErrInvalidUpgradeTimeout, "either timeout height or timeout timestamp must be non-zero")
}
```

We could add a new channel error `ErrInvalidUpgradeTimeout`

AdityaSripal (approved):
Nice work!

modules/core/04-channel/types/msgs.go
seantking:
```
So how would it be possible to reach this check on chanUpgradeTry then?
```

@crodriguezvega I guess technically the source chain could have a different/forked implementation for the `UPGRADE_INIT` step.

damiannolan (approved):
Nice work!

seantking (approved):

PR #1907, opened 2022-08-08
fix: prevent blocked addresses from sending ICS 20 transfers

colin-axner:
## Description

Prevent IBC transfers from being sent by a blocked module account address. If the module account address is blocked it would not be capable of receiving a refund and thus should not be allowed to send IBC transfers. Module accounts which would like to interfact with ICS 20 should ensure they do not need to be blocked

closes: #XXXX

---

write a little note why.


modules/apps/transfer/keeper/relay.go
```diff
@@ -48,6 +48,8 @@ import (
 // 4. A -> C : sender chain is sink zone. Denom upon receiving: 'C/B/denom'
 // 5. C -> B : sender chain is sink zone. Denom upon receiving: 'B/denom'
 // 6. B -> A : sender chain is sink zone. Denom upon receiving: 'denom'
+//
+// Note: An IBC Transfer must be initiated using a MsgTransfer via the Transfer rpc handler
 func (k Keeper) SendTransfer(
```
colin-axner, line 53: I will do a followup pr to make this private for the next major release

damiannolan (approved):

modules/apps/transfer/keeper/relay.go
```diff
@@ -62,6 +64,10 @@ func (k Keeper) SendTransfer(
 		return types.ErrSendDisabled
```
crodriguezvega, line 64: This is a nit and no need to block the PR on this, but now that you added tests for the msg server, maybe we can also had a test case for the situation where send is disabled.

crodriguezvega (approved):
Thank you, @colin-axner!

modules/apps/transfer/keeper/relay.go
colin-axner: looks like this test case is missing for receive as well. Would prefer it be done in a separate pr

chatton (approved):
LGTM

seantking (approved):

Issue #1963, opened 2022-08-10
E2E: Successful token transfer


chatton:
## Summary

Add a basic E2E test which ensures a basic token transfer is successful.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1973, opened 2022-08-10
feat(test): add successful e2e transfer test

colin-axner:
## Description

closes: #1963

---

write a little note why.

### DIFF
--- a/e2e/testsuite/testsuite.go
+++ b/e2e/testsuite/testsuite.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/cosmos/ibc-go/e2e/testconfig"
 	feetypes "github.com/cosmos/ibc-go/v5/modules/apps/29-fee/types"
+	channeltypes "github.com/cosmos/ibc-go/v5/modules/core/04-channel/types"
 )
 
 const (
@@ -46,7 +47,8 @@ type E2ETestSuite struct {
 // These should typically be used for query clients only. If we need to make changes, we should
 // use E2ETestSuite.BroadcastMessages to broadcast transactions instead.
 type GRPCClients struct {
-	FeeQueryClient feetypes.QueryClient
+	FeeQueryClient     feetypes.QueryClient
+	ChannelQueryClient channeltypes.QueryClient
 }
 
 // path is a pairing of two chains which will be used in a test.
@@ -274,7 +276,8 @@ func (s *E2ETestSuite) initGRPCClients(chain *cosmos.CosmosChain) {
 	}
 
 	s.grpcClients[chain.Config().ChainID] = GRPCClients{
-		FeeQueryClient: feetypes.NewQueryClient(grpcConn),
+		ChannelQueryClient: channeltypes.NewQueryClient(grpcConn),
+		FeeQueryClient:     feetypes.NewQueryClient(grpcConn),
 	}
 }
 
--- a/e2e/testsuite/testsuite.go
+++ b/e2e/testsuite/testsuite.go
@@ -292,6 +292,13 @@ func (s *E2ETestSuite) AssertValidTxResponse(resp sdk.TxResponse) {
 	s.Require().NotEmpty(resp.Data, respLogsMsg)
 }
 
+// AssertPacketRelayed asserts that the packet commitment does not exist on the sending chain.
+// The packet commitment will be deleted upon a packet acknowledgement or timeout.
+func (s *E2ETestSuite) AssertPacketRelayed(ctx context.Context, chain *cosmos.CosmosChain, portID, channelID string, sequence uint64) {
+	commitment, _ := s.QueryPacketCommitment(ctx, chain, portID, channelID, sequence)
+	s.Require().Empty(commitment)
+}
+
 // createCosmosChains creates two separate chains in docker containers.
 // test and can be retrieved with GetChains.
 func (s *E2ETestSuite) createCosmosChains(chainOptions testconfig.ChainOptions) (*cosmos.CosmosChain, *cosmos.CosmosChain) {
--- a/e2e/transfer_test.go
+++ b/e2e/transfer_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cosmos/ibc-go/e2e/testvalues"
 	transfertypes "github.com/cosmos/ibc-go/v5/modules/apps/transfer/types"
 	clienttypes "github.com/cosmos/ibc-go/v5/modules/core/02-client/types"
+	channeltypes "github.com/cosmos/ibc-go/v5/modules/core/04-channel/types"
 )
 
 func TestTransferTestSuite(t *testing.T) {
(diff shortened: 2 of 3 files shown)


e2e/transfer_test.go
```diff
@@ -0,0 +1,132 @@
...
+func (s *TransferTestSuite) Transfer(ctx context.Context, chain *cosmos.CosmosChain, user *ibctest.User,
+	portID, channelID string, token sdk.Coin, sender, receiver string, timeoutHeight clienttypes.Height, timeoutTimestamp uint64,
+) (sdk.TxResponse, error) {
+	msg := transfertypes.NewMsgTransfer(portID, channelID, token, sender, receiver, timeoutHeight, timeoutTimestamp)
+	return s.BroadcastMessages(ctx, chain, user, msg)
+}
+
+// TestMsgTransfer_Succeeds_Nonincentivized will test sending successful IBC transfers from chainA to chainB.
+// The transfer will occur over a basic transfer channel (non incentivized) and both native and non-native tokens
+// will be sent forwards and backwards in the IBC transfer timeline (both chains will act as source and receiver chains).
```
colin-axner, line 39: I thought about writing a test where chainA and chainB both send do the native denomination and non-native denomination send back, but I think the test suite is currently setup to expect tests to happen via chainA -> chainB interaction and then we should ensure during cross compatiblity that we test both sides ie
v5 -> v2
v2 -> v5

chatton: there's nothing stopping us from doing it whichever way we want, in this case it would just mean that we would run one test and it would cover both directions. 

We should be able to send from chainB -> chainA just as easily as chainA -> chainB unless I'm missing something!

colin-axner: Yup, I think the main area of uncertainty for me is why channelB isn't returned. I think that's what made me think the test suite was meant to be unidirectional rather than bidirectional. 

I think it might be best to aim for all tests to be written bidirectional. It's a bit extra work, but then we won't miss a bug because we forget to run an e2e test
v2 -> v5
v5 -> v2

Do we think we could modify the e2e test suite to have some sort of structure that wraps channelA/channelB. In the testing package we have [path](https://github.com/cosmos/ibc-go/blob/main/testing/path.go#L11) which contains references to an endpoint which has the (client/connection/channel info)

colin-axner: After some more thought, I think it is better to write a unidirectional testing pattern. By that I mean, focusing on the interaction of chainA to chainB and forcing the tests to be run in both directions (if relevant)

So for this test:
v2 -> v5 should be run
v5 -> v2 should be run

If we add a ics20-2 version in the future. We will likely add two sorts of tests with this v2
- tests expecting chainB to support ics20-2
- tests expecting chainB not to support ics20-2

I think we could introduce logic to automatically configure this when writing the test suite (ie indicate the version range chainB should use, try running both sides with the v2/v5 binaries and skip the test if the version range check isn't met), but we could also do this manually in the short term

I think it makes more sense to indicate what versions chainA should have and what versions chainB should have and then run every combination rather than to perform the chainA/chainB logic twice, as it gets more convoluted when you trying adding a third chain or more specific version handling

chatton: Yeah I think I like the idea of each test focusing on one direction. Re the channelB not being returned, the use cases I had found so far had been satisfied by just using `channelA.CounterParty` but I think it's completely reasonable to return both channelA and channelB if that is needed.

If tests are unidirectional, we can do any sort of configuration of permutations outside of the tests themselves which is quite powerful. We can define the rules however we like and run the same tests with different versions as appropriate.

e2e/transfer_test.go
```diff
@@ -0,0 +1,152 @@
...
+	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+	chainAAddress := chainAWallet.Bech32Address(chainA.Config().Bech32Prefix)
+
+	chainBWallet := s.CreateUserOnChainB(ctx, testvalues.StartingTokenAmount)
+	chainBAddress := chainBWallet.Bech32Address(chainB.Config().Bech32Prefix)
+
+	t.Run("relayer wallets recovered", func(t *testing.T) {
+		err := s.RecoverRelayerWallets(ctx, relayer)
+		s.Require().NoError(err)
+	})
```
chatton, line 73: I believe we don't need this. In the Fee Middleware tests, we recover the relayer wallets as we want to broadcast messages on behalf of relayer users recovering wallets adds the corresponding entries to the test keychain and so enables this. 

In this test we don't need to do that as we only broadcast transactions from the created wallets, not the relayer ones.

e2e/transfer_test.go
colin-axner: Access to channelB just allows us to use `ibc.ChannelOutputs` as a function argument instead of passing `channelA.Counterparty.PortID, channelA.Counterparty.ChannelID`

seantking (changes requested):
I think we should also add an entry [here](
https://github.com/cosmos/ibc-go/pull/1992/files#diff-c753b66269ec4c6adfc34edb8f894676a39b4d1b2d4b95521ba132427540fc9eR14 ) for manual testing.

e2e/transfer_test.go
chatton: Great point, I'm happy to have the signature updated to return both `ibc.ChannelOutputs`

seantking (approved):

e2e/testsuite/testsuite.go
```diff
@@ -292,6 +292,13 @@ func (s *E2ETestSuite) AssertValidTxResponse(resp sdk.TxResponse) {
 	s.Require().NotEmpty(resp.Data, respLogsMsg)
 }
 
+// AssertPacketRelayed asserts that the packet commitment does not exist on the sending chain.
+// The packet commitment will be deleted upon a packet acknowledgement or timeout.
+func (s *E2ETestSuite) AssertPacketRelayed(ctx context.Context, chain *cosmos.CosmosChain, portID, channelID string, sequence uint64) {
+	commitment, _ := s.QueryPacketCommitment(ctx, chain, portID, channelID, sequence)
+	s.Require().Empty(commitment)
+}
```
chatton, line 300: 👍

chatton (approved):
LGTM! Great work @colin-axner

e2e/transfer_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+	chainAAddress := chainAWallet.Bech32Address(chainA.Config().Bech32Prefix)
+
+	chainBWallet := s.CreateUserOnChainB(ctx, testvalues.StartingTokenAmount)
+	chainBAddress := chainBWallet.Bech32Address(chainB.Config().Bech32Prefix)
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 1, chainA, chainB), "failed to wait for blocks")
+
+	t.Run("native IBC token transfer from chainA (source) to chainB, sender is source", func(t *testing.T) {
```
damiannolan, line 61: nit: minor improvement maybe? feel free to disregard

```suggestion
	t.Run("native IBC token transfer from chainA (source) to chainB (destination), sender is source", func(t *testing.T) {
```

e2e/transfer_test.go
```diff
@@ -0,0 +1,129 @@
...
+		s.Require().Equal(expected, actualBalance)
+	})
+}
+
+// transferChannelOptions configures both of the chains to have non-incentivized transfer channels.
+func transferChannelOptions() func(options *ibc.CreateChannelOptions) {
+	return func(opts *ibc.CreateChannelOptions) {
+		opts.Version = "ics20-1"
+		opts.SourcePortName = "transfer"
+		opts.DestPortName = "transfer"
```
damiannolan, line 122: Should we use `transfertypes.Version` and `transfertypes.PortID` here?

e2e/transfer_test.go
damiannolan: Thanks for this thread!

damiannolan (approved):
Looks great! I just left some very minor suggestions but feel free to disregard them! LGTM 🚀

crodriguezvega (approved):
😎

e2e/transfer_test.go
colin-axner: great call!

e2e/transfer_test.go
colin-axner: source was in reference to the tokens origination, not where the tokens were being sent from

Issue #1660, opened 2022-07-05
Allow callback to IBC app callers


AdityaSripal:
IBC was designed with callbacks between core IBC and IBC applications. IBC apps would send a packet to core IBC. When the result of the packet lifecycle eventually resolved into either an acknowledgement or a timeout, the core IBC called a callback on the IBC application so that the IBC application could take action on the basis of the result (e.g. unescrow tokens for ICS-20)

We are now seeing the desire for secondary applications to call into IBC apps as part of their state machine logic and then do some actions on the basis of the packet result.

E.g. Send an ICS-20 packet, and if it is successful, then send an ICA-packet to swap tokens on LP and return funds to sender

This requires a second layer of callbacks. The IBC application already gets the result of the packet from core IBC, but currently there is no standardized way to pass this information on to a caller module/smart contract.

Of course this isn't an issue in the IBC protocol since it is a matter of IBC applications getting this information to their own callers (it is one layer above IBC). However standardizing on the best approach will allow callers to implement a standard interface to interact with all IBC apps.

---

Proposal:

Introduce an interface that accounts can optionally implement:

```go
interface IBCCallerAccount {
    onAcknowledgement(portID, channelID string, sequence uint64, ack exported.Acknowledgement)
    onTimeoutPacket(portID, channelID string, sequence uint64)
}
```

So far, IBC applications have different APIs for their packet sending logic.

Transfer has `Transfer`, ICA has `SendTx`. I think this is fine since IBC apps do not want to nakedly pass through packets, and they will need to have different APIs for their different purposes.

However, I think we should standardize on the `MsgResponse` which should return `channelID`, `portID`, `sequence` of the packet that gets sent by IBC.

```go
MsgResponse {
    packetID: PacketID,
    // app-custom fields
}
```

IBC applications are also encouraged to accept a sender address field in all of their APIs, and to store this for a particular packet.

The caller application can then call the desired API with their sender address. They can then get back the `packetID` and store any information they desire keyed on the packetID.

Later, the packet lifecycle resolves. Core IBC calls the IBC app callback. And if the sender is a `IBCCallerAccount`, then the IBC application can call the callback `onAcknowledgmentPacket` or `onTimeoutPacket` with the relevant arguments.

The IBC caller, gets the packet ID and the result of the packet. They can use this along with any information they stored to execute further logic.

This solves many use cases of the form: `I want to transfer X tokens, and then do Y after programmatically`

----

Open Questions:

[...]

Issue #1918, opened 2022-08-08
make `SendTransfer` private


colin-axner:
## Summary

We should make the rpc handler `Transfer` be the only entry point for sending ibc transfers

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #1976, opened 2022-08-10
ADR 008: IBC Actor Callbacks

AdityaSripal:
## Description

The following is a standardized interface for IBC actors to expose an interface that can be called by IBC applications/middleware to route packet/channel responses back to the original actor so that it can perform subsequent logic

closes: #1660 

---

write a little note why.


docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+// this is the module-defined SendPacket function. It may differ from application to application
+// e.g. For ICS20 this would be the SendTransfer function
+func SendPacket(ctx sdk.Context, packet channeltypes.Packet, caller IBCaller) {
+    // do custom logic and send packet
+
+    // store a mapping of the packetID to the caller
+    k.setCaller(packetID, caller)
+}
+
+func OnAcknowledgementPacket(
```
crodriguezvega, line 66: Thought it would be good for extra clarification...

```suggestion
func (im IBCModule) OnAcknowledgementPacket(
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+    relayer string,
+) {
+    // application-specific onAcknowledgmentPacket logic
+
+    // unmarshal ack bytes into the acknowledgment interface
+    var ack exported.Acknowledgement
+    unmarshal(acknowledgement, ack)
+
+    // send acknowledgement to original caller
+    caller := k.getCaller(packetID)
```
crodriguezvega, line 79:
```suggestion
    caller := im.keeper.getCaller(packetID)
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+    // store a mapping of the packetID to the caller
+    k.setCaller(packetID, caller)
+}
+
+func OnAcknowledgementPacket(
+    ctx sdk.Context,
+    packet channeltypes.Packet,
+    acknowledgement []byte,
+    relayer string,
+) {
```
crodriguezvega, line 71:
```suggestion
) error {
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+    // unmarshal ack bytes into the acknowledgment interface
+    var ack exported.Acknowledgement
+    unmarshal(acknowledgement, ack)
+
+    // send acknowledgement to original caller
+    caller := k.getCaller(packetID)
+    caller.OnAcknowledgmentPacket(ctx, packet, ack)
+}
+
+func OnTimeoutPacket(
```
crodriguezvega, line 83:
```suggestion
func (im IBCModule) OnTimeoutPacket(
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+    ctx sdk.Context,
+    packet channeltypes.Packet,
+    relayer string,
+) {
```
crodriguezvega, line 87:
```suggestion
) error {
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+    // application-specific onTimeoutPacket logic
+
+    // call timeout callback on original caller
+    caller := k.getCaller(packetID)
```
crodriguezvega, line 91:
```suggestion
    caller := im.keeper.getCaller(packetID)
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+    caller.OnTimeoutPacket(ctx, packet)
+}
+```
+
+## Consequences
+
+### Positive
+
+- IBC Callers can now programatically execute logic that involves sending a packet and then performing some additional logic once the packet lifecycle is complete
```
crodriguezvega, line 100:
```suggestion
- IBC callers can now programatically execute logic that involves sending a packet and then performing some additional logic once the packet lifecycle is complete.
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+- Leverages the same callback architecture used between core IBC and IBC applications
```
crodriguezvega, line 101:
```suggestion
- Leverages the same callback architecture used between core IBC and IBC applications.
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+    )
+}
+```
+
+IBC Apps or middleware can then call the IBCCaller callbacks like so in their own callbacks:
+
+```go
+// this is the module-defined SendPacket function. It may differ from application to application
+// e.g. For ICS20 this would be the SendTransfer function
+func SendPacket(ctx sdk.Context, packet channeltypes.Packet, caller IBCaller) {
```
crodriguezvega, line 59:
```suggestion
func SendPacket(ctx sdk.Context, packet channeltypes.Packet, caller IBCCaller) {
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+    // OnTimeoutPacket will be called on the IBC Caller
+    // after the IBC Application handles its own OnTimeoutPacket callback
+    OnTimeoutPacket(
+        ctx sdk.Context,
+        packet channeltypes.Packet
+    )
+}
+```
+
+IBC Apps or middleware can then call the IBCCaller callbacks like so in their own callbacks:
```
crodriguezvega, line 54:
```suggestion
IBC Apps or middleware can then call the `IBCCaller` callbacks like so in their own callbacks:
```

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+
+```go
+// this is the module-defined SendPacket function. It may differ from application to application
+// e.g. For ICS20 this would be the SendTransfer function
```
crodriguezvega, line 58: So would you mind elaborating how this would be wired for `SendTransfer` for example? How is the caller passed in the argument? An app would be built and would call `SendTransfer` with the caller argument? This means we need to adjust the signature of `SendTransfer`, right? And then we wouldn't be able to [make it private](https://github.com/cosmos/ibc-go/issues/1918) anymore, I guess?

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+This requires a second layer of callbacks. The IBC application already gets the result of the packet from core IBC, but currently there is no standardized way to pass this information on to a caller module/smart contract.
+
+## Decision
+
+Create a standardized callback interface that callers can implement. IBC applications (or middleware that wraps IBC applications) can now call this callback to route the result of the packet/channel handshake from core IBC to the IBC application to the original caller.
+
+IBC callers may implement the following interface:
+
+```go
+type IBCCaller interface {
```
crodriguezvega, line 27: Should we add and `error` return type to these functions in case something happens?

crodriguezvega (commented):
Nice, @AdityaSripal!

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+) {
+    // application-specific onAcknowledgmentPacket logic
+
+    // unmarshal ack bytes into the acknowledgment interface
+    var ack exported.Acknowledgement
+    unmarshal(acknowledgement, ack)
+
+    // send acknowledgement to original caller
+    caller := k.getCaller(packetID)
+    caller.OnAcknowledgmentPacket(ctx, packet, ack)
```
crodriguezvega, line 80: Would it be possible to do something like `deleteCaller(packetID)` if the packet lifecycle completes and we don't need the caller for this particular packet to be stored in state anymore? Maybe that would help with cleaning up? 

Same thing for the timeout callback...

docs/architecture/adr-008-app-caller-cbs.md
asalzmann: Afaik, `SendPacket` isn't executed for all IBC modules. For example, ICA calls `SendTx` directly, which calls the underlying channel keeper's `SendPacket`. 

https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/ibc_middleware.go#L180

One way to get around this is to have modules implementing this set the caller at the call site (e.g. below `SendTx`, which returns a sequence number). This gets slightly tricky though, because `SendTransfer` doesn't return a sequence number, so you would have to set the caller in `SendPacket`. This is an issue we ran into designing this as well, so curious to hear your thoughts!

docs/architecture/adr-008-app-caller-cbs.md
puneet2019: Can we also have `OnRecvPacket`? 

One use I can think of which I might want is..
User ibc-transfers atom from cosmoshub to persistence chain - without xprt on the origin wallet the user cannot do anything with the atom.
If there were a callback that uses some sort of swap on persistence for transaction fees worth of xprt OnRecvPacket, that could make UX easier

AdityaSripal: Agreed that this would be useful! It's not quite in scope for this ADR since the receiver address is not the caller of the IBC application.

Also, there is no generic definition of receiver address for IBC packets since they may not include one

AdityaSripal: > User ibc-transfers atom from cosmoshub to persistence chain - without xprt on the origin wallet the user cannot do anything with the atom.
If there were a callback that uses some sort of swap on persistence for transaction fees worth of xprt OnRecvPacket, that could make UX easier

This usecase in particular does not require this ADR. Please see [middleware guide](https://github.com/cosmos/ibc-go/blob/main/docs/ibc/middleware/develop.md)

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,111 @@
...
+IBC Apps or middleware can then call the IBCCaller callbacks like so in their own callbacks:
+
+```go
+// this is the module-defined SendPacket function. It may differ from application to application
+// e.g. For ICS20 this would be the SendTransfer function
+func SendPacket(ctx sdk.Context, packet channeltypes.Packet, caller IBCaller) {
+    // do custom logic and send packet
+
+    // store a mapping of the packetID to the caller
+    k.setCaller(packetID, caller)
```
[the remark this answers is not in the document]
AdityaSripal: Fixed so this is no longer a concern

docs/architecture/adr-008-app-caller-cbs.md
AdityaSripal: I changed the semantics of this ADR so that it is back in scope and specified in this ADR. Thanks for the feedback @puneet2019 !

docs/architecture/adr-008-app-caller-cbs.md
AdityaSripal: done in the new model

AdityaSripal:
We can very easily just add the interfaces in ibc-go and break the necessary APIs.

This will then unblock CosmWasm team and evmos team from writing middleware to implement ADR-8 on their ends which will be the main usecase for the foreseeable future

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,249 @@
...
+The Packet interface will get extended to add `GetSender` and `GetReceiver` methods. These may return an address
+or they may return the empty string. The address may reference an IBCActor or it may be a regular user address. Any 
+IBC application or middleware that uses these methods must handle these cases.
+
+```go
+type Packet interface {
+    // existing Packet methods
+
+    // may return the empty string
+    GetSender() string
```
colin-axner, line 77: Where does the non-empty value come from?

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,249 @@
...
+func OnChanOpenInit(
+    ctx sdk.Context,
+    order channeltypes.Order,
+    connectionHops []string,
+    portID string,
+    channelID string,
+    channelCap *capabilitytypes.Capability,
+    counterparty channeltypes.Counterparty,
+    version string,
+    actor string,
```
colin-axner, line 100: Where does this value come from? MsgChannelOpenInit sender?

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,249 @@
...
+) (string, error) {
+    acc := k.getAccount(ctx, actor)
+    ibcActor, ok := acc.(IBCActor)
+    if ok {
+        k.setActor(ctx, portID, channelID, actor)
+    }
```
colin-axner, line 106: Can you provide justification for why core IBC doesn't do this. It makes sense to me, to avoid creating too much core logic, but should still be discussed. It may not be apparent to the future reader

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,249 @@
...
+    // run any necessary logic first
+    // negotiate final version
+
+    actor := k.getActor(ctx, portID, channelID)
+    if actor != "" {
+        ibcActor, _ := acc.(IBCActor)
+        ibcActor.OnChanOpen(ctx, portID, channelID, version)
+    }
+    // cleanup state
+    k.deleteActor(ctx, portID, channelID)
```
colin-axner, line 127: Why do you delete the actor?

docs/architecture/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,249 @@
...
+    ctx sdk.Context,
+    portID,
+    channelID,
+    actor string,
+) error {
+    acc := k.getAccount(ctx, actor)
+    ibcActor, ok := acc.(IBCActor)
+    if ok {
+        k.setActor(ctx, portID, channelID, actor)
+    }
```
colin-axner, line 157: I don't understand this? OnChanCloseConfirm will never be called on the same chain that calls OnChanCloseConfirm. The setting of the actor seems unnecessary

docs/architecture/adr-008-app-caller-cbs.md
colin-axner, line 106: Can you provide a use case here? This is defining a single actor to handle channel creation and many actors to handle packet sends. Would like to understand why we should support the single actor cases on channel handshakes

colin-axner (commented):
Ay! LGTM, I don't understand how this actually supposed to be implemented though. The addition of `packet.GetSender()` and `packet.GetReceiver()` seems more involved than noted in the ADR. There's no existing structure for us to implement these interface functions

Do you suggest adding new fields to the packet type? Or are you suggesting adding interface functions to packet data? I think the latter would actually be quite difficult due to marshalling interfaces as Any's and the fact that core IBC currently expects packet data as already marshalled

aidan-calculated:
Does this PR enable callbacks for an ibc transfer so an action can be taken on a successful transfer of a token?

womensrights:
yes @aidan-calculated that is the intended use case

docs/architecture/adr-008-app-caller-cbs.md
aidan-calculated: Wouldn't the non-empty value come from the address that sent the IBC message, and the address that's going to receive the IBC message?

docs/architecture/adr-008-app-caller-cbs.md
aidan-calculated: Actors are a concern at the app level, it would be mixing conerns if this logic was put into core.

aidan-calculated (commented):
This feature is really vital to the growth of the app-chain and interchain future of the cosmos eco system.

More people will develop contracts than app-chains, so we should work on amazing feature asap!

colin-axner:
I believe it is possible to implement ADR 008 now but with the caveat that it must add a switch statement for the supported packet data types. It was determined that the generalized approach required changes to the middleware structure

@AdityaSripal could you update this ADR with the short term approach?

AdityaSripal:
Updated from discussions with Osmosis.

The Packet actor and packet method callbacks are way more relevant.

The handshake callback error handling has been left aside for the moment, since there aren't current users to inform the full design

colin-axner (commented):
Thanks for all the work you've done for ADR 8 :heart: 

Conceptually it all makes sense to me and seems rock solid/thorough design. I've jotted down several questions that came to mind rereviewing this ADR 8, but most my comments are clarifications/code hygenie

It might make sense to split up this ADR into 2 sections:
- changes ibc-go needs
- implementation guide for ADR 8 implementors 

For the most part, this is already the case, but it might make sense to make it explicit.

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,374 @@
...
+- Actor: an actor is an on-chain module (this may be a hardcoded module in the chain binary or a smart contract) that wishes to execute custom logic whenever IBC receives a packet flow that it has either sent or received. It **must** be addressable by a string value.
+
+## Decision
+
+Create a standardized callback interface that actors can implement. IBC applications (or middleware that wraps IBC applications) can now call this callback to route the result of the packet/channel handshake from core IBC to the IBC application to the original actor on the sending chain. IBC applications can route the packet receipt to the destination actor on the receiving chain.
+
+IBC actors may implement the following interface:
+
+```go
+type IBCActor interface {
```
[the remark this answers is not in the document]
AdityaSripal: Yes eventually. The channel handshake callbacks are not going to be implemented in the short term

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    acc := k.getAccount(ctx, cbPacketData.GetDstCallbackAddress())
+    ibcActor, ok := acc.(IBCActor)
+    if ok {
+        // set gas limit for callback
+        gasLimit := getGasLimit(ctx, cbPacketData)
+        cbCtx = ctx.WithGasLimit(gasLimit)
+
+        err := ibcActor.OnRecvPacket(cbCtx, packet, relayer)
```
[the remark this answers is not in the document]
AdityaSripal: The packetData should contain all information necessary. It is up to the contract in question to unmarshal and retrieve the information appropriately if it expects the data to be in a particular format

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,374 @@
...
+        // create cached context with gas limit
+        cacheCtx, writeFn := ctx.CacheContext()
+        cacheCtx = cacheCtx.WithGasLimit(gasLimit)
+        
+        defer func() {
+            if e := recover(); e != nil {
+                log("ran out of gas in callback. reverting callback state")
+            } else {
+                // only write callback state if we did not panic during execution
+                writeFn()
```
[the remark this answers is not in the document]
AdityaSripal: No I think the pseudocode is written such that it would only write if the error is nil?

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,374 @@
...
+            }
+        }
+
+        err := ibcActor.OnAcknowledgementPacket(cacheCtx, packet, ack, relayer)
+
+        // deduct consumed gas from original context
+        ctx = ctx.WithGasLimit(ctx.GasMeter().RemainingGas() - cbCtx.GasMeter().GasConsumed())
+
+        setAckCallbackError(ctx, packet, err)
```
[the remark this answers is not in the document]
AdityaSripal: It is set so that the user/smart contract can retrieve it. It is not used by the protocol itself

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,374 @@
...
+    ) error
+
+    // OnTimeoutPacket will be called on the IBC Actor
+    // after the IBC Application handles its own OnTimeoutPacket callback
+    OnTimeoutPacket(
+        ctx sdk.Context,
+        packet channeltypes.Packet,
+        relayer string
+    ) error
+}
```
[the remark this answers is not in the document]
AdityaSripal: These interfaces don't actually exist in #3287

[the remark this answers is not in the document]
damiannolan: I am all for reducing the [IBCModule](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/types/module.go#L14) interface down into smaller composable interfaces! 👍

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,374 @@
...
+    // Otherwise, we will set the gas limit passed into the callback to the `min(ctx.GasLimit, UserDefinedGasLimit())`
+    UserDefinedGasLimit() uint64
+}
+```
+
+IBC Apps or middleware can then call the IBCActor callbacks like so in their own callbacks:
+
+### Handshake Callbacks
+
+The handshake init callbacks (`OnChanOpenInit` and `OnChanCloseInit`) will need to include an additional field so that the initiating actor can be tracked and called upon during handshake completion.
```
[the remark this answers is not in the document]
damiannolan: Hmm, could be possible to reuse the `signer` field if it was such that the `actor` is the sender of the msg. Would be nicer than adding more fields and non api breaking..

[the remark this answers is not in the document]
damiannolan: actually would still require breaking the api at the 04-channel level as we don't pass along the signer field atm

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,376 @@
...
+    ctx sdk.Context,
+    portID,
+    channelID,
+    actor string,
+) error {
+    acc := k.getAccount(ctx, actor)
+    ibcActor, ok := acc.(IBCActor)
+    if ok {
+        k.setActor(ctx, portID, channelID, actor)
+    }
```
damiannolan, line 175: seems odd to set a different actor for the closing handshake steps rather than just reusing what's already there. i.e. not deleting the state entry in ACK/CONFIRM.

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,376 @@
...
+) (ack exported.Acknowledgement) {
+    // run any necesssary logic first
+    // IBCActor logic will postprocess
+
+    // unmarshal packet data into expected interface
+    var cbPacketData callbackPacketData
+    unmarshalInterface(packet.GetData(), cbPacketData)
+    if cbPacketData == nil {
+        return
+    }
```
damiannolan, line 218: shouldn't this `return` rather be `return app.OnRecvPacket(...)`? maybe it's supposed to be implicit?

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,376 @@
...
+### Positive
+
+- IBC Actors can now programatically execute logic that involves sending a packet and then performing some additional logic once the packet lifecycle is complete
+- Middleware implementing ADR-8 can be generally used for any application
+- Leverages the same callback architecture used between core IBC and IBC applications
+
+### Negative
+
+- Callbacks may now have unbounded gas consumption since the actor may execute arbitrary logic. Chains implementing this feature should take care to place limitations on how much gas an actor callback can consume.
+- Application packets that want to support ADR-8 must additionally have their packet data implement the `CallbackPacketData` interface and register their implementation on the chain codec
```
damiannolan, line 370: Personally, I'd move this to the `Neutral` section 🤷

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,376 @@
...
+    acknowledgement []byte,
+    relayer string,
+) error {
+    // application-specific onAcknowledgmentPacket logic
+
+    // unmarshal packet data into expected interface
+    var cbPacketData callbackPacketData
+    unmarshalInterface(packet.GetData(), cbPacketData)
+    if cbPacketData == nil {
+        return
```
damiannolan, line 256: ditto on implicit/explicit about passing through to app rather than silently returning

damiannolan (approved):
Going to leave my approval now, would be great to get this merged. I think it would be fine to make any amendments in follow ups as the middleware implementation is fleshed out.

Thank you @AdityaSripal for all of the time and effort spent on this! ❤️

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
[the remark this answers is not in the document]
chatton: Big fan of splitting up larger interfaces into smaller ones and accepting narrower scoped interfaces in method signatures where possible!

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,376 @@
...
+
+```go
+// Implemented by any packet data type that wants to support
+// PacketActor callbacks
+type CallbackPacketData interface {
+    // may return the empty string
+    GetSrcCallbackAddress() string
+
+    // may return the empty string
+    GetDestCallbackAddress() string
```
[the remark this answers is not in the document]
chatton: I think this is worth doing

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,376 @@
...
+    ibcActor, ok := acc.(IBCActor)
+    if ok {
+        gasLimit := getGasLimit(ctx, cbPacketData)
+
+        // create cached context with gas limit
+        cacheCtx, writeFn := ctx.CacheContext()
+        cacheCtx = cacheCtx.WithGasLimit(gasLimit)
+        
+        defer func() {
+            if e := recover(); e != nil {
```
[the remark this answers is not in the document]
chatton: will panicking revert state in a way that returning an error won't?

chatton (approved):
Meant to leave approval in the original review!

Thanks again for all the effort you've put into this, looking forward to getting this merged 🥇

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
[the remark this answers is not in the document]
colin-axner: The way the pseudo code is currently written is that state is always written if the call to `OnAcknowledgePacket` does not panic (ie if gas does not run out). The error returned by `OnAcknowledgePacket` is not checked

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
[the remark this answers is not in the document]
colin-axner: Okay so this is state within the adr 8 module? I guess there's no plan for state cleanup? Why would the contract return an error and then need to retrieve it later?

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
```diff
@@ -0,0 +1,374 @@
...
+// Call the IBCActor recvPacket callback after processing the packet
+// if the recvPacket callback exists and returns an error
+// then return an error ack to revert all packet data processing
+func OnRecvPacket(
+    ctx sdk.Context,
+    packet channeltypes.Packet,
+    relayer sdk.AccAddress,
+) (ack exported.Acknowledgement) {
+    // run any necesssary logic first
+    // IBCActor logic will postprocess
```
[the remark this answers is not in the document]
colin-axner: So if the underlying app fails it's receive, a failed acknowledgement should be returned instead of performing post processing?

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
colin-axner: I believe that is what:
`// application-specific onAcknowledgmentPacket logic`
is

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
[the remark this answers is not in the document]
colin-axner: I guess the idea is that there is a possibility the `cacheCtx` runs out of gas, but the regular context still has some execution left. I think we need a bit more logic to handle this scenario as you need to propogate the panic if the regular context is also out of gas

colin-axner (approved):
Ay, let's get this merged. I can follow up with pr's/issues for all the points brought up in review

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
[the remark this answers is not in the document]
colin-axner: Opened #3317 to track this suggestion

docs/architecture/adr-008-app-caller-cbs/adr-008-app-caller-cbs.md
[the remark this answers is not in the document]
colin-axner: #3318

PR #1979, opened 2022-08-11
Use single registry for all images

chatton:
## Description

Initially we used two registries to store images, `ghcr.io/cosmos/ibc-go-simd` for tags, e.g. `v2.0.0` and `ghcr.io/cosmos/ibc-go-simd-e2e` which stored images for `prs`. I think this actually makes things more convoluted and it would be simpler and easier to maintain to just store the images in a single registry.

All images will go [here](https://github.com/cosmos/ibc-go/pkgs/container/ibc-go-simd/versions) after this PR is merged.

---

write a little note why.

### DIFF
--- a/.github/workflows/e2e-fork.yml
+++ b/.github/workflows/e2e-fork.yml
@@ -25,7 +25,7 @@ jobs:
   e2e:
     env:
       CHAIN_A_SIMD_TAG: latest
-      CHAIN_A_SIMD_IMAGE: ibc-go-simd-e2e
+      CHAIN_A_SIMD_IMAGE: ibc-go-simd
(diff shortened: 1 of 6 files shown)


colin-axner (approved):
Nice! Should we consider switching to a repository name that doesn't use "simd" in case we want to change it in the future? cc @crodriguezvega 

I just figure if we want to eventually make the change, it might be nice if we have all historical built binaries in the repo. Is it possible to change the repository name without losing previously built tags?

chatton:
@colin-axner we would simply just need to re-tag every image and push them to a new registry. This can be done with a simple  bash script without too much effort. All of the images are re-buildable based on tags if we find ourselves needing to rebuild them for some reason.

Although the point about the binary name is interesting, we would need to account for test chains having different binary names. I hadn't considered that yet but something we'll need to watch out for.

crodriguezvega (approved):
Looks good to me! 👍

Issue #1971, opened 2022-08-10
Make IBCtesting GetStakingKeeper be based on an interface


ValarDragon:
## Summary

The IBCtesting `GetStakingKeeper` required interface method (https://github.com/cosmos/ibc-go/blob/main/testing/app.go#L39) should not depend on the literal staking keeper in the hub.

This should instead have an interface dependency, to account for apps doing weirder things with their staking modules. E.g. Osmosis superfluid staking, we really need to look at what IBC depends on, to determine what we should be returning.

Also would recommend renaming to `GetIBCTestingStakingKeeper` as part of this.

## Problem Definition

The current design inhibits modularity of ibctesting, for a small reason.

## Proposal

Define an interface for what keeper methods IBC depends on from the staking keeper, and then define an interface accordingly. Use that interface in the required ibctesting app interface.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
@ValarDragon if the API is changed to:
```go
GetStakingKeeper() ibctestingtypes.StakingKeeper
```

do you still want this change:
> Also would recommend renaming to GetIBCTestingStakingKeeper as part of this.

PR #2028, opened 2022-08-17
chore: set up IBCTestStakingKeeper interface 

charleenfei:
## Description

closes: #1971 

---

write a little note why.


testing/types/expected_keepers.go
```diff
@@ -0,0 +1,22 @@
+package types
```
colin-axner, line 1: I'm not so sure we need a separate directory for this. I think it makes more sense just to add to the top level `ibc-go/testing` directory

testing/chain_test.go
```diff
@@ -23,11 +23,11 @@ func TestChangeValSet(t *testing.T) {
...
 	require.True(t, ok)
 
-	val := chainA.App.GetStakingKeeper().GetValidators(chainA.GetContext(), 4)
+	val := chainA.App.GetIBCTestStakingKeeper().GetValidators(chainA.GetContext(), 4)
 
-	chainA.App.GetStakingKeeper().Delegate(chainA.GetContext(), chainA.SenderAccounts[1].SenderAccount.GetAddress(),
+	chainA.App.GetIBCTestStakingKeeper().Delegate(chainA.GetContext(), chainA.SenderAccounts[1].SenderAccount.GetAddress(),
 		amount, types.Unbonded, val[1], true)
-	chainA.App.GetStakingKeeper().Delegate(chainA.GetContext(), chainA.SenderAccounts[3].SenderAccount.GetAddress(),
+	chainA.App.GetIBCTestStakingKeeper().Delegate(chainA.GetContext(), chainA.SenderAccounts[3].SenderAccount.GetAddress(),
```
colin-axner, line 30: Since this is the test file, maybe we can use `chainA.GetSimApp().StakingKeeper`

testing/types/expected_keepers.go
```diff
@@ -0,0 +1,22 @@
...
+type IBCTestingStakingKeeper interface {
+	GetValidators(ctx sdk.Context, maxRetrieve uint32) (validators []stakingtypes.Validator)
+	Delegate(
+		ctx sdk.Context,
+		delAddr sdk.AccAddress,
+		bondAmt math.Int,
+		tokenSrc stakingtypes.BondStatus,
+		validator stakingtypes.Validator,
+		subtractAccount bool,
+	) (newShares sdk.Dec, err error)
```
colin-axner, line 20: lets try removing these functions

colin-axner (commented):
LGTM, you'll need to update the `GetStakingKeeper` function in `simapp/app.go` as well. The readme needs to be updated too. Left a few comments

testing/README.md
```diff
@@ -48,7 +48,7 @@ type TestingApp interface {
 
 	// ibc-go additions
 	GetBaseApp() *baseapp.BaseApp
-	GetStakingKeeper() stakingkeeper.Keeper
+	GetIBCTestStakingKeeper() ibctestingtypes.StakingKeeper
```
crodriguezvega, line 51: I honestly liked the name `GetStakingKeeper`, since the name of the type already indicates that this is for testing, so I consider that is redundant to add `Test` in the name of the function... But no problem if we decide to rename it.

testing/chain_test.go
```diff
@@ -23,11 +23,11 @@ func TestChangeValSet(t *testing.T) {
 	amount2, ok := sdk.NewIntFromString("30000000000000000000")
 	require.True(t, ok)
 
-	val := chainA.App.GetStakingKeeper().GetValidators(chainA.GetContext(), 4)
+	val := chainA.GetSimApp().StakingKeeper.GetValidators(chainA.GetContext(), 4)
```
crodriguezvega, line 26: Is it not possible to keep using `App.GetIBCTestStakingKeeper()` and add `GetValidators` and `Delegate` to the `StakingKeeper` interface?

crodriguezvega (commented):
Thanks for picking this up, @charleenfei!

testing/README.md
charleenfei: i renamed on request from @ValarDragon in the issue, i don't have a particularly strong opinion either way tho. I can see the value add in being explicit, but it is true that the type is also quite informative.

testing/chain_test.go
charleenfei: i think the idea is that in our simApp, we would continue using the staking module's staking keeper so we don't actually need to add these two methods as they are already there. 

it's just that in the testing chain and testing app setup, we would like to enable osmosis to swap out their customised interfluid staking keeper so they can use our testing setup. this is why we only need that one method which is used in the testing/chain.go file.

testing/types/expected_keepers.go
charleenfei: left in order to avoid circular imports

testing/README.md
colin-axner: I think now that the interface returns the `ibctestingtypes.StakingKeeper` it can be redundant. I do agree `GetStakingKeeper` is cleaner, but I understand this function will be implemented in `app.go` for a production chain. 

@ValarDragon do you still prefer `GetIBCTestStakingKeeper` given the new return value

testing/chain_test.go
colin-axner: I actually requested those functions to not be added. I think we should keep our interfaces as small as possible (ideally we wouldn't even need the staking keeper interface)

testing/simapp/app.go
```diff
@@ -748,7 +749,7 @@ func (app *SimApp) GetBaseApp() *baseapp.BaseApp {
 }
 
 // GetStakingKeeper implements the TestingApp interface.
```
colin-axner, line 751: update godoc

colin-axner (approved):
LGTM. Could you add a changelog entry?

I'm also okay leaving the `GetStakingKeeper` name as is and doing in a followup pr if it is still desired. I don't have strong preferences though

testing/simapp/app.go
charleenfei: > LGTM. Could you add a changelog entry?
> 
> I'm also okay leaving the `GetStakingKeeper` name as is and doing in a followup pr if it is still desired. I don't have strong preferences though

then it's no longer API breaking, correct?

testing/types/expected_keepers.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+)
+
+// IBCTestingStakingKeeper defines the expected staking keeper interface used in the
```
colin-axner, line 8:
```suggestion
// StakingKeeper defines the expected staking keeper interface used in the
```

colin-axner (approved):
followup ACK

damiannolan (approved):
Nice work!

seantking (approved):

colin-axner:
We should probably add a note about the API change in the migration docs. Maybe link the simapp change we have as well?

Issue #2052, opened 2022-08-19
Add `SendTx` rpc and proto message definitions


damiannolan:
## Summary

- [x] Add `SendTx` rpc and message definitions
- [x] Add empty boilerplate func to `controller` keeper
- [x] Register codec types

```proto
service Msg {
  ...
  rpc SubmitTx(MsgSubmitTx) returns (MsgSubmitTxResponse);
}

message MsgSendTx {
  option (gogoproto.equal)           = false;
  option (gogoproto.goproto_getters) = false;

  string owner                       = 1;
  string connection_id               = 2 [(gogoproto.moretags) = "yaml:\"connection_id\""];
  string timeout                     = 3;
  google.protobuf.Any msg            = 4;
}

message MsgSendTxResponse {}
```
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #2086, opened 2022-08-23
chore: adding basic Msg service and SubmitTx rpc boilerplate

charleenfei:
## Description

Please review #2133 first!

NOTE: i updated the tx name to be SubmitTx in order to avoid name collision with the current SendTx method in the ICA controller keeper. Open to other suggestions!

closes: #2052 

---

write a little note why.


docs/ibc/proto-docs.md
```diff
@@ -1555,6 +1557,34 @@ MsgRegisterAccountResponse defines the response for Msg/RegisterAccount
...
+
+### MsgSubmitTx
+MsgSubmitTx defines the payload for MsgSubmitTx
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| `owner` | [string](#string) |  |  |
+| `connection_id` | [string](#string) |  |  |
+| `timeout` | [string](#string) |  |  |
```
charleenfei, line 1571: should timeout be Height here?

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
```diff
@@ -26,3 +29,17 @@ message MsgRegisterAccount {
...
+
+// MsgSubmitTx defines the payload for MsgSubmitTx
+message MsgSubmitTx {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string   owner                   = 1;
+  string   connection_id           = 2 [(gogoproto.moretags) = "yaml:\"connection_id\""];
+  string   timeout                 = 3;
+  repeated google.protobuf.Any msg = 4;
```
colin-axner, line 41:
```suggestion
  repeated google.protobuf.Any msgs = 4;
```

docs/ibc/proto-docs.md
colin-axner: I think we should support both. Should be similar to `MsgTransfer`

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
colin-axner, line 41: @damiannolan what's the reason for using `Any`s here? Shouldn't we just be taking in the ics27 [packet data](https://github.com/cosmos/ibc-go/blob/0fa2978b0edad338f995790b5ab3a517ff302351/modules/apps/27-interchain-accounts/types/packet.pb.go#L57)

colin-axner:
> NOTE: i updated the tx name to be SubmitTx in order to avoid name collision with the SendTx method in the transfer app. Open to other suggestions!

I like `SubmitTx`. As a side note, I don't think the name would conflict with ics20 since a user would submit a `MsgTransfer` and calling the ics20 function directly (it really should be private) would be transferkeeper.SendTx() compared to icacontrollerkeeper.SendTx()

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
damiannolan: I think if an end user wants to use `x/group` or `x/gov` they can't really construct the packet data type as it expects a proto encoded `[]byte` of message data. If we use `Any` it can be defined in JSON and submitted for proposal execution. Essentially like it's done [here](https://github.com/cosmos/cosmos-sdk/blob/main/x/group/client/cli/tx.go#L559). We would have a proto-JSON encoded `MsgSubmitTx` which has a nested array of ICA messages to be executed on the host.

docs/ibc/proto-docs.md
damiannolan: I agree, I think we should support both going forward but the current implementation uses [`clienttypes.ZeroHeight`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/keeper/relay.go#L66).
I think we can only support the timeout height via the new execution path - i.e. existing auth modules using the `SendTx` go API directly won't be able to specify a height, right? 

For `SubmitTx` we can use `uint64` for `timeout_timestamp` and `ibc.core.client.v1.Height` for `timeout_height` like [`MsgTransfer`](https://github.com/cosmos/ibc-go/blob/main/proto/ibc/applications/transfer/v1/tx.proto#L36-L40) as @colin-axner  suggests! :)

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
colin-axner: Okay so this is an any which is decoded into packet data?

We cannot assume the counterparty uses any's for message delivery (this is SDK behaviour). I also suspect folks will want to make use of the memo field (it allows for a short term hack of enabling top level middleware for application handling, ie memo field: "swap:lp" would indicate, swap tokens then lp)

colin-axner: I think it'd be problematic to use Any to represent the messages delivered to the counterparty chain. These messages would need to be registered on the controller chain since the SDK will verify the Any URL is registered on the interface registry. Does that make sense?

charleenfei:
> 

ah sorry, i misread the error, there is already a [SendTx ](https://github.com/cosmos/ibc-go/blob/0fa2978b0edad338f995790b5ab3a517ff302351/modules/apps/27-interchain-accounts/controller/keeper/relay.go#L18) in the controller keeper. If i understand this function correctly, this functionality is actually what should now be triggered by the new SubmitTx msg_server handler right?

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
damiannolan: > Okay so this is an any which is decoded into packet data?

I was thinking that this would be an array `Any`s which would be encoded into packet data using [`SerializeCosmosTx`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/types/codec.go#L32). 

> These messages would need to be registered on the controller chain since the SDK will verify the Any URL is registered on the interface registry. Does that make sense?

This does make a lot of sense! It's a good point, do you know where in the SDK this would fail if for example you tried to provide an osmosis specific message type (swap) via a cosmos-hub(controller) group proposal?
I think this may be kind of problematic from a ux perspective...

colin-axner:
> ah sorry, i misread the error, there is already a [SendTx ](https://github.com/cosmos/ibc-go/blob/0fa2978b0edad338f995790b5ab3a517ff302351/modules/apps/27-interchain-accounts/controller/keeper/relay.go#L18)in the controller keeper. If i understand this function correctly, this functionality is actually what should now be triggered by the new SubmitTx msg_server handler right? except we would take the Msg and parse the messages from that rather than the InterchainAccountPacketData

Ah yea, those names will collide as two functions for the same keeper. We want to leave the old function for backwards compatibility purposes. The new msg handler should perform the same functionality as the existing function but take in the arguments in form of MsgSubmitTx

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
colin-axner: I believe the issue occurs because [`UnpackAny`](https://github.com/cosmos/cosmos-sdk/blob/main/codec/types/interface_registry.go#L217) will call [`UnpackInterfaces`](https://github.com/cosmos/cosmos-sdk/blob/main/codec/types/interface_registry.go#L294) which will try to [`unpackany`](https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/types/client.go#L42) the nested Any, but this will [fail to resolve](https://github.com/cosmos/cosmos-sdk/blob/main/codec/types/interface_registry.go#L243)

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
```diff
@@ -26,3 +30,23 @@ message MsgRegisterAccount {
...
+  string owner         = 1;
+  string connection_id = 2 [(gogoproto.moretags) = "yaml:\"connection_id\""];
+  // Timeout height relative to the current block height.
+  // The timeout is disabled when set to 0.
+  ibc.core.client.v1.Height timeout_height = 3
+      [(gogoproto.moretags) = "yaml:\"timeout_height\"", (gogoproto.nullable) = false];
+  // Timeout timestamp in absolute nanoseconds since unix epoch.
+  // The timeout is disabled when set to 0.
+  uint64   timeout_timestamp = 4 [(gogoproto.moretags) = "yaml:\"timeout_timestamp\""];
+  repeated ibc.applications.interchain_accounts.v1.InterchainAccountPacketData msgs = 5;
```
damiannolan, line 48: This type doesn't need to be `repeated`. One or many messages can be encoded in the `InterchainAccountPacketData` byte slice. 

Could we also rename from `msgs` to `packet_data` maybe?

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
```diff
@@ -26,3 +30,23 @@ message MsgRegisterAccount {
...
+}
+
+// MsgSubmitTxResponse defines the response for MsgSubmitTx
+message MsgSubmitTxResponse {}
```
damiannolan, line 52: I think we want to return the `uint64` sequence here in the response

modules/apps/27-interchain-accounts/controller/types/codec.go
```diff
@@ -8,4 +8,5 @@ import (
 // RegisterInterfaces registers the interchain accounts controller message types using the provided InterfaceRegistry
 func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
 	registry.RegisterImplementations((*sdk.Msg)(nil), &MsgRegisterAccount{})
+	registry.RegisterImplementations((*sdk.Msg)(nil), &MsgSubmitTx{})
```
damiannolan, line 11: I think we can consolidate these registrations

```suggestion
	registry.RegisterImplementations(
	    (*sdk.Msg)(nil), 
	    &MsgRegisterAccount{},
	    &MsgSubmitTx{},
	)
```

damiannolan (commented):
Nice work, looks good so far. Will review again once the genesis types refactor frees up the diffs!

modules/apps/27-interchain-accounts/controller/types/msgs.go
```diff
@@ -46,3 +48,24 @@ func (msg MsgRegisterAccount) GetSigners() []sdk.AccAddress {
 
 	return []sdk.AccAddress{accAddr}
 }
+
+// NewMsgSubmitTx creates a new instance of MsgSubmitTx
+func NewMsgSubmitTx(connectionID, owner string, timeoutHeight clienttypes.Height, timeoutTimestamp uint64, msgs []*icatypes.InterchainAccountPacketData) *MsgSubmitTx {
+	return &MsgSubmitTx{
+		ConnectionId:     connectionID,
+		Owner:            owner,
```
AdityaSripal, line 56: This should be an AccAddress I believe because it is on the controller chain, we know it should be a valid address

modules/apps/27-interchain-accounts/controller/types/msgs.go
```diff
@@ -46,3 +48,24 @@ func (msg MsgRegisterAccount) GetSigners() []sdk.AccAddress {
...
+}
+
+// ValidateBasic implements sdk.Msg
+func (msg MsgSubmitTx) ValidateBasic() error {
+	return nil
+}
+
+// GetSigners implements sdk.Msg
+func (msg MsgSubmitTx) GetSigners() []sdk.AccAddress {
+	return []sdk.AccAddress{}
```
AdityaSripal, line 70:
```suggestion
	return []sdk.AccAddress{owner}
```

This needs to be authenticated by owner signature

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
```diff
@@ -26,3 +30,23 @@ message MsgRegisterAccount {
 message MsgRegisterAccountResponse {
   string channel_id = 1 [(gogoproto.moretags) = "yaml:\"channel_id\""];
 }
+
+// MsgSubmitTx defines the payload for MsgSubmitTx
+message MsgSubmitTx {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string owner         = 1;
```
AdityaSripal, line 39: This can also be cast to sdk.Address

AdityaSripal (changes requested):
It's ok to have owner be an address on the controller chain. It also needs to be a signer, so that user-controlled ICAs are secure

AdityaSripal:
Realize my requests are addressed in #2135 . Ignore the above!

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
charleenfei: is it better to cast it in the proto or check if its a valid addr in ValidateBasic?

modules/apps/27-interchain-accounts/controller/types/msgs.go
charleenfei: yea i would fill out this function in the next PR along with validate basic :)

modules/apps/27-interchain-accounts/controller/types/msgs.go
charleenfei: is it better to cast it in the proto or check if its a valid addr in ValidateBasic?

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
colin-axner: I think it is fine to leave as is, primarily because the owner is also the signer and all of our Msg types have the signer as an uncasted string (albeit I'm not sure why)

colin-axner: I think it makes sense just to apply the casting change to all our messages at once. The SDK does some [different](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/group/v1/tx.proto#L315) things nowadays anyways

seantking (approved):
:rocket: :rocket:

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
```diff
@@ -26,3 +30,25 @@ message MsgRegisterAccount {
...
+  string connection_id = 2 [(gogoproto.moretags) = "yaml:\"connection_id\""];
+  // Timeout height relative to the current block height.
+  // The timeout is disabled when set to 0.
+  ibc.core.client.v1.Height timeout_height = 3
+      [(gogoproto.moretags) = "yaml:\"timeout_height\"", (gogoproto.nullable) = false];
+  // Timeout timestamp in absolute nanoseconds since unix epoch.
+  // The timeout is disabled when set to 0.
+  uint64 timeout_timestamp = 4 [(gogoproto.moretags) = "yaml:\"timeout_timestamp\""];
+  ibc.applications.interchain_accounts.v1.InterchainAccountPacketData packet_data = 5;
```
damiannolan, line 48: We should add yaml tags and I think we can probs make this a value type instead of a pointer:
`(gogoproto.nullable) = false`

modules/apps/27-interchain-accounts/controller/types/msgs.go
```diff
@@ -46,3 +48,24 @@ func (msg MsgRegisterAccount) GetSigners() []sdk.AccAddress {
 
 	return []sdk.AccAddress{accAddr}
 }
+
+// NewMsgSubmitTx creates a new instance of MsgSubmitTx
+func NewMsgSubmitTx(connectionID, owner string, timeoutHeight clienttypes.Height, timeoutTimestamp uint64, packetData *icatypes.InterchainAccountPacketData) *MsgSubmitTx {
```
colin-axner, line 53:
```suggestion
func NewMsgSubmitTx(connectionID, owner string, timeoutHeight clienttypes.Height, timeoutTimestamp uint64, packetData icatypes.InterchainAccountPacketData) *MsgSubmitTx {
```

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
```diff
@@ -26,3 +30,26 @@ message MsgRegisterAccount {
...
 }
+
+// MsgSubmitTx defines the payload for MsgSubmitTx
+message MsgSubmitTx {
+  option (gogoproto.equal)           = false;
+  option (gogoproto.goproto_getters) = false;
+
+  string owner         = 1;
+  string connection_id = 2 [(gogoproto.moretags) = "yaml:\"connection_id\""];
+  // Timeout height relative to the current block height.
```
colin-axner, line 41: I'm not sure we should be taking in a relative height here (especially since the timestamp is labeled as absolute). I think I'd prefer external UX to just figure out the correct timeout height

Is there a specific use case we can think of which needs a relative timeout?

modules/apps/27-interchain-accounts/controller/types/msgs.go
colin-axner, line 53: I find it mildly odd that the argument order doesn't match the protobuf numbering order. I don't have a strong preference. Typically timeout info is last though

colin-axner (approved):
LGTM, left minor suggestions

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
colin-axner: Looks like it needs the yaml tag as well

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
charleenfei: i just grabbed the timeout parameters from the transfer app -- i guess the relative height might make sense if you want a time limit on the action relative to when you've submitted the tx (ie: you want to trigger a swap only within the next 24 hours)

proto/ibc/applications/interchain_accounts/controller/v1/tx.proto
colin-axner: I looked into this some more. Looks like the SDK intentionally chose to go with a string in 2020 (during amino -> proto migration). Might have been some concern with the gRPC gateway, nevertheless there's talk about removing reliance on the global bech32 config and moving it to be a param of the auth keeper, so you'd need access to the auth keeper to get the acc address (they want to change get signers to return a string instead of an address)

Lets keep with what we have. It'll be easier to deal with everything uniformly

Issue #2042, opened 2022-08-18
Add SetupClients to e2e testsuite


colin-axner:
## Summary

Add function `SetupClients` which calls relayer.CreateClients. This should be similar to the testing package [`SeutpClients`](https://github.com/cosmos/ibc-go/blob/main/testing/coordinator.go#L87)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #1965, opened 2022-08-10
E2E: Successful client recovery proposal


colin-axner:
## Summary

Add a test which successfully updates a client using the update client proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #2130, opened 2022-08-25
e2e: add client recovery proposal test

colin-axner:
## Description

closes: #1965
closes: #2042 

---

write a little note why.


e2e/testsuite/testsuite.go
```diff
@@ -157,6 +156,18 @@ func (s *E2ETestSuite) SetupChainsRelayerAndChannel(ctx context.Context, channel
...
+func (s *E2ETestSuite) getPathName() string {
+	pathName := fmt.Sprintf("%s-path", s.T().Name())
+	return strings.ReplaceAll(pathName, "/", "-")
+}
+
+// SetupClients creates clients on chainA and chainB using the provided create client options
+func (s *E2ETestSuite) SetupClients(ctx context.Context, relayer ibc.Relayer, opts ibc.CreateClientOptions) {
+	err := relayer.CreateClients(ctx, s.GetRelayerExecReporter(), s.getPathName(), opts)
+	s.Require().NoError(err)
+}
```
colin-axner, line 169: I think this is currently not working. My guess is it is no-oping because a client already exists for the given path name. I think we need to generate/add a path for every client/connection/channel (aka endpoint) stack. My vote would be to add the endpoint type and when you call NewEndpoint, it adds the ibctest path name

boojamya: Soo looks like you're right, that new client is not being created. 
The easy solution here is to wire up the `--override` flag in ibctest for the relayer.

The client is not begin created b/c in the relayer config, the client already exists so it does not create a new client.
This is what the `--override` flag is used for; to create a new client even though the config is showing a client-id already exists. But, it is currently not working. Issue here: https://github.com/cosmos/relayer/issues/952

In the meantime, one dirty solution that I *think* will work is to create another relayer instances with a different home folder path. Then create the new "substituteClient" with the bad trusting period that way.

e2e/testsuite/testsuite.go
```diff
@@ -159,6 +161,39 @@ func (s *E2ETestSuite) SetupChainsRelayerAndChannel(ctx context.Context, channel
 	return r, chainAChannels[len(chainAChannels)-1]
 }
 
+// generatePathName generates the path name using the test suites name
+func (s *E2ETestSuite) generatePathName() string {
+	pathName := fmt.Sprintf("%s-path-%d", s.T().Name(), s.pathNameIndex)
+	s.pathNameIndex++
+	return strings.ReplaceAll(pathName, "/", "-")
```
colin-axner, line 168: for every endpoint (client/connection/channel) stack we need a new path name, so I added an index kept by the `E2ETestSuite`

e2e/testsuite/testsuite.go
```diff
@@ -159,6 +161,39 @@ func (s *E2ETestSuite) SetupChainsRelayerAndChannel(ctx context.Context, channel
...
+}
+
+// SetupClients creates clients on chainA and chainB using the provided create client options
+func (s *E2ETestSuite) SetupClients(ctx context.Context, relayer ibc.Relayer, opts ibc.CreateClientOptions) {
+	pathName := s.generatePath(ctx, relayer)
+	err := relayer.CreateClients(ctx, s.GetRelayerExecReporter(), pathName, opts)
+	s.Require().NoError(err)
+}
+
+// UpdateClients updates clients on chainA and chainB using the provided create client options
```
colin-axner, line 191:
```suggestion
// UpdateClients updates clients on chainA and chainB
```

e2e/tests/core/client_test.go
```diff
@@ -0,0 +1,121 @@
...
+	t.Run("create substitute client with correct trusting period", func(t *testing.T) {
+		relayer, _ = s.SetupChainsRelayerAndChannel(ctx)
+
+		// TODO: update when client identifier created is accessible
+		// currently assumes first client is 07-tendermint-0
+		substituteClientID = clienttypes.FormatClientIdentifier(ibcexported.Tendermint, 0)
+
+		// TODO: replace with better handling of path names
+		pathName = fmt.Sprintf("%s-path-%d", s.T().Name(), 0)
+		pathName = strings.ReplaceAll(pathName, "/", "-")
```
colin-axner, line 63: These are a bit hacky, but I think it is better to just do this for now and figure out a better solution later. I think adding the endpoint type should solve the issue in both cases

e2e/testsuite/testsuite.go
colin-axner: Generating a new path (with a new path name) worked

colin-axner:
`--- PASS: TestClientTestSuite/TestClientUpdateProposal_Succeeds (159.51s)`

:sweat_smile: Maybe we should only build some of these tests on commits to main and get notifications of failure? I don't think client recovery proposal e2e tests need to be run on every pr

chatton:
@colin-axner maybe just manually maintaining the list of tests and where they run will be the best option. We're already starting to have tests we don't want to run all the time. These, icad (and soon upgrade)

We can maybe just keep a simple json file in the repo and pass that to different github actions. What do you think?

colin-axner:
> We can maybe just keep a simple json file in the repo and pass that to different github actions. What do you think?

Sounds good, maybe we could run them nightly or something?

.github/workflows/e2e-fork.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Run e2e Test
         env:
           # see images here https://github.com/cosmos/relayer/pkgs/container/relayer/versions
-          RLY_TAG: "v2.0.0-rc2"
+          RLY_TAG: "v2.0.0"
```
colin-axner, line 51: Can we add a var for this? Is that possible?

chatton: this file is the tests that run on forks only, is this the one you want to paramaterize? If not, the existing `workflow_call` (a workflow re-usable in other workflows) has the relayer tag as a paramter. See [here](https://github.com/cosmos/ibc-go/blob/6922700c19c2b78409dbb2b6157c59ae890f8a4e/.github/workflows/e2e-test-workflow-call.yml#L34)

This is configurable in the [manually triggerable workflows](https://github.com/cosmos/ibc-go/blob/6922700c19c2b78409dbb2b6157c59ae890f8a4e/.github/workflows/e2e-manual-simd.yaml#L44)


The task that runs on every pr is [this one](https://github.com/cosmos/ibc-go/blob/6922700c19c2b78409dbb2b6157c59ae890f8a4e/.github/workflows/e2e.yaml#L44)

In the end the env var `RLY_TAG` is what's used to configure it.

damiannolan (approved):
Nice work! 🚀

seantking (approved):

.github/workflows/e2e-fork.yml
colin-axner: I'd like a single var to update when bumping the default relayer tag

chatton: as it currently is there isn't one place to specify the relayer version. Unfortunately github actions doesn't really support sharing variables across workflows like that (from within the workflow files).

We could move this to a secret which would be configured via the UI, or we could do something like [this](https://arinco.com.au/blog/github-actions-share-environment-variables-across-workflows/)

and just maintain a file that we use to add env vars to all the runs.

chatton: Although even this might not work, as we need to manually list items to provide the drop down (in the manual tasks) and as far as I know the only way to provide this is with an explicit list.

colin-axner: Ah no worries, it was a minor nit. Not worth the extra hassle

Issue #635, opened 2021-12-15
use go.mod for testing package


colin-axner:
## Summary

Breaking changes to the testing package should not affect the versioning of core IBC. We should use a go.mod for the testing package, this also gives us a way to test out using a separate go.mod for each IBC application

ref: #614 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #2132, opened 2022-08-26
feat: adding testing package to separate go.mod and using within ibc-go

seantking:
## Description

closes: [#635](https://github.com/cosmos/ibc-go/issues/635)

---

write a little note why.


seantking:
The issues are the following:
  1. In order to manage releases of multiple modules within a single repo from what I can tell we would need to change our scheme for how we tag releases. Maybe each tag would be prepended with the module name i.e. testing/v1.0.0 etc. I think we could maybe do more research here and see how other monorepos handle this.
 2.  Breaking changes in between tags have the effect of breaking the testing package.  i.e. changing solomachine client types has a downstream effect of breaking the testing package. This creates the scenario where we cannot fix the testing package until we have a tag of ibc-go with the breaking changes that can be imported in the testing package go.mod. Alternatively, we can pin the testing package to the local main branch. However, pinning locally to main means the testing package cannot be imported by external parties. I actually ran into this very issue as we have changes on main that are not part of a tag/release yet which is why the e2e module and the testing module are pinned to the local main branch. In this PR we're importing a version of the new testing module by the commit hash on the open PR itself (otherwise we would have to pin to the local package with a replace statement).

I think it was worthwhile doing a spike on this but I think before we would continue down this path it would be worthwhile to consider how much value this would really add compared to the overhead and potential side effects.

seantking:
After speaking to @colin-axner we agreed to close this PR for now and revisit it later. Perhaps we can take a look at https://go.dev/doc/tutorial/workspaces and see if this could be of some help here.

PR #2133, opened 2022-08-26
chore: refactor genesis types to separate directory, mv controller interfaces registration to module.go

charleenfei:
## Description

prior to this pr, the genesis protos were in interchain_account/v1/types but imported controllertypes in the testing file, and the controller interfaces registration was also being handled in interchain account types dir. 

this caused a circular import problem when trying to import the InterchainAccountPacketData type from the ICA types directory to the controller types directory, as part of the PR addressing #2052 (https://github.com/cosmos/ibc-go/issues/2052.) 

this pr moves the genesis proto, genesis.go and genesis_testing file to it's own directory in a structure similar to `host` and `controller`

---

write a little note why.


damiannolan (approved):
I'm happy enough with this layout if it solves the issues of importing `InterchainAccountPacketData` :)

modules/apps/27-interchain-accounts/genesis/types/genesis.go
```diff
@@ -90,7 +91,7 @@ func (gs ControllerGenesisState) Validate() error {
 // DefaultHostGenesis creates and returns the default interchain accounts HostGenesisState
 func DefaultHostGenesis() HostGenesisState {
 	return HostGenesisState{
-		Port:   PortID,
+		Port:   icatypes.PortID,
 		Params: hosttypes.DefaultParams(),
```
colin-axner, line 95: Sorry for just noticing this, but maybe we can just move this function to hosttypes? Same for controller types. I think that should remove the host/controller imports

proto/ibc/applications/interchain_accounts/genesis/v1/genesis.proto
```diff
@@ -1,8 +1,8 @@
 syntax = "proto3";
 
-package ibc.applications.interchain_accounts.v1;
+package ibc.applications.interchain_accounts.genesis.v1;
```
colin-axner, line 3: We don't want to change the package name ideally since it can have unintended side affects (it is used in decoding Anys)

modules/apps/27-interchain-accounts/genesis/types/genesis.go
charleenfei: but then we still need to import controllertypes/hosttypes into genesis types to use the function, which if we leave the genesis in ica will trigger the original import cycle issue

proto/ibc/applications/interchain_accounts/genesis/v1/genesis.proto
charleenfei: wouldnt changing this back to just the ica v1 package cause proto compile errors?

modules/apps/27-interchain-accounts/genesis/types/genesis.go
colin-axner: Oh yea because this is going to be called by the `DefaultGenesis()` function

proto/ibc/applications/interchain_accounts/genesis/v1/genesis.proto
colin-axner: No, the go option specifies where the go files are generated

colin-axner (commented):
LGTM, I think the proto package name should remain unchanged though. The go option can be modified to generate the proto files in the new genesis directory as you have done. Can you add a changelog indicating that the genesis types have been moved to their own directory?

damiannolan:
> The go option can be modified to generate the proto files in the new genesis directory as you have done.

As far as I remember you can't specify two different `go_package` options in the same proto pkg directory. I think we had issues generating the `.pb.go` files when I looked at this with @charleenfei

charleenfei:
> > The go option can be modified to generate the proto files in the new genesis directory as you have done.
> 
> As far as I remember you can't specify two different `go_package` options in the same proto pkg directory. I think we had issues generating the `.pb.go` files when I looked at this with @charleenfei

```{"path":"proto/ibc/applications/interchain_accounts/genesis/v1/genesis.proto","start_line":3,"start_column":1,"end_line":3,"end_column":49,"type":"PACKAGE_DIRECTORY_MATCH","message":"Files with package \"ibc.applications.interchain_accounts.v1\" must be within a directory \"ibc/applications/interchain_accounts/v1\" relative to root but were in directory \"ibc/applications/interchain_accounts/genesis/v1\"."}``` this is the error i get

File: proto/ibc/applications/interchain_accounts/genesis/v1/genesis.proto (hunk 2 above)
Comment (charleenfei) [reply proto/ibc/applications/interchain_accounts/genesis/v1/genesis.proto:3]: i think we can't only change the go option without changing the package name, that causes the matching error i pasted below:

Comment (colin-axner) [reply proto/ibc/applications/interchain_accounts/genesis/v1/genesis.proto:3]: yea that's correct, looks like proto doesn't allow multiple go package options in the same package. Did not know that

File: CHANGELOG.md [hunk 3]
```diff
@@ -71,6 +71,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### Improvements
 
+* (apps/27-interchain-accounts) [\#2133](https://github.com/cosmos/ibc-go/pull/2133) generates genesis protos in a separate directory to avoid circular import errors
```
colin-axner, line 74:
```suggestion
* (apps/27-interchain-accounts) [\#2133](https://github.com/cosmos/ibc-go/pull/2133) Generates genesis protos in a separate directory to avoid circular import errors. The protobuf package name has changed for the genesis types. 
```

colin-axner (approved):
ACK, just a few notes. This is protobuf breaking since we are changing the package name. We are moving the types though which is why we aren't changing the `v1` to `v2`. So long as the genesis types are not packed to an Any (no reason to do so), then this should not be a problem. 

There isn't really a good alternative to avoid this situation if we are to have separate message/query servers for controller and host. It doesn't make too much sense to combine the message/query server into one keeper and could potentially lead to more issues. This seems like the least disruptive approach

PR #2134, opened 2022-08-26
chore: ics27 channel capability migrations

damiannolan:
## Description

- Adding upgrade handler for ICS27 channel capabilities migrations
- Adding tests

TODO:
- Add migration docs

closes: #XXXX

---

write a little note why.


modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
```diff
@@ -0,0 +1,63 @@
...
+	host "github.com/cosmos/ibc-go/v5/modules/core/24-host"
+)
+
+// MigrateICS27ChannelCapability performs a search on a prefix store using the provided store key and module name.
+// It retrieves the associated channel capability index and reassigns ownership to the ICS27 controller submodule.
+func MigrateICS27ChannelCapability(
+	ctx sdk.Context,
+	memStoreKey storetypes.StoreKey,
+	capabilityKeeper *capabilitykeeper.Keeper,
+	module string,
```
colin-axner, line 24: This is the module name of the scoped keeper for the underlying app module?

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
```diff
@@ -0,0 +1,63 @@
...
+) error {
+	// construct a prefix store using the x/capability reverse lookup key: {module}/rev/{name} -> index
+	keyPrefix := capabilitytypes.RevCapabilityKey(module, fmt.Sprintf("%s/%s/%s", host.KeyChannelCapabilityPrefix, host.KeyPortPrefix, icatypes.PortPrefix))
```
colin-axner, line 27: might be good to note why you end on `icatypes.PortPrefix`

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
```diff
@@ -0,0 +1,63 @@
...
+	prefixStore := prefix.NewStore(ctx.KVStore(memStoreKey), keyPrefix)
```
colin-axner, line 28: As discussed, this needs to be replaced with iteration over the indicies (instead of the mem store which will be uninitialized)

colin-axner (changes requested):

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations_test.go
```diff
@@ -0,0 +1,147 @@
...
+	suite.Require().True(found)
+
+	isAuthenticated := suite.chainA.GetSimApp().ScopedICAMockKeeper.AuthenticateCapability(suite.chainA.GetContext(), cap, capName)
+	suite.Require().True(isAuthenticated)
+
+	cap, found = suite.chainA.GetSimApp().ScopedICAControllerKeeper.GetCapability(suite.chainA.GetContext(), capName)
+	suite.Require().Nil(cap)
+	suite.Require().False(found)
+
+	suite.ResetMemStore() // empty the x/capability in-memory store
```
damiannolan, line 110: I decided to add this to the tests to mock/simulate a new chain binary running the migration handler against persisted state only. I think with this I feel better about the assertions done below after `capabilityKeeper.InitMemStore(ctx)` is called in the migration handler.

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
damiannolan: I guess this code is no longer relevant.

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
seantking: maybe we can just leave a comment?

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
```diff
@@ -0,0 +1,58 @@
...
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	capabilitykeeper "github.com/cosmos/cosmos-sdk/x/capability/keeper"
+	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
+
+	"github.com/cosmos/ibc-go/v5/modules/apps/27-interchain-accounts/controller/types"
+)
+
+// MigrateICS27ChannelCapability performs a search on a prefix store using the provided store key and module name.
+// It retrieves the associated channel capability index and reassigns ownership to the ICS27 controller submodule.
```
seantking, line 15: Might be worth linking to an issue describing why the migration is necessary :thinking:

seantking (approved):
Super clean :handshake:

colin-axner (approved):
Wow, excellent! Super clean and very easy to follow. Could you add a changelog entry?

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
```diff
@@ -0,0 +1,58 @@
...
+	prefixStore := prefix.NewStore(ctx.KVStore(storeKey), capabilitytypes.KeyPrefixIndexCapability)
+	iterator := sdk.KVStorePrefixIterator(prefixStore, nil)
+	defer iterator.Close()
+
+	for ; iterator.Valid(); iterator.Next() {
+		// unmarshal the capability index value and set of owners
+		index := capabilitytypes.IndexFromKey(iterator.Key())
+
+		var owners capabilitytypes.CapabilityOwners
+		cdc.MustUnmarshal(iterator.Value(), &owners)
```
chatton, line 33: if the function already has an error as a method signature, should be return an error instead? (or do we want the panic to be handled elsewhere?)

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
```diff
@@ -0,0 +1,58 @@
...
+	module string,
+) error {
+	// construct a prefix store using the x/capability index prefix: index->capability owners
+	prefixStore := prefix.NewStore(ctx.KVStore(storeKey), capabilitytypes.KeyPrefixIndexCapability)
+	iterator := sdk.KVStorePrefixIterator(prefixStore, nil)
+	defer iterator.Close()
+
+	for ; iterator.Valid(); iterator.Next() {
+		// unmarshal the capability index value and set of owners
+		index := capabilitytypes.IndexFromKey(iterator.Key())
```
chatton, line 30: what's the significance of this `index` rather than using the index in the loop below, are these separate?

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
```diff
@@ -0,0 +1,58 @@
...
+		var owners capabilitytypes.CapabilityOwners
+		cdc.MustUnmarshal(iterator.Value(), &owners)
+
+		for _, owner := range owners.GetOwners() {
+			if owner.Module == module {
+				// remove the owner from the set
+				owners.Remove(owner)
+
+				// reassign the owner module to icacontroller
+				owner.Module = types.SubModuleName
```
chatton, line 41: do we need to remove, update, and re-add? Can we just do

```golang
owners[i].Module = types.SubModuleName
capabilityKeeper.SetOwners(ctx, index, owners)
```

Or will this not behave how I think it will?

chatton (commented):
Looks great! Just left a few comments/questions

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
AdityaSripal: This error would be very unexpected. It would imply a bug in the code, so I think a panic makes sense

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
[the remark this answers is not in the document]
damiannolan: Yep. In tests it looks like this for the module where's interested in: 
`"capabilities/ports/icacontroller-cosmos17dtl0mjt3t77kpuhg2edqzjpszulwhgzuj9ljs/channels/channel-0"`

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
damiannolan: Yeah, I can update this to return the err. I would expect the panic to be caught by the sdk, but it's probably cleaner to return the error. I just grabbed this from some code in `x/capability`.

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
damiannolan: This `index` refers to the capability index itself rather than the loop index. 
The `x/capability` module essentially maintains a monotonically increasing `uint64` sequence or index which it uses to maintain the capability. This index is a field on the capability pointer which is generated per node. So the index should always be the same but obviously there will be different pointer values on each node. 

Is this a fair description @AdityaSripal?

AdityaSripal (approved):
Excellent test!!

modules/apps/27-interchain-accounts/controller/migrations/v5/migrations.go
damiannolan: Seems like the APIs don't play nicely with this approach. I've tried rejigging the code a bit but I think its easier to leave as is 😅 

tldr; is the `CapabilityOwners` type encapsulates the slice and it makes it awkward to use index referencing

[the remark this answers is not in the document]
colin-axner: I think it is fine to assume the controller module name for now. If someone has changed it, the controller migration assertion will catch the issue and we can post a fixed version for them specifically

Issue #2121, opened 2022-08-25
Add e2e test for modifying 03-connection params


colin-axner:
## Summary

Add a test which modifies the [`MaxExpectedBlockTime`](https://github.com/cosmos/ibc-go/blob/main/modules/core/03-connection/types/params.go#L14) param

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

seantking:
For now, this test is just checking that the param is successfully changed. In the future, we can verify that the block delay is successful (this requires investigating if the golang relayer supports this right now).

Issue #2012, opened 2022-08-16
Update e2e test harness to allow for n chains 


seantking:
## Summary

Currently, the helper fns in our TestSuite allows for only [two chains](https://github.com/cosmos/ibc-go/blob/main/e2e/testsuite/testsuite.go#L95) to be run as part of the network. 

[Some tests](https://github.com/cosmos/ibc-go/issues/1968) will require 3+ chains with channels created between them. 

- [ ] Update the helper fns to allow for a user-defined amount of chains to be spun up as part of the test network. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chatton:
Worth nothing that we also want to be able to run tests with a single chain, so we should make sure to account for n >= 1

chatton:
this is now supported :partying_face:

PR #2139, opened 2022-08-29
test: 03-connection e2e test param change

seantking:
## Description

closes: #2121 

For now this test is just checking that the param is successfully changed. In the future we can verify that the block delay is successful (this requires investigating if the golang relayer supports this right now). 

---

write a little note why.


e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	return time
+}
+
+// TestMaxExpectedTimePerBlock tests changing the MaxExpectedTimePerBlock param using a governance proposal
+func (s *ConnectionTestSuite) TestMaxExpectedTimePerBlock() {
+	t := s.T()
+	ctx := context.TODO()
+
+	_, _ = s.SetupChainsRelayerAndChannel(ctx, transferChannelOptions())
```
seantking, line 53: This step is overkill as the test only requires one chain but the tests fail without the setup.

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,89 @@
...
+func (s *ConnectionTestSuite) QueryMaxExpectedTimePerBlockParam(ctx context.Context, chain ibc.Chain) uint64 {
+	queryClient := s.GetChainGRCPClients(chain).ParamsQueryClient
+	res, err := queryClient.Params(ctx, &paramsproposaltypes.QueryParamsRequest{
+		Subspace: "ibc",
+		Key:      string(connectiontypes.KeyMaxExpectedTimePerBlock),
+	})
+	s.Require().NoError(err)
+
+	// removing additional strings that are used for amino
+	delay := strings.ReplaceAll(res.Param.Value, "\"", "")
```
seantking, line 41: I wasn't exactly sure why the Value is double wrapped in quotes but based on an error I received when I don't add them it is something to do with the legacy amino implementation. 

cc: @colin-axner

colin-axner: That's a good question. I'm not sure either (didn't realized this was the case). I guess we can fix this during param migrations

e2e/tests/core/03-connection/connection_test.go
colin-axner: Yea that's one limitation I've noticed with the testing package. It assumes all tests require 2 chains to do the same functionality

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,89 @@
...
+	_, _ = s.SetupChainsRelayerAndChannel(ctx, transferChannelOptions())
+	chainA, chainB := s.GetChains()
+	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 1, chainA, chainB), "failed to wait for blocks")
+
+	t.Run("ensure delay is set to the default of 30 seconds", func(t *testing.T) {
+		expectedDelay := fmt.Sprintf("\"%d\"", 30*time.Second)
+		delay := fmt.Sprintf("\"%d\"", s.QueryMaxExpectedTimePerBlockParam(ctx, chainA))
+		s.Require().Equal(expectedDelay, delay)
```
colin-axner, line 62: Why not just compare the time.Duration types?

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,89 @@
...
+		proposal := paramsproposaltypes.NewParameterChangeProposal(ibctesting.Title, ibctesting.Description, changes)
+		s.ExecuteGovProposal(ctx, chainA, chainAWallet, proposal)
+	})
+
+	t.Run("validate the param was successfully changed", func(t *testing.T) {
+		expectedDelay := fmt.Sprintf("\"%d\"", 60*time.Second)
+		delay := s.QueryMaxExpectedTimePerBlockParam(ctx, chainA)
+		s.Require().Equal(expectedDelay, fmt.Sprintf("\"%d\"", delay))
+	})
+}
```
colin-axner, line 80: Maybe we should receive a packet on chainA. The [`GetMaxExpectedTimePerBlock`](https://github.com/cosmos/ibc-go/blob/main/modules/core/03-connection/keeper/verify.go#L427) will get called (at least ensuring it can still be queried correctly)

colin-axner (commented):
Looks great to me, left a few suggestions. I think I'd like to see a packet received before I add an approval

e2e/tests/core/03-connection/connection_test.go
chatton: There's an [issue](https://github.com/cosmos/ibc-go/issues/2012) to make this more flexible with the intention of supporting an environment for multi-hop, I think maybe we could extend that to provide arbitrary chain configs which would make it easy to just create a single chain for tests where that's all we need.

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,89 @@
...
+	ctx := context.TODO()
+
+	_, _ = s.SetupChainsRelayerAndChannel(ctx, transferChannelOptions())
+	chainA, chainB := s.GetChains()
+	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 1, chainA, chainB), "failed to wait for blocks")
+
+	t.Run("ensure delay is set to the default of 30 seconds", func(t *testing.T) {
+		expectedDelay := fmt.Sprintf("\"%d\"", 30*time.Second)
```
chatton, line 60: rather than escaping double quotes `\"` characters, I like to use template strings to make it more readable.

backticks don't render nicely in markdown (or at least I can't figure out how to make them render)

see [this example](https://go.dev/play/p/KqYnXetLlj8)

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+func TestConnectionTestSuite(t *testing.T) {
+	suite.Run(t, new(ConnectionTestSuite))
+}
+
+type ConnectionTestSuite struct {
+	testsuite.E2ETestSuite
+}
+
+// QueryConnectionEnabledParam queries the on-chain connection enabled param for 03-connection
```
chatton, line 31: doc string doesn't match the method name

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,89 @@
...
+	s.Require().NoError(test.WaitForBlocks(ctx, 1, chainA, chainB), "failed to wait for blocks")
+
+	t.Run("ensure delay is set to the default of 30 seconds", func(t *testing.T) {
+		expectedDelay := fmt.Sprintf("\"%d\"", 30*time.Second)
+		delay := fmt.Sprintf("\"%d\"", s.QueryMaxExpectedTimePerBlockParam(ctx, chainA))
+		s.Require().Equal(expectedDelay, delay)
+	})
+
+	t.Run("change the delay to 60 seconds", func(t *testing.T) {
+		delay := fmt.Sprintf("\"%d\"", 60*time.Second)
```
chatton, line 66:
```suggestion
		delay := fmt.Sprintf("\"%d\"", 1*time.Minute)
```

e2e/tests/core/03-connection/connection_test.go
chatton: if we're ignoring both return values maybe we can just call

```golang
s.SetupChainsRelayerAndChannel(ctx, transferChannelOptions())
```
without explicitly discarding them. I don't have a strong opinion about this, I can see the value in explicitly showing that there is a return value we're not using if that is the intent.

chatton (commented):
Looks great, I left a couple of comments / suggestions 👍

e2e/tests/core/03-connection/connection_test.go
crodriguezvega: We end up using both chains to test that a packet can be received.

e2e/tests/core/03-connection/connection_test.go
crodriguezvega: I changed it to compare `uint64`.

e2e/tests/core/03-connection/connection_test.go
crodriguezvega: I added a token transfer from chain B to chain A.

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+type ConnectionTestSuite struct {
+	testsuite.E2ETestSuite
+}
+
+// QueryMaxExpectedTimePerBlockParam queries the on-chain max expected time per block param for 03-connection
+func (s *ConnectionTestSuite) QueryMaxExpectedTimePerBlockParam(ctx context.Context, chain ibc.Chain) uint64 {
+	queryClient := s.GetChainGRCPClients(chain).ParamsQueryClient
+	res, err := queryClient.Params(ctx, &paramsproposaltypes.QueryParamsRequest{
+		Subspace: "ibc",
```
damiannolan, line 35: We should use `host.ModuleName` from `24-host`

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// removing additional strings that are used for amino
+	delay := strings.ReplaceAll(res.Param.Value, "\"", "")
+	time, err := strconv.ParseUint(delay, 10, 64)
+	s.Require().NoError(err)
+
+	return time
+}
+
+// TestMaxExpectedTimePerBlock tests changing the MaxExpectedTimePerBlock param using a governance proposal
+func (s *ConnectionTestSuite) TestMaxExpectedTimePerBlock() {
```
damiannolan, line 49: We should change this method name to have a `Param` suffix to align with others in transfer tests for example

```suggestion
func (s *ConnectionTestSuite) TestMaxExpectedTimePerBlockParam() {
```

e2e/tests/core/03-connection/connection_test.go
```diff
@@ -0,0 +1,131 @@
...
+	t.Run("ensure delay is set to the default of 30 seconds", func(t *testing.T) {
+		expectedDelay := uint64(30 * time.Second)
+		delay := s.QueryMaxExpectedTimePerBlockParam(ctx, chainA)
+		s.Require().Equal(expectedDelay, delay)
+	})
+
+	t.Run("change the delay to 60 seconds", func(t *testing.T) {
+		delay := fmt.Sprintf(`"%d"`, 1*time.Minute)
+		changes := []paramsproposaltypes.ParamChange{
+			paramsproposaltypes.NewParamChange("ibc", string(connectiontypes.KeyMaxExpectedTimePerBlock), delay),
```
damiannolan, line 76: Same comment re. `host.ModuleName` for ibc substore

damiannolan (approved):
Looks good to me, just left some nits

colin-axner (approved):

PR #2185, opened 2022-09-05
chore: remove duplicate imports

damiannolan:
## Description

- Cleanup import aliases in ICS27 `msg_server_test.go`

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/controller/keeper/msg_server_test.go
+++ b/modules/apps/27-interchain-accounts/controller/keeper/msg_server_test.go
@@ -4,11 +4,9 @@ import (
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	sdktypes "github.com/cosmos/cosmos-sdk/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 
 	"github.com/cosmos/ibc-go/v5/modules/apps/27-interchain-accounts/controller/types"
-	controllertypes "github.com/cosmos/ibc-go/v5/modules/apps/27-interchain-accounts/controller/types"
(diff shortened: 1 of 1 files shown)


modules/apps/27-interchain-accounts/controller/keeper/msg_server_test.go
```diff
@@ -18,7 +16,7 @@ import (
 
 func (suite *KeeperTestSuite) TestRegisterAccount() {
 	var (
-		msg               *controllertypes.MsgRegisterAccount
+		msg               *types.MsgRegisterAccount
```
chatton, line 19: if we have `icatypes`, `clienttypes` `channeltypes` etc. does it make sense to have a package imported as just `types`? It's not immediately obvious which sort of types it is, I would lean in favour of keeping it called `controllertypes` and removing the unaliased one WDYT?

damiannolan: In a lot of other packages we normally import the relative `types` pkg as is, and then for any `types` pkgs from other modules we alias them. 

I don't really feel strongly about it here. It's also a test file so.. 🤷

damiannolan: For interchain accounts it may help to always alias tho. Considering there is the shared `icatypes` and then a separate `types` for both controller and host submodules

chatton: ah I see okay, so we generally just do external types with the alias, that seems fine with me 👍

chatton (approved):

colin-axner (approved):

PR #2216, opened 2022-09-07
Interchain Queries Implemenation

jackzampolin:


docs/apps/icq/overview.md
```diff
@@ -0,0 +1,22 @@
+<!--
+order: 1
+-->
+
```
danburck, line 4: could add a list of prerequisite learnings, e.g. `whats an acknowledgement?`

danburck (commented):
Consider splitting this PR up into smaller pieces, starting with only the specs and docs. This way we can agree on the overall implementation and it will be easier to review the following PRs.  You could split up into: 

- specs and docs
- proto and types
- keeper and module wiring

E.g. when I opened this PR I thought of: 
* PR Description is missing or link to specification can be added to provide more context
* Lifecycle of the ibc query is not very clear in the documentation

georgelombardi97:
I will need this for a product that I am planning for. If there are no issues, would be great if we could have this merged in ibc-go so we chains will deploy it by default. Asking and convincing each chain to add this later would be hard imho.

georgelombardi97:
Can someone please explain why this is not merged? Maybe I can help?

adamewozniak:
+1 on prioritizing this PR ❤️

womensrights:
We have written up our findings on Interchain / Cross-chain queries in [this discussion](https://github.com/cosmos/ibc-go/discussions/2866), please let me know your thoughts. Just to note that having a queries implementation in the ibc-go repo would not mean chains deploy it by default cc @georgelombardi97

PR #2221, opened 2022-09-08
fix: add length validation of recipient address

crodriguezvega:
## Description

Validates that recipient address is not longer than 2048 characters. Happy to use a smaller or larger number as limit; I just chose one that would be large enough. 

See this [transaction](https://www.mintscan.io/osmosis/txs/F506105FAB8A7D26D950043B732C935A2896733C7AEB695F729CDC0631963252) for reference.

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/apps/transfer/types/msgs.go
+++ b/modules/apps/transfer/types/msgs.go
@@ -12,7 +12,8 @@ import (
 
 // msg types
 const (
-	TypeMsgTransfer = "transfer"
+	TypeMsgTransfer           = "transfer"
+	MaxRecipientAddressLength = 2048
 )
 
 // NewMsgTransfer creates a new MsgTransfer instance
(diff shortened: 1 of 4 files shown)


modules/apps/transfer/types/msgs.go
```diff
@@ -69,6 +70,10 @@ func (msg MsgTransfer) ValidateBasic() error {
 	if strings.TrimSpace(msg.Receiver) == "" {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, "missing recipient address")
 	}
+	receiver := []rune(msg.Receiver)
```
colin-axner, line 73: is the []rune necessary?

crodriguezvega: I convert to `[]rune` so that `len` gives me the number of characters that the string has. Otherwise, if I do `len` directly on `msg.Receiver`, it gives me the number of bytes of the string, which maybe or may not be the same as the number of characters.

colin-axner: But we want to check size right? It'd be preferable to ensure the address does not go above a certain size as opposed to character length?

crodriguezvega: For UTF-8 encoded strings one character could be up to 4 bytes, so if we count character and, in the worst case scenario all of them use 4 bytes, then the max byte length that we would allow would be 4*2048 bytes. I thought of counting characters instead of bytes, because the error message would be maybe nicer for the user: you can tell the user that the string is too long for so many characters, but if you tell the user that the string is too long in number of bytes, then it's harder to know by how many characters is too long... Although it would be strange to have non-alpha numeric characters in the recipient address, and in that case they are encoded with 1 byte... So maybe I am just overthinking this too much and we can just count bytes... :D

colin-axner: My preference would be to stick with the normal way we do length validation, which is by counting bytes. We shouldn't need to enforce length validation on the receiver address in the first place

PR #2270, opened 2022-09-11
modules/core/02-client/types: fix clientID validation regex to conform closer to spec

### LINKED ISSUE #2269: modules/core/02-client/types: IsClientIDFormat regex permits invalid client types due to misinterpretation of regex syntax for character sets and also doesn't respect the specification which requires entirely ASCII
## Summary of Bug
The definition of an identifier is specified loosely at https://github.com/cosmos/ibc/tree/main/spec/core/ics-024-host-requirements#paths-identifiers-separators and it should
* contain entirely ASCII
* not have the path separator "/" in it
Screen Shot 2022-09-10 at 5 22 50 PM

If we look at this validation regex https://github.com/cosmos/ibc-go/blob/37aa2ebec4f63a178659db2de2761f3cbafb583c/modules/core/02-client/types/keys.go#L35-L37 and read the comment above, it says it would like to ensure that client IDs are in the format `{client-type}-{N}` but if we look closely we can see a few things, the regex rule is `^.*[^\n-]-[0-9]{1,20}`
What that means:
* It can be preceded with anything but before we hit the hyphen "-" for the value of N, we can't have a newline "\n" nor hyphen "-"
* we can have non-ASCII in there

Unfortunately that means that now client-type can be "\t\t\t\t\t\t\t\t" (repeated as many times) then say "a" plus hyphen "-" then a number and the following values will erroneously be accepted:
* "\t\t\ta-100"
* "-----a-100"
* "                  a-100"
* "🚨😎-100"
and plenty more invalid patterns left to the imagination

## Remedy
If we were to correctly write that regex, we can simply eliminate and reduce it to `^\w+([\w-]+\w)?-[0-9]{1,20}$` to make it conform to specification

## Expected Behaviour

Rejection of the above listed patterns and non-ASCII values

## Version
a4be5614fb2fc92cfb430c82f18e7a9f67383c6e

____

#### For Admin Use

- [x] Not duplicate issue

odeke-em:
Rejects non-ASCII plus whitespaces and slashes to get it much closer to the specifications per
https://github.com/cosmos/ibc/tree/main/spec/core/ics-024-host-requirements#paths-identifiers-separators

while here added test vectors that were previously failing. However, we still need much better specifications for how long of values are accepted.

Fixes #2269

### DIFF
--- a/modules/core/02-client/types/keys.go
+++ b/modules/core/02-client/types/keys.go
@@ -34,7 +34,7 @@ func FormatClientIdentifier(clientType string, sequence uint64) string {
 
 // IsClientIDFormat checks if a clientID is in the format required on the SDK for
 // parsing client identifiers. The client identifier must be in the form: `{client-type}-{N}
-var IsClientIDFormat = regexp.MustCompile(`^.*[^\n-]-[0-9]{1,20}$`).MatchString
+var IsClientIDFormat = regexp.MustCompile(`^\w+([\w-]+\w)?-[0-9]{1,20}$`).MatchString
 
 // IsValidClientID checks if the clientID is valid and can be parsed into the client
 // identifier format.
--- a/modules/core/02-client/types/keys_test.go
+++ b/modules/core/02-client/types/keys_test.go
@@ -35,22 +35,31 @@ func TestParseClientIdentifier(t *testing.T) {
 		{"negative sequence", "tendermint--1", "tendermint", 0, false},
 		{"invalid format", "tendermint-tm", "tendermint", 0, false},
 		{"empty clientype", " -100", "tendermint", 0, false},
+		{"with in the middle tabs", "a\t\t\t-100", "tendermint", 0, false},
+		{"leading tabs", "\t\t\ta-100", "tendermint", 0, false},
+		{"with whitespace", "                  a-100", "tendermint", 0, false},
+		{"leading hyphens", "-----a-100", "tendermint", 0, false},
+		{"with slash", "tendermint/-100", "tendermint", 0, false},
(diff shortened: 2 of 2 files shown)


modules/core/02-client/types/keys_test.go
```diff
@@ -35,22 +35,31 @@ func TestParseClientIdentifier(t *testing.T) {
 		{"negative sequence", "tendermint--1", "tendermint", 0, false},
 		{"invalid format", "tendermint-tm", "tendermint", 0, false},
 		{"empty clientype", " -100", "tendermint", 0, false},
+		{"with in the middle tabs", "a\t\t\t-100", "tendermint", 0, false},
+		{"leading tabs", "\t\t\ta-100", "tendermint", 0, false},
+		{"with whitespace", "                  a-100", "tendermint", 0, false},
+		{"leading hyphens", "-----a-100", "tendermint", 0, false},
+		{"with slash", "tendermint/-100", "tendermint", 0, false},
+		{"non-ASCII:: emoji", "🚨😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎😎-100", "tendermint", 0, false},
```
crodriguezvega, line 43: Gotta love this test case :)

crodriguezvega:
@colin-axner in what version do you think we should release this? And could it be back ported?

crodriguezvega:
@odeke-em Can you please add a changelog entry in the `Bug fixes` section?

damiannolan (approved):
🙏

AdityaSripal (approved):

modules/core/02-client/types/keys.go
```diff
@@ -34,7 +34,7 @@ func FormatClientIdentifier(clientType string, sequence uint64) string {
 
 // IsClientIDFormat checks if a clientID is in the format required on the SDK for
 // parsing client identifiers. The client identifier must be in the form: `{client-type}-{N}
-var IsClientIDFormat = regexp.MustCompile(`^.*[^\n-]-[0-9]{1,20}$`).MatchString
+var IsClientIDFormat = regexp.MustCompile(`^\w+([\w-]+\w)?-[0-9]{1,20}$`).MatchString
```
AdityaSripal, line 37: Could you document what this regex is doing in the godocs? Thanks!

odeke-em: I've updated it in the PR #2510, thank you for the request @AdityaSripal!

PR #1826, opened 2022-07-28
docs: migration docs v4 to v5

crodriguezvega:
## Description

closes: #1781
closed: #2083 

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #2302, opened 2022-09-19
fix: ica handshake reopening channel - use `GetAppVersion` in favour of `channel.Version`

damiannolan:
## Description

- Adding unwrapping of channel version to ica controller handshake reopening flow

closes: #XXXX

---

write a little note why.


crodriguezvega (commented):
Can we add a changelog entry, @damiannolan? :)

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -70,7 +70,12 @@ func (k Keeper) OnChanOpenInit(
 			return "", sdkerrors.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s is already OPEN", activeChannelID, portID)
 		}
 
-		if !icatypes.IsPreviousMetadataEqual(channel.Version, metadata) {
+		appVersion, found := k.GetAppVersion(ctx, portID, activeChannelID)
+		if !found {
+			panic(fmt.Sprintf("active channel mapping set for %s, but channel does not exist in channel store", activeChannelID))
```
chatton, line 75: looks like this panic introduces multiple conventions in the function (error return + panic) does it make more sense to return an error instead?

colin-axner (commented):
Is it possible to add a test case for the exact issue we had?

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
colin-axner: Do we need to panic?

AdityaSripal (approved):
Nice work!

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
damiannolan: As mentioned earlier in person I just followed essentially the same used above [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L64-L67) - as in some places in ibc-go code we just `panic` because it's essentially unreachable code.

damiannolan:
I've been modified the existing the test `TestControlAccountAfterChannelClose` to use an ICS29/ICS27 channel version.

It looks like we need to handle the same unwrapping of the channel version on the host in the `OnChanOpenTry` callback: https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L62.

This requires us to compose the `hostKeeper` with an `ics4Wrapper` and pass the `feeKeeper` similarly to how we do with the controller: https://github.com/cosmos/ibc-go/blob/main/testing/simapp/app.go#L401
This has to be done to unwrap the ICS27 app version from the fee enabled version correctly. 

I'll make the changes locally and see if it resolves the failing test before pushing an update.

jackzampolin:
Thank you for this fix ibc-go team!

crodriguezvega:
@damiannolan Could you please document this API change in the [PR for the migration docs of v5](https://github.com/cosmos/ibc-go/pull/1826)?

colin-axner (approved):

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
chatton: maybe we could add a comment indicating as such? The message we panic with seems like it would be a regular error that we could run into.

chatton (approved):
LGTM just left a very minor nit!

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
charleenfei, line 75: a nit: `channel does not exist in channel store` is kind of confusing if the error is `GetAppVersion` returning not found -- maybe something like channel version does not exist?

charleenfei (approved):
nice!

AdityaSripal (approved):

Issue #2245, opened 2022-09-08
Add port prefix iterator function to 04-channel


colin-axner:
## Summary

Add a function to 04-channel that takes in a port prefix and returns a list of all the portID, channelID pairings that have that port prefix. This should be used in place of `GetAllActiveChannels` in the controller module migrations

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #2304, opened 2022-09-19
Adding GetChannelsWithPortPrefix function to 04-channel

chatton:
## Description

closes: #2245

---

write a little note why.


modules/apps/27-interchain-accounts/controller/keeper/migrations.go
```diff
@@ -31,7 +33,7 @@ func (m Migrator) AssertChannelCapabilityMigrations(ctx sdk.Context) error {
 
 		isAuthenticated := m.keeper.scopedKeeper.AuthenticateCapability(ctx, cap, name)
 		if !isAuthenticated {
-			return sdkerrors.Wrapf(capabilitytypes.ErrCapabilityNotOwned, "expected capability owner: %s", types.SubModuleName)
+			return sdkerrors.Wrapf(capabilitytypes.ErrCapabilityNotOwned, "expected capability owner: %s", commitmenttypes.SubModuleName)
```
damiannolan, line 36: Not sure how this ended up as `commitmenttypes` I see the import previously would've been incorrect also. 
It should be `controllertypes.SubModuleName`.

modules/core/04-channel/keeper/keeper.go
```diff
@@ -469,3 +488,8 @@ func (k Keeper) iterateHashes(_ sdk.Context, iterator db.Iterator, cb func(portI
 		}
 	}
 }
+
+// filteredPortPrefix returns the prefix key for the given port prefix.
+func filteredPortPrefix(portPrefix string) []byte {
+	return []byte(fmt.Sprintf("%s/ports/%s", host.KeyChannelEndPrefix, portPrefix))
```
damiannolan, line 494: small nit:

```suggestion
	return []byte(fmt.Sprintf("%s/%s/%s", host.KeyChannelEndPrefix, host.KeyPortPrefix, portPrefix))
```

modules/apps/27-interchain-accounts/controller/keeper/migrations.go
chatton: Ah right, I was wondering about that actually, I changed the import to be inline with other imports in the file. I can update this now.

modules/core/04-channel/keeper/keeper.go
chatton: good call

modules/apps/27-interchain-accounts/controller/keeper/migrations.go
chatton: I think the dangers of `types` as the import alias.

modules/apps/27-interchain-accounts/controller/keeper/migrations.go
```diff
@@ -22,16 +23,17 @@ func NewMigrator(keeper *Keeper) Migrator {
 // AssertChannelCapabilityMigrations checks that all channel capabilities generated using the interchain accounts controller port prefix
 // are owned by the controller submodule and ibc.
 func (m Migrator) AssertChannelCapabilityMigrations(ctx sdk.Context) error {
-	for _, ch := range m.keeper.GetAllActiveChannels(ctx) {
+	filteredChannels := m.keeper.channelKeeper.GetChannelsWithPortPrefix(ctx, icatypes.PortPrefix)
+	for _, ch := range filteredChannels {
 		name := host.ChannelCapabilityPath(ch.PortId, ch.ChannelId)
-		cap, found := m.keeper.scopedKeeper.GetCapability(ctx, name)
+		capacity, found := m.keeper.scopedKeeper.GetCapability(ctx, name)
```
chatton, line 29: I renamed this to avoid shadowing the built in `cap` function.

modules/core/04-channel/keeper/keeper.go
```diff
@@ -469,3 +488,8 @@ func (k Keeper) iterateHashes(_ sdk.Context, iterator db.Iterator, cb func(portI
 		}
 	}
 }
+
+// filteredPortPrefix returns the prefix key for the given port prefix.
+func filteredPortPrefix(portPrefix string) []byte {
```
crodriguezvega, line 493: Would it make sense to place this function [in `types/keys.go`](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/keys.go) instead?

crodriguezvega, line 493: Would it make sense to place this function [in `types/keys.go`](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/keys.go) instead?

chatton: good suggestion, done!

modules/core/04-channel/keeper/keeper.go
```diff
@@ -384,6 +384,24 @@ func (k Keeper) IterateChannels(ctx sdk.Context, cb func(types.IdentifiedChannel
 	}
 }
 
+// GetChannelsWithPortPrefix returns all channels with the specified port prefix.
+func (k Keeper) GetChannelsWithPortPrefix(ctx sdk.Context, portPrefix string) []types.IdentifiedChannel {
```
colin-axner, line 388: Could you add a basic test in 04-channel keeper for this function? It'd be nice to know prefixes and full portIDs are supported

colin-axner (commented):
LGTM, just would like a little testing in 04-channel as well. Nice work thus far!

modules/apps/27-interchain-accounts/controller/keeper/migrations_test.go
```diff
@@ -16,15 +20,23 @@ func (suite *KeeperTestSuite) TestAssertChannelCapabilityMigrations() {
 			func() {},
 			true,
 		},
+		{
+			"channel with different port is filtered out",
+			func() {
+				portIdWithOutPrefix := ibctesting.MockPort
+				suite.chainA.GetSimApp().IBCKeeper.ChannelKeeper.SetChannel(suite.chainA.GetContext(), portIdWithOutPrefix, ibctesting.FirstChannelID, channeltypes.Channel{
+					ConnectionHops: []string{ibctesting.FirstConnectionID},
+				})
```
damiannolan, line 29: We normally favour using `ID` over `Id` in naming (except proto generated code).

```suggestion
				portIDWithOutPrefix := ibctesting.MockPort
				suite.chainA.GetSimApp().IBCKeeper.ChannelKeeper.SetChannel(suite.chainA.GetContext(), portIDWithOutPrefix, ibctesting.FirstChannelID, channeltypes.Channel{
					ConnectionHops: []string{ibctesting.FirstConnectionID},
				})
```

modules/core/04-channel/keeper/keeper.go
damiannolan, line 388: Should we consider naming this function to align with `GetAllChannels` below which returns `[]types.IdentifiedChannel` also?

Maybe `GetAllChannelsWithPortPrefix` or `GetAllPortPrefixedChannels`? 

I don't feel particularly strongly but just a thought. cc. @crodriguezvega @colin-axner

chatton: I like `GetAllChannelsWithPortPrefix` to stay consistent

modules/core/04-channel/keeper/keeper.go
```diff
@@ -384,6 +384,28 @@ func (k Keeper) IterateChannels(ctx sdk.Context, cb func(types.IdentifiedChannel
 	}
 }
 
+// GetAllChannelsWithPortPrefix returns all channels with the specified port prefix. If an empty prefix is provided
+// all channels will be returned.
+func (k Keeper) GetAllChannelsWithPortPrefix(ctx sdk.Context, portPrefix string) []types.IdentifiedChannel {
+	if strings.TrimSpace(portPrefix) == "" {
+		return k.GetAllChannels(ctx)
+	}
```
chatton, line 392: missed this condition before the tests!

colin-axner: Nice! This is a good addition

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -81,6 +82,73 @@ func (suite *KeeperTestSuite) TestGetAppVersion() {
...
+		name             string
+		prefix           string
+		allChannels      []types.IdentifiedChannel
+		expectedChannels []types.IdentifiedChannel
+	}{
+		{
+			name:   "transfer channel is retrieved with prefix",
+			prefix: "tra",
+			allChannels: []types.IdentifiedChannel{
+				types.NewIdentifiedChannel(transfertypes.TypeMsgTransfer, ibctesting.FirstChannelID, types.Channel{}),
```
colin-axner, line 97: haha didn't realize  "TypeMsgTransfer" still exists! It was historically used for to fulfill a "Type()" interface function on sdk.Msg, but since the SDK now uses a gRPC router for routing messages, the type can be inferred from the proto url

I'd use the transfertypes.PortID, we should probably remove unused constants (since "TypeMsgTransfer" has no intended meaning right now)

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -81,6 +82,73 @@ func (suite *KeeperTestSuite) TestGetAppVersion() {
...
+				types.NewIdentifiedChannel("different-portid", "channel-1", types.Channel{}),
```
colin-axner, line 98: should we make a local constant for "channel-1"?

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -81,6 +82,73 @@ func (suite *KeeperTestSuite) TestGetAppVersion() {
...
+			},
+			expectedChannels: []types.IdentifiedChannel{types.NewIdentifiedChannel(transfertypes.TypeMsgTransfer, ibctesting.FirstChannelID, types.Channel{})},
+		},
+		{
+			name:   "matches port with full name as prefix",
+			prefix: transfertypes.TypeMsgTransfer,
+			allChannels: []types.IdentifiedChannel{
+				types.NewIdentifiedChannel(transfertypes.TypeMsgTransfer, "channel-0", types.Channel{}),
+				types.NewIdentifiedChannel("different-portid", "channel-1", types.Channel{}),
```
colin-axner, line 107: looks like these get reused, should we specify them as local variables "transferChannel" "differentPortIDChannel"?

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -81,6 +82,73 @@ func (suite *KeeperTestSuite) TestGetAppVersion() {
...
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest()
+
+			for _, ch := range tc.allChannels {
+				suite.chainA.GetSimApp().GetIBCKeeper().ChannelKeeper.SetChannel(suite.chainA.GetContext(), ch.PortId, ch.ChannelId, types.Channel{})
+			}
```
colin-axner, line 141: Maybe we should just preset a transfer channel and a different portID channel for all test cases? Then just change the prefix and the expected channels?

colin-axner (approved):
LGTM! Left some suggestions for the tests, mostly nits. Happy to go with what you think would be best

modules/core/04-channel/keeper/keeper.go
colin-axner: Sounds good to me. I have no preference

modules/apps/27-interchain-accounts/controller/keeper/migrations_test.go
colin-axner: We only favor "Id" in proto naming because using "ID" messes with the grpc gateway generation (I'm not sure why nor if this is still an issue, but it was when we initially switched to proto)

modules/core/04-channel/keeper/keeper_test.go
chatton: personally I think it's more readable to have the actual and expected values side by side in a given test case, without needing to look down at the test to see which channels are added there as initial values.

There is some duplication this way, but I would lean in favour of readability vs de-deduplication for a test suite like this.

WDYT?

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -81,6 +83,97 @@ func (suite *KeeperTestSuite) TestGetAppVersion() {
...
+			actualChannels := suite.chainA.GetSimApp().GetIBCKeeper().ChannelKeeper.GetAllChannelsWithPortPrefix(ctxA, tc.prefix)
+
+			suite.Require().True(containsAll(tc.expectedChannels, actualChannels))
+		})
+	}
+}
+
+// containsAll verifies if all elements in the expected slice exist in the actual slice
+// independent of order.
+func containsAll(expected, actual []types.IdentifiedChannel) bool {
```
chatton, line 161: I added this function as I saw that the order of the values returned in `GetAllChannels` is not necessarily in the order they were added.

modules/core/04-channel/keeper/keeper_test.go
colin-axner: That makes sense to me :+1: The only other change I might do is:
```go
allChannels: []types.IdentifiedChannel{
	types.NewIdentifiedChannel(transferChannel),
	types.NewIdentifiedChannel(differentPortIDChannel),
},
expectedChannels: []types.IdentifiedChannel{
	types.NewIdentifiedChannel(transferChannel}),
	types.NewIdentifiedChannel(differentPortIDChannel),
},
```
			
Just so that there is less to read

colin-axner (approved):
Followup ACK

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -81,6 +83,87 @@ func (suite *KeeperTestSuite) TestGetAppVersion() {
...
+	const (
+		secondChannelID        = "channel-1"
+		differentChannelPortID = "different-portid"
+	)
+
+	allChannels := []types.IdentifiedChannel{
+		types.NewIdentifiedChannel(transfertypes.PortID, ibctesting.FirstChannelID, types.Channel{}),
+		types.NewIdentifiedChannel(differentChannelPortID, secondChannelID, types.Channel{}),
+	}
+
```
chatton, line 98: great suggestion @colin-axner , this makes things much cleaner.

PR #2305, opened 2022-09-19
added optional packet metadata to the packet and message types

nicolaslara:
## Description

Allow packets to include an optional metadata field. 

This PR introduces the data changes. A new PR will add the integration and tests.

---

write a little note why.


agouin:
Which version is this targeted for? Would be great to have it in v6

proto/ibc/applications/transfer/v2/packet.proto
```diff
@@ -16,4 +16,6 @@ message FungibleTokenPacketData {
   string sender = 3;
   // the recipient address on the destination chain
   string receiver = 4;
+  // optional metadata
+  bytes metadata = 5;
```
colin-axner, line 20: What's the reasoning for bytes instead of string? Just curious since all the other packet data fields are strings

modules/apps/transfer/types/packet.go
```diff
@@ -34,6 +34,20 @@ func NewFungibleTokenPacketData(
 	}
 }
 
+func NewFungibleTokenPacketDataWithMetadata(
```
colin-axner, line 37: We should break the API of NewFungibleTokenPacketData. In backports we could include this function, but maybe it has no benefit if we remove it in v6. Thoughts?

modules/apps/transfer/types/msgs.go
```diff
@@ -34,6 +34,24 @@ func NewMsgTransfer(
 	}
 }
 
+func NewMsgTransferWithMetadata(
```
colin-axner, line 37: lets break API, same comment as for packet data

colin-axner (commented):
Concept ACK. Will review your testing pr when I have time. Thanks for splitting the changes up. I think it makes sense to merge this first when ready and then do the testing changes. It'll make backporting easier as well

We should also add a changelog entry and add an entry to the migration docs for docs/migrations/v5-to-v6.md

proto/ibc/applications/transfer/v2/packet.proto
nicolaslara: I figured someone may want to use protobufs for metadata, so they'd encode it as bytes

modules/apps/transfer/types/packet.go
nicolaslara: that sounds good to me. Not sure what the process for this is, do we create different PRs for each version?

proto/ibc/applications/transfer/v2/packet.proto
crodriguezvega, line 20: Is the `metadata` something we should also include in the events emitted [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go#L191-L209), [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go#L235-L262) and [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go#L282-L290). If yes, then we should also update [the docs](https://github.com/cosmos/ibc-go/blob/main/docs/apps/transfer/events.md#events).

crodriguezvega (commented):
Thanks for this useful addition, @nicolaslara. I left a couple of initial comments.

proto/ibc/applications/transfer/v1/tx.proto
```diff
@@ -38,6 +38,8 @@ message MsgTransfer {
   // Timeout timestamp in absolute nanoseconds since unix epoch.
   // The timeout is disabled when set to 0.
   uint64 timeout_timestamp = 7 [(gogoproto.moretags) = "yaml:\"timeout_timestamp\""];
+  // optional metadata
+  bytes metadata = 8;
```
crodriguezvega, line 42: We should update [the docs](https://github.com/cosmos/ibc-go/blob/main/docs/apps/transfer/messages.md#msgtransfer) with this addition.

modules/apps/transfer/types/packet.go
colin-axner: We will add the backport (for a new minor version on each supported major release line) tags to this pr. Once merged, mergify will try to automatically backport this to the release line. At this point we can manually modify each backport to remove any API breaking changes. This is something we can handle

nicolaslara:
Updated the the code to have the changes break the API and added the events and docs.

modules/apps/transfer/ibc_module.go
```diff
@@ -194,6 +194,7 @@ func (im IBCModule) OnRecvPacket(
 		sdk.NewAttribute(types.AttributeKeyReceiver, data.Receiver),
 		sdk.NewAttribute(types.AttributeKeyDenom, data.Denom),
 		sdk.NewAttribute(types.AttributeKeyAmount, data.Amount),
+		sdk.NewAttribute(types.AttributeKeyMetadata, string(data.Metadata)),
```
AdityaSripal, line 197: We may have to emit this in hex format? I believe we ran. into this issue before with the packet data:

https://github.com/cosmos/ibc-go/blob/33639176c35fa421c4affbbc26f7f6fab10cdfe3/modules/core/04-channel/keeper/events.go#L135

https://github.com/cosmos/ibc-go/pull/144

modules/apps/transfer/ibc_module.go
```diff
@@ -240,6 +241,7 @@ func (im IBCModule) OnAcknowledgementPacket(
 			sdk.NewAttribute(types.AttributeKeyReceiver, data.Receiver),
 			sdk.NewAttribute(types.AttributeKeyDenom, data.Denom),
 			sdk.NewAttribute(types.AttributeKeyAmount, data.Amount),
+			sdk.NewAttribute(types.AttributeKeyMetadata, string(data.Metadata)),
```
AdityaSripal, line 244: ditto other events

modules/apps/transfer/types/events.go
```diff
@@ -18,4 +18,5 @@ const (
 	AttributeKeyAck            = "acknowledgement"
 	AttributeKeyAckError       = "error"
 	AttributeKeyTraceHash      = "trace_hash"
+	AttributeKeyMetadata       = "amount"
```
AdityaSripal, line 21:
```suggestion
	AttributeKeyMetadata       = "metadata"
```

modules/apps/transfer/types/packet.go
AdityaSripal, line 37: I believe @colin-axner meant that this function should be removed and the above function should take in a `metadata` field

AdityaSripal (changes requested):
Looks great. Main point about events should be noted

modules/apps/transfer/types/packet.go
nicolaslara: fixed

modules/apps/transfer/types/events.go
nicolaslara: fixed

modules/apps/transfer/ibc_module.go
nicolaslara: fixed

modules/apps/transfer/types/events.go
colin-axner: Looks like maybe the fixing commit needs to be pushed?

modules/apps/transfer/keeper/relay.go
```diff
@@ -59,29 +59,7 @@ func (k Keeper) SendTransfer(
...
-// sendTransfer handles transfer sending logic.
-func (k Keeper) sendTransfer(
-	ctx sdk.Context,
-	sourcePort,
-	sourceChannel string,
-	token sdk.Coin,
-	sender sdk.AccAddress,
-	receiver string,
-	timeoutHeight clienttypes.Height,
-	timeoutTimestamp uint64,
```
colin-axner, line 84: This should be added back, this will help with backports. The public `SendTransfer` function will soon be deleted

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -19,8 +18,9 @@ func (k Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types.
 		return nil, err
 	}
 
-	if err := k.sendTransfer(
+	if err := k.SendTransfer(
```
colin-axner, line 21: This should be reverted

modules/apps/transfer/client/cli/tx.go
```diff
@@ -22,6 +22,7 @@ const (
 	flagPacketTimeoutHeight    = "packet-timeout-height"
 	flagPacketTimeoutTimestamp = "packet-timeout-timestamp"
 	flagAbsoluteTimeouts       = "absolute-timeouts"
+	flagMetadata               = "metadata"
```
colin-axner, line 25: I think this flag needs to be added below, see line 120. The default value can be an empty string

colin-axner:
Hey @nicolaslara, thanks for handling this work. Just checking in so we all know what is left to be done before we can merge and include the changes in a backport:

For this pr I think you just need to handle the merge conflicts and apply both @AdityaSripal's and my suggestions. I think you might have an unpushed commit? We are happy to handle any of this work, just let us know

For backporting, we will remove the API breaking changes. External users should be expected to use code similar to:
```go
msgTransfer := transfertypes.NewMsgTransfer(...)
msgTransfer.Metadata = metadata
```

We will handle the backports

The other pr you have should just be the e2e test (after we merge this). I will look it over in detail tomorrow. I can open a pr to your branch if I have several requested changes. With regards to the testing package bumping, we will handle this in a separate pr

Let us know if you have any concerns/questions

nicolaslara:
Definitely have some unpushed commits. I'll push those as soon as I can (travelling right now). I'll ping you back when those are in. The tests should work on v5 once this is merged (and backported), but I'll adapt them if needed

nicolaslara:
Definitely have some unpushed commits. I'll push those as soon as I can (travelling right now). I'll ping you back when those are in. The tests should work on v5 once this is merged (and backported), but I'll adapt them if needed

nicolaslara:
@colin-axner added the missing commits

modules/apps/transfer/client/cli/tx.go
```diff
@@ -111,7 +117,7 @@ corresponding to the counterparty channel. Any timeout set to 0 is disabled.`),
 			}
 
 			msg := types.NewMsgTransfer(
-				srcPort, srcChannel, coin, sender, receiver, timeoutHeight, timeoutTimestamp,
+				srcPort, srcChannel, coin, sender, receiver, timeoutHeight, timeoutTimestamp, []byte(metadataStr),
```
crodriguezvega, line 120: Would it be possible to pass in as metadata some serialized protocol buffer data, for example? Or is it expected that the metadata should only be string information?

nicolaslara: Maybe it should be bytes on those functions too. That give users more flexibility. I can make the change

crodriguezvega: Thanks in advance! Yes, I think that generalization will be good to have.

nicolaslara: ohh wait, looked at this in detail now. We're already using bytes everywhere else, and I think it makes more sense to keep it as string in the CLI so that it's easy to write/use. We could use hex encoded bytes, but for 99% of the cases we'd end up making it harder for the users. 

I think if they really need bytes, they can create the message manually and pass it via grpc.

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -143,6 +161,7 @@ func (suite *KeeperTestSuite) TestSendTransfer() {
 			err = suite.chainA.GetSimApp().TransferKeeper.SendTransfer(
 				suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, amount,
 				sender, suite.chainB.SenderAccount.GetAddress().String(), suite.chainB.GetTimeoutHeight(), 0,
+				metadata,
```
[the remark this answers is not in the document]
nicolaslara: That makes sense. I made that change.

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -219,7 +245,7 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
 			if tc.recvIsSource {
```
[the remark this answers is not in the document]
nicolaslara: same as above. Added the reset.

nicolaslara:
Added the changes based on the reviews above, merged main and resolved the conflicts. The only outstanding discussion is wether the CLI should use String or Bytes (from hex) as its input for metadata. I lean towards leaving it as is (string) but if you feel strongly about it I'll change it to bytes. 

I'm promoting this from draft to regular PR since I think it's ready to merge/backport.

crodriguezvega:
> Added the changes based on the reviews above, merged main and resolved the conflicts. The only outstanding discussion is wether the CLI should use String or Bytes (from hex) as its input for metadata. I lean towards leaving it as is (string) but if you feel strongly about it I'll change it to bytes.

Fine by leaving it like it is then. @colin-axner, do you agree? But if we leave it like this, should some comment be added to the help of CLI to make it clear that the CLI only works with strings? Just to avoid any potential confusion...

colin-axner:
Let's leave it as a string, we can always modify later there's a good reason to do so

nicolaslara:
Added the flag with a comment about using bytes instead of strings

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -2,7 +2,6 @@ package keeper
 
 import (
 	"context"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
```
colin-axner, line 5: `make lint-fix` will fix this

colin-axner:
Looks like there is a minor error in the tests:
```
Error: modules/apps/29-fee/transfer_test.go:33:232: not enough arguments in call to transfertypes.NewMsgTransfer
```

damiannolan (approved):
Awesome, looks good to me! Thanks @nicolaslara 

LGTM pending minor fix in ics29 transfer test 🚀

modules/apps/transfer/keeper/msg_server.go
nicolaslara: Thanks! ran it now.

nicolaslara:
fixed the failing test

AdityaSripal:
Nice work @nicolaslara and team!

PR #2369, opened 2022-09-22
Support full compatibility test runs through a GitHub action

chatton:
## Description

See this [workflow](https://github.com/cosmos/ibc-go/actions/runs/3104251129) as an example.

closes: #XXXX

---

write a little note why.

### DIFF
--- a/.github/workflows/e2e-compatibility.yaml
+++ b/.github/workflows/e2e-compatibility.yaml
@@ -27,12 +27,6 @@ jobs:
           echo "::set-output name=docker-tag::$docker_tag"
         id: set-docker-tag
 
-  log-docker-tag:
-    needs: determine-docker-tag
-    runs-on: ubuntu-latest
-    steps:
-      - run: echo "${{ needs.determine-docker-tag.outputs.docker-tag }}"
-
   load-test-matrix:
     outputs:
       test-matrix: ${{ steps.set-test-matrix.outputs.test-matrix }}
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-compatibility.yaml
```diff
@@ -0,0 +1,88 @@
+name: Compatibility E2E
+on:
+  workflow_dispatch:
+    release-branch:
+      description: 'Release branch to test'
+      required: true
+      type: choice
+      options:
+        - release/v5.0.x
```
chatton, line 9: this workflow won't become available until it is merged to main, the sample workflow I linked hard coded values rather than receiving them from UI

crodriguezvega: So here we would add other release branches to have them as options in the GitHub UI, right?

crodriguezvega (approved):
🪄

crodriguezvega:
If we want to see the results of the tests we would need to go to the `e2e-tests` job and check at the bottom of `Run e2e test`, right? 

Maybe we can think of creating a nice visual dashboard for this?

.github/workflows/e2e-compatibility.yaml
chatton: yes, anything in this list will show up in a drop down menu. There's nothing stopping us to add a new field which allows anything, but I think the dropdown is a nice UX. Perhaps we could add an optional override which would take precedence in the case of wanting to run a custom image.

colin-axner (approved):
Amazing! Can't wait to give it a shot!

PR #2377, opened 2022-09-25
MsgTransferResponse add sequence

### LINKED ISSUE #1969: MsgTransferResponse should include the packet sequence
## Summary

`MsgTransferResponse` should include the sequence of the packet the transfer is sent in. The value derived in `SendTransfer` within `relay.go` should be returned in the message response. The SendTransfer API will need to be modified. 

#### For Admin Use

- [ ] Not duplicate issue

agouin:
## Description

Returns sequence from `sendTransfer`, and returns it with the `MsgTransferResponse`. This is not an API breaking change.

Retrieving the sequence at the time of creating the transfer is necessary in the packet forward middleware for correlation with multihop packet flows.

https://github.com/strangelove-ventures/packet-forward-middleware/pull/33
https://github.com/strangelove-ventures/ibctest/pull/306

Closes #1969 

---

Existing test coverage exercises this new method due to the re-routing of `SendTransfer` through `SendPacketTransfer`


crodriguezvega:
Thanks for opening this PR, @agouin.

We already had [this issue](https://github.com/cosmos/ibc-go/issues/1969) to add the sequence of the packet in `MsgTransferResponse`. This is an API breaking change, but once this is implemented the `SendTransfer` function would also return the packet sequence.

But we also have [this other issue](https://github.com/cosmos/ibc-go/issues/1918) to make `SendTransfer` private. So our idea is eventually to make the `Transfer` RPC handler be the only entry point for sending IBC transfers. That would mean that the packet forward middleware cannot use `SendTransfer`, but does the packet forward middleware strictly need to directly use the `SendTransfer` function or would it be possible for it to make use of the [`Transfer` RPC handler](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/msg_server.go#L14) and get the packet sequence from the `MsgTransferResponse`? If that is possible, then we could move on the direction that we had planned with those 2 issues mentioned above.

agouin:
> We already had https://github.com/cosmos/ibc-go/issues/1969 to add the sequence of the packet in MsgTransferResponse. This is an API breaking change, but once this is implemented the SendTransfer function would also return the packet sequence.

Thank you for pointing this out @crodriguezvega . I updated this PR to use the breaking change strategy which can close #1969.

> does the packet forward middleware strictly need to directly use the `SendTransfer` function or would it be possible for it to make use of the [Transfer RPC handler](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/msg_server.go#L14) and get the packet sequence from the `MsgTransferResponse`?

It should be fine for us to migrate to the Transfer RPC handler with the sequence in the `MsgTransferResponse`. With this change merged in we will be unblocked and then #1918 will not break our implementation once it is added.

crodriguezvega:
> Thank you for pointing this out @crodriguezvega . I updated this PR to use the breaking change strategy which can close https://github.com/cosmos/ibc-go/issues/1969.

Thanks a lot for adjusting the implementation, @agouin. Really appreciate it!

crodriguezvega:
@agouin We just talked about this change and we are going to try to release it in v6. Timeline for final release is probably end of October. Would that be ok for you?

We are also going to try to implement https://github.com/cosmos/ibc-go/issues/1918 so that we can bundle and release together a few API breaking changes.

agouin:
> Timeline for final release is probably end of October/beginning of November. Would that be ok for you?

Yes this sounds good, we are going to attempt to use a workaround for v3 and v5 by fetching the packet sequence with `GetNextSequenceSend(...) - 1` after it is sent, so we should be able to wait for v6 with the fix.

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -140,7 +140,7 @@ func (suite *KeeperTestSuite) TestSendTransfer() {
 				suite.Require().NoError(err) // message committed
 			}
 
-			err = suite.chainA.GetSimApp().TransferKeeper.SendTransfer(
+			_, err = suite.chainA.GetSimApp().TransferKeeper.SendTransfer(
```
damiannolan, line 143: This is a super nit, but maybe we could not discard the return value and assert not zero in the `expPass` cases?
I always lean towards minimising as many discards as possible.

damiannolan (approved):
Excellent work, just left one nit! 
Thanks 🙏

modules/apps/transfer/keeper/relay_test.go
agouin: good call, updated

modules/apps/transfer/keeper/relay.go
```diff
@@ -59,18 +59,18 @@ func (k Keeper) SendTransfer(
 	receiver string,
 	timeoutHeight clienttypes.Height,
 	timeoutTimestamp uint64,
-) error {
+) (uint64, error) {
```
crodriguezvega, line 62: Could we mention in the godoc of this function something like `The first return value is the sequence number of the packet sent. It will be greater than zero in case of success, zero otherwise.`?

crodriguezvega (commented):
Thanks, @agouin! I just left a couple of nits.

modules/apps/transfer/keeper/relay.go
agouin: Good catch, updated.

agouin:
@crodriguezvega @colin-axner For this and #2305, there are discussions of backporting to previous versions as far back as `v3`. If we go with that plan, I'd assume we will want an API breaking implementation for v6, but a non-breaking for prior versions. That would really help us with the packet-forward-middleware also so that we can have a more consistent implementation across ibc-go versions.

Let me know if we can move forward with this and I can get an additional PR up for a non-breaking backport of this.

colin-axner:
@agouin do you need this backported as a patch release or minor release? That is, are the changes you are making on packet forwarding middleware state machine breaking? Thus requiring chains to upgrade to a new version anyways? I am wondering if we plan to backport to v3.3 and v4.1 or just to a new minor release

agouin:
> do you need this backported as a patch release or minor release? That is, are the changes you are making on packet forwarding middleware state machine breaking? Thus requiring chains to upgrade to a new version anyways? I am wondering if we plan to backport to v3.3 and v4.1 or just to a new minor release

Chains will need to upgrade to the new version of the packet forward middleware, yes. So a new minor release of ibc-go v3/v4 with this and the memo field addition would be great.

AdityaSripal (approved):
Thanks for the addition!

AdityaSripal:
Should we merge this first: https://github.com/cosmos/ibc-go/pull/1703

And then rebase this PR against it?

cc: @colin-axner

CHANGELOG.md
```diff
@@ -93,6 +93,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * (light-clients/06-solomachine) Moving `verifyMisbehaviour` function from update.go to misbehaviour_handle.go.
 * (apps/27-interchain-accounts) [\#2297](https://github.com/cosmos/ibc-go/pull/2297) Adding cli command to generate ICS27 packet data.
 * (modules/core/keeper) [\#1728](https://github.com/cosmos/ibc-go/pull/2399) Updated channel callback errors to include portID & channelID for better identification of errors.
+* (transfer) [\#2377](https://github.com/cosmos/ibc-go/pull/2377) Adding `sequence` to `MsgTransferResponse`.
```
colin-axner, line 96: Thanks for moving this out of API breaking. Could you switch it to state machine breaking? The addition of a field in the response type should change the block hash

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -39,5 +40,5 @@ func (k Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types.
 		),
 	})
 
-	return &types.MsgTransferResponse{}, nil
+	return &types.MsgTransferResponse{Sequence: sequence}, nil
```
colin-axner, line 43: Could you assert the res.Sequence is non 0 on passing and 0 on failing tests (for msg_server_test.go)?

colin-axner (approved):
Thanks @agouin! We will handle the backports. I left two minor suggestions. Could you address these? Then I will merge the pr

modules/apps/transfer/keeper/msg_server.go
agouin: sure! updated.

CHANGELOG.md
agouin: Done

colin-axner (approved):
Thanks @agouin! :rocket:

modules/apps/transfer/keeper/msg_server_test.go
```diff
@@ -61,9 +61,11 @@ func (suite *KeeperTestSuite) TestMsgTransfer() {
 		res, err := suite.chainA.GetSimApp().TransferKeeper.Transfer(sdk.WrapSDKContext(suite.chainA.GetContext()), msg)
 
 		if tc.expPass {
+			suite.Require().NotEqual(res.Sequence, uint64(0))
 			suite.Require().NoError(err)
 			suite.Require().NotNil(res)
 		} else {
+			suite.Require().Equal(res.Sequence, uint64(0))
```
damiannolan, line 68: Looks like there's a `nil` pointer happening here in tests. We can't make the assertion at this level as `if err != nil` we return `nil` for the `res`.

agouin: Whoops, thanks. Fixed.

PR #2370, opened 2022-09-22
Add compatibility tests for v5.0.x

chatton:
## Description

While the json file is a bit noisy and has a lot of duplication, this is something we could spend time to generate with tooling if it becomes an issue. For now I think it will be fine to maintain by hand.

~[This workflow](https://github.com/cosmos/ibc-go/actions/runs/3106285592) was triggered by the UI.~

~[After bumping versions for client test](https://github.com/cosmos/ibc-go/actions/runs/3111705623)~

[bump v2 and v3](https://github.com/cosmos/ibc-go/actions/runs/3112040088)

You can test this by going to 

`Actions` -> `Compatibility E2E` -> `run workflow` (choose this branch - not main) -> click `Run Workflow`

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Review `Codecov Report` in the comment section below once CI passes

PR #2396, opened 2022-09-26
Add compatibility tests for v5.0.x #2

chatton:
## Description

This PR replaces #2370

This is a re-work of the referenced PR. Instead of using a large json file per release branch, this method allows us to define smaller, more maintainable files per release branch / test suite pair.

Sample workflow: [here](https://github.com/cosmos/ibc-go/actions/runs/3129300394)

Update: One additional benefit, each job is limited to creating 256 tasks. A github workflow can have multiple jobs, i.e. transfer, client, incentivized etc. While we are not at that level yet (transfer test is creating 80), splitting up  like this leaves more room for growth without refactoring how the tests run in the future.

closes #2370

---

write a little note why.


.github/workflows/e2e-compatibility-workflow-call.yaml
```diff
@@ -0,0 +1,50 @@
+on:
+  workflow_call:
```
chatton, line 2: this is a re-usable workflow that we can call for each test suite.

chatton:
@colin-axner & @charleenfei thank you for reviewing #2370 after some discussion with @crodriguezvega and looking some more into Colin's suggestion, I refactored the workflow to be (much) more maintainable. The workflows are largely the same, it's mostly just the smaller config file we need to manage.

.github/workflows/e2e-compatibility.yaml
```diff
@@ -47,42 +48,29 @@ jobs:
...
-          # to create the matrix in the following job.
-          test_matrix="$(cat .github/compatibility-test-matrices/${{ needs.determine-docker-tag.outputs.docker-tag }}.json | jq -c)"
-          echo $test_matrix
-          echo "::set-output name=test-matrix::$test_matrix"
-        id: set-test-matrix
+  transfer:
+    needs:
+      - build-release-image
+      - determine-docker-tag
+    uses: ./.github/workflows/e2e-compatibility-workflow-call.yaml
```
chatton, line 55: this is how workflows can be re-used in github actions.

.github/workflows/e2e-compatibility.yaml
```diff
@@ -47,42 +48,29 @@ jobs:
...
+    with:
+      docker-tag: "${{ needs.determine-docker-tag.outputs.docker-tag }}"
+      test-suite: "transfer"
```
chatton, line 58: these strings map to sub directories, I'm open to alternatives if there are any suggestions!

damiannolan: Will we also need to add blocks for `interchain-accounts` here?

.github/workflows/e2e-compatibility-workflow-call.yaml
```diff
@@ -0,0 +1,50 @@
...
+          echo $test_matrix
+          echo "::set-output name=test-matrix::$test_matrix"
+        id: set-test-matrix
+
+  e2e:
+    runs-on: ubuntu-latest
+    needs: load-test-matrix
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.load-test-matrix.outputs.test-matrix) }}
```
damiannolan, line 34: Aha, so this magic uses `load-test-matrix` to parse the JSON files and set the values in `matrix` which is used below to run the e2e test and populate the env?

damiannolan (approved):
Just left a few questions, I'm getting a better grasp on the workflows, slowly but surely its becoming easier to parse.

The yaml gods have spoken! ⚒️⚡ 

L: 
  G:
    T:
      M: 

❤️

.github/workflows/e2e-compatibility.yaml
```diff
@@ -1,12 +1,13 @@
...
-      type: choice
-      options:
-        - release/v5.0.x
+    inputs:
+      release-branch:
+        description: 'Release branch to test'
+        required: true
+        type: choice
+        options:
+          - release/v5.0.x
```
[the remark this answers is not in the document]
chatton: yes - this is just for a dropdown, if we prefer, we can have a text box and enter anything we like. I just think the dropdown is better UX

.github/workflows/e2e-compatibility-workflow-call.yaml
chatton: yes the previous job `load-test-matrix` stores a json string as its output (the contents of a json file depending on args) and the `matrix` field accepts dynamic input (or statically defined in the workflow)

chatton:
@damiannolan 

> Will we also need to add blocks for interchain-accounts here?

Currently all the interchain accounts tests are running using the icad, so the state of the release branch in `ibc-go` should have no impact here right?

> I'm not sure I understand fully here. Could you provide me with the verbose output

Right maybe I can add some clarification, the tests themselves exist under `e2e/...` but the matrix files for the workflow live separately (the new dir added in this PR) `compatibility-test-matrices`. 

This additional field e.g. `test-suite: "transfer"` is literally just a convention I'm using to pass arguments to the reusable workflow so that we can run any test suite the same way by adhering to a directory structure.

`compatibility-test-matrices/<release-<version>/<test-suite>.json`

.github/workflows/e2e-compatibility.yaml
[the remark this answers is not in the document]
damiannolan: Agree!

.github/workflows/e2e-compatibility.yaml
chatton: maybe we could rename the value to be something explicit like `matrix-filename` WDYT?

damiannolan: I'm easy. I think its fine to leave as is

.github/workflows/e2e-compatibility-workflow-call.yaml
colin-axner: This is indeed magic :sweat_smile:

.github/workflows/e2e-compatibility.yaml
colin-axner: Whatever is fine by me!

.github/compatibility-test-matrices/release-v5.0.x/client.json
```diff
@@ -0,0 +1,10 @@
+{
+  "chain-a": ["release-v5.0.x", "v4.1.0", "v3.3.0", "v2.4.0"],
+  "chain-b": ["release-v5.0.x", "v4.1.0", "v3.3.0", "v2.4.0"],
```
colin-axner, line 3: so clean! :heart:

colin-axner (approved):
absolutely fabulous! :crown: 

These e2e test additions are like a dream come true!

colin-axner:
> Currently all the interchain accounts tests are running using the icad, so the state of the release branch in ibc-go should have no impact here right?

I guess the question is if we want to be able to trigger icad compatibility tests from ibc-go as well (long term no, but maybe in the short term?)

chatton:
@colin-axner 

> I guess the question is if we want to be able to trigger icad compatibility tests from ibc-go as well (long term no, but maybe in the short term?)

Right, we can already do this with the `Manual E2E (Icad)` test. But I think long term will be able to use simd and it can be added as a new file here.

.github/workflows/e2e-compatibility-workflow-call.yaml
chatton: the process of passing variables between steps is quite cumbersome unfortunately 😓

Issue #2268, opened 2022-09-10
modules/core/keeper: NewKeeper blank value detection assumes interface values are all non-pointer receivers and will fail to detect blank pointer receivers


odeke-em:
## Summary of Bug
The code inside https://github.com/cosmos/ibc-go/blob/a4be5614fb2fc92cfb430c82f18e7a9f67383c6e/modules/core/keeper/keeper.go#L52-L60 checks if values are Zero values using reflect.ValueOf. That'll work in cases where value receivers are passed in. However, for a general purpose package that receives either pointer or value receivers, this forces odd behavior in which only value receivers can be checked. For example if we examine https://go.dev/play/p/zfFZRDjqz-X or inlined below
```go

package main

import (
	"io"
	"reflect"
)

type dt struct {
	i int
	s string
}

func (dt) Write(b []byte) (int, error) { return len(b), nil }

func main() {
	println(reflect.ValueOf(dt{}).IsZero())
	println(reflect.ValueOf(&dt{}).IsZero())
	println(reflect.ValueOf((io.Writer)(dt{})).IsZero())
	println(reflect.ValueOf((io.Writer)(&dt{})).IsZero())
}
```

Which prints out

```shell
true
false
true
false
```

But really if zero value checking were implemented correctly by dereferencing pointers, every line should print true.

## Expected Behaviour
Catching blank pointer values

Just an FYI for my colleagues @elias-orijtech @kirbyquerby @willpoint

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Maybe I am missing something, but the `stakingKeeper` and `upgradeKeeper` parameters need to be passed by value ([as the signature of NewKeeper indicates](https://github.com/cosmos/ibc-go/blob/a4be5614fb2fc92cfb430c82f18e7a9f67383c6e/modules/core/keeper/keeper.go#L39-L43)), so if in the upstream caller function they were pointers, they would need to be dereferenced before being passed to `NewKeeper`, right?

odeke-em:
@crodriguezvega it expects values that conform to an interface. Anything that conforms to that interface can be passed in and they could either be pointers or non-pointers.

crodriguezvega:
> @crodriguezvega it expects values that conform to an interface. Anything that conforms to that interface can be passed in and they could either be pointers or non-pointers.

Ah, yes, I see now. Thanks!

[...]

PR #2403, opened 2022-09-27
added switch case to detect blank pointers in keeper

alizahidraja:
## Description

Added switch case to detect blank pointers in the keeper.

closes: #2268

---

write a little note why.

- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`) (not applicable)


modules/core/keeper/keeper.go
```diff
@@ -50,12 +50,28 @@ func NewKeeper(
...
+
+	// switch case to detect blank pointers
+	switch reflect.TypeOf(stakingKeeper).Kind() {
+	case reflect.Ptr:
+		if reflect.ValueOf(&stakingKeeper).IsZero() {
+			panic(fmt.Errorf("cannot initialize IBC keeper: empty staking keeper"))
+		}
+	default:
+		if reflect.ValueOf(stakingKeeper).IsZero() {
+			panic(fmt.Errorf("cannot initialize IBC keeper: empty staking keeper"))
```
crodriguezvega, line 62: Could we maybe have a slight different error message, so that in case the error occurs is immediately obvious what the problem was (i.e. either empty value or empty pointer)?
```suggestion
			panic(fmt.Errorf("cannot initialize IBC keeper: empty staking keeper value"))
```

modules/core/keeper/keeper.go
```diff
@@ -50,12 +50,28 @@ func NewKeeper(
...
 
 	// panic if any of the keepers passed in is empty
-	if reflect.ValueOf(stakingKeeper).IsZero() {
-		panic(fmt.Errorf("cannot initialize IBC keeper: empty staking keeper"))
+
+	// switch case to detect blank pointers
+	switch reflect.TypeOf(stakingKeeper).Kind() {
+	case reflect.Ptr:
+		if reflect.ValueOf(&stakingKeeper).IsZero() {
+			panic(fmt.Errorf("cannot initialize IBC keeper: empty staking keeper"))
```
crodriguezvega, line 58:
```suggestion
			panic(fmt.Errorf("cannot initialize IBC keeper: empty staking keeper pointer"))
```

crodriguezvega (commented):
Thanks for fixing this, @alizahidraja!

Could you also please try to add some tests for the empty pointer situations [here](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/keeper_test.go#L77)?

modules/core/keeper/keeper.go
```diff
@@ -50,12 +50,28 @@ func NewKeeper(
 	}
 
 	// panic if any of the keepers passed in is empty
-	if reflect.ValueOf(stakingKeeper).IsZero() {
```
damiannolan, line 53: I think [`ValueOf`](https://pkg.go.dev/reflect#ValueOf) returns the zero value for `nil`, i.e. blank pointers.
Perhaps we could add some test cases before adapting this code to use the switch case.

damiannolan (commented):
Thanks for your contributions @alizahidraja, much appreciated! 🙏 

I have some doubts over this code needing to change, but perhaps some additional tests could get to the bottom of it!

modules/core/keeper/keeper.go
alizahidraja: Hey @damiannolan! 

In the code from the [issue](https://github.com/cosmos/ibc-go/issues/2268) raised, the behavior of a pointer value was not as predicted with `ValueOf`

Will try to add additional test cases to confirm the need for switch case.

Another approach can be to add another `or` statement in the same line, but I thought switch case would have more readability plus a different error message for value & pointers.

alizahidraja:
@crodriguezvega @damiannolan I added 2 new test cases and ran them on both, the switch case code and previous code; the previous code fail the new test cases with pointers

Please do have a look and lmk your thoughts

crodriguezvega (approved):
Thanks, @alizahidraja!

modules/core/keeper/keeper.go
colin-axner: Could we modify the if statement to be:
```go
if stakingKeeper == nil || reflect.ValueOf(stakingKeeper).IsZero() {
```

I don't really want to over complicate this code. It is just a defensive check to prevent the chain from starting if the app.go is misconfigured. We simply want to know if the arguments provided are empty. I feel like there should be simple way of doing this? Otherwise we should add a private function to perform the switch so it just reads:

```go
assertNotEmpty(stakingKeeper)
assertNotEmpty(upgradeKeeper)
```

alizahidraja: I understand the concern but running the following code

```fmt.Println(stakingKeeper == nil)```

when given a pointer, returns `false`

I tried to solve it in a simpler way but failed. I can make a separate function in the same file that would check if an interface object(struct or pointer) is empty if you think that approach is better here

Comment (colin-axner) [reply modules/core/keeper/keeper.go:53]: I see the nil check won't catch the &stakingtypes.StakingKeeper{} issue. 

I think my preference would be to move the switch to a private assertNotEmpty function which returns an error which we panic on wrapping the name of the keeper:
```go
if isEmpty(stakingKeeper) {
    panic(fmt.Errorf("cannot initialize IBC keeper: empty staking keeper"))
}
```

colin-axner (commented):
Thanks for picking this up @alizahidraja. It's a bit of a tricky issue so I appreciate your patience. See my [latest comment](https://github.com/cosmos/ibc-go/pull/2403/files#r983395304). I think it makes sense to add this to the bottom of the keeper.go file

alizahidraja:
No worries @colin-axner, I'll make a function to check for that at the bottom

modules/core/keeper/keeper.go
alizahidraja: As @crodriguezvega mentioned, it's good to know if the error was due to value or pointer, if we make a function as you suggest, we won't be able to know the difference

Is that okay or do you suggest we think of something else?

alizahidraja:
Added an isEmpty function at the bottom of the keeper

The final concern would be that the error message cannot specify if the keeper object is a struct or a pointer, dk if that is important, lmk what you guys think

colin-axner (approved):
Thanks @alizahidraja!

colin-axner:
> The final concern would be that the error message cannot specify if the keeper object is a struct or a pointer, dk if that is important, lmk what you guys think

No, I think as you have it now is great. We mostly just want to panic if something is misconfigured so the chain developer knows to look into the code

PR #2419, opened 2022-09-28
Allow chains that can't introspect their own `ConsensusState` support IBC

seunlanlege:
ref https://github.com/cosmos/ibc/pull/839


proto/ibc/core/connection/v1/tx.proto
```diff
@@ -67,6 +67,8 @@ message MsgConnectionOpenTry {
   ibc.core.client.v1.Height consensus_height = 11
       [(gogoproto.moretags) = "yaml:\"consensus_height\"", (gogoproto.nullable) = false];
   string signer = 12;
+  // extra data for state machaines that can't introspect their own consensus state.
```
crodriguezvega, line 70:
```suggestion
  // extra data for state machines that can't introspect their own consensus state.
```

proto/ibc/core/connection/v1/tx.proto
```diff
@@ -67,6 +67,8 @@ message MsgConnectionOpenTry {
...
+  bytes host_consensus_state_info = 13;
```
crodriguezvega, line 71: To match the [spec changes that were merged](https://github.com/cosmos/ibc/pull/839)...

```suggestion
  bytes host_consensus_state_proof = 13;
```

proto/ibc/core/connection/v1/tx.proto
```diff
@@ -94,6 +96,8 @@ message MsgConnectionOpenAck {
   ibc.core.client.v1.Height consensus_height = 9
       [(gogoproto.moretags) = "yaml:\"consensus_height\"", (gogoproto.nullable) = false];
   string signer = 10;
+    // extra data for state machaines that can't introspect their own consensus state.
+    bytes host_consensus_state_info = 11;
```
crodriguezvega, line 100:
```suggestion
    bytes host_consensus_state_proof = 11;
```

AdityaSripal (approved):
Thanks @seunlanlege

en:
Hi @seunlanlege and @AdityaSripal,
Why don't we define this field as optional? I think the cosmos-sdk based chain should not need this field.

PR #2502, opened 2022-10-10

colin-axner:
## Description

closes: #XXXX

---

write a little note why.


e2e/testsuite/semver_test.go
```diff
@@ -0,0 +1,53 @@
...
+		{"non semantic version", "1.minor.1", true},
+		{"supported semantic version", "v2.5.0", true},
+		{"supported semantic version", "v3.4.0", true},
+		{"supported semantic version", "v4.2.0", true},
+		{"supported semantic version", "v5.1.0", true},
+		{"supported semantic version", "v6.0.0", true},
+		{"supported semantic version", "v6.1.0", true},
+		{"supported semantic version", "v7.1.0", true},
+		{"supported semantic version", "v22.5.1", true},
+		{"supported semantic version with v", "2.5.0", true},
```
crodriguezvega, line 40:
```suggestion
		{"supported semantic version without v", "2.5.0", true},
```

e2e/testsuite/semver.go
```diff
@@ -0,0 +1,75 @@
...
+
+// NewVersion constructs a new version given the major and minor version.
+func NewVersion(major, minor uint64) Version {
+	return Version{
+		Major: major,
+		Minor: minor,
+	}
+}
+
+func ParseChainVersion(version string) (Version, bool) {
```
crodriguezvega, line 22: Is this function is only used in `IsSupported`, can it be unexported? If it cannot, can we add a godoc then?

e2e/testsuite/semver.go
```diff
@@ -0,0 +1,75 @@
...
+	MajorVersion  uint64
+	MinorVersions []Version
+}
+
+// IsSupported returns whether the version contains the feature.
+// This is true if the version is greater than or equal to the major version it was released in
+// or is greater than or equal to the list of minor releases it was included in.
+func (fr FeatureReleases) IsSupported(versionStr string) bool {
+	// assume any non semantic version formatted version support the feature
+	// this will be useful during development of the e2e test with the new feature
```
crodriguezvega, line 58: So this is to support images with names of working branches, right?

e2e/testsuite/semver.go
```diff
@@ -0,0 +1,75 @@
...
+
+	minor, err := strconv.ParseUint(versions[1], 10, 64)
+	if err != nil {
+		return Version{}, false
+	}
+
+	return Version{
+		Major: major,
+		Minor: minor,
+	}, true
```
chatton, line 44: The majority of the parsing logic here seems to be duplicating  `semver.Major` and `semver.Minor`

e2e/testsuite/semver.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if version.Major >= fr.MajorVersion {
+		return true
+	}
+
+	for _, mv := range fr.MinorVersions {
+		if mv.Major == version.Major {
+			return version.Minor >= mv.Minor
+		}
+	}
```
chatton, line 72: as the input to this funciton is the version string, we can use `semver.Compare` to handle the version comparisons.

chatton (commented):
The test cases look great ❤️ 

I think a lot of the custom logic that we have here can be replaced with the `semver` package. Especially since our inputs are strings.

e2e/testsuite/semver.go
colin-axner: ah! I was looking for that, wasn't sure it existed, looks to be importable under "semver"?

chatton:
After the refactor, the custom types to represent versions were removed in favour of strings which are natively supported by the `semver` package. A few small functions were added to aid readability. All test cases remain the same.

colin-axner (commented):
ACK, thanks @chatton! Feel free to merge for me whenever you feel ready

e2e/testsuite/semver.go
crodriguezvega: All right... maybe this is an edge case, but if we have a PR for a branch that does not support the feature (for example, a backport to `release/v2.4.x`) then the test will fail, right? Because this will wrongly assume that the PR supports the feature... Although maybe this is not a problem if the metadata tests are only run as part of the compatibility check.

charleenfei (commented):
nice work @colin-axner @chatton !

colin-axner (commented):
Approve! (can't click approve button on github, feel free to do it on my behalf)

chatton (approved):
(on behalf of @colin-axner  ❤️ )

Issue #2020, opened 2022-08-16
change type of `scopedKeeper` in all `Keeper`s to interface


crodriguezvega:
The `scopedKeeper` in all `Keeper`s in ibc-go has at the moment a concrete type (i.e. `capabilitykeeper.ScopedKeeper`), but there is no good reason why we should not use an interface instead. Other keepers in the `Keeper` types are already using interfaces.

### Proposal

1. Add `ScopedKeeper` interface.

For 05-port create an `expected_keepers.go` file in package `types` and add the following interface:
 
```go
type ScopedKeeper interface {
  NewCapability(ctx sdk.Context, name string) (*capabilitytypes.Capability, error)
  GetCapability(ctx sdk.Context, name string) (*capabilitytypes.Capability, bool)
  AuthenticateCapability(ctx sdk.Context, cap *capabilitytypes.Capability, name string) bool
  LookupModules(ctx sdk.Context, name string) ([]string, *capabilitytypes.Capability, error)
}
```

For `04-channel` add the following interface to [`expected_keepers.go`](https://github.com/cosmos/ibc-go/blob/v4.0.0/modules/core/04-channel/types/expected_keepers.go):

```go
type ScopedKeeper interface {
  NewCapability(ctx sdk.Context, name string) (*capabilitytypes.Capability, error)
  AuthenticateCapability(ctx sdk.Context, cap *capabilitytypes.Capability, name string) bool
  LookupModules(ctx sdk.Context, name string) ([]string, *capabilitytypes.Capability, error)
}
```

For `transfer` add the following interface to [`expected_keepers.go`](https://github.com/cosmos/ibc-go/blob/v4.0.0/modules/apps/transfer/types/expected_keepers.go):

```go
type ScopedKeeper interface {
  GetCapability(ctx sdk.Context, name string) (*capabilitytypes.Capability, bool)
  AuthenticateCapability(ctx sdk.Context, cap *capabilitytypes.Capability, name string) bool
  LookupModules(ctx sdk.Context, name string) ([]string, *capabilitytypes.Capability, error)
}
```

For `27-interchain-accounts` add the following interface to [`expected_keepers.go`](https://github.com/cosmos/ibc-go/blob/v4.0.0/modules/apps/27-interchain-accounts/types/expected_keepers.go):

```go
type ScopedKeeper interface {
  GetCapability(ctx sdk.Context, name string) (*capabilitytypes.Capability, bool)
[...]

PR #2638, opened 2022-10-30
Add scopedkeeper as an interface to exported package

Anmol1696:
## Description

Move all expected scoped keepers to a unified interface in `core/exported` package.

closes: #2020 

---

write a little note why.


colin-axner (approved):
LGTM thanks @Anmol1696!

modules/core/exported/expected_keepers.go
```diff
@@ -1,4 +1,4 @@
-package types
+package exported
```
damiannolan, line 1: nit:

@colin-axner wdyt about putting this into `exported/channel.go` instead of a new `expected_keepers.go` file?

damiannolan (approved):
Thank you @Anmol1696, much appreciated! 🚀

modules/core/exported/expected_keepers.go
colin-axner: I have no preference. That makes sense to me since the scoped keeper is for channel/port capabilities

Anmol1696: My concern with moving this to `channel.go` would be, `CapabilityKeeper` which is `ScopedKeeper` inferface has nothing to do with `channel/port`, and it is the other way around, channel/port uses this interface.

Again since this will be anyways be exposed as the `exported` package, i will be ok moving it, reducing the files.

PR #2640, opened 2022-10-31
e2e: add transfer with memo compatibility tests

crodriguezvega:
## Description

~~I am not sure if the tests will run for version < v6.0.0 since we use [the new API for `NewMsgTransfer` in the e2e code](https://github.com/cosmos/ibc-go/blob/main/e2e/testsuite/testsuite.go#L309)?~~

closes: #XXXX

---

write a little note why.


.github/compatibility-test-matrices/release-v3.4.x/transfer.json
```diff
@@ -0,0 +1,15 @@
+{
+  "chain-a": ["release-v3.4.x", "v3.3.1", "release-v2.5.x", "v2.4.2"],
```
crodriguezvega, line 2: I am adding to tips of release branches here, so that we can test that transfer with memo works between `v3.4.x` and `v2.5.x`, but I am unsure if this is supported by the e2e framework, so let's give it a try and see what happens...

.github/compatibility-test-matrices/release-v2.5.x/transfer.json
```diff
@@ -0,0 +1,15 @@
...
+  "chain-b": ["release-v2.5.x", "v2.4.2"],
+  "entrypoint": ["TestTransferTestSuite"],
+  "test": [
+    "TestMsgTransfer_Succeeds_Nonincentivized",
+    "TestMsgTransfer_Fails_InvalidAddress",
+    "TestMsgTransfer_Timeout_Nonincentivized",
+    "TestMsgTransfer_WithMemo",
+    "TestSendEnabledParam",
+    "TestReceiveEnabledParam"
+  ],
```
colin-axner, line 12: It might be nice if in the future we can just have test exclusions, I forget why it was difficult to list all possible tests

.github/compatibility-test-matrices/release-v2.5.x/transfer.json
```diff
@@ -0,0 +1,15 @@
+{
+  "chain-a": ["release-v2.5.x", "v2.4.2"],
+  "chain-b": ["release-v2.5.x", "v2.4.2"],
```
colin-axner, line 3: All compatibility tests should test against all supported versions. If we were only doing a v2.5 release (and not v3.4, 4.2 etc), this would not test against the other releases (and we likely would run compatibility tests for the other releases)

colin-axner (commented):
LGTM! I think each `transfer.json` should use the same array of all versions though

.github/compatibility-test-matrices/release-v3.4.x/transfer.json
```diff
@@ -0,0 +1,15 @@
+{
+  "chain-a": ["release-v3.4.x", "release-v6.0.x", "release-v5.1.x", "v5.0.1", "release-v4.2.x", "v4.1.1", "v3.3.1", "release-v2.5.x", "v2.4.2"],
+  "chain-b": ["release-v3.4.x", "release-v6.0.x", "release-v5.1.x", "v5.0.1", "release-v4.2.x", "v4.1.1", "v3.3.1", "release-v2.5.x", "v2.4.2"],
```
chatton, line 3: in our v3.4.x I do we want to explicitly test again the tip of other release branches? I think it is fine to test only against all other released versions. The compatibility tests for those other branches should handle those cases.

.github/compatibility-test-matrices/release-v2.5.x/transfer.json
```diff
@@ -0,0 +1,15 @@
+{
+  "chain-a": ["release-v2.5.x", "release-v6.0.x", "release-v5.1.x", "v5.0.1", "release-v4.2.x", "v4.1.1", "release-v3.4.x", "v3.3.1", "v2.4.2"],
+  "chain-b": ["release-v2.5.x", "release-v6.0.x", "release-v5.1.x", "v5.0.1", "release-v4.2.x", "v4.1.1", "release-v3.4.x", "v3.3.1", "v2.4.2"],
```
chatton, line 3: same comment about not needing to test against all other release branches (just released versions)

.github/compatibility-test-matrices/release-v3.4.x/transfer.json
chatton: ah I left my below comments before seeing this. This one might be tricky as the first part of the compatibility test process is building the the `release-x.y.x` image. In this case some of those images won't exist. As a work around you could launch those compatibility tests for the images that don't exist manually then cancel them after the image is built.

colin-axner (approved):
LGTM, but it still seems a little off to me. 

My expected behaviour, running compatibility tests for release branch X, all relevant tests for that release branch are run against every supported version (and to be released version, ie v2.5 should test against v3.4)

I guess the ordering of our releases right now are a little odd

chatton:
> all relevant tests for that release branch are run against every supported version (and to be released version, ie v2.5 should test against v3.4)

@colin-axner do you mean you think we should run the tip of each release branch against one another e.g. `release-v2.5.x` against `release-v3.4.x`? Or do you think it is enough to run the release branch of the release in question against all already released versions?

PR #2709, opened 2022-11-08
docs: refactor contributing guidelines - add go style guide

crodriguezvega:
## Description

Based off [Tendermint's](https://github.com/tendermint/tendermint/blob/f58ba4d2f9c71071554dfb0273252edf90aa85e5/STYLE_GUIDE.md) 

closes: #XXXX

---

write a little note why.


docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+- Functions that return functions should have the suffix `Fn`.
+- Names should not [stutter](https://blog.golang.org/package-names). For example, a struct generally shouldn’t have a field named after itself; e.g., this shouldn't occur:
+
+``` golang
+type middleware struct {
+	middleware Middleware
+}
+```
+
+* In comments, use "iff" to mean, "if and only if".
```
chatton, line 49: I think this kind of contradicts the suggestion in the `Comments` section about ensuring all comments are sentences. I personally am not familiar with using `iff` and would actually interpret this as a typo.

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+* Acronyms are all capitalized, like "RPC", "gRPC", "API".  "MyID", rather than "MyId".
+* Prefer errors.New() instead of fmt.Errorf() unless you're actually using the format feature with arguments.
```
chatton, line 51: I agree with this, however this is not always safe for us to do, in a large number of cases we need to use sdk errors. However the point is valid for cases when you would use `fmt.Errorf`. Maybe it's worth adding a small note about ensuring it is safe to use a regular go error as opposed to an sdk error? If people think that is excessive I'm happy to leave this as is 👍

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+
+* Use `gofumpt` to format all code upon saving it (or run `make format`).
+* Think about documentation, and try to leave godoc comments, when it will help new developers.
+* Every package should have a high level doc.go file to describe the purpose of that package, its main functions, and any other relevant information.
+* Applications (e.g. clis/servers) *should* panic on unexpected unrecoverable errors and print a stack trace.
+
+## Comments
+
+* Use a space after the comment deliminter (ex. `// your comment`).
+* Many comments are not sentences. These should begin with a lower case letter and end without a period.
```
chatton, line 31: I wonder if we need to include things like this that can easily automated with a linter?

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+- Options (only if they are seen as critical to the struct else they should be placed in another file)
+- Initialization / Start and stop of the service
+- Msgs/Events
+- Public Functions (In order of most important)
+- Private/helper functions
+- Auxiliary structs and function (can also be above private functions or in a separate file)
+
+## General
+
+* Use `gofumpt` to format all code upon saving it (or run `make format`).
```
chatton, line 23: I think we don't need to change anything now, but I think we should establish some sort of pre-commit hook that will run all this stuff. Relying on people to run `make format` manually each time isn't ideal.

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
+
```
chatton, line 1: should we link this somewhere directly on the main README? It might be nice to highlight this somewhere with more visibility. Perhaps even a PR template?

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+- Make use of table driven testing where possible and not-cumbersome. Read [this blog post](https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) for more information.
+- Make use of [assert](https://godoc.org/github.com/stretchr/testify/assert) and [require](https://godoc.org/github.com/stretchr/testify/require)
+- When using mocks, it is recommended to use Testify [mock](https://pkg.go.dev/github.com/stretchr/testify/mock) along with [Mockery](https://github.com/vektra/mockery) for autogeneration.
+
+## Errors
+
+- Ensure that errors are concise, clear and traceable.
+- Use stdlib errors package.
+- For wrapping errors, use `fmt.Errorf()` with `%w`.
+- Panic is appropriate when an internal invariant of a system is broken, while all other cases (in particular, incorrect or invalid usage) should return errors.
```
chatton, line 74: I agree with this in general, but I feel like it is a special case with the SDK where returning errors can sometimes reset state changes and panics are caught and as a result will not result in the shutdown of the system.

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+## Testing
+
+- Make use of table driven testing where possible and not-cumbersome. Read [this blog post](https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) for more information.
+- Make use of [assert](https://godoc.org/github.com/stretchr/testify/assert) and [require](https://godoc.org/github.com/stretchr/testify/require)
+- When using mocks, it is recommended to use Testify [mock](https://pkg.go.dev/github.com/stretchr/testify/mock) along with [Mockery](https://github.com/vektra/mockery) for autogeneration.
+
+## Errors
+
+- Ensure that errors are concise, clear and traceable.
+- Use stdlib errors package.
```
chatton, line 72: same point as above, in most cases we will use sdkerrors

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Testing
+
+- Make use of table driven testing where possible and not-cumbersome. Read [this blog post](https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) for more information.
+- Make use of [assert](https://godoc.org/github.com/stretchr/testify/assert) and [require](https://godoc.org/github.com/stretchr/testify/require)
+- When using mocks, it is recommended to use Testify [mock](https://pkg.go.dev/github.com/stretchr/testify/mock) along with [Mockery](https://github.com/vektra/mockery) for autogeneration.
+
+## Errors
+
+- Ensure that errors are concise, clear and traceable.
```
chatton, line 71: ++++

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+
+We expect all contributors to be familiar with [Effective Go](https://golang.org/doc/effective_go.html) (and it's recommended reading for all Go programmers anyways). Additionally, we generally agree with the suggestions in [Uber's style guide](https://github.com/uber-go/guide/blob/master/style.md) and use that as a starting point.
+
+## Code Structure
+
+Perhaps more key for code readability than good commenting is having the right structure. As a rule of thumb, try to write in a logical order of importance, taking a little time to think how to order and divide the code such that someone could scroll down and understand the functionality of it just as well as you do. A loose example of such order would be:
+
+- Constants, global and package-level variables
+- Main Struct
```
chatton, line 13: what does this mean exactly?

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+- Auxiliary structs and function (can also be above private functions or in a separate file)
+
+## General
+
+* Use `gofumpt` to format all code upon saving it (or run `make format`).
+* Think about documentation, and try to leave godoc comments, when it will help new developers.
+* Every package should have a high level doc.go file to describe the purpose of that package, its main functions, and any other relevant information.
+* Applications (e.g. clis/servers) *should* panic on unexpected unrecoverable errors and print a stack trace.
+
+## Comments
```
chatton, line 28: I would like to make an addition here.

In most cases comments should explain `why` something is being done rather than `what` the code is doing. I.e


Maybe a contrived example but the comment

```
// assign a variable foo
f := foo
// assign f to b
b := f
```
has little value.

but the following is more useful
```
f := foo
// we copy the variable f because we want to preserve the state at time of initialization
b := f

**Review by chatton — COMMENTED** [review]
Thanks for adding this! I have a lot of small questions and suggestions!

File: docs/dev/go-style-guide.md [hunk 11]
```diff
@@ -0,0 +1,74 @@
...
+Perhaps more key for code readability than good commenting is having the right structure. As a rule of thumb, try to write in a logical order of importance, taking a little time to think how to order and divide the code such that someone could scroll down and understand the functionality of it just as well as you do. A loose example of such order would be:
+
+- Constants, global and package-level variables
+- Main Struct
+- Options (only if they are seen as critical to the struct else they should be placed in another file)
+- Initialization / Start and stop of the service
+- Msgs/Events
+- Public Functions (In order of most important)
+- Private/helper functions
+- Auxiliary structs and function (can also be above private functions or in a separate file)
```
colin-axner, line 19: I'm not sure this section is very reflective of our existing code. Msgs/Events are split into separate files so I don't understand how there is an ordering here?

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+
+## General
+
+* Use `gofumpt` to format all code upon saving it (or run `make format`).
+* Think about documentation, and try to leave godoc comments, when it will help new developers.
+* Every package should have a high level doc.go file to describe the purpose of that package, its main functions, and any other relevant information.
```
colin-axner, line 25: We don't follow this practice at the moment..

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+* Many comments are not sentences. These should begin with a lower case letter and end without a period.
+* Conversely, sentences in comments should be sentenced-cased and end with a period.
+
+## Linting
+
+* Run `make lint-fix` to fix any linting errors.
+
+## Various
+
+- Functions that return functions should have the suffix `Fn`.
```
colin-axner, line 40: I'm not sure we apply this

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+* In comments, use "iff" to mean, "if and only if".
+* Acronyms are all capitalized, like "RPC", "gRPC", "API".  "MyID", rather than "MyId".
+* Prefer errors.New() instead of fmt.Errorf() unless you're actually using the format feature with arguments.
+
+## Importing libraries
+
+* Use [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports).
+* Separate imports into blocks - one for the standard lib, one for external libs and one for application libs.
```
colin-axner, line 56: We should provide an example here

docs/dev/go-style-guide.md
crodriguezvega: Will add a link to it in the PR template.

docs/dev/go-style-guide.md
crodriguezvega: We do in some places, but I can open a separate PR to update all occurrences.

docs/dev/go-style-guide.md
crodriguezvega: I would say a `struct` definition...

docs/dev/go-style-guide.md
crodriguezvega: Done.

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
...
+## Dependencies
+
+* Dependencies should be pinned by a release tag, or specific commit, to avoid breaking `go get` when external dependencies are updated.
+* Refer to the [contributing](./development.md#dependencies) document for more details
+
+## Testing
+
+- Make use of table driven testing where possible and not-cumbersome. Read [this blog post](https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) for more information.
+- Make use of [assert](https://godoc.org/github.com/stretchr/testify/assert) and [require](https://godoc.org/github.com/stretchr/testify/require)
+- When using mocks, it is recommended to use Testify [mock](https://pkg.go.dev/github.com/stretchr/testify/mock) along with [Mockery](https://github.com/vektra/mockery) for autogeneration.
```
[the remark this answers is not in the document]
crodriguezvega: We can if we ever need to use mocks!

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,116 @@
...
+- Initialization/start and stop of the service functions.
+- Public functions (in order of most important).
+- Private/helper functions.
+- Auxiliary structs and function (can also be above private functions or in a separate file).
+
+## General
+
+- Use `gofumpt` to format all code upon saving it (or run `make format`).
+- Think about documentation, and try to leave godoc comments, when it will help new developers.
+- Every package should have a high level doc.go file to describe the purpose of that package, its main functions, and any other relevant information.
```
damiannolan, line 24: While we don't have this at the moment I think it would be nice for apps, lightclients and core submodules to have a `doc.go` with a brief description.

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,74 @@
+
+# Go style guide
+
+In order to keep our code looking good with lots of programmers working on it, it helps to have a "style guide", so all the code generally looks quite similar. This doesn't mean there is only one "right way" to write code, or even that this standard is better than your style.  But if we agree to a number of stylistic practices, it makes it much easier to read and modify new code. Please feel free to make suggestions if there's something you would like to add or modify.
+
+We expect all contributors to be familiar with [Effective Go](https://golang.org/doc/effective_go.html) (and it's recommended reading for all Go programmers anyways). Additionally, we generally agree with the suggestions in [Uber's style guide](https://github.com/uber-go/guide/blob/master/style.md) and use that as a starting point.
```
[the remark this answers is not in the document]
damiannolan: I've had a scroll threw it once or twice before, but not read in-depth. 

In summary, it more or less just had pretty good, reasonable and sane suggestions wrt code style and things that are considered more "idiomatic go".

docs/dev/go-style-guide.md
```diff
@@ -0,0 +1,116 @@
...
+- Error messages should be formatted as following:
+
+	```golang
+	sdkerrors.Wrapf(
+		<most specific error type possible>,
+		"<optional text description ended by colon and space>expected %s, got %s",
+		<value 1>,
+		<value 2>
+	)
+	```
```
damiannolan, line 116: Seems like the formatting of this is a little off? Sentence starts with "ke", should this be "Take"?

damiannolan (approved):
I'm happy enough to merge this and we can revisit and tweak it as needs be. 
Thanks for all of these nice new docs @crodriguezvega! 🚀

docs/dev/go-style-guide.md
crodriguezvega: Let's do that in a follow-up PR!

docs/dev/go-style-guide.md
crodriguezvega: It's indented to the right a bit so that aligns with the text in the bullet point that it belongs to.

Issue #1884, opened 2022-08-04
Use commitmenttypes.MerklePath in VerifyMembership/VerifyNonMembership arguments


colin-axner:
## Summary

In `VerifyMembership` and `VerifyNonMembership` we should use the merkle path interface instead of bytes to reduce unnecessary marshal/unmarshling. Use casting, open issue for removing path interface

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
I think we are using `[]byte` as this would result in an import cycle, right? 

We register `exported` interfaces in [`23-commitment/types/codec.go`](https://github.com/cosmos/ibc-go/blob/main/modules/core/23-commitment/types/codec.go) against the concrete types, so using `MerklePath` would result in an import cycle.

PR #2736, opened 2022-11-14
chore: updating `VerifyMembership` and `VerifyNonMembership` methods to use `Path` interface

damiannolan:
## Description

- Updating `VerifyMembership` and `VerifyNonMembership` client state methods to use `Path` interface in favour of `[]byte`. This removes the need for marshalling and unmarshalling the key path in `03-connection` and lightclient modules respectively.

closes: #1884 

---

write a little note why.

### DIFF
--- a/modules/light-clients/06-solomachine/client_state.go
+++ b/modules/light-clients/06-solomachine/client_state.go
@@ -112,7 +112,7 @@ func (cs *ClientState) VerifyMembership(
 	delayTimePeriod uint64,
 	delayBlockPeriod uint64,
 	proof []byte,
-	path []byte,
+	path exported.Path,
 	value []byte,
 ) error {
 	publicKey, sigData, timestamp, sequence, err := produceVerificationArgs(cdc, cs, height, proof)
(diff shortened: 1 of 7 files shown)


modules/light-clients/06-solomachine/client_state.go
```diff
@@ -112,17 +112,17 @@ func (cs *ClientState) VerifyMembership(
...
 		return err
 	}
 
-	var merklePath commitmenttypes.MerklePath
-	if err := cdc.Unmarshal(path, &merklePath); err != nil {
-		return sdkerrors.Wrap(commitmenttypes.ErrInvalidProof, "failed to unmarshal path into ICS 23 commitment merkle path")
+	merklePath, ok := path.(commitmenttypes.MerklePath)
+	if !ok {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidType, "expected %T, got %T", commitmenttypes.MerklePath{}, path)
 	}
```
damiannolan, line 126: Code coverage both here and in `07-tendermint` will likely be reduced. I can cover the error case here by adding a mock path type to the tests like so:

```go
type mockPath struct{}

func (mockPath) String() string {
	return ""
}

func (mockPath) Empty() bool {
	return false
}
```

damiannolan: fwiw I don't feel super strongly about covering this case...

damiannolan:
I think we are missing a changelog for the addition of the generic membership verification funcs. 
Should we add one now or later?

AdityaSripal (approved):
Nice cleanup!

modules/light-clients/06-solomachine/client_state.go
colin-axner: No preferences, probably doesn't hurt to add. Hopefully one day it can be removed. I'm fine if it is skipped as well

colin-axner:
> I think we are missing a changelog for the addition of the generic membership verification funcs.

Either is fine by me

Issue #2009, opened 2022-08-15
Create cannonical place to register light client interfaces


colin-axner:
## Summary

Currently we register light client interfaces in [RegisterInterfaces](https://github.com/cosmos/ibc-go/blob/main/modules/core/module.go#L94) within the core IBC module set as an sdk module. We should expose some API to register interfaces without adding code to the core IBC module. My first inclination is to add this function to the app module basic, so app.go files will be required to provide a function registering all their desired light clients. An alternative is to use the go `init()` function

In the short term, chains can simply add a call to the RegisterInterfaces for external light clients within the construction of the App in `app.go`

#### Decision

Add a field to `AppModuleBasic` to register interfaces, the list of functions can be taken in via `NewAppModuleBasic`. In `AppModuleBasic.RegisterInterfaces` should iterate over the functions provided

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
After discussion with @damiannolan we will go with the app module basic approach. IBC clients as SDK modules allows for additional benefits such as having external information (metadata) being controlled outside of 02-client (migrations, genesis). The benefits require SDK intermodule routing, but this is a step in the long term direction that will eventually be needed. I will write an ADR outlining the problems/eventual solution and why the `AppModuleBasic` approach is aligned in that direction

colin-axner:
Closed! AppModuleBasics have been added for tendermint and solo machine. #2008 will add documentation to show external devs how to create an app module basic

PR #2739, opened 2022-11-14
Create canonical place to register light client interfaces

chatton:
## Description

closes: #2009

---

write a little note why.


modules/core/types/codec.go
```diff
@@ -7,16 +7,12 @@ import (
 	connectiontypes "github.com/cosmos/ibc-go/v6/modules/core/03-connection/types"
 	channeltypes "github.com/cosmos/ibc-go/v6/modules/core/04-channel/types"
 	commitmenttypes "github.com/cosmos/ibc-go/v6/modules/core/23-commitment/types"
-	solomachinetypes "github.com/cosmos/ibc-go/v6/modules/light-clients/06-solomachine"
-	ibctm "github.com/cosmos/ibc-go/v6/modules/light-clients/07-tendermint"
 )
```
damiannolan, line 11: What do you think about adding an alias to the func signature for context? 
Not sure about naming, if it makes sense etc...

```suggestion

// ClientRegistrar handles registration of light-client interface types
type ClientRegistrar func(registry codectypes.InterfaceRegistry)

```

chatton: I was thinking of doing something like this, technically this functionality isn't limited in scope to registration functions for clients, it would let you register any interface types. That's why I used `additionalRegistration` rather than `clientRegistration`.  We could do something like `type InterfaceRegistrar func(registry codectypes.InterfaceRegistry)`?

modules/core/module.go
```diff
@@ -35,7 +35,17 @@ var (
...
 
 // AppModuleBasic defines the basic application module used by the ibc module.
-type AppModuleBasic struct{}
+type AppModuleBasic struct {
+	additionalInterfaceRegistrationFns []func(registry codectypes.InterfaceRegistry)
+}
+
+// NewAppModuleBasic creates a new AppModuleBasic object.
+// all functions provided will be executed in AppModuleBasic.RegisterInterfaces.
+func NewAppModuleBasic(additionalInterfaceRegistrationFns []func(codectypes.InterfaceRegistry)) AppModuleBasic {
```
colin-axner, line 44: Does it make sense to use the `...` here?
```suggestion
func NewAppModuleBasic(additionalInterfaceRegistrationFns func(codectypes.InterfaceRegistry)...) AppModuleBasic {
```

modules/core/types/codec.go
damiannolan: Aye, that's kind of what I was getting at with using an alias for the func signature here. I know you could technically pass anything, but it would be nice to infer that the intension is to provide register interfaces funcs for client modules.

I like `type InterfaceRegistrar func(registry codectypes.InterfaceRegistry)`, but we could then name the variable something more client specific.

chatton: I'm happy with your original suggestion of `ClientRegistar` @damiannolan , I've implemented that change, we can always rename if we like a different name better!

chatton:
Closing this PR as this is not the approach we want to take. We will instead explore the use of `AppModulBasic` as the mechanism for client registration.

Issue #2069, opened 2022-08-22
Add Params gRPC for 03-connection


colin-axner:
## Summary

Same reasoning as #2066. Should be added for 03-connection params

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
For reference: https://github.com/cosmos/ibc-go/issues/1628#issuecomment-1253528327

crodriguezvega:
The implementation for transfer can serve as reference. See this [comment](https://github.com/cosmos/ibc-go/issues/2066#issuecomment-1272935612) for details.

damiannolan:
I think we need to go with some alternative naming. e.g. rename the `Params` grpc in 02-client to `ClientParams` and use `ConnectionParams` here.

Otherwise we cause `duplicate method: Params` compiler errors due to: https://github.com/cosmos/ibc-go/blob/main/modules/core/types/query.go#L15

What do you think?

PR #2697, opened 2022-11-07
Add git cliff toml to support generation of changelog

chatton:
## Description

closes: #2106

Going forward, we can adhere to the conventional commit structure and ensure that every commit that gets merged to main is a descriptive changelog entry. 

Currently this only generates a changelog of unreleased changes. If we want to maintain a static changelog file we can decide how we want to do that and implement in a follow up PR. 

Update about release branches:

1. In order to generate the changelog for e.g. 5.1.0, we need to checkout `release/v5.1.x` and then run

```bash
make changelog tag=v5.1.0
```
(once this PR is backported)

With this example in particular, the output will also include changelog entries for 5.0.0 and 5.0.0-rc2. I suspect this has something to do with out v5.1.x branched from v5.0.x and are considered `unreleased` in some way. These are grouped into a separate category and can be safely ignored.

2. All commits we wish to include **must** be conventional. Ie. `chore: something i did`, `fix: fixed a bug`. If this format is not met, they will not be included in the changelog.

3. We should ensure that all PRs include the proposed commit message (changelog entry) for approval in the PR. As there would not necessarily be a file that is reviewable in the Github UI.

E.g. 

`chore: integrating git cliff into the code base to automate generation of changelogs` 

**We can start using this with the next set of releases** as the current commit history will not generate a clean changelog.

The following was generated by running

```bash
make changelog tag=v6.0.0
```

# Changelog

All notable changes to this project will be documented in this file.

### Dependencies

* Bump cosmossdk.io/math from 1.0.0-beta.2 to 1.0.0-beta.3 ([#1842](https://github.com/cosmos/ibc-go/issues/1842))
* Bump docker/build-push-action from 3.1.0 to 3.1.1 ([#1910](https://github.com/cosmos/ibc-go/issues/1910))
* Bump google.golang.org/grpc from 1.48.0 to 1.49.0 ([#2107](https://github.com/cosmos/ibc-go/issues/2107))
* Bump github.com/spf13/viper from 1.12.0 to 1.13.0 ([#2199](https://github.com/cosmos/ibc-go/issues/2199))
* Bump github.com/armon/go-metrics from 0.4.0 to 0.4.1 ([#2227](https://github.com/cosmos/ibc-go/issues/2227))
* Bump github.com/cosmos/cosmos-sdk from 0.46.1 to 0.46.2 ([#2458](https://github.com/cosmos/ibc-go/issues/2458))
[...]

PR #2745, opened 2022-11-14
feat: adding connection params grpc and cli query

damiannolan:
## Description

- Rename query `Params` to `ClientParams` to avoid naming collisions
- Adding `ConnectionParams` grpc and cli query
- Adding missing top level wiring for client query `UpgradedConsensusState`

closes: #2069 

### Commit / Changelog

```
imp: (core/03-connection) Adding `ConnectionParams` grpc query and CLI to 03-connection.
fix: (modules/core/keeper) Fix request wiring for `UpgradedConsensusState` in core query server.
```

---

write a little note why.


damiannolan:
Note this undos the changes from #2573 in order to avoid naming collisions. We can keep consistent naming of `Params` but this requires the removal of the [this code](https://github.com/cosmos/ibc-go/blob/main/modules/core/types/query.go#L14-L26) and for the query servers to be bound to their respective keepers [here](https://github.com/cosmos/ibc-go/blob/main/modules/core/module.go#L136-L140). 

This would also result in the removal of [modules/core/keeper/grpc_query.go](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/grpc_query.go)

Thoughts?

crodriguezvega (commented):
Thank you, @damiannolan. 

The renaming of `ClientParams` to `Params` is actually already in v6.0.x, so if we should keep it as `ClientParams`, should we revert the PR that did the renaming or should we just backport this PR (and then we get `ConnectionParams` and also the fix for `UpgradedConsensusState` for free)? Any preference?

Since this PR contains actually a feature (the addition of the `ConnectionParams` gRPC) and a bug fix (for the `UpgradedConsensusState` gRPC). How should we handle that with the automated changelog generation? Can we have multiple lines in the squash commit and gitcliff will interpret each? cc: @chatton

chatton:
@crodriguezvega very good question, in the current state of #2697 we will just be looking at the first line of the commit message, so a single entry. It is possible to configure it it to treat every line as an entry, I think this is an excellent use case. Non conventional commits are filtered out, so I believe it should just look at the lines that have the `type: message`.

damiannolan:
I'm happy to pick out the fix as a separate PR if needs @crodriguezvega 
Otherwise I'll add two changelogs for this PR. Let me know what you think? I do think it would be good to backport this to v6

colin-axner (approved):

chatton (commented):
Looks great, since we do now support multiple changelog entries, can we include those in PR description before merge ❤️

damiannolan:
> Looks great, since we do now support multiple changelog entries, can we include those in PR description before merge ❤️

Does it look good to you? Should they be added as the squashed commit msg?
I've also manually added the changelogs here, as I had to remove an old one too!

chatton:
> Does it look good to you? Should they be added as the squashed commit msg? I've also manually added the changelogs here, as I had to remove an old one too!

@damiannolan looks good! You can put the two lines in as the description in the squashed pit, so you can have

PR name

Additional Description
```
imp(core/03-connection): Adding `ConnectionParams` grpc query and CLI to 03-connection.
fix(modules/core/keeper): Fix request wiring for `UpgradedConsensusState` in core query server.
```

One thing to keep in mind is that this message `imp(core/03-connection): Adding ConnectionParams grpc query and CLI to 03-connection.` will render as `Adding ConnectionParams grpc query and CLI to 03-connection.`. i.e. not include `core/03-connection` in the message. You may have intended to do `imp: (core/03-connection) Adding ConnectionParams grpc query and CLI to 03-connection.` if the goal is to include core/03-connection in the changelog entry.

chatton (approved):
Looks great thank you!

Issue #1874, opened 2022-08-04
Enforce that proof height is 0 for solo machine


colin-axner:
## Summary

Due to addition of the generic `VerifyMembership` and `VerifyNonMembership`, the solo machine special case handling of batched verifications (ConnOpenTry and ConnOpenAck) will fail. Instead of the solo machine using the proof height to match the sequence (which is used for replay protection), the proof height should always be 0 for the solo machine and for verification it should use its latest sequence

Tests should be added which would have caught the ConnOpenTry issue

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

Issue #121, opened 2021-04-14
Add end to end solo machine tests


colin-axner:
## Summary

an end to end solo machine test would have caught #119. It would be great if we could utilize v2 of the testing package to quickly write integration tests

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
@colin-axner Can this issue be considered done with the tests added as part of the 02-client refactor?

colin-axner:
Yup!

PR #2746, opened 2022-11-14
fix: allow zero proof height, solo machine discards provided proof height in favor of sequence

colin-axner:
## Description

closes: #1874 
partially: #121

---

write a little note why.


modules/light-clients/06-solomachine/solomachine_test.go
```diff
@@ -46,6 +46,18 @@ func TestSoloMachineTestSuite(t *testing.T) {
 	suite.Run(t, new(SoloMachineTestSuite))
 }
 
+func (suite *SoloMachineTestSuite) TestConnectionHandshake() {
```
colin-axner, line 49: when channel handshake tests are added, this func can be renamed to `Setup()` and it can be extended to finish a channel handshake

colin-axner, line 49: These tests couldn't be split out because they actually will fail on main (main is currently broken for solo machines)

damiannolan: This isn't being used anywhere yet, right?
Where do you think we will use it? And should we mock the counterparty(TRY/CONFIRM) handlers too?

testing/solomachine.go
```diff
@@ -13,10 +13,18 @@ import (
...
+	connectiontypes "github.com/cosmos/ibc-go/v6/modules/core/03-connection/types"
 	commitmenttypes "github.com/cosmos/ibc-go/v6/modules/core/23-commitment/types"
 	host "github.com/cosmos/ibc-go/v6/modules/core/24-host"
 	"github.com/cosmos/ibc-go/v6/modules/core/exported"
 	solomachine "github.com/cosmos/ibc-go/v6/modules/light-clients/06-solomachine"
+	ibctm "github.com/cosmos/ibc-go/v6/modules/light-clients/07-tendermint"
+)
+
+var (
+	clientIDSolomachine     = "client-on-solomachine"     // clientID generated on solo machine side
```
damiannolan, line 25: nit: should we use `06-solomachine-0`? so it follows convention?

damiannolan (approved):
Awesome, nice to see all those red diffs!

testing/solomachine.go
```diff
@@ -13,10 +13,18 @@ import (
...
+	connectionIDSolomachine = "connection-on-solomachine" // connectionID generated on solo machine side
```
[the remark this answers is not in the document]
colin-axner: I did this initially, but I actually like making the identifiers completely different. These are supposed to be identifiers on the solomachine (not on the tendermint chain). It gets kind of confusing reading `FirstConnectionID` everywhere, so I thought it'd make more sense to make a unique identifier for the client and connection so when it is printed, you can understand which connectionID belongs to tendermint and which belongs to solomachine. Does that make sense?

testing/solomachine.go
colin-axner: No, that identifier will be used on the tendermint chain (chainA). I would find it very hard to debug if both sides had the same identifier, I find it easier to work through the code with unique identifiers for each side. Does that make sense?

modules/light-clients/06-solomachine/solomachine_test.go
colin-axner: Currently it is acting as an integration test for setting up connections. We will use it in integration tests for setting up channels and sending/receiving packets. Currently without the fixes in this pr, main will not pass these tests (solo machine has no integration tests with core IBC and several bugs have historically slipped through because of that)

testing/solomachine.go
damiannolan: Ah okay, I see your reasoning! So chainA would be a regular tm chain and the counterparty would be a solomachine implementation with a completely different ID. I agree for debugging its better to keep like this.

In practice, it would need to follow the standard client ID structure to pass [`ValidateClientType`](https://github.com/cosmos/ibc-go/blob/1b7745f30ff1e35640d121d449087aa5b8910793/modules/core/02-client/types/client.go#L90) tho, right? That is for the solomachine client on the tm chain. 
Doesn't that mean that the solomachine(chainB) must have a `07-tendermint` client for tracking its counterparty (chainA)?

testing/solomachine.go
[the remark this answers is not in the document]
damiannolan: Yeah, that's fair 👍 
From reading the other comment also, I agree.

testing/solomachine.go
colin-axner: For the solomachine on the tm chain that is correct. The identifier in question is the tendermint client on the solomachine though. 

> Doesn't that mean that the solomachine(chainB) must have a 07-tendermint client for tracking its counterparty (chainA)?

Correct, but we just mock this behaviour (pretend it exists). We convince chainA exists in the connection handshake (generation of the tm client state/consensus state)

docs/migrations/v6-to-v7.md
```diff
@@ -45,6 +45,8 @@ The `CheckMisbehaviourAndUpdateState` function has been removed from `ClientStat
 
 The function `GetTimestampAtHeight` has been added to the `ClientState` interface. It should return the timestamp for a consensus state associated with the provided height.
 
+A zero proof height is now allowed by core IBC and may be passed into `VerifyMembership` and `VerifyNonMembership`. Light clients are responsible for returning an error if a zero proof height is invalid behaviour.
```
AdityaSripal, line 48: Should we also mention relayers are must pass zero height for solomachines?

AdityaSripal (approved):

docs/migrations/v6-to-v7.md
colin-axner: Relayers can pass non zero heights. Solo machine simply ignores the provided value

PR #2780, opened 2022-11-17
e2e: remove irrelevant compatibility tests

crodriguezvega:
## Description

After discussion with @chatton this is a quick-and-~~dirty~~-simple proposal to remove irrelevant compatibility tests (i.e. tests run between tags that have already been released, like v4.2.0 and v3.4.0). Each json file has now 2 variants: a `-chain-a.json` and a `-chain-b.json`. The `chain-a` suffix means that the json file has the tip of the release branch only as the binary acting as chain a (and the same thing for `chain-b`). This way we run all the combinations that we are interested for testing compatibility of the release branch, and nothing else.

The advantage of this is that we can group all the transfer tests again in a single file (since it will be harder to hit the 256 GitHub job limit) and that we hopefully get less test failures (and if we get them they are relevant to the compatibility for the release branch and not related to two tags that have already been released).

closes: #XXXX

```bash
e2e: remove irrelevant compatibility tests
```

---

write a little note why.


chatton (commented):
Thank you for the ~~dirty~~ simple fix! @crodriguezvega ! ❤️ 

config files look good to me, have you run a full set of compatibility tests from this branch to make sure everything works okay?

I believe we still need to update [the workflow itself](https://github.com/cosmos/ibc-go/blob/8c07cb891ff4d42a4a525409b0c9abc8e2de85b3/.github/workflows/e2e-compatibility.yaml#L60-L85)

crodriguezvega:
> config files look good to me, have you run a full set of compatibility tests from this branch to make sure everything works okay?

Good point. I will run them.

> I believe we still need to update [the workflow itself](https://github.com/cosmos/ibc-go/blob/8c07cb891ff4d42a4a525409b0c9abc8e2de85b3/.github/workflows/e2e-compatibility.yaml#L60-L85)

Good catch. I will fix it now.

.github/workflows/e2e-compatibility-unreleased.yaml
```diff
@@ -0,0 +1,106 @@
...
+          docker build . -t "${REGISTRY}/${ORG}/${IMAGE_NAME}:$docker_tag"
+          docker push "${REGISTRY}/${ORG}/${IMAGE_NAME}:$docker_tag"
+
+  transfer:
+    needs:
+      - build-release-images
+      - determine-test-directory
+    uses: ./.github/workflows/e2e-compatibility-workflow-call.yaml
+    with:
+      test-file-directory: "${{ needs.determine-test-directory.outputs.test-directory }}"
```
chatton, line 78: in this case the test directory should be something like `unreleased` so we don't need to dynamically determine it. We can just hard code it for everything.

.github/workflows/e2e-compatibility-unreleased.yaml
```diff
@@ -0,0 +1,106 @@
...
+  determine-test-directory:
+    runs-on: ubuntu-latest
+    outputs:
+      test-directory: ${{ steps.set-test-dir.outputs.test-directory }}
+    steps:
+      - run: |
+          test_dir="$(echo $RELEASE_BRANCH | sed 's/\//-/')"
+          echo $test_dir
+          echo "::set-output name=test-directory::$test_dir"
+        id: set-test-dir
```
chatton, line 28: we should be able to remove this task and all the references to it.

.github/workflows/e2e-compatibility-unreleased.yaml
```diff
@@ -0,0 +1,106 @@
+name: Compatibility E2E
+on:
+  workflow_dispatch:
+    inputs:
+      release-branch:
+        description: 'Select unreleased to run all release branches against each other'
+        required: true
+        type: choice
+        options:
+          - unreleased
```
chatton, line 10: We can leave this as is, or remove it since we have one option, will leave it to you!

chatton (commented):
Looks great I think these changes should do the trick. I have one concern though, we are adding  jobs that don't exist for e.g. `release-v2.5.x` does not have a `client-chain-a`. I believe this will have the effect of the workflow running but this job failing. This is not a huge deal, but something we will run into later. I'm happy
if we are okay leaving this and fix it later when we re-merge these workflows.

It would be nice to also try to run both workflows from this branch to ensure there are no issues with the changes!

crodriguezvega:
[Test run](https://github.com/cosmos/ibc-go/actions/runs/3487750960) for the compatibility tests of `v6.0.x`.
[Test run](https://github.com/cosmos/ibc-go/actions/runs/3487826300/jobs/5836112535) for the compatibility tests of `v2.5.x` (it shows that it skips the tests for the json files that are not present).
I haven't been able to run the compatibility tests among unreleased branches, since the new workflow doesn't show on the left sidebar (because I think it also takes the workflows from main?).

chatton (approved):
LGTM thanks @crodriguezvega

damiannolan (approved):

Issue #2751, opened 2022-11-14
Allow zero proof height on packet receive, ack, timeout (with solomachine test)


colin-axner:
## Summary

Add solomachine packet receive, ack, timeout test. Tests should fail on zero proof height, then remove proofHeight.IsZero() check from receive, ack, timeout. May be split up if you only have time to implement one of the 3 tests

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #2781, opened 2022-11-17
Allow zero proof height packet recv ack timeout

chatton:
## Description

closes: #2751

```bash
type: commit message
```

---

write a little note why.


modules/light-clients/06-solomachine/client_state_test.go
```diff
@@ -802,11 +802,6 @@ func (suite *SoloMachineTestSuite) TestGetTimestampAtHeight() {
 			expValue:    suite.solomachine.ClientState().ConsensusState.Timestamp,
 			expPass:     true,
 		},
-		{
-			name:        "get timestamp at height not exists",
-			clientState: suite.solomachine.ClientState(),
-			height:      suite.solomachine.ClientState().GetLatestHeight().Increment(),
-		},
```
chatton, line 809: this test case was testing the behaviour of `GetTimestampAtHeight` which would error out when `!cs.GetLatestHeight().EQ(height)` , which means a zero proof would never be valid. 

I'm not sure exactly how to go ahead with this test case or if it's still valid. cc @damiannolan @colin-axner

modules/light-clients/06-solomachine/client_state.go
```diff
@@ -46,9 +46,6 @@ func (cs ClientState) GetTimestampAtHeight(
 	cdc codec.BinaryCodec,
 	height exported.Height,
 ) (uint64, error) {
-	if !cs.GetLatestHeight().EQ(height) {
-		return 0, sdkerrors.Wrapf(ErrInvalidSequence, "not latest height (%s)", height)
-	}
```
colin-axner, line 51: this is correct :+1:

modules/light-clients/06-solomachine/solomachine_test.go
```diff
@@ -46,7 +54,7 @@ func TestSoloMachineTestSuite(t *testing.T) {
 	suite.Run(t, new(SoloMachineTestSuite))
 }
 
-func (suite *SoloMachineTestSuite) TestSolomachineSetup() {
+func (suite *SoloMachineTestSuite) SetupSolomachine() string {
```
damiannolan, line 57: I think we should maybe return a `SolomachineConfig` struct from this function which encapsulates the client/connection/channel/port IDs.

damiannolan:
I think this is more or less r4r. Pushed some additional changes for `MsgTimeoutOnClose` and grouped the functions together logically in `testing/solomachine.go`.

I think we should add a `SolomachineConfig` struct of some sort to encapsulate the ibc identifiers for the tm chain and solomachine together. I'll create an issue for adding a test for `UpdateClient` and that can be done in a separate PR.

testing/solomachine.go
```diff
@@ -231,119 +346,74 @@ func (solo *Solomachine) ChanCloseConfirm(chain *TestChain, channelID string) {
...
-	solo.PrivateKeys = newPrivKeys
-	solo.PublicKeys = newPubKeys
-	solo.PublicKey = newPubKey
-	solo.Diversifier = newDiversifier
-
-	return header
+// SendPacket mocks sending a packet by setting a packet commitment directly.
+func (solo *Solomachine) SendPacket(chain *TestChain, packet channeltypes.Packet) {
+	commitmentHash := channeltypes.CommitPacket(chain.Codec, packet)
+	chain.GetSimApp().IBCKeeper.ChannelKeeper.SetPacketCommitment(chain.GetContext(), packet.GetSourcePort(), packet.GetSourceChannel(), packet.GetSequence(), commitmentHash)
```
colin-axner, line 352: Ideally we actually send like an ics20 transfer. There was a bug in ibc core channel keeper in relation to solution machine which wouldn't be caught by this, but would be caught by calling ics20 send. I think other clients might look at these tests as inspiration, so might be useful to have a full test. Totally fine with creating a followup issue

colin-axner (approved):
Yea, wahoo!! :rocket: This is so cool to have :)

AdityaSripal (approved):

testing/solomachine.go
chatton: Great suggestion, I'll create a follow up issue for this 👍

modules/light-clients/06-solomachine/solomachine_test.go
chatton: looking into this now, is looks like the endpoints in all the other tests are constructed using a `suite.chainA` and `suite.chainB`, for this will we need to fully create instances of `ibctesting.Endpoint` that are referencing `solomachine` and `chainA`?

chatton: @colin-axner do you think we should do this in this PR or are you happy with a follow up issue? This PR is already pretty big!

colin-axner: I think any change to use endpoint can be made in a followup pr. I would not recommend modifying the testing package to handle solo machine endpoints, I was primarily thinking that you could manually fill in the endpoint with the correct values without supporting the ability to do `suite.coordinator.Setup(path)`. I see the difficulty though since when you call `NewEndpoint` you provide quite a bit of information. What do you think about having a `NewSolomachineEndpoint`, also very open to just creating a new solomachine endpoint struct which contains less information (if that makes sense). I figure we can always consolidate the two types in the future if there is enough overlap

colin-axner:
@chatton You may merge into my branch whenever you feel comfortable (also feel free to set my branch to automerge if I'm away when you do). If you'd like me to merge my branch into main first, just let me know.

modules/light-clients/06-solomachine/solomachine_test.go
chatton: I think a new type may work well here, I will create a follow up issue for this and look into after this is merged into your branch! 👍

PR #2820, opened 2022-11-23
feat: add helper function for parsing clientID from client state path

colin-axner:
## Description

Will be used in the 02-client migrations prs as well

```bash
feat: add `MustParseClientStatePath` which parses the clientID from a client state key path
```

---

write a little note why.

### DIFF
--- a/.github/workflows/e2e-compatibility-workflow-call.yaml
+++ b/.github/workflows/e2e-compatibility-workflow-call.yaml
@@ -50,4 +50,4 @@ jobs:
           CHAIN_A_TAG: "${{ matrix.chain-a }}"
           CHAIN_B_TAG: "${{ matrix.chain-b }}"
           CHAIN_BINARY: "${{ matrix.chain-binary }}"
-          RLY_TAG: "v2.0.0"
+          RLY_TAG: "v2.1.2"
--- a/.github/workflows/e2e-fork.yml
+++ b/.github/workflows/e2e-fork.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Run e2e Test
         env:
           # see images here https://github.com/cosmos/relayer/pkgs/container/relayer/versions
-          RLY_TAG: "v2.0.0"
+          RLY_TAG: "v2.1.2"
         run: |
           cd e2e
           make e2e-test entrypoint=${{ matrix.entrypoint }} test=${{ matrix.test }}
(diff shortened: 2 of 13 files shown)


modules/core/24-host/parse.go
```diff
@@ -32,6 +32,40 @@ func ParseIdentifier(identifier, prefix string) (uint64, error) {
...
+	if err != nil {
+		panic(err.Error())
+	}
+
+	return clientID
+}
+
+// parseClientStatePath returns the client ID from a client state path. It returns
+// an error if the provided path is invalid.
+func parseClientStatePath(path string) (string, error) {
```
colin-axner, line 48: I debated whether to make this public or not. I opted for only making `MustParseClientStatePath` public. My reasoning is that when receiving an error from this function, it is difficult to tell whether the error was because this is not a client state path or because the key path itself is corrupted (clientID is malformed). I think it is safer to expect the caller to ensure that it only calls this function on paths it expects to be a client state path, thus only exposing `MustParseClientStatePath` makes sense as any error implies state corruption

modules/core/02-client/keeper/keeper.go
```diff
@@ -359,15 +359,16 @@ func (k Keeper) IterateClients(ctx sdk.Context, cb func(clientID string, cs expo
 
 	defer iterator.Close()
 	for ; iterator.Valid(); iterator.Next() {
-		keySplit := strings.Split(string(iterator.Key()), "/")
-		if keySplit[len(keySplit)-1] != host.KeyClientState {
+		path := string(iterator.Key())
+		if !strings.Contains(path, host.KeyClientState) {
+			// skip non client state keys
 			continue
```
damiannolan, line 365: will this ever be hit?

damiannolan (approved):
Nice addition! LGTM 🚀

modules/core/02-client/keeper/keeper.go
colin-axner: Yes for consensus state keys and metadata keys

colin-axner: Since client state keys are at the end of the key, you cannot prefix iterate on them

charleenfei (approved):
nice! lgtm

chatton (approved):
LGTM

Issue #1863, opened 2022-08-02
Add migrations for solo machine protobuf types


colin-axner:
## Summary

Todo:
- [x] add optional migrations #2800 
- [x] add automatic in-place store migrations #2819 
- [x] add genesis migrations #2824 
- [ ] remove legacy migrations
- [x] review documentation

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
remove `Allow` boolean in solo machine client state as well

colin-axner:
I did some initial work for this since the issue lacked sufficient information to get started. https://github.com/cosmos/ibc-go/tree/colin/1863-v7-migrations

What I have done:
- reused v1 to v2 consensus version migrations and renamed them to v2 to v3
- moved localhost migration to v2 to v3 migrations and made it automatic
- moved all migrations into `02-client/migrations/v7`
- removed addition of metadata (used in v1 to v2)
- updated solomachine migration to be v2 to v3 proto definition

What's left?
- decide if we want to support genesis migrations (it's really messy), if so review the tendermint expired client migration code (I only removed the addition of the extra metadata, but it might be possible to simplify)
- fix broken tests 
- probably move localhost_test.go into store_test.go
- potentially add genesis migration test for localhost
- remove localhost client state in genesis migration (not currently happening because it is a bit tricky)

Note: It will not be possible to upgrade from stargate -> v7

colin-axner:
Further thoughts:

Let's move tendermint expired consensus state pruning into an optional migration which chains can call directly. I like the idea of optimizations being optional and required migrations being automatic. Iterating over all clients/consensus states can actually take quite a bit of time (a year ago it took like an hour), so I think it makes sense to skip as automatic. It could also be reused during any upgrade. Only iterating over solo machine/localhost clients should be a lot faster!

Lets also add logging to the expired tendermint pruning. Might be nice to log the number of consensus states pruned. This can inform a chain whether it is worth running the migration or not since even if 2 consensus states are pruned, all consensus states will be iterated

PR #2824, opened 2022-11-24
feat: add genesis migrations for v7

colin-axner:
## Description

ref: #1863 

```bash
feat: Add genesis migrations for v6 to v7. The migration migrates the solo machine client state definition, removes all solo machine consensus states and removes the localhost client.
```

---

write a little note why.

- [x] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#testing).


modules/core/02-client/migrations/v7/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/cosmos/ibc-go/v6/modules/core/02-client/types"
+	"github.com/cosmos/ibc-go/v6/modules/core/exported"
+)
+
+// MigrateGenesis accepts an exported IBC client genesis file and migrates it to:
+//
+// - Update solo machine client state protobuf definition (v2 to v3)
+// - Remove all solo machine consensus states
+// - Remove localhost client
+func MigrateGenesis(clientGenState *types.GenesisState, cdc codec.ProtoCodecMarshaler) (*types.GenesisState, error) {
```
colin-axner, line 16: A lot of this work is ported from v1 legacy migrations. I cleaned up a lot of it, but there might be some more cleanup. Generally working with genesis isn't very clean though

modules/core/migrations/v7/genesis_test.go
```diff
@@ -0,0 +1,156 @@
...
+func (suite *MigrationsV7TestSuite) TestMigrateGenesisSolomachine() {
+	// create tendermint clients
+	for i := 0; i < 3; i++ {
+		path := ibctesting.NewPath(suite.chainA, suite.chainB)
+
+		suite.coordinator.SetupClients(path)
+
+		err := path.EndpointA.UpdateClient()
+		suite.Require().NoError(err)
+		err = path.EndpointA.UpdateClient()
```
charleenfei, line 52: repeated codeblock here?

colin-axner: it was intentional to update twice to generate a little more state. I can add a comment

damiannolan (approved):
Nice! I can create an issue to add an e2e for the v7 upgrade

edit: one already exists!

modules/core/02-client/migrations/v7/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+		switch clientType {
+		case exported.Solomachine:
+			var clientState ClientState
+			if err := cdc.Unmarshal(client.ClientState.Value, &clientState); err != nil {
+				return nil, sdkerrors.Wrap(err, "failed to unmarshal client state bytes into solo machine client state")
+			}
+
+			updatedClientState := migrateSolomachine(clientState)
+
+			any, err := types.PackClientState(&updatedClientState)
```
chatton, line 40: `any` is now a golang keyword (alias for `interface{}` as of golang 1.18) can we change this to maybe `anyCodec` or similar?

modules/core/02-client/migrations/v7/genesis.go
```diff
@@ -0,0 +1,82 @@
+package v7
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/cosmos/ibc-go/v6/modules/core/02-client/types"
```
chatton, line 7: should we alias this as `clienttypes`? I think having a `types` import without the alias adds a little overhead in reading the code. (more so if other types imorts exist which is not the case here!)

modules/core/02-client/migrations/v7/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+		clients          []types.IdentifiedClientState
+	)
+
+	for _, client := range clientGenState.Clients {
+		clientType, _, err := types.ParseClientIdentifier(client.ClientId)
+		if err != nil {
+			return nil, err
+		}
+
+		// update solo machine client state defintions
```
chatton, line 30: [nit] should we move this comment inside the solomachine case? The way we do for the localhost case.

modules/core/02-client/migrations/v7/genesis.go
```diff
@@ -0,0 +1,82 @@
...
+			// add all other client states
+			clients = append(clients, client)
+		}
+
+		// iterate consensus states by client
+		for _, clientConsensusStates := range clientGenState.ClientsConsensus {
+			// look for consensus states for the current client
+			if clientConsensusStates.ClientId == client.ClientId {
+				switch clientType {
+				case exported.Solomachine:
```
chatton, line 63: [nit] we don't need multiple switch cases here for solomachine/localhost

```go
switch clientType {
  case exported.Solomachine, Localhost:
    // noop
  case default:
      clientsConsensus = append(clientsConsensus, clientConsensusStates)
}
```

modules/core/02-client/migrations/v7/genesis_test.go
```diff
@@ -0,0 +1,134 @@
...
+			ConsensusState: &v7.ConsensusState{
+				PublicKey:   clientState.ConsensusState.PublicKey,
+				Diversifier: clientState.ConsensusState.Diversifier,
+				Timestamp:   clientState.ConsensusState.Timestamp,
+			},
+			AllowUpdateAfterProposal: true,
+		}
+
+		// set client state
+		any, err := codectypes.NewAnyWithValue(legacyClientState)
```
chatton, line 57: [nit] same point about `any`

modules/core/02-client/migrations/v7/genesis_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	bz, err := cdc.MarshalJSON(&expectedClientGenState)
+	suite.Require().NoError(err)
+
+	// Indent the JSON bz correctly.
+	var jsonObj map[string]interface{}
+	err = json.Unmarshal(bz, &jsonObj)
+	suite.Require().NoError(err)
+	expectedIndentedBz, err := json.MarshalIndent(jsonObj, "", "\t")
+	suite.Require().NoError(err)
```
chatton, line 122: this actually seems like a useful utility function that would be nice to have any time we are doing a json comparison. WDYT?

modules/core/migrations/v7/genesis.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/cosmos/ibc-go/v6/modules/core/types"
+)
+
+// MigrateGenesis accepts an exported IBC client genesis file and migrates it to:
+//
+// - Update solo machine client state protobuf definition (v2 to v3)
+// - Remove all solo machine consensus states
+// - Remove any localhost clients
+func MigrateGenesis(appState genutiltypes.AppMap, cdc codec.ProtoCodecMarshaler) (genutiltypes.AppMap, error) {
+	if appState[host.ModuleName] != nil {
```
chatton, line 18: [nit] I think this could be a little cleaner by reducing the indentation level of the core logic (i.e. early return of `appState` when `appState[host.ModuleName] == nil`)

modules/core/migrations/v7/genesis.go
```diff
@@ -0,0 +1,40 @@
...
+		// ensure legacy solo machines are registered
+		clientv7.RegisterInterfaces(cdc.InterfaceRegistry())
```
chatton, line 20: will `RegisterInterfaces` panic if the interface registry is already registered? I remember seeing some behaviour where duplicate interface registration panicked.

modules/core/migrations/v7/genesis.go
```diff
@@ -0,0 +1,40 @@
...
+
+		// unmarshal relative source genesis application state
+		ibcGenState := &types.GenesisState{}
+		cdc.MustUnmarshalJSON(appState[host.ModuleName], ibcGenState)
```
chatton, line 24: [q] is it safe to panic during migrations like this?

modules/core/migrations/v7/genesis_test.go
```diff
@@ -0,0 +1,158 @@
...
+			ConsensusState: &clientv7.ConsensusState{
+				PublicKey:   clientState.ConsensusState.PublicKey,
+				Diversifier: clientState.ConsensusState.Diversifier,
+				Timestamp:   clientState.ConsensusState.Timestamp,
+			},
+			AllowUpdateAfterProposal: true,
+		}
+
+		// set client state
+		any, err := codectypes.NewAnyWithValue(legacyClientState)
```
chatton, line 84: same comment about `any`

modules/core/migrations/v7/genesis_test.go
```diff
@@ -0,0 +1,158 @@
+package v7_test
```
chatton, line 1: should we add a test to make sure migration passes with the interfaces already registered (only if it is a possible concern ofc!)

chatton (commented):
Great work! I mostly had some nits but also a few questions! The migrations themselves are really clean 🧹

modules/core/02-client/migrations/v7/genesis.go
damiannolan: nice catch, we could use `protoAny` to avoid shadowing maybe? I reckon `any` is used widely across ibc-go and the sdk as well. 

I think I noticed before we also. use `cap` for capability in places which is the "capacity" function in Go. I don't think its a super big deal tho and shouldn't cause any issues

modules/core/02-client/migrations/v7/genesis.go
colin-axner: Yea I agree. The reasoning for why it is types is because it is within the `02-client` subdirectory, but it is a little removed from that so I think it makes sense to use `clienttypes`

modules/core/02-client/migrations/v7/genesis_test.go
colin-axner: I could see that. Do we do json comparison anywhere else? Genesis migrations are quite the special case

modules/core/migrations/v7/genesis.go
colin-axner: Yes it is, this is being run while the chain is halted

modules/core/migrations/v7/genesis.go
colin-axner: We don't want the user to call `RegisterInterfaces` before hand. The migration registration should not persist onto the chain codec, it should only be used to perform the genesis migrations. I can see if I can add a test to ensure it panics

modules/core/02-client/migrations/v7/genesis_test.go
chatton: yeah I think it might just be a special case here, I'm happy if you'd prefer to hold out on needing it an additional time before we implement a helper 👍

modules/core/migrations/v7/genesis_test.go
colin-axner: this is already the case. I added a comment in the testing code

modules/core/migrations/v7/genesis.go
colin-axner: The codec will not panic if the types are already registered. It only panics if you try to register duplicate types (same URL, different generated files)

modules/core/02-client/migrations/v7/genesis.go
colin-axner: updated

modules/core/02-client/migrations/v7/genesis.go
colin-axner: updated

modules/core/migrations/v7/genesis_test.go
chatton: ah thank you I missed that! 👍

modules/core/02-client/migrations/v7/genesis_test.go
colin-axner: Yea lets go with that. We at least know where to find the code :wink:

modules/core/migrations/v7/genesis.go
chatton: thanks for the clarification!

modules/core/02-client/migrations/v7/genesis.go
colin-axner: I removed the comment. I don't think it added much information

modules/core/migrations/v7/genesis.go
chatton: great that makes sense thanks

Issue #2823, opened 2022-11-24
Events unit testing across the board


Anmol1696:
## Summary

Currently there is no tests for events emitted by any of the msgs. The following proposal is a way to have easy testablity in unit tests for events emitted per msg.

### Why do we need this feature?
* Events are slowly becoming source of truth for many downstream applications, we need a way to make sure there are no regressions happening on events
* Events produced is untested code, we will also increase the code coverage

### What benefits does the ibc-go stand to gain by including this feature?
* Programatic way to catch regressions on events faster

### Are there any disadvantages of including this feature?
* More overhead for developers as testing events has more boiler plate
* Benifits might be minimal compared to the efforts

## Proposal

The proposal is to add a testutil function which can assert events.
```go
package testutil

assertEvents(suite suite.Suite, expected map[string]map[string]string, actual sdk.Events)
```
To make this function importable, I propose we create a new package `testutil` at the root level, where we can have multiple test utils that can then be imported. Something like [`testutil`](https://github.com/cosmos/cosmos-sdk/tree/main/testutil). package in `cosmos-sdk`

In the test functions where the events are emmited we can call the above function as:
```go
// Create expected event
expEvents := map[string]map[string]string{
  "ibc_transfer": {
	  "sender":   suite.chainA.SenderAccount.GetAddress().String(),
	  "receiver": suite.chainB.SenderAccount.GetAddress().String(),
	  "amount":   coin.Amount.String(),
	  "denom":    coin.Denom,
  },
  "send": {
    ...
  }
}

// Actual event
events := ctx.EventManager().Events()

// Perform the assert
assertEvents(suite, expEvents, events)
```

## Points of disucssion
* As per the current proposal we hardcode the event types and attribute keys per test, so as to make the test case itself brittle, if the default strings changes the test should fail as well. But this is a point of dicussion, open to changing it to less hard coded.
* For this proposal we need to make `assertEvents` function an importable function which can be used by various functions. Creating a new package might be an overkill, but can go a long way in the future to have common test utils.

## PRs
- [x]   assert function and `app/transfer` - #2829 
- [x]  `04-channel` - #2830 
- [x] `03-connection` #5682
- [x] `02-client` #5686
- [x]  `apps/27-ica` #5687
- [ ]  `apps/29-fee` #5706

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

[...]

PR #2829, opened 2022-11-26
imp: event assertion tests function

Anmol1696:
## Description

Add assertion function in `ibctesting` package, and use it in the transfer app tests

part of: #2823

```bash
imp: Add `AssertEvents` which asserts events against expected event map.
```

---

write a little note why.


testing/events.go
```diff
@@ -129,3 +130,33 @@ func ParseAckFromEvents(events sdk.Events) ([]byte, error) {
 	}
 	return nil, fmt.Errorf("acknowledgement event attribute not found")
 }
+
+// AssertEvents asserts that expected events are present in the actual events
+// expected map needs to be a subset of actual events to pass
+func AssertEvents(
+	suite *suite.Suite,
+	expected map[string]map[string]string,
```
chatton, line 138: is there a reason to use `map[string]map[string]string` as opposed to taking an `sdk.Events` directly?

chatton (approved):
This is a fantastic additional thanks @Anmol1696 , just had one question, but the implementation looks great 👍

testing/events.go
Anmol1696: Just wanted the map of expected events being readable as a map for the test itself.

We can change this to `sdk.Events` as well, but i chose readability of expected events. Do you think we should change it?

Anmol1696: Cool, this snipet is the reason, to me this just makes intutive sense of what to expect from the event. https://github.com/cosmos/ibc-go/pull/2829/files#diff-ae66bd29b0c68b16e3fc64fb2d8c88dcc90ef4ed6c1f67be36baf8d693473f8eR150-R156

modules/apps/transfer/keeper/msg_server_test.go
```diff
@@ -131,14 +139,26 @@ func (suite *KeeperTestSuite) TestMsgTransfer() {
...
+					"sender":   suite.chainA.SenderAccount.GetAddress().String(),
+					"receiver": suite.chainB.SenderAccount.GetAddress().String(),
+					"amount":   coin.Amount.String(),
+					"denom":    coin.Denom,
+					"memo":     "memo",
+				},
+			}
+
+			if hasEvents {
+				ibctesting.AssertEvents(&suite.Suite, expEvents, events)
```
crodriguezvega, line 160: I am wondering if it would be possible to move this assertion to the happy path case where `tc.expPass` would be true and then we wouldn't need the extra boolean `hasEvents`... Because are they both (and `tc.expPass` and `hasEvents`) true when we would like to assert events or would there be any situation where `tc.expPass` would be false but `hasEvents` is true?

testing/events.go
crodriguezvega: Would it be possible though to make a type definition for `map[string]map[string]string` just to improve readability? Like `eventsMap` or something like that? Something that signals the purpose of the type more clearly than the mouthful of `map[string]map[string]string`. :)

testing/events.go
```diff
@@ -129,3 +130,33 @@ func ParseAckFromEvents(events sdk.Events) ([]byte, error) {
 	}
 	return nil, fmt.Errorf("acknowledgement event attribute not found")
 }
+
+// AssertEvents asserts that expected events are present in the actual events
+// expected map needs to be a subset of actual events to pass
```
crodriguezvega, line 135:
```suggestion
// Expected map needs to be a subset of actual events to pass.
```

testing/events.go
```diff
@@ -129,3 +130,33 @@ func ParseAckFromEvents(events sdk.Events) ([]byte, error) {
 	}
 	return nil, fmt.Errorf("acknowledgement event attribute not found")
 }
+
+// AssertEvents asserts that expected events are present in the actual events
```
crodriguezvega, line 134:
```suggestion
// AssertEvents asserts that expected events are present in the actual events.
```

crodriguezvega (commented):
Thanks a lot, @Anmol1696. I left a couple of questions, let us know what you think about them!

testing/events.go
Anmol1696: makes sense will do

modules/apps/transfer/keeper/msg_server_test.go
Anmol1696: Also some more places where this was the case: https://github.com/cosmos/ibc-go/pull/2830/files#diff-ce312850968149b915fa31bf5158acb380a7efe3931048052422b0c6bd92ab80R749-R789

Anmol1696:
Could we please resolve the comments and move forward.
@crodriguezvega @chatton 

If you can also review @colin-axner @damiannolan @charleenfei

modules/apps/transfer/keeper/msg_server_test.go
```diff
@@ -131,14 +139,26 @@ func (suite *KeeperTestSuite) TestMsgTransfer() {
...
+			events := ctx.EventManager().Events()
+			expEvents := ibctesting.EventsMap{
+				"ibc_transfer": {
+					"sender":   suite.chainA.SenderAccount.GetAddress().String(),
+					"receiver": suite.chainB.SenderAccount.GetAddress().String(),
+					"amount":   coin.Amount.String(),
+					"denom":    coin.Denom,
+					"memo":     "memo",
+				},
+			}
```
damiannolan, line 157: I think we can use the event attributes here for the keys.

https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/events.go

damiannolan (approved):
Thanks for this @Anmol1696. Apologies for the delay in review. 
Left one suggestion regarding attribute keys

charleenfei (approved):
thank you for this, lgtm!

modules/apps/transfer/keeper/msg_server_test.go
```diff
@@ -131,14 +139,26 @@ func (suite *KeeperTestSuite) TestMsgTransfer() {
 				suite.Require().NoError(err)
 				suite.Require().NotNil(res)
 				suite.Require().NotEqual(res.Sequence, uint64(0))
-
-				events := ctx.EventManager().Events()
-				suite.assertTransferEvents(events, coin, "memo")
```
colin-axner, line 136: did this function get removed? Or is it still being used?

modules/apps/transfer/keeper/msg_server_test.go
colin-axner: See [discussion](https://github.com/cosmos/ibc-go/issues/2823#issuecomment-1326286697)

modules/apps/transfer/keeper/msg_server_test.go
Anmol1696: Cool makes sense, then will use `hasEvents` wherever necessary

modules/apps/transfer/keeper/msg_server_test.go
Anmol1696: This function is removed in favour of new function, which we can use everywhere

modules/apps/transfer/keeper/msg_server_test.go
Anmol1696: Done @colin-axner sorry for the delay :)

crodriguezvega (approved):
Great work, @Anmol1696! And thank for your patience and addressing all the feedback.

modules/apps/transfer/keeper/msg_server_test.go
colin-axner: Will it be removed in a separate pr? It is still in the [codebase](https://github.com/Anmol1696/ibc-go/blob/anmol/event-tests-transfer/modules/apps/transfer/keeper/msg_server_test.go#L11)?

Anmol1696: ahh sorry about that, i thought i removed it. will do asap

Anmol1696: Done removed it.

Issue #2844, opened 2022-11-29
Add iteration function to 02-client keeper which returns all clientIDs of a specific client type


colin-axner:
## Summary

Same iteration used in #2800. The function should take in a clientType and return a slice of clientIDs that correspond to the client type provided. It should be added to the 02-client keeper

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #2856, opened 2022-11-30
refactor: rename IterateClients to IterateClientStates, add a prefix

colin-axner:
## Description

After discussion with @chatton, we decided every key/value set in our stores should have a corresponding `Iterate` function. That is `IterateClientStates`, `IterateConsensusStates`, etc. These should take in a prefix to determine the starting point of the iteration (allows to limit unnecessary iterations) and a callback which can accrue the desired values and exit iteration early based on optimizations. This PR renames `IterateClients` to `IterateClientStates` and adds the prefix necessary for optimized iteration

closes: #2844 

```bash
feat(api)!: rename `IterateClients` to `IterateClientStates`. The function now takes a prefix argument which may be used for prefix iteration over the client store.
feat: add `PrefixedClientStorePath` and `PrefixedClientStoreKey` functions to 24-host
```

---

write a little note why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### API Breaking
 
-* (core/02-client) [\#2856](https://github.com/cosmos/ibc-go/pull/2856) Rename `IterateClients` to `IterateClientStates`. The function now takes a prefix argument which may be used for prefix iteration over client the client store.
+* (core/02-client) [\#2856](https://github.com/cosmos/ibc-go/pull/2856) Rename `IterateClients` to `IterateClientStates`. The function now takes a prefix argument which may be used for prefix iteration over the client store.
 * (apps/27-interchain-accounts) [\#2607](https://github.com/cosmos/ibc-go/pull/2607) `SerializeCosmosTx` now takes in a `[]proto.Message` instead of `[]sdk.Msg`.
 * (apps/transfer) [\#2446](https://github.com/cosmos/ibc-go/pull/2446) Remove `SendTransfer` function in favor of a private `sendTransfer` function. All IBC transfers must be initiated with `MsgTransfer`.
 * (apps/29-fee) [\#2395](https://github.com/cosmos/ibc-go/pull/2395) Remove param space from ics29 NewKeeper function. The field was unused.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### API Breaking
 
+* (core/02-client) [\#2856](https://github.com/cosmos/ibc-go/pull/2856) Rename `IterateClients` to `IterateClientStates`. The function now takes a prefix argument which may be used for prefix iteration over client the client store.
 * (apps/27-interchain-accounts) [\#2607](https://github.com/cosmos/ibc-go/pull/2607) `SerializeCosmosTx` now takes in a `[]proto.Message` instead of `[]sdk.Msg`.
 * (apps/transfer) [\#2446](https://github.com/cosmos/ibc-go/pull/2446) Remove `SendTransfer` function in favor of a private `sendTransfer` function. All IBC transfers must be initiated with `MsgTransfer`.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### API Breaking
 
+* (core/02-client) [\#2856](https://github.com/cosmos/ibc-go/pull/2856) Rename `IterateClients` to `IterateClientStates`. The function now takes a prefix argument which may be used for prefix iteration over client the client store.
```
colin-axner, line 45:
```suggestion
* (core/02-client) [\#2856](https://github.com/cosmos/ibc-go/pull/2856) Rename `IterateClients` to `IterateClientStates`. The function now takes a prefix argument which may be used for prefix iteration over the client store.
```

modules/core/02-client/keeper/keeper.go
```diff
@@ -147,7 +147,7 @@ func (k Keeper) IterateConsensusStates(ctx sdk.Context, cb func(clientID string,
 // GetAllGenesisClients returns all the clients in state with their client ids returned as IdentifiedClientState
 func (k Keeper) GetAllGenesisClients(ctx sdk.Context) types.IdentifiedClientStates {
 	var genClients types.IdentifiedClientStates
-	k.IterateClients(ctx, func(clientID string, cs exported.ClientState) bool {
+	k.IterateClientStates(ctx, nil, func(clientID string, cs exported.ClientState) bool {
```
chatton, line 150: I'm not a huge fan of method signatures where explicitly passing `nil` an intended/regular workflow.  Maybe this is overkill, but I think I prefer the idea of an explicit sentinel value like

```golang
var EmptyPrefx []byte
```

And use this rather than `nil`.  A value of nil could mean a lot of different things, with a value like this it is quite obvious what we are doing.

modules/core/24-host/keys.go
```diff
@@ -55,6 +55,20 @@ func FullClientKey(clientID string, path []byte) []byte {
 	return []byte(FullClientPath(clientID, string(path)))
 }
 
+// PrefixedClientStorePath returns a key path which can be used for prefixed
+// key store iteration. The prefix may be a clientType, clientID, or any
+// valid key prefix which may be concatenated with the client store constant.
+func PrefixedClientStorePath(prefix []byte) string {
```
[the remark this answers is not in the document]
chatton: ^^ this was written before the suggestion for a sentinel value.

chatton (commented):
Looks great this is a nice way of handling iterations 👍 

I had some suggestions about the usage of `nil`. Let me know what you think!

modules/core/02-client/keeper/keeper.go
colin-axner: That sounds fine to me. I'm impartial to this decision. The SDK already makes use of nil with [iterators](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/store.go#L237) so it doesn't feel out of the normal for me

 Where would you place the `EmptyPrefix`, 24-host?

modules/core/24-host/keys.go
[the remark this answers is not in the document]
colin-axner: Would you want any differing behaviour if we used a sentinel value?

modules/core/02-client/keeper/keeper.go
chatton: Ah I see, if it's already an existing pattern to pass nil to iterators then it probably is better to just stay consistent.

chatton (approved):
Happy to stick with `nil` for iterators to stay consistent with the sdk 👍

modules/core/02-client/keeper/keeper.go
colin-axner: Yup. I'm still happy to go with a sentinel value if we think that is cleaner. I guess we can use nil for now an update in the future?

colin-axner:
Going to merge, I agree with the `EmptyPrefix` approach, but I don't know where we should put such a type (as really it should be within the SDK). I think it makes sense to handle in a followup, maybe we can consider discussing with the SDK to change the design pattern in general for nil to mean iterate all?

Issue #2845, opened 2022-11-29
Modify optional tendermint migration to take in 02-client keeper instead of store key


colin-axner:
## Summary

Once #2844 is implemented, we can modify the optional tendermint pruning migration to take in the IBC client keeper instead of the store key. We can replace the usage of [this](https://github.com/cosmos/ibc-go/blob/9720607dadd8d258e36311d84ed4628f2516aa6b/modules/light-clients/07-tendermint/migrations.go#L27-#L40) with #2844 and [this](https://github.com/cosmos/ibc-go/blob/9720607dadd8d258e36311d84ed4628f2516aa6b/modules/light-clients/07-tendermint/migrations.go#L47-#L48) with a call to the `clientKeeper.ClientStore()`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #2862, opened 2022-12-01
refactor: simplify optional tendermint pruning migrations

colin-axner:
…grations

## Description

skipping changelog as this modifies unreleased changes

ref: #2845 

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/light-clients/07-tendermint/expected_keepers.go
+++ b/modules/light-clients/07-tendermint/expected_keepers.go
@@ -1,0 +1,16 @@
+package tendermint
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/tendermint/tendermint/libs/log"
+
+	"github.com/cosmos/ibc-go/v6/modules/core/exported"
+)
+
+// ClientKeeper expected account IBC client keeper
+type ClientKeeper interface {
+	GetClientState(ctx sdk.Context, clientID string) (exported.ClientState, bool)
+	IterateClientStates(ctx sdk.Context, prefix []byte, cb func(string, exported.ClientState) bool)
+	ClientStore(ctx sdk.Context, clientID string) sdk.KVStore
+	Logger(ctx sdk.Context) log.Logger
+}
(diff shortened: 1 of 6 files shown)


modules/light-clients/07-tendermint/expected_keepers.go
```diff
@@ -0,0 +1,16 @@
+package tendermint
```
colin-axner, line 1: we can optionally move `expected_keepers.go`, `migrations.go`, and `migrations_test.go` to its own package. 

Then the import would be:
```go
ibctmmigrations.PruneTendermintConsensusStates
```

Maybe we can rename the function to `PruneExpiredConsensusStates`?

damiannolan: I like the idea of moving these to their own package. I do agree, I'm not a huge fan of `ibctmmigrations` but it would just be another import alias and it could be aliased however a chain dev sees fit. 
I think the trade off for us is better because we can maintain the migration code and expected keepers file separately, at the expense of a kind of ugly import alias which we don't really have to worry about on a day to day.

I'd be leaning towards the migrations pkg personally.

chatton: If we don't expect these functions to be used outside the context of migrations I think a migrations package makes sense. Also not a huge fan of import alias but it's not a huge deal breaker.

colin-axner: I think pruning which would be performed outside of migrations would need to be optimized (not just iterating over all clients, but maybe a specific set of clients), so I think it makes sense to add this to a separate package and implement the optimized approach (which could maybe be executed via a governance proposal) at some other point, especially since `PruneAllExpiredConsensusStates` will continue to be a funciton of 07-tendermint

colin-axner: updated

damiannolan (approved):
I left a minor suggestion! LGTM! 🚀

chatton (approved):
LGTM!

Issue #330, opened 2021-08-12
Use module account and sub accounts for ICS20 escrow addresses


colin-axner:
## Summary

Currently, ICS20 escrow addresses use a prefix and hash of the port/channel to generate the address. Careful consideration was made to ensure these addresses would not conflict with one another. Now that v0.43 of the SDK has implemented ADR 28, we should eventually migrate our escrow addresses to using a module account and sub accounts for each escrow address.

This solution is low priority since the current implementation was determined to be secure, but only with the portID "transfer". I'm unsure if using the ADR 28 solution would allow us to lift the maximum channel sequence requirement

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
see [comment](https://github.com/cosmos/ibc-go/pull/2871#issuecomment-1341177770) for more details

gjermundgaraba:
Closing this for now. We might reopen this in the future, but this is not currently planned.

PR #2871, opened 2022-12-02
chore: use module accounts to escrow transferred coins

GAtom22:
## Description

At the moment, escrow accounts used on IBC transfers are regular accounts instead of ModuleAccounts. The proposed changes in this PR migrate escrow accounts to the `ModuleAccount` type

closes: [#330](https://github.com/cosmos/ibc-go/issues/330)

---

write a little note why.


GAtom22:
Hi @colin-axner, thanks for your comment! Having module accounts for the escrow accounts makes more sense in my opinion. Leaving them as simple accounts makes them vulnerable to a birthday attack. Quoting  [ADR-028](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-028-public-key-addresses.md)

>1. simple accounts: represented by a regular public key (ie: secp256k1, sr25519)
>2. naive multisig: accounts composed by other addressable objects (ie: naive multisig)
>3. composed accounts with a native address key (ie: bls, group module accounts)
>4. module accounts: basically any accounts which cannot sign transactions and which are managed internally by modules

I believe the module account description is more aligned with the escrow account than the simple account description. These accounts are managed by the transfer module.

Regarding the address derivation as module sub-accounts, I have no problem with it, seems it's a better design. It is true that this will be a breaking change and needs to be done carefully. But I believe these should be module accounts for the reasons explained. LMK your thoughts

GAtom22:
@colin-axner thanks for the clarification. I understand your point, I'm aware that the `GetModuleAccount` would not succeed on getting the escrow accounts even if they are ModuleAccounts. 

I'm putting emphasis on the account type because this situation has resulted in undesired behavior on our side. There's logic in our codebase where we check if an account is a ModuleAccount. In some cases we do this to perform actions only on BaseAccounts, as users accounts. But being the escrow addresses BaseAccounts, they are treated as if they were users accounts. This generated issues. So, changing the address type to ModuleAddress solves this type of bugs in a scalable way.

colin-axner:
Why not just check if the account has a public key set?

GAtom22:
That could work in some cases, but in other cases, being the escrow account a BaseAccount still generates undesired behavior. See [this code section](https://github.com/evmos/evmos/blob/445d08074298afa7c04097e47c96a9b506736103/x/recovery/keeper/ibc_callbacks.go#L86-L96) for example

colin-axner:
@GAtom22 thanks for linking the code. The current handling also does not account for interchain account types, though it seems quite difficult to me to find the key which corresponds to an account managed by a on chain module and thus bypass this [check](https://github.com/evmos/evmos/blob/445d08074298afa7c04097e47c96a9b506736103/x/recovery/keeper/ibc_callbacks.go#L73). Are there concrete concerns regarding this?

I think there's another unconsidered edge case in the existing code. What happens if I send funds to an address which supports or doesn't support the evmos keys? That account will not have a public key set (as I sent it funds), but it might still support evmos keys, but all its funds will be sent to the connected chain regardless. I don't see a safety concern, but it is undocumented behaviour. This case may be possible for regular genesis accounts as well since their public key may be nil. Are all genesis account presumed to have a nil or non-nil public key?

---

Getting back to ibc-go. I remain unconvinced we should be making any changes to ICS20. Here's my perspective. 

The current solution is works and is safe. The issue linked in the pr #330 aims to increase the hardness by ensuring the existing solution is safe for any variant of the ICS 20 portID name, though it is incredibly unclear why anyone would modify the "transfer" portID string and I have yet to see a single chain do so. It would also likely have external UX repercussions and thus is probably ill-advised. 

Increasing the hardness of ICS20 security is quite sensitive though and risky. Because funds will be moved to different addresses, there is no room for a single bug. Thus I would strongly recommend no change be made without strong reason to do so. 

Changing the account type of the escrow addresses also does not increase functionality for ibc-go. The module account is a base account, with extra information our escrow addresses do not need. 

The linked code is trying to ask the account keeper the question "Is this account a user account? If so, does it have a public key supported by us?" and it is unclear to me whether this is a question the SDK can answer this and if they plan to so in the future. In the existing state, the answer appears to be "no" to me. There is no restriction that BaseAccounts cannot be module managed accounts. ModuleAccounts themselves are intended to be statically defined with a module "name", thus it doesn't seem logical to me for a module to instantiate many accounts, especially at runtime, as that would break the existing `GetModuleAccount` API and potentially other uninvestigated issues. 

To bypass the lack of flexibility by the SDK, the request is to update all dependencies to set all module managed accounts as `ModuleAccount`s. This is an unsustainable approach in my opinion as you will need to constantly vet all dependencies to ensure this assumption is constantly met. As noted in the beginning of this post, the code doesn't account for interchain accounts either. Since this solution is a surface level patch, you will likely face several security incidences over time regarding this same issue  

For these reasons, I don't believe ibc-go should set its escrow accounts as `ModuleAccount`s, at least not now. I would recommend requesting the SDK to tighten up the restriction on what a user account is and isn't, in addition to creating stronger guidelines for how module managed accounts should be depicted, as I believe base accounts are perfectly fine for ICS 20 escrow accounts (maybe it could be enhanced with the restriction that only ics20 can send to/from the accounts)

--- 

Assuming genesis accounts you wish to recover funds from do not have a public key set, my recommendation as a short term hack would be to set the escrow address of every authorized channel. Escrow addresses are stateless and thus could be set into state at the same time you set the authorize channels. You can extend your authorized channel type to also set the escrow address, as you iterate through your authorized channels, ensure the escrow address does not match the recipient

crodriguezvega:
@GAtom22 Following up here after @colin-axner's latest comment: is there anything you would like to reply? Do you think we should still keep this PR open?

Issue #2813, opened 2022-11-22
e2e: add v6 -> v7 e2e upgrade tests


colin-axner:
## Summary

Add an e2e test which tests the upgrade from a v6 to v7 chain.

#### Pre-upgrade logic
- Create multiple tendermint clients, ideally update both at least once
- Create a channel, send a transfer and receive a transfer
- Create a solo machine client (no need to perform actions with the solo machine client)

#### Post upgrade logic
- Send a packet on the existing channel
- Query the tendermint clients
- Query the solo machine client
- Optional: send a MsgUpdate in reference to the solo machine client (proof can be invalid, mostly want to ensure the msg does not panic when referencing a solo machine clientID)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #2902, opened 2022-12-07
e2e: adding e2e upgrade test for ibc-go v7

charleenfei:
## Description

closes: #2813 

```bash
e2e: adding e2e upgrade test for ibc-go v7
```

---

write a little note why.

### DIFF
--- a/e2e/tests/upgrades/upgrade_test.go
+++ b/e2e/tests/upgrades/upgrade_test.go
@@ -9,6 +9,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	v7upgrades "github.com/cosmos/ibc-go/v6/testing/simapp/upgrades/v7"
 	v6upgrades "github.com/cosmos/interchain-accounts/app/upgrades/v6"
 	intertxtypes "github.com/cosmos/interchain-accounts/x/inter-tx/types"
 	"github.com/gogo/protobuf/proto"
@@ -401,41 +402,87 @@ func (s *UpgradeTestSuite) TestV6ToV7ChainUpgrade() {
 	// Optional: send a MsgUpdate in reference to the solo machine client (proof can be invalid, mostly want to ensure the msg does not panic when referencing a solo machine clientID)
 
 	t := s.T()
-	// testCfg := testconfig.FromEnv()
+	testCfg := testconfig.FromEnv()
 
 	ctx := context.Background()
-	relayer, _ := s.SetupChainsRelayerAndChannel(ctx)
+	relayer, channelA := s.SetupChainsRelayerAndChannel(ctx)
 	chainA, chainB := s.GetChains()
 
 	createClientOptions := ibc.CreateClientOptions{
 		TrustingPeriod: time.Duration(time.Hour * 24 * 7).String(),
 	}
 
-	// s.SetupClients(ctx, relayer, createClientOptions)
+	// create second tendermint client
 	s.SetupClients(ctx, relayer, createClientOptions)
 
-	status, err := s.Status(ctx, chainA, chainA.Config().ChainID)
+	status, err := s.Status(ctx, chainA, ibctesting.FirstClientID)
 	s.Require().NoError(err)
-	fmt.Println(status)
+	s.Require().Equal("Active", status)
 
-	// create separate user specifically for the upgrade proposal to more easily verify starting
-	// and end balances of the chainA users.
-	// chainAUpgradeProposalWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+	status, err = s.Status(ctx, chainA, "07-tendermint-1")
+	s.Require().NoError(err)
+	s.Require().Equal("Active", status)
+
+	var (
+		chainADenom    = chainA.Config().Denom
+		chainBIBCToken = testsuite.GetIBCToken(chainADenom, channelA.Counterparty.PortID, channelA.Counterparty.ChannelID) // IBC token sent to chainB
+
+		// chainBDenom    = chainB.Config().Denom
+		// chainAIBCToken = testsuite.GetIBCToken(chainBDenom, channelA.PortID, channelA.ChannelID) // IBC token sent to chainA
+	)
+
+	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+	chainAAddress := chainAWallet.Bech32Address(chainA.Config().Bech32Prefix)
+
+	chainBWallet := s.CreateUserOnChainB(ctx, testvalues.StartingTokenAmount)
+	chainBAddress := chainBWallet.Bech32Address(chainB.Config().Bech32Prefix)
 
 	s.Require().NoError(test.WaitForBlocks(ctx, 1, chainA, chainB), "failed to wait for blocks")
+
+	t.Run("native IBC token transfer from chainA to chainB, sender is source of tokens", func(t *testing.T) {
+		transferTxResp, err := s.Transfer(ctx, chainA, chainAWallet, channelA.PortID, channelA.ChannelID, testvalues.DefaultTransferAmount(chainADenom), chainAAddress, chainBAddress, s.GetTimeoutHeight(ctx, chainB), 0, "")
+		s.Require().NoError(err)
+		s.AssertValidTxResponse(transferTxResp)
+	})
+
+	t.Run("tokens are escrowed", func(t *testing.T) {
+		actualBalance, err := s.GetChainANativeBalance(ctx, chainAWallet)
+		s.Require().NoError(err)
+
+		expected := testvalues.StartingTokenAmount - testvalues.IBCTransferAmount
+		s.Require().Equal(expected, actualBalance)
+	})
 
 	t.Run("start relayer", func(t *testing.T) {
 		s.StartRelayer(relayer)
 	})
 
-	// t.Run("upgrade chainA", func(t *testing.T) {
-	// 	s.UpgradeChain(ctx, chainA, chainAUpgradeProposalWallet, v6upgrades.UpgradeName, testCfg.ChainAConfig.Tag, testCfg.UpgradeTag)
-	// })
+	t.Run("packets are relayed", func(t *testing.T) {
+		s.AssertPacketRelayed(ctx, chainA, channelA.PortID, channelA.ChannelID, 1)
 
-	// t.Run("restart relayer", func(t *testing.T) {
-	// 	s.StopRelayer(ctx, relayer)
-	// 	s.StartRelayer(relayer)
-	// })
+		actualBalance, err := chainB.GetBalance(ctx, chainBAddress, chainBIBCToken.IBCDenom())
+		s.Require().NoError(err)
+
+		expected := testvalues.IBCTransferAmount
+		s.Require().Equal(expected, actualBalance)
+	})
+
+	// create separate user specifically for the upgrade proposal to more easily verify starting
+	// and end balances of the chainA users.
+	chainAUpgradeProposalWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+
+	t.Run("upgrade chainA", func(t *testing.T) {
+		s.UpgradeChain(ctx, chainA, chainAUpgradeProposalWallet, v7upgrades.UpgradeName, testCfg.ChainAConfig.Tag, testCfg.UpgradeTag)
+	})
+
+	status, err = s.Status(ctx, chainA, ibctesting.FirstClientID)
+	s.Require().NoError(err)
+	s.Require().Equal("Active", status)
+
+	status, err = s.Status(ctx, chainA, "07-tendermint-1")
+	s.Require().NoError(err)
+	s.Require().Equal("Active", status)
+
 }
 
 // RegisterInterchainAccount will attempt to register an interchain account on the counterparty chain.
--- a/e2e/tests/upgrades/upgrade_test.go
+++ b/e2e/tests/upgrades/upgrade_test.go
@@ -9,7 +9,6 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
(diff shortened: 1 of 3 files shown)


e2e/tests/upgrades/upgrade_test.go
```diff
@@ -70,6 +73,8 @@ func (s *UpgradeTestSuite) UpgradeChain(ctx context.Context, chain *cosmos.Cosmo
 	err = chain.StartAllNodes(ctx)
 	s.Require().NoError(err, "error starting upgraded node(s)")
 
+	s.InitGRPCClients(chain)
```
chatton, line 76: maybe we can add a small comment saying why we're doing this?

testing/values.go
```diff
@@ -22,6 +22,8 @@ const (
 	FirstChannelID    = "channel-0"
 	FirstConnectionID = "connection-0"
 
+	SecondClientID = "07-tendermint-1"
```
chatton, line 25: now that solomachine clients are in the mix, what do you think about instead of this constant, we have a function like

```go
secondClientID := TendermintClientID(2)
```

e2e/testsuite/testsuite.go
```diff
@@ -363,7 +363,7 @@ func (s *E2ETestSuite) GetChainGRCPClients(chain ibc.Chain) GRPCClients {
 
 // initGRPCClients establishes GRPC clients with the given chain.
```
chatton, line 364: [nit] can we update the docstring too to indicate it's now exported?

e2e/testsuite/testsuite.go
```diff
@@ -76,8 +76,8 @@ type GRPCClients struct {
 	InterTxQueryClient intertxtypes.QueryClient
 
 	// SDK query clients
-	GovQueryClient   govtypesv1beta1.QueryClient
-	GovQueryClientV1 govtypesv1.QueryClient
+	GovQueryClient    govtypesv1beta1.QueryClient
+	GovQueryClientV1  govtypesv1.QueryClient
```
chatton, line 80: oh weird this kinda seems like we're not auto formatting the test code if linting was passing without the correct indentation.

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -405,3 +551,29 @@ func getICAVersion(chainAVersion, chainBVersion string) string {
...
+	})
+	if err != nil {
+		return "", err
+	}
+
+	return res.Status, nil
+}
+
+// ClientState queries the current ClientState by clientID
+func (s *UpgradeTestSuite) ClientState(ctx context.Context, chain ibc.Chain, clientID string) (*clienttypes.QueryClientStateResponse, error) {
```
chatton, line 569: fetching client state might be something we want to do in other tests, I'm happy to leave it here for now, but it might be worth moving this to the `E2ETestSuite` WDYT?

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -405,3 +551,29 @@ func getICAVersion(chainAVersion, chainBVersion string) string {
 	// explicitly set the version string because the host chain might not yet support incentivized channels.
 	return icatypes.NewDefaultMetadataString(ibctesting.FirstConnectionID, ibctesting.FirstConnectionID)
 }
+
+// Status queries the status of the client by clientID
+func (s *UpgradeTestSuite) Status(ctx context.Context, chain ibc.Chain, clientID string) (string, error) {
```
chatton, line 556: maybe this can be renamed to `ClientStatus` to stay consistent with the `ClientState` function?

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
...
+		s.Require().NoError(err)
+
+		expected := testvalues.IBCTransferAmount * 2
+		s.Require().Equal(expected, actualBalance)
+	})
+
+	// check that the v2 solo machine clientstate has been updated to the v3 solo machine clientstate
+	res, err = s.ClientState(ctx, chainA, soloMachineClientID)
+	s.Require().NoError(err)
+	s.Require().Equal(res.ClientState.TypeUrl, "/ibc.lightclients.solomachine.v3.ClientState")
```
chatton, line 532: is `"/ibc.lightclients.solomachine.v3.ClientState"` a constant somewhere we can use? Maybe we should create one if not.

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
...
+	s.Require().Equal("Active", status)
+
+	status, err = s.Status(ctx, chainA, ibctesting.SecondClientID)
+	s.Require().NoError(err)
+	s.Require().Equal("Active", status)
+
+	var (
+		chainADenom         = chainA.Config().Denom
+		chainBIBCToken      = testsuite.GetIBCToken(chainADenom, channelA.Counterparty.PortID, channelA.Counterparty.ChannelID) // IBC token sent to chainB
+		soloMachineClientID = "06-solomachine-2"
```
chatton, line 420: I left a comment later about a function to name client IDs, I think this is a good place to use it!

e2e/tests/upgrades/upgrade_test.go
colin-axner, line 76: just a note for the future, it might make sense to bundle these into a `restartChain` function. If we ever have another function which stops a chain, we will need to realize to initialize the grpc clients again. I'm very fine with this for now

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
...
+	)
+
+	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+	chainAAddress := chainAWallet.Bech32Address(chainA.Config().Bech32Prefix)
+
+	chainBWallet := s.CreateUserOnChainB(ctx, testvalues.StartingTokenAmount)
+	chainBAddress := chainBWallet.Bech32Address(chainB.Config().Bech32Prefix)
+
+	// create solo machine client
```
chatton, line 429: I don't think this comment is very useful. Maybe we can outline **why** we're using the solomachine type from `ibctesting` instead. i.e. we don't have a real implementation so we can use this to simulate it.

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
...
+	createClientOptions := ibc.CreateClientOptions{
+		TrustingPeriod: time.Duration(time.Hour * 24 * 7).String(),
+	}
+
+	// create second tendermint client
+	s.SetupClients(ctx, relayer, createClientOptions)
+
+	status, err := s.Status(ctx, chainA, ibctesting.FirstClientID)
+	s.Require().NoError(err)
+	s.Require().Equal("Active", status)
```
colin-axner, line 411: Could we use the ibcexported.Active string?

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -24,8 +24,11 @@ import (
 	"github.com/cosmos/ibc-go/e2e/testvalues"
 	controllertypes "github.com/cosmos/ibc-go/v6/modules/apps/27-interchain-accounts/controller/types"
 	icatypes "github.com/cosmos/ibc-go/v6/modules/apps/27-interchain-accounts/types"
+	v7 "github.com/cosmos/ibc-go/v6/modules/core/02-client/migrations/v7"
```
chatton, line 27: maybe the alias should be `v7migrations` instead of just `v7` to make it explicit in the tests. The same way we have `v7upgrades`

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
...
+func (s *UpgradeTestSuite) TestV6ToV7ChainUpgrade() {
+	t := s.T()
+	testCfg := testconfig.FromEnv()
+
+	ctx := context.Background()
+	relayer, channelA := s.SetupChainsRelayerAndChannel(ctx)
+	chainA, chainB := s.GetChains()
+
+	createClientOptions := ibc.CreateClientOptions{
+		TrustingPeriod: time.Duration(time.Hour * 24 * 7).String(),
```
colin-axner, line 403: Is there a DefaultTrustingPeriod we could make use of? Maybe in the testing package?

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
...
+		chainAWallet,
+		msgCreateSoloMachineClient,
+	)
+
+	s.AssertValidTxResponse(resp)
+	s.Require().NoError(err)
+
+	status, err = s.Status(ctx, chainA, soloMachineClientID)
+	s.Require().NoError(err)
+	s.Require().Equal("Active", status)
```
chatton, line 460: I think there is a constant we can use for `Active`

chatton, line 460: this seems like a good candidate for a subtest rather than being at the root level of the test

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
...
+
+	res, err := s.ClientState(ctx, chainA, soloMachineClientID)
+	s.Require().NoError(err)
+	s.Require().Equal(res.ClientState.TypeUrl, "/ibc.lightclients.solomachine.v2.ClientState")
```
chatton, line 464: another great candidate for t.Run !

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
...
+		Diversifier: solo.ConsensusState().Diversifier,
+		Timestamp:   solo.ConsensusState().Timestamp,
+	}
+
+	legacyClientState := &v7.ClientState{
+		Sequence:                 solo.ClientState().Sequence,
+		IsFrozen:                 solo.ClientState().IsFrozen,
+		ConsensusState:           legacyConsensusState,
+		AllowUpdateAfterProposal: true,
+	}
```
colin-axner, line 443: could we add a comment explaining why we need to construct the legacy types? That is, the ibc testing package uses the newest version of the solo machine which isn't supported by the chain we are broadcasting the message to, so we need to construct the legacy type in order to setup the pre upgraded chain with legacy state

chatton (commented):
Great work so far! 🥇

e2e/tests/upgrades/upgrade_test.go
colin-axner, line 464: I think these should be swapped? Usually the expected value comes first. We could potentially use v7.ClientState{}.TypeURL()? (not sure if that compiles? or how to access the type url given a concrete type, might need to be casted to a proto.Message). I'm okay using a string but then we should make a const at the top of this file for v2 and v3 and state that the upgrade from v6 to v7 migrates the solo machine from v2 to v3 (assume the test reader has no context on the migrations performed)

e2e/tests/upgrades/upgrade_test.go
colin-axner, line 569: I believe this function exists in `testsuite/grpc_query.go`

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -386,6 +391,147 @@ func (s *UpgradeTestSuite) TestV5ToV6ChainUpgrade() {
 	})
 }
 
+func (s *UpgradeTestSuite) TestV6ToV7ChainUpgrade() {
```
colin-axner, line 394: It might help to add a godoc asserting what we are testing. Maybe something like:

```go
// TestV6ToV7ChainUpgrade will test that an upgrade from a v6 ibc-go binary to a v7 ibc-go binary is successful
// and that the automatic migrations associated with the 02-client module are performed. Namely that the solo machine
// proto definition is migrated in state from the v2 to v3 definition. This is checked by creating a solo machine client
// before the upgrade and asserting that its TypeURL has been changed after the upgrade. The test also ensure packets
// can be sent before and after the upgrade without issue
```

colin-axner (approved):
Excellent work! The test looks great to me. Left code improvement suggestions. I suppose this pr needs to wait to bump ibctest? Maybe we do that in a separate pr first?

testing/values.go
colin-axner: I like this idea!

e2e/tests/upgrades/upgrade_test.go
damiannolan: Normally we could use [`MsgTypeURL`](https://github.com/cosmos/cosmos-sdk/blob/main/types/tx_msg.go#L83) from the sdk and pass a new/empty default instance of the type, but this won't work I think because `ClientState` doesn't implement the sdk.Msg interface. 

You could use what the impl of that function does tho, something like:
```
fmt.Sprint("/", proto.MessageName(&solomachine.ClientState{}))
```

e2e/tests/upgrades/upgrade_test.go
charleenfei: i could move this one and Status as well, I think they are generic enough functions

charleenfei: so I am leaving this one, bc the function in grpc query unpacks the client state there which runs into the issue that the ClientState cannot be Unpacked bc it is trying to unpack it from a legacy type which is unsupported

e2e/testvalues/values.go
```diff
@@ -34,3 +36,11 @@ func DefaultFee(denom string) feetypes.Fee {
 func DefaultTransferAmount(denom string) sdk.Coin {
 	return sdk.Coin{Denom: denom, Amount: sdk.NewInt(IBCTransferAmount)}
 }
+
+func TendermintClientID(id int) string {
+	return fmt.Sprint("07-tendermint-" + strconv.Itoa(id))
```
chatton, line 41: [super nit] fmt.Sprintf handles formatting without doing it ourselves.


`return fmt.Sprintf("07-tendermint-%d", id)`

chatton (approved):
LGTM! Left a few tiny nits / suggestions, great work on this test!

Issue #2880, opened 2022-12-05
Replacing the concrete keeper structs with interfaces in IBC keeper


SebastianElvis:
## Summary

This issue proposes to replace the concrete keeper structs with interfaces in the IBC keeper, in order to allow IBC-enabled applications to add extra functionalities atop these keepers.

## Problem Definition

The IBC keeper struct has fields of [concrete types of other keepers](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/keeper.go#L31-L34), thus does not allow to drop in any other structs that implement the same set of functions.

This rules out the possibility for IBC-enabled applications to add extra functionalities atop these keepers (without forking IBC-Go). For example, possible extensions of these keepers include
- adding hooks to these keepers, so that other modules can learn internal states of IBC
- defining and emitting new events upon certain state changes in the IBC module
- implementing new queries and messages for these internal keepers

In addition, the implementation also differs from the practice in Cosmos SDK, where each module contains a set of "expected keeper" interfaces, and the module keeper implements these expected keeper interfaces (e.g., [in staking keeper](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/types/expected_keepers.go)).

## Proposal

This issue proposes to replace the concrete keeper structs with interfaces in the IBC keeper. Specifically, one possible and arguably least intrusive approach is as follows
- for each of the {client, connection, channel, port} keepers, add a file `exported/exported.go`. The file provides an interface that includes all public functions of the concrete keeper implementation.
- replace the concrete keeper structs with these interfaces [in the IBC keeper definition](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/keeper.go#L31-L34).

This shall not break any existing functionality or unit tests, and is a practice as done in Cosmos SDK (e.g., [staking keeper's exported interface](https://github.com/cosmos/cosmos-sdk/blob/main/x/staking/exported/exported.go)).

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Hi! Thanks for opening an issue! It was well written and easy to follow. Apologies for the time delay in following up.

[...]

PR #2967, opened 2022-12-22
feat: replacing concrete client keeper struct with an interface

SebastianElvis:
## Description

This PR replaces the concrete `ClientKeeper` struct with an interface. This allows IBC-enabled applications to add extra functionalities atop the client keeper (without forking IBC-Go), including

- adding hooks to these keepers, so that other modules can learn internal states of IBC
- defining and emitting new events upon certain state changes in the IBC module
- implementing new queries and messages for these internal keepers

closes: #2880 

---

write a little note why.

### DIFF
--- a/modules/core/02-client/exported/exported.go
+++ b/modules/core/02-client/exported/exported.go
@@ -1,0 +1,68 @@
+package exported
+
+import (
+	"context"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	clienttypes "github.com/cosmos/ibc-go/v6/modules/core/02-client/types"
+	"github.com/cosmos/ibc-go/v6/modules/core/exported"
+)
+
+// ClientKeeper defines the expected interface of the light client used for IBC
+type ClientKeeper interface {
+	CreateClient(ctx sdk.Context, clientState exported.ClientState, consensusState exported.ConsensusState) (string, error)
+	UpdateClient(ctx sdk.Context, clientID string, clientMsg exported.ClientMessage) error
+	UpgradeClient(ctx sdk.Context, clientID string, upgradedClient exported.ClientState, upgradedConsState exported.ConsensusState, proofUpgradeClient, proofUpgradeConsState []byte) error
+	GenerateClientIdentifier(ctx sdk.Context, clientType string) string
+	SetClientState(ctx sdk.Context, clientID string, clientState exported.ClientState)
+	GetClientState(ctx sdk.Context, clientID string) (exported.ClientState, bool)
+	IterateClientStates(ctx sdk.Context, prefix []byte, cb func(string, exported.ClientState) bool)
+	GetClientConsensusState(ctx sdk.Context, clientID string, height exported.Height) (exported.ConsensusState, bool)
+	SetClientConsensusState(ctx sdk.Context, clientID string, height exported.Height, consensusState exported.ConsensusState)
+	GetNextClientSequence(ctx sdk.Context) uint64
+	SetNextClientSequence(ctx sdk.Context, sequence uint64)
+	IterateConsensusStates(ctx sdk.Context, cb func(clientID string, cs clienttypes.ConsensusStateWithHeight) bool)
+	GetAllGenesisClients(ctx sdk.Context) clienttypes.IdentifiedClientStates
+	GetAllClientMetadata(ctx sdk.Context, genClients []clienttypes.IdentifiedClientState) ([]clienttypes.IdentifiedGenesisMetadata, error)
+	SetAllClientMetadata(ctx sdk.Context, genMetadata []clienttypes.IdentifiedGenesisMetadata)
+	GetAllConsensusStates(ctx sdk.Context) clienttypes.ClientsConsensusStates
+	HasClientConsensusState(ctx sdk.Context, clientID string, height exported.Height) bool
+	GetLatestClientConsensusState(ctx sdk.Context, clientID string) (exported.ConsensusState, bool)
+	GetSelfConsensusState(ctx sdk.Context, height exported.Height) (exported.ConsensusState, error)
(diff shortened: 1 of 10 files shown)


crodriguezvega:
Thanks for this contribution, @SebastianElvis! We will discuss and review in the coming weeks. 

Do you already have a concrete use case to use this? Do you have any urgency to have this released soon?

SebastianElvis:
Thanks for the response!

> Do you already have a concrete use case to use this?

Yep, we indeed have a concrete use case -- adding hooks over `ClientKeeper` for checkpointing light client headers.

Our project [Babylon](https://babylonchain.io/) aims at building a blockchain Babylon that checkpoints other PoS chains to Bitcoin. The checkpointing process from PoS chains to Babylon leverages the IBC light client, where Babylon receives PoS chain's headers via IBC. In order to do the checkpointing, Babylon has to use hooks to get notification at the application layer upon receiving headers from IBC.

At the moment, we are maintaining [a fork of IBC-Go](https://github.com/babylonchain/ibc-go) that implements a hook [AfterHeaderWithValidCommit](https://github.com/babylonchain/ibc-go/blob/babylon-v5.0.0/modules/core/02-client/keeper/extended_keeper_hooks.go#L33-L37). The hook is invoked upon a header with a valid quorum certificate from IBC. The most code of Babylon is still WIP and we hope to open-source everything as soon as possible.

> Do you have any urgency to have this released soon?

Yeah we definitely hope to have this feature in the official IBC-Go implementation. Nevertheless, we respect your priorities. Before having this feature in IBC-Go we can still use our fork :) Thanks for asking.

testing/simapp/app.go
```diff
@@ -893,13 +894,18 @@ func (app *SimApp) setupUpgradeHandlers() {
...
 	)
 
+	clientKeeper, ok := app.IBCKeeper.ClientKeeper.(clientkeeper.Keeper)
+	if !ok {
+		panic("failed to assert app.IBCKeeper.ClientKeeper into type clientkeeper.Keeper")
+	}
+
 	app.UpgradeKeeper.SetUpgradeHandler(
 		v7.UpgradeName,
 		v7.CreateUpgradeHandler(
```
colin-axner, line 904: same with this one

modules/core/02-client/exported/exported.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
+	clienttypes "github.com/cosmos/ibc-go/v6/modules/core/02-client/types"
+	"github.com/cosmos/ibc-go/v6/modules/core/exported"
+)
+
+// ClientKeeper defines the expected interface of the light client used for IBC
```
colin-axner, line 12: Maybe we could add some short documentation explaining why this interface exists? That is to extend the ClientKeeper implementation (the ibc module client keeper implementation should not have its logic modified, only extended)

colin-axner: Maybe we could add additional documentation [here](https://github.com/cosmos/ibc-go/blob/main/docs/ibc/integration.md), via a section "Extending the client keeper"

colin-axner (commented):
Thanks @SebastianElvis! The changes look good! Left some minor comments, let me know if you have any questions

SebastianElvis:
Thanks @colin-axner for the feedback, which is really helpful! I have rebased the PR to the latest `main` branch, and fixed the comments. Specifically, this PR 

- changes the signatures of `NewMigrator`, `PruneExpiredConsensusStates`, and `SetUpgradeHandler`
- removes unnecessary type assertions given the updated function signatures
- adds documentations for `clientexported.ClientKeeper` and integration

Please feel free to have a look again.

chatton:
Hi @SebastianElvis 

Thanks for all the work you've been doing here, looking at the code, I am a little concerned about the size of the interface that ends up being required here, and a lot of these interface functions are actually implementation details (un/marshalling functions, client id generation etc. )

I was investigating an [alternative approach](https://github.com/cosmos/ibc-go/pull/3214) which would allow custom logic to be executed at different client events (creation/upgrading/updating) through the use of hooks. I believe this should be enough to enable any sort of custom logic that is required while also making sure that the client keeper logic is guaranteed to execute as normal.

Please let me know what you think of this approach and if it is enough to implement the logic you want to add.

cc @colin-axner

SebastianElvis:
Hi @chatton , thanks for the feedback!

> Thanks for all the work you've been doing here, looking at the code, I am a little concerned about the size of the interface that ends up being required here, and a lot of these interface functions are actually implementation details (un/marshalling functions, client id generation etc. )

Yep I agree with that. It's indeed not good to have an interface that includes a number of functions that concern implementation details.

> I was investigating an [alternative approach](https://github.com/cosmos/ibc-go/pull/3214) which would allow custom logic to be executed at different client events (creation/upgrading/updating) through the use of hooks. I believe this should be enough to enable any sort of custom logic that is required while also making sure that the client keeper logic is guaranteed to execute as normal.

Thanks for your investigation and work. I believe adding relevant hooks directly to the client keeper suffices for our use cases and is much cleaner than the interface approach.

One thing is that our use case needs to distinguish between a header in the canonical chain and in the fork (i.e., has a valid quorum certificate but has non-monotonic timestamps or conflicting header). What we did is to provide the hook with a boolean value which indicates whether the header is canonical or fork (see [here](https://github.com/babylonchain/ibc-go/blob/generalize-babylon-callback/modules/core/02-client/keeper/extended_keeper.go#L82) and [here](https://github.com/babylonchain/ibc-go/blob/generalize-babylon-callback/modules/core/02-client/keeper/extended_keeper.go#L110)). This is difficult to be done in an external module subscribing to client keeper hooks since the external module does not have access to `clientState`s. Is it possible for you to implement something similar? Thanks!

crodriguezvega:
Closing for now, since the use case can be accomplished with SDK post handlers.

Issue #2007, opened 2022-08-15
Move non-verification misbehaviour checks to CheckForMisbehaviour


colin-axner:
## Summary

In 07-tendermint, we are checking whether something is misbehaviour in `VerifyClientMessage`, we should instead be doing this in `CheckForMisbehaviour`. Move this batch of [code](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/misbehaviour_handle.go#L26-#L46) to `CheckForMisbehaviour

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

Daniyal98:
Hello, I would like to start working on this issue. Thank you!

charleenfei:
@Daniyal98 are you still working on this?

Daniyal98:
@charleenfei It wasn't assigned to me so I didn't start working on it

charleenfei:
sorry about that! i am assigning you now :)

colin-axner:
Apologies for the delayed reply @Daniyal98. Will you be able to work on this issue now, or is it not the right time to pickup the issue? This issue is apart of our next iteration (Nov 28 - Dec 11) and we would like to complete it soon

Daniyal98:
@colin-axner yeah sure. I will try my best :)

Daniyal98:
@colin-axner the statement of this issue seems a bit misleading. The part that you mentioned is already in CheckForMisbehaviour. Am I missing something?

colin-axner:
Hi @Daniyal98 apologies again for the late reply! When I made the issue, I linked to main, but the code got moved around. The part of the code that should be moved is [here](https://github.com/cosmos/ibc-go/blob/7d60efa07b7d564e5d4f531a0398d3f670fc973f/modules/light-clients/07-tendermint/misbehaviour_handle.go#L71-#L93), it should be moved into the switch statement case `Misbehaviour` in `CheckForMisbehaviour`

The idea is that when we check for misbehaviour, we have three cases to check for:
- if it is a header, check if we already had a valid update for this height (and that the update isn't a duplicate)
- if it is misbehaviour, check for time misbehaviour (produced two headers which disobey BFT time restrictions
- if it is misbheaviour, check for fork misbehaviour (two headers at the same height)

[...]

PR #2971, opened 2022-12-22
moved non-verification misbehaviour checks to checkForMisbehaviour

Daniyal98:
## Description

closes: #2007

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/light-clients/07-tendermint/misbehaviour_handle.go
+++ b/modules/light-clients/07-tendermint/misbehaviour_handle.go
@@ -55,13 +55,13 @@ func (cs ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCode
 		// Thus, here we can return true, as ClientMessage is of type Misbehaviour
 		// if heights are equal check that this is valid misbehaviour of a fork
 		// otherwise if heights are unequal check that this is valid misbehavior of BFT time violation
-		if misbehaviour.Header1.GetHeight().EQ(misbehaviour.Header2.GetHeight()) {
-			blockID1, err := tmtypes.BlockIDFromProto(&misbehaviour.Header1.SignedHeader.Commit.BlockID)
+		if msg.Header1.GetHeight().EQ(msg.Header2.GetHeight()) {
+			blockID1, err := tmtypes.BlockIDFromProto(&msg.Header1.SignedHeader.Commit.BlockID)
 			if err != nil {
 				return true
 			}
 
-			blockID2, err := tmtypes.BlockIDFromProto(&misbehaviour.Header2.SignedHeader.Commit.BlockID)
+			blockID2, err := tmtypes.BlockIDFromProto(&msg.Header2.SignedHeader.Commit.BlockID)
 			if err != nil {
 				return true
 			}
@@ -71,7 +71,7 @@ func (cs ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCode
 				return true
 			}
 
-		} else if misbehaviour.Header1.SignedHeader.Header.Time.After(misbehaviour.Header2.SignedHeader.Header.Time) {
+		} else if msg.Header1.SignedHeader.Header.Time.After(msg.Header2.SignedHeader.Header.Time) {
 			// Header1 is at greater height than Header2, therefore Header1 time must be less than or equal to
 			// Header2 time in order to be valid misbehaviour (violation of monotonic time).
 			return true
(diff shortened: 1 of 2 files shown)


charleenfei:
hey @Daniyal98, the pr looks ok but there are some failing tests in the 07 tendermint client, i think it's because the test is checking to see if `VerifyClientMessage` will catch the valid misbehaviour cases but since the logic was moved out of that function it is no longer erroring as expected. We should make sure that the test `CheckForMisbehaviour` function catches these cases :)

modules/light-clients/07-tendermint/misbehaviour_handle.go
```diff
@@ -53,6 +53,29 @@ func (cs ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCode
...
+		if msg.Header1.GetHeight().EQ(msg.Header2.GetHeight()) {
+			blockID1, err := tmtypes.BlockIDFromProto(&msg.Header1.SignedHeader.Commit.BlockID)
+			if err != nil {
+				return true
+			}
+
+			blockID2, err := tmtypes.BlockIDFromProto(&msg.Header2.SignedHeader.Commit.BlockID)
+			if err != nil {
+				return true
+			}
```
damiannolan, line 67: Given the method signature only returns a `bool` I think it would be more appropriate to return false in both of these conditionals. In reality, `BlockIDFromProto` should never return errors here as its checked in [`ValidateBasic`](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/misbehaviour.go#L91-L98) before this code would be reached.

The alternative option is to `panic` given its unreachable code, but I think I would rather opt for returning false.

modules/light-clients/07-tendermint/misbehaviour_handle.go
```diff
@@ -53,6 +53,29 @@ func (cs ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCode
...
+
+			// Ensure that Commit Hashes are different
+			if bytes.Equal(blockID1.Hash, blockID2.Hash) {
+				return true
+			}
```
damiannolan, line 72: If both headers are at the same height then the hashes must be different in order to be valid fork misbehaviour.

```suggestion
			if !bytes.Equal(blockID1.Hash, blockID2.Hash) {
				return true
			}
```

modules/light-clients/07-tendermint/misbehaviour_handle.go
```diff
@@ -53,6 +53,29 @@ func (cs ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCode
...
+
+		} else if msg.Header1.SignedHeader.Header.Time.After(msg.Header2.SignedHeader.Header.Time) {
+			// Header1 is at greater height than Header2, therefore Header1 time must be less than or equal to
+			// Header2 time in order to be valid misbehaviour (violation of monotonic time).
+			return true
```
damiannolan, line 77:
```suggestion
		} else if !msg.Header1.SignedHeader.Header.Time.After(msg.Header2.SignedHeader.Header.Time) {
			// Header1 is at greater height than Header2, therefore Header1 time must be less than or equal to
			// Header2 time in order to be valid misbehaviour (violation of monotonic time).
			return true
```

[`ValidateBasic` ensures that header1 height is greater than or equal to header2 height](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/misbehaviour.go#L86-L89). Therefore, header1 time must be after header2. If it is not, then we violate time monotonicity and should return true as this is considered misbehaviour.

modules/light-clients/07-tendermint/misbehaviour_handle.go
```diff
@@ -53,6 +53,29 @@ func (cs ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCode
...
+		}
 		return true
```
damiannolan, line 79: If we do not hit a case for valid misbehaviour above we should return false

damiannolan (commented):
Thanks for getting this started @Daniyal98! 

Left some comments regarding the logic as we are now expecting to return a boolean value rather than an error case. I think some of the test cases would also need to be refactored to accommodate the restructuring of the code.

damiannolan:
Hey @Daniyal98, if its okay for you we can push some updates to address some of the comments here if you don't have the capacity.

charleenfei:
hey @Daniyal98 it looks like there are still failing tests -- would you mind taking a look and fixing? or let us know if we should take over this pr. thank you!

Daniyal98:
Hey @charleenfei, I am unable to find the time nowadays to work on this. Could you please take over this PR? Thank you

damiannolan:
> Hey @charleenfei, I am unable to find the time nowadays to work on this. Could you please take over this PR? Thank you

Thanks for letting us know @Daniyal98, we can take this over. Thanks for the work you've done on it, much appreciated! ❤️

charleenfei:
closing in favour of #3046

Issue #2961, opened 2022-12-20
authz/ics20: unit tests for `Accept` method of `TransferAuthorization`


crodriguezvega:
Refactor [existing test](https://github.com/cosmos/ibc-go/blob/feat/authz-ics20/modules/apps/transfer/types/transfer_authz_test.go) and convert it into table tests covering the following scenarios:

- WHEN `TransferAuthorization` contains 1 allocation AND source port ID/channel ID in `MsgTransfer` matches the source port and channel of the allocation AND the recipient address in `MsgTransfer` is found in the `allow_list`:
  - AND the amount in `MsgTransfer` is greater than the available funds, THEN error is returned.
  - AND the amount in `MsgTransfer` is equal to available funds, THEN the updated `TransferAuthorization` is `nil`.
  - AND the amount in `MsgTransfer` is smaller than the available funds, THEN the spend limit in the updated `TransferAuthorization` is reduced by the amount in `MsgTransfer`.

- WHEN `TransferAuthorization` contains 1 allocation AND the source port ID/channel ID in combination `MsgTransfer` does not match the source port and channel of the allocation, THEN an error is returned.

- WHEN `TransferAuthorization` contains 1 allocation AND source port ID/channel ID in `MsgTransfer` matches the source port and channel of the allocation AND the `allow_list` is empty AND the amount in `MsgTransfer` is smaller than the available funds, THEN the spend limit in the updated `TransferAuthorization` is reduced by the amount in `MsgTransfer`.

- WHEN `TransferAuthorization` contains 1 allocation AND the recipient address is not in found in the `allow_list`, THEN an error is returned.

~- WHEN `TransferAuthorization` contains 1 allocation with empty source port and source channel AND the recipient address in `MsgTransfer` is found in the `allow_list` AND the amount in `MsgTransfer` is smaller than the available funds, then the spend limit in the updated `TransferAuthorization` is reduced by the amount in `MsgTransfer`.~
  
- WHEN `TransferAuthorization` contains 2 allocations AND source port ID/channel ID in `MsgTransfer` matches the source port and channel of one of the allocations AND the recipient address in `MsgTransfer` is found in the `allow_list`:
  - AND the amount in `MsgTransfer` is greater than the available funds for one allocation, THEN an error is returned.
  - AND the amount in `MsgTransfer` is equal to available funds for one allocation, THEN the updated `TransferAuthorization` contains only the other allocation.
[...]

PR #2998, opened 2023-01-11
chore: refactor transfer authz tests to use table tests

damiannolan:
## Description

- Renaming `ack_test.go` -> `types_test.go`
  - no acknowledgement tests previously existed, only the test suite setup funcs
- Refactor transfer authz `Accept` tests
- Added test for transfer authz `MsgTypeURL`

closes: #2961 

```bash
chore: refactor transfer authz tests to use table tests
```

---

write a little note why.


modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -1,132 +1,161 @@
...
-	}
-	authorization = types.NewTransferAuthorization(allocation)
-	transfer = types.NewMsgTransfer(sourcePort2, sourceChannel2, coin500, fromAddr.String(), toAddr.String(), timeoutHeight, 0, "")
-	_, err = authorization.Accept(ctx, transfer)
-	require.Error(t, err)
+	testCases := []struct {
+		name      string
+		malleate  func()
+		expPass   bool
+		expResult func(res authz.AcceptResponse)
```
chatton, line 22: [nit] the name `expResult` to me indicates that a function should return the result which will later be used in some sort of assertion, but this function is actually doing the assertions. Maybe `assertResult` or similar is more appropriate WDYT?

chatton: As a follow up, if this function is renamed, we no longer really need the boolean `expPass` as that assertion could be done in the function too, in this case maybe the method signature could change to `func(res authz.AcceptResponse, err error)`

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -1,132 +1,161 @@
...
-		SpendLimit:    coins1000,
-		AllowList:     []string{toAddr.String()},
-	}
-	authorization = types.NewTransferAuthorization(allocation)
-	transfer = types.NewMsgTransfer(sourcePort2, sourceChannel2, coin500, fromAddr.String(), toAddr.String(), timeoutHeight, 0, "")
-	_, err = authorization.Accept(ctx, transfer)
-	require.Error(t, err)
+	testCases := []struct {
+		name      string
+		malleate  func()
```
chatton, line 20: [nit] for the malleate function, we use `func()` in almost all our tests I think, but what do you think about limiting the scope of what we can change and removing the test suite level variables 

```golang
var (
		msgTransfer   types.MsgTransfer
		transferAuthz types.TransferAuthorization
	)
```

altogether?

Something like

```golang
malleate func(msgTransfer types.MsgTransfer, transferAuthz types.TransferAuthorization)
```

I think this makes the intent far more clear than when no arguments are supplied and eliminates the potential for one test to accidentally effect another.

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -1,132 +1,161 @@
...
-	transfer = types.NewMsgTransfer(sourcePort, sourceChannel, coin500, fromAddr.String(), toAddr.String(), timeoutHeight, 0, "")
-	_, err = authorization.Accept(ctx, transfer)
-	require.Error(t, err)
 }
 
-func TestTransferAuthorizationValidateBasic(t *testing.T) {
+func (suite *TypesTestSuite) TestTransferAuthorizationMsgTypeURL() {
+	var transferAuthz types.TransferAuthorization
+	suite.Require().Equal(sdk.MsgTypeURL(&types.MsgTransfer{}), transferAuthz.MsgTypeURL(), "invalid type url for transfer authorization")
+}
```
chatton, line 156: nice addtion!

chatton (approved):
LGTM Looking great, I like the look of the new table tests. Left some suggestions, will leave it to you if you think they are worth implementing!

modules/apps/transfer/types/transfer_authorization_test.go
damiannolan: I've applied the suggested changes here, but I think I'll leave the malleate func as is for now. I agree it would be good to limit the scope of what can be mutated in the malleate function but for now I'll just keep it as is to be consistent with other tests

Issue #2935, opened 2022-12-13
Paginating over ClientStates causes endless requests


webmaster128:
## Summary of Bug

I am trying to use the `ClientStates` query with pagination. However, most result pages contain no elements and nextKey is always non-empty.

## Expected Behaviour

Each page of a `ClientStates` query contains dozens of results by default and after a few pages, the pagination `nextKey` is unset.

## Version

wasmd 0.29.0-rc2
cosmos-sdk 0.45.8
[ibc-go 3.3.0](https://github.com/CosmWasm/wasmd/blob/v0.29.0-rc2/go.mod#L9)

## Steps to Reproduce

I query against nois-testnet-003, a network with 48 clients (https://ibc.nois.network/connections). When I do that in CosmJS, I get the following pages. My script stopped after 60 seconds of requests.

```
Page 1, response count 1, next key: /07-tendermint-0/consensusStates/1-1422008/processedTime
Page 2, response count 0, next key: /07-tendermint-0/consensusStates/1-1422337
Page 3, response count 0, next key: /07-tendermint-0/consensusStates/1-1422584/processedHeight
Page 4, response count 0, next key: /07-tendermint-0/consensusStates/1-1422771/processedTime
Page 5, response count 0, next key: /07-tendermint-0/consensusStates/1-1423200
Page 6, response count 0, next key: /07-tendermint-0/consensusStates/1-1423446/processedHeight
Page 7, response count 0, next key: /07-tendermint-0/consensusStates/1-1423606/processedTime
Page 8, response count 0, next key: /07-tendermint-0/consensusStates/1-1423812
Page 9, response count 0, next key: /07-tendermint-0/consensusStates/1-1424017/processedHeight
Page 10, response count 0, next key: /07-tendermint-0/consensusStates/1-1424206/processedTime
```

Full output with 816 pages: [test.log](https://github.com/cosmos/ibc-go/files/10222440/test.log).

This output is generated from a CosmJS test: https://github.com/cosmos/cosmjs/pull/1352/files.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

webmaster128:
In [a different test case](https://github.com/confio/ts-relayer/pull/262) performing the same query it looks better:

```
# Beginning of the test
Page 1, response count: 7, next key: /07-tendermint-6/consensusStates/0-621
Page 2, response count: 3, next key: 

# Client is added here

[...]

PR #3010, opened 2023-01-13
fix: update `Paginate` to use `FilterPaginate` in `ClientStates` and `ConnectionChannels` grpc queries

charleenfei:
## Description

closes: #2935 

```bash
fix: update `Paginate` to use `FilterPaginate` in `ClientStates` and `ConnectionChannels` grpc queries
```

---

write a little note why.


modules/core/02-client/keeper/grpc_query.go
```diff
@@ -63,25 +63,26 @@ func (q Keeper) ClientStates(c context.Context, req *types.QueryClientStatesRequ
 	clientStates := types.IdentifiedClientStates{}
 	store := prefix.NewStore(ctx.KVStore(q.storeKey), host.KeyClientStorePrefix)
 
-	pageRes, err := query.Paginate(store, req.Pagination, func(key, value []byte) error {
+	pageRes, err := query.FilteredPaginate(store, req.Pagination, func(key, value []byte, accumulate bool) (bool, error) {
```
crodriguezvega, line 66: Would it be possible to modify [the existing test](https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/keeper/grpc_query_test.go#L131-L154) to check for the `total` value in the `pagination` field of the response (i.e. `total` doesn't account for `consensusStates`)?

Maybe could also add another test with two clients that sets `limit` to 1 and checks that the `nextKey` in `pagination` points to a key with the format `/<client-id>/clientState`. But I understand that maybe this is a bit too much, and anyway maybe it's testing functionality from `Paginate`, not really our gRPC endpoint, so we can also skip it.

modules/core/04-channel/keeper/grpc_query.go
```diff
@@ -96,26 +96,27 @@ func (q Keeper) ConnectionChannels(c context.Context, req *types.QueryConnection
 	channels := []*types.IdentifiedChannel{}
 	store := prefix.NewStore(ctx.KVStore(q.storeKey), []byte(host.KeyChannelEndPrefix))
 
-	pageRes, err := query.Paginate(store, req.Pagination, func(key, value []byte) error {
+	pageRes, err := query.FilteredPaginate(store, req.Pagination, func(key, value []byte, accumulate bool) (bool, error) {
```
crodriguezvega, line 99: Same comment: maybe it's possible to modify one of the existing [tests](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/grpc_query_test.go#L203)?

crodriguezvega (commented):
Thank you for fixing this, @charleenfei!

damiannolan (commented):
Looks great, agree with @crodriguezvega, a couple of test cases before we merge would be awesome tho! 
Thanks for picking this up! :)

modules/core/02-client/keeper/grpc_query_test.go
```diff
@@ -377,11 +387,11 @@ func (suite *KeeperTestSuite) TestQueryConsensusStates() {
 			if tc.expPass {
 				suite.Require().NoError(err)
 				suite.Require().NotNil(res)
+				suite.Require().Equal(tc.expectedConsensusStates, res.Pagination.Total)
 				suite.Require().Equal(len(expConsensusStates), len(res.ConsensusStates))
```
damiannolan, line 391: Aren't these checks essentially the same?

charleenfei: true actually, i guess this test was passing bc we were reconstructing the situation and did not have other items in the query resp, whereas in real life without `FilterPaginate` other items were being returned in addition.

charleenfei: i can add some other items maybe to the update

modules/core/04-channel/keeper/grpc_query_test.go
```diff
@@ -193,6 +197,7 @@ func (suite *KeeperTestSuite) TestQueryChannels() {
 			if tc.expPass {
 				suite.Require().NoError(err)
 				suite.Require().NotNil(res)
+				suite.Require().Equal(tc.expectedChannels, res.Pagination.Total)
 				suite.Require().Equal(expChannels, res.Channels)
 				suite.Require().Equal(len(expChannels), int(res.Pagination.Total))
```
damiannolan, line 202: Sorry for slowing this down but I think this pretty much the same check as below too, right?
`suite.Require().Equal(len(expChannels), int(res.Pagination.Total))`

We can probably remove the `tc.expectedChannels`.

I think as long as the scenario from the original issue has been tested and covered then I'm happy to get this merged.

crodriguezvega: Yeah, I agree with @damiannolan and I think we can remove `tc.expectedChannels`.

modules/core/02-client/keeper/grpc_query_test.go
```diff
@@ -328,6 +328,9 @@ func (suite *KeeperTestSuite) TestQueryConsensusStates() {
 				path := ibctesting.NewPath(suite.chainA, suite.chainB)
 				suite.coordinator.SetupClients(path)
 
+				path1 := ibctesting.NewPath(suite.chainA, suite.chainB)
```
crodriguezvega, line 331: Do we need this extra path in the test?

modules/core/02-client/keeper/grpc_query_test.go
```diff
@@ -163,10 +163,10 @@ func (suite *KeeperTestSuite) TestQueryClientStates() {
 			ctx := sdk.WrapSDKContext(suite.chainA.GetContext())
 
 			res, err := suite.chainA.QueryServer.ClientStates(ctx, req)
-
 			if tc.expPass {
 				suite.Require().NoError(err)
 				suite.Require().NotNil(res)
+				suite.Require().Equal(uint64(len(expClientStates)), res.Pagination.Total)
```
crodriguezvega, line 169: Really a super nit, but my consistency freak would write the check the same way as in this [other test](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/grpc_query_test.go#L197). And maybe also put it as the last one (after `suite.Require().Equal(expClientStates.Sort(), res.ClientStates)`).

```suggestion
				suite.Require().Equal(len(expClientStates)), int(res.Pagination.Total))
```

crodriguezvega (approved):

Issue #2664, opened 2022-11-02
Make a state entry to find total amount of native assets IBC'd out


ValarDragon:
## Summary

Currently every channel has its own escrow bank account, which is a great design! So its very easy to see the amount of native assets IBC'd in or out on any IBC channel. 

However, there are use cases that would like us to be able to reason about the total amount of a native asset that has been IBC'd out. 

So if there are say three channels from Osmosis to Juno:
```
channel-1: 10 Osmo -> Juno
channel-2: 20 Osmo -> Juno
channel-3: 30 Osmo -> Juno
```
I would like to be find that its been `60 osmo` IBC'd out, with no iteration.

## Problem Definition

One such use case is rate limiting, where we would like to limit the amount of osmo in total that can flow out of osmosis in a given time interval. For IBC"d in assets, this is quite easy as all incoming assets for it on are on the same channel, so we can easily find "total IBC'd to Osmosis".

For native assets though, there is no efficient way to get the total amount IBC'd out. We'd like to make our rate limits reason about this, rather than the total supply of osmo on Osmosis. (As much of the Osmo should be illiquid due to staking and LP'ing) We can't iterate over all channels, as there are thousands of these on Osmosis.

## Proposal

* Add a new state entry for `total_native_ibc_out/{denom}` which then stores the amount IBC'd out across all channels.
* Add a query that allows us to find this value quickly
* Add a safety check if this value is ever attempted to be serialized to a negative value

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
I'm in favor. You technically could do this as a middleware if it was needed tomorrow, but I think transfer should do this anyways. I guess this will need to run a migration which calculates the native asset value in escrow when the functionality is added. This is something that could be done with #330 (though I wouldn't block on #330 being implemented)

colin-axner:
Just to clear up terminology:
- native assets (tokens born on this chain)
- IBC assets (tokens sent from another chain)
- IBC'd in assets (tokens received from another chain)
- IBC'd out assets (native or IBC assets sent to another chain)

The proposal here is to track the IBC'd out of all assets, not just native ones, correct? 

> For native assets though, there is no efficient way to get the total amount IBC'd out.

This is true for all types of tokens? Not just native ones? (using my definition of native)

stackman27:
[...]

Issue #2867, opened 2022-12-01
Document iteration function code structure we intend to use


colin-axner:
## Summary

Document the format we intend to have for our keeper iteration functions which allows for optmization and reuse outside of the keeper. 

The format decided on was implemented for `IterateClientStates` in this [pr](https://github.com/cosmos/ibc-go/pull/2856). See the top comment:
> After discussion with @chatton, we decided every key/value set in our stores should have a corresponding Iterate function. That is IterateClientStates, IterateConsensusStates, etc. These should take in a prefix to determine the starting point of the iteration (allows to limit unnecessary iterations) and a callback which can accrue the desired values and exit iteration early based on optimizations. This PR renames IterateClients to IterateClientStates and adds the prefix necessary for optimized iteration

The idea is that for every key path we store, there should be an associated iteration function. This function should take in a prefix and a callback. It should begin iteration at the top level of the store. It should parse any necessary information from the key and value and then perform a callback (given a provided function by the caller). This will allow for optimized iteration in a generic enough format that the caller can perform specific iteration (such as iterating over a client type) without adding an additional function into the keeper

Each function should look similar to:
```go
func (k Keeper) Iterate<KeyName>(ctx sdk.Context, prefix []byte, cb func(key Key, value Value) bool) {
	store := ctx.KVStore(k.storeKey)
	iterator := sdk.KVStorePrefixIterator(store, PrefixedStoreKey(prefix))

	defer iterator.Close()
	for ; iterator.Valid(); iterator.Next() {
		path := string(iterator.Key())
		if !strings.Contains(path, <KeyName>) {
			// skip non non matching keys
			continue
		}

		key := host.MustParseKeyh(path)
		value := k.MustUnmarshalValue(iterator.Value())

		if cb(key, value) {
			break
		}
	}
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Going to close as stale. Issue hasn't arisen since we ran into this

PR #3019, opened 2023-01-16
Chore: Make a state entry to find total amount of native assets IBC'd out

stackman27:
## Description

Currently every channel has its own escrow bank account, which is a great design! So its very easy to see the amount of native assets IBC'd in or out on any IBC channel.

However, there are use cases that would like us to be able to reason about the total amount of a native asset that has been IBC'd out.

So if there are say three channels from Osmosis to Juno:
```
channel-1: 10 Osmo -> Juno
channel-2: 20 Osmo -> Juno
channel-3: 30 Osmo -> Juno
```
I would like to be find that its been 60 osmo IBC'd out, with no iteration.

closes: https://github.com/cosmos/ibc-go/issues/2664
closes: https://github.com/cosmos/ibc-go/issues/3187

```bash
Chore: Make a state entry to find total amount of native assets IBC'd out
```

---

write a little note why.


damiannolan:
Awesome! Thanks for the PR @stackman27, will try to review this soon!

modules/apps/transfer/client/cli/query.go
```diff
@@ -167,3 +167,36 @@ func GetCmdQueryDenomHash() *cobra.Command {
 	flags.AddQueryFlagsToCmd(cmd)
 	return cmd
 }
+
+// GetCmdQueryDenomHash defines the command to query a denomination hash from a given trace.
+func GetCmdQueryIBCTokenOut() *cobra.Command {
```
nicolaslara, line 172: Can we rename this (and other similar names) to something more explicit? Maybe `QueryTotalEscrowForDenom`?

nicolaslara (commented):
Added some comments about name changes for clarity.

Other than that and getting CI to pass, this looks good to me

modules/apps/transfer/keeper/relay.go
```diff
@@ -109,6 +109,12 @@ func (k Keeper) sendTransfer(
 		); err != nil {
 			return 0, err
 		}
+
+		// get the existing amount in escrow.
+		existingAmount := k.bankKeeper.GetBalance(ctx, escrowAddress, token.Denom)
+
+		// store the token about to be IBC'd Out here
+		k.SetIBCOutDenomAmount(ctx, token.GetDenom(), existingAmount.Amount)
```
crodriguezvega, line 117: Maybe I'm missing something, but I think that this is setting the total amount escrowed for `token.Denom` on a _particular_ channel, not across all channels, right? In the example you mention in the description of the PR, I think this code would set the total amount in escrow in `channel-1`, `channel-2` or `channel-3` depending on what channel the transfer is going out at any given time. For example, if I do a transfer on `channel-1` it sets the total amount in escrow for `channel-1`, but if afterwards I do a transfer on `channel-2` then the previous vale would get overwritten with the total amount in escrow for `channel-2`...

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -35,6 +35,11 @@ service Query {
   rpc EscrowAddress(QueryEscrowAddressRequest) returns (QueryEscrowAddressResponse) {
     option (google.api.http).get = "/ibc/apps/transfer/v1/channels/{channel_id}/ports/{port_id}/escrow_address";
   }
+
+  // TotalNativeIBCOut returns the total native token that's IBC'd out.
+  rpc TotalNativeIBCOut(QueryTotalNativeIBCOutRequest) returns (QueryTotalNativeIBCOutResponse) {
+    option (google.api.http).get = "/ibc/apps/transfer/v1/total_native_ibc_out/{denom}";
```
crodriguezvega, line 41: Maybe for naming consistency and in case in the future we want to have other queries related to denoms?

```suggestion
    option (google.api.http).get = "/ibc/apps/transfer/v1/denoms/{denom}/total_escrow";
```

And would this work fine if the denom contains slashes?

crodriguezvega: Wondering also if the `port_id` should be part of the URI...

crodriguezvega (commented):
Thanks for this contribution, @stackman27! I have a few of questions/remarks:

- It would be great to have some tests for this!
- The escrowed amount for a particular denom would be set on the first transfer of that denom. So until a transfer is done, the query would return 0, right? Would you consider adding a migration that calculates the amount in escrow and sets it beforehand?
- When native tokens return back to the source chain, the total amount in escrow would not be updated until there's a transfer. Should you maybe update the total amount in escrow in the `OnRecvPacket` callback as well? Similar comment in case the transfer times out or the acknowledgement received is an error, then the tokens are refunded, so maybe it's good to update the value as well [in `refundTokens`](https://github.com/cosmos/ibc-go/blob/05685b3f1bb4241ec04c4768dbb7152273f83a96/modules/apps/transfer/keeper/relay.go#L338). In general, I would try to think in what places this value needs to be updated so it doesn't get out of sync with the actual amount in escrow.
- Reading the [original issue](https://github.com/cosmos/ibc-go/issues/2664), I think the implementation is missing one requirement: `Add a safety check if this value is ever attempted to be serialized to a negative value`.
- Following on @nicolaslara comment about naming, I think it would be nice to keep it consistent and use `TotalEscrow` instead of `IBCOut` everywhere.

modules/apps/transfer/keeper/relay.go
stackman27: aah i see, i went with the assumption that all channels share same escrowAccount. So here, i'll probably have to **get all the channel's escrow account sum it up and then store in every transfer** right?

stackman27: Also i was thinking, are all transfers happening under "channels in one port"? Do i also have check for different ports that contains different channels for transfers?

proto/ibc/applications/transfer/v1/query.proto
nicolaslara: Since the tracking is only happening for the transfer port, I think it makes sense to keep the port out of the uri or hard-code it to "transfer".

stackman27:
Thank you @crodriguezvega for the feedback. I am a little confused with this feedback 

> When native tokens return back to the source chain, the total amount in escrow would not be updated until there's a transfer. Should you maybe update the total amount in escrow in the OnRecvPacket callback as well? 

I thought we were only storing the amount that's getting IBC'd out of the source chain. If we account for native tokens that's coming into the source chain, then we are just storing all the channels escrow account amount, not specifically ibc'd amount
cc @nicolaslara

stackman27:
Thanks you for your feedback @crodriguezvega. I have few questions/comments

> * It would be great to have some tests for this!
Finalizing some tests. Want to make sure I have approach ACK 

> * The escrowed amount for a particular denom would be set on the first transfer of that denom. So until a transfer is done, the query would return 0, right? Would you consider adding a migration that calculates the amount in escrow and sets it beforehand?
Addressed! 

> * When native tokens return back to the source chain, the total amount in escrow would not be updated until there's a transfer. Should you maybe update the total amount in escrow in the `OnRecvPacket` callback as well? Similar comment in case the transfer times out or the acknowledgement received is an error, then the tokens are refunded, so maybe it's good to update the value as well [in `refundTokens`](https://github.com/cosmos/ibc-go/blob/05685b3f1bb4241ec04c4768dbb7152273f83a96/modules/apps/transfer/keeper/relay.go#L338). In general, I would try to think in what places this value needs to be updated so it doesn't get out of sync with the actual amount in escrow.
Addressed! I originally thought we were only tracking total channel's IBC out instead of tracking channel's total escrow amount so it got me a little confused

> * Reading the [original issue](https://github.com/cosmos/ibc-go/issues/2664), I think the implementation is missing one requirement: `Add a safety check if this value is ever attempted to be serialized to a negative value`.
Addressed! 

> * Following on @nicolaslara comment about naming, I think it would be nice to keep it consistent and use `TotalEscrow` instead of `IBCOut` everywhere.
Working on this, I still am unsure if native denoms contains slash("/") or not. I imagine it to be a standalone denomination like uosmo, uatom etc

> Another thing that maybe deserves some thinking: in case there are other applications or middleware (e.g. rate limiting) relying on this value, we should be careful that the value cannot be tampered with to fool those applications. For example: one possible situation that comes to mind is that a malicious actor could transfer native tokens directly to one of the escrow accounts for the denom, then the value that you calculate based on the balance doesn't exactly reflect the true amount of native tokens that have been transferred out of the chain.
I didn't really think about this beforehand because i thought this security was already in place. If a random actor transfers $$ to channel-0 escrowAddress without proper transfer/refund, doesnot that violate current security gurantee as well

modules/apps/transfer/keeper/migrations.go
```diff
@@ -53,6 +54,34 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
...
+	getExistingChannels := m.keeper.channelKeeper.GetAllChannelsWithPortPrefix(ctx, types.PortID)
+
+	var nativeDenom string
+	escrowAmount := sdk.ZeroInt()
+
+	for _, channel := range getExistingChannels {
+		escrowAddress := types.GetEscrowAddress(types.PortID, channel.ChannelId)
+		getEscrowBalances := m.keeper.bankKeeper.GetAllBalances(ctx, escrowAddress)
+
+		for _, escrowBalance := range getEscrowBalances {
```
stackman27, line 68: Note: very hacking way of finding sourceChain native denom. Please let me know if there is a better way of retrieving it :))

modules/apps/transfer/keeper/migrations.go
```diff
@@ -53,6 +54,34 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
...
+			if !strings.Contains(escrowBalance.Denom, "ibc") {
+				// native tokens
+				nativeDenom = escrowBalance.Denom
```
stackman27, line 71: Note: very hacky way of finding sourceChain native denom. Please let me know if there is a better way of retrieving it :))

modules/apps/transfer/keeper/relay.go
crodriguezvega: Yes, exactly. Tokens can be transferred out of a chain to the same destination chain through different channels, so you'll have add up the the balances of potentially multiple escrow accounts.

stackman27: sounds great. I have addressed most of it as per this [comment](https://github.com/cosmos/ibc-go/pull/3019#issuecomment-1399208014)

crodriguezvega:
Thanks very much for addressing the comment so quickly, @stackman27! ❤️ I will give the PR another review in the next couple of days. To reply to your last comment:

> I originally thought we were only tracking total channel's IBC out instead of tracking channel's total escrow amount so it got me a little confused

This is a good question. We both had different interpretations of the requirements :) I assumed that we wanted to track the total amount of native assets that are outside of the chain (and hence the total balance of the escrow accounts would give us), but your interpretation might as well be correct (track how many tokens have flowed out of the chain regardless of whether they came back). I think it would be good to hear @ValarDragon thoughts on what he originally wanted to track when he wrote the issue. I am also curious to hear @womensrights opinion on this.

> Working on this, I still am unsure if native denoms contains slash("/") or not. I imagine it to be a standalone denomination like uosmo, uatom etc

Native denoms might contain slashes, so we should support this. We had an issue in the past that we didn't support denom with slashes correctly and it was not possible to transferred them.

> If a random actor transfers $$ to channel-0 escrowAddress without proper transfer/refund, doesnot that violate current security gurantee as well

I think at the moment this doesn't pose any risk. The only thing is that the tokens will stay in the escrow account, probably locked forever (or until governance decides to take them out) since there will be no counterparty vouchers created for them in any other chain. So basically the user will waste their tokens without causing any risk to the application. But I think my point still stands: applications monitoring the total amount escrowed, might be fooled if this value is tampered with. I will discuss this with the rest of the team.

stackman27:
> This is a good question. We both had different interpretations of the requirements :) I assumed that we wanted to track the total amount of native assets that are outside of the chain (and hence the total balance of the escrow accounts would give us), but your interpretation might as well be correct (track how many tokens have flowed out of the chain regardless of whether they came back). I think it would be good to hear @ValarDragon thoughts on what he originally wanted to track when he wrote the issue. I am also curious to hear @womensrights opinion on this.

Hi @crodriguezvega had a chat with @nicolaslara about this and we are tracking "total amount of native assets that are outside of the chain". The new commit i pushed highlights those changes. Please lmk what you think about it

modules/apps/transfer/keeper/migrations.go
```diff
@@ -53,6 +54,34 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
 	return nil
 }
 
+// MigrateTotalEscrowOut migrates the total escrow amount to calculate total IBC'd out.
+func (m Migrator) MigrateTotalEscrowOut(ctx sdk.Context) error {
```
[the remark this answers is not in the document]
stackman27: yup it assumes that there is only one native denom. Can a chain have multiple native denom?

modules/apps/transfer/keeper/migrations.go
```diff
@@ -53,6 +54,34 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
...
+
+	var nativeDenom string
+	escrowAmount := sdk.ZeroInt()
+
+	for _, channel := range getExistingChannels {
+		escrowAddress := types.GetEscrowAddress(types.PortID, channel.ChannelId)
+		getEscrowBalances := m.keeper.bankKeeper.GetAllBalances(ctx, escrowAddress)
+
+		for _, escrowBalance := range getEscrowBalances {
+			if !strings.Contains(escrowBalance.Denom, "ibc") {
```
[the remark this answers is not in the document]
stackman27: So, there are 3 denom possibilities? 
```
// Denom possibilities:
// - "atom" = native denom
// - "ibc/atom" = non native denom
// - "atom/ibc/osmo" = native denom
```

modules/apps/transfer/keeper/migrations.go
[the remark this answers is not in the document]
nicolaslara: yes! a chain can have multiple native denoms (for example: osmo, ion, and anything created via tokenfactory)

modules/apps/transfer/keeper/relay.go
```diff
@@ -109,6 +109,14 @@ func (k Keeper) sendTransfer(
 		); err != nil {
 			return 0, err
 		}
+
+		// get the existing escrow amount in store.
+		existingToken := k.GetIBCOutDenomAmount(ctx, token.Denom)
+		existingTokenWithNewTokens := existingToken.Add(token.Amount)
+
+		// store the token about to be IBC'd Out here
+		k.SetIBCOutDenomAmount(ctx, token.GetDenom(), existingTokenWithNewTokens)
```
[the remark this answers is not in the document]
stackman27: addressed!

modules/apps/transfer/keeper/migrations.go
```diff
@@ -53,6 +54,34 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
...
+				// native tokens
+				nativeDenom = escrowBalance.Denom
+				escrowAmount = escrowAmount.Add(escrowBalance.Amount)
+			}
+		}
+	}
+
+	if len(nativeDenom) == 0 {
+		return fmt.Errorf("invalid native denom")
```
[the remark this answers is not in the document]
stackman27: fixed!

modules/apps/transfer/keeper/migrations.go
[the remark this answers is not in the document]
stackman27: addressed migration for more than one denom!

stackman27:
@crodriguezvega @nicolaslara apologies for the delay, i was busy with some other work. 

However, I've addressed all the comments and this PR is ready for re-review :))

crodriguezvega:
> @crodriguezvega @nicolaslara apologies for the delay, i was busy with some other work.
> 
> However, I've addressed all the comments and this PR is ready for re-review :))

Hi @stackman27, thanks for addressing the comments! I've started reviewing again and there's a few small changes that I would like to do, but if you don't mind, I think it would help to move this PR faster if I push the changes directly to your branch. Is that ok?

There's also a scenario that I think it's not covered yet. Imagine the following situation for tokens traveling from chain A (where the tokens are native) to chain C:

chains:  chain A (channel-0) -> (channel-0) chain B (channel-1) -> (channel-1) chain C
denom: stake                               transfer/channel-0/stake             -> transfer/channel-1/transfer/channel-0/stake

When the tokens are transferred from chain B to chain C, [the `SendChainIsSource` check on `sendTransfer`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#L100) will pass for tokens with denom `transfer/channel-0/stake`, because the source channel is `channel-1` (since tokens are now being transferred to chain C). And because this check passes, then an escrow amount will be stored in state for denom `transfer/channel-0/stake`. But this would not be desired behavior, since we only want to track native denoms. I think similar behavior is happening also for the packet callbacks.

I am writing some tests to confirm this, and if my reasoning is not wrong, then I will implement the necessary changes in the code to handle this scenario and I will also push them to your branch. I hope this is also ok. I will comment on the PR, as soon as I push the changes, so that you can also review.

I think I will be able to push to your branch, in case I can, I will also comment here to ask you to give me permissions. 

Thanks a lot for all this work so far!

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -32,12 +32,15 @@ func (k Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types.
 		return nil, sdkerrors.Wrapf(sdkerrors.ErrUnauthorized, "%s is not allowed to send funds", sender)
 	}
 
+	// Use a cached context to prevent accidental state changes when seting the total amount in escrow
+	cacheCtx, writeFn := ctx.CacheContext()
```
crodriguezvega, line 36: I would like @colin-axner or @damiannolan to check that this is the best way to achieve this.

damiannolan: Looks unnecessary. The context is [already cached by baseapp](https://github.com/cosmos/cosmos-sdk/blob/main/baseapp/baseapp.go#L729), so if an error is returned by `sendTransfer` it will bubble up and state will not be commited by `writeCache()` at the sdk level.

crodriguezvega: All right, thanks, @damiannolan. I added it because when running [this test](https://github.com/cosmos/ibc-go/blob/0903989307e2fe0b1d272eb3c0fd8ee87b42e731/modules/apps/transfer/keeper/relay_test.go#L102) the state seemed to not be reverted... I will check once more, though.

damiannolan: In the testcase its calling the keeper function directly, so the msg is never being routed through baseapp. 
You could use `chain.SendMsgs(msgTransfer)` to route it through baseapp, but I think that'd require a good bit of refactoring to the testcases

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -149,16 +157,92 @@ func (suite *KeeperTestSuite) TestSendTransfer() {
 	}
 }
 
+func (suite *KeeperTestSuite) TestSendTransferDoesNotSetTotalEscrowAmountForSourceIBCToken() {
```
stackman27, line 160: This function name got me thinking for a bit can we change it to something like `TestSendTransferNativeTotalEscrowAmount` or anything shorter?

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -149,16 +157,92 @@ func (suite *KeeperTestSuite) TestSendTransfer() {
...
+	/*
+		Given the following flow of tokens:
+
+		chain A (channel 0) -> (channel-0) chain B (channel-1) -> (channel-1) chain A
```
stackman27, line 164: nit: wouldn't this be channel-0?
```suggestion
		chain A (channel 0) -> (channel-0) chain B (channel-1) -> (channel-0) chain A
```

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -261,17 +406,102 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
 	}
 }
 
+func (suite *KeeperTestSuite) TestOnRecvPacketDoesNotSetTotalEscrowAmountForSourceIBCToken() {
+	/*
+		Given the following flow of tokens:
+
+		chain A (channel 0) -> (channel-0) chain B (channel-1) -> (channel-1) chain A (channel-1)             -> (channel-1) chain B
```
stackman27, line 413: Why can't we just do chainA -> chanB and check onRecv on chainB instead of doing chainA->ChainB->ChainA->ChainB to test this feature?

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -149,16 +157,92 @@ func (suite *KeeperTestSuite) TestSendTransfer() {
...
+		suite.chainA.GetTimeoutHeight(), 0, "",
+	)
+
+	res, err := suite.chainB.GetSimApp().TransferKeeper.Transfer(sdk.WrapSDKContext(suite.chainB.GetContext()), msg)
+	suite.Require().NoError(err)
+	suite.Require().NotNil(res)
+
+	// check total amount in escrow of sent token on sending chain
+	totalEscrow := suite.chainB.GetSimApp().TransferKeeper.GetTotalEscrowForDenom(suite.chainB.GetContext(), coin.GetDenom())
+	suite.Require().Equal(math.ZeroInt(), totalEscrow)
```
stackman27, line 232: Can we check chainA totalEscrow and see if it's 100stake?

stackman27 (commented):
Thank you so much @crodriguezvega for all the changes!! I added few feedbacks, if you think its relevant i can push commit and address them as well :))

modules/apps/transfer/keeper/relay_test.go
stackman27, line 413: wouldn't this be channel 0 because chainA only only established ch-0 to transfer to chainB? 

```suggestion
		chain A (channel 0) -> (channel-0) chain B (channel-1) -> (channel-1) chain A (channel-1)             -> (channel-0) chain B
```

modules/apps/transfer/keeper/relay.go
```diff
@@ -109,6 +109,16 @@ func (k Keeper) sendTransfer(
 		); err != nil {
 			return 0, err
 		}
+
+		denomTrace := types.ParseDenomTrace(fullDenomPath)
+		if denomTrace.GetPath() == "" {
```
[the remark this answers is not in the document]
crodriguezvega: I believe this check is needed because you could have a denom like `transfer/channel-0/uatom` and send it on `channel-1` and then `SenderChainIsSource` will still pass, but the denom is not native. Does it make sense?

modules/apps/transfer/keeper/relay_test.go
crodriguezvega: Yes, I think we could that check here as well, although that scenario should have been checked in the test `successful transfer with native token` of `TestSendTransfer` above.

modules/apps/transfer/keeper/relay_test.go
crodriguezvega: Because we should test the scenario of receiving non-native IBC tokens in the source chain. If you only do chainA -> chanB, the vouchers received on B are non-native IBC, but the chain is not the source. Do you agree?

crodriguezvega: The test establishes 2 channels between chain A and B. If we send the tokens over channel-0 the receiving chain would actually mint tokens with denom `transfer/channel-0/transfer/channel-1/transfer-channel-0/stake`.

modules/apps/transfer/keeper/relay.go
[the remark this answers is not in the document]
colin-axner: @crodriguezvega is correct. `SenderChainIsSource` is just determining if the tokens being sent originate from the sending chain or if they originate from the receiving chain (path unwinding)

The terminology `native token` should imply the token is not an IBC token or another token that was originally transferred from another chain. An IBC token is not native, but sender is source when it is being sent on a channel it didn't arrive from. Tokens are native regardless of where they are being transferred to

modules/apps/transfer/keeper/relay.go
```diff
@@ -229,6 +238,15 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 			return sdkerrors.Wrap(err, "unable to unescrow tokens, this may be caused by a malicious counterparty module or a bug: please open an issue on counterparty module")
 		}
 
+		if denomTrace.GetPath() == "" {
```
colin-axner, line 241: I think we actually want to do this for all tokens being sent/received https://github.com/cosmos/ibc-go/issues/2664#issuecomment-1439845851

modules/apps/transfer/keeper/relay.go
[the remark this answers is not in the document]
stackman27: aaah this makes sense. Just to clarify in the example below 
```
chain A (channel 0) -> (channel-0) chain B (channel-1) -> (channel-1) chain A
stake                  transfer/channel-0/stake           transfer/channel-1/transfer/channel-0/stake
```

if we donot have this `GetPath()` check ChainB ---> ChainA  transfer will pass the function `SenderChainIsSource` but it is not a native denom right?

modules/apps/transfer/keeper/relay_test.go
stackman27: yup i agree this makes sense

stackman27: hmm still a bit confused

so we have 2 channels from ChainA -> ChainB 
```
channel-0:  ChainA -> ChainB
channel-1:  ChainA -> ChainB

chainA (sends stake denom over ch-0)  -----> ChainB (receives transfer/ch-0/stake)
chainB (sends transfer/ch-0/stake denom over ch-1 ----> ChainA (receives transfer/ch-1/transfer/ch-0/stake) 
```

My question is wouldn't ChainB just send stake as (transfer/ch-0/stake) to chainA Why does it have to send it over a different channel ch-1 instead of ch-1?

modules/apps/transfer/keeper/relay.go
crodriguezvega: Let's check with @ValarDragon and @womensrights to agree on the requirement. :)

modules/apps/transfer/keeper/relay.go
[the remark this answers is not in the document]
crodriguezvega: Yes, that's right. When sending tokens of denom `transfer/channel-0/stake` from chain B, you have these two situations:

- If you send them over `channel-0` (the channel from where they were received, since `channel-0` is part of the trace path of the denom), then `SenderChainIsSource` will return false, since it is the receiving chain (chain A) the source of the tokens.
- If you send them over a different channel (for example `channel-1`), then chain B becomes the source chain and `SenderChainIsSource` will return true. In this case, and assuming a definition of native tokens that does not include IBC denoms, then we should not store in state the total escrow for these tokens.

I hope this helps to clarify!

modules/apps/transfer/keeper/relay_test.go
crodriguezvega: If we send `transfer/channel-0/stake` from chain B to chain A over `channel-1` then chain A will not be the source chain (`ReceiverChainIsSource` will return false). We need to send `transfer/channel-1/transfer/channel-0/stake` from chain A to chain B over `channel-1`, so that chain B is the source chain for tokens with denom `transfer/channel-0/stake` and `ReceiverChainIsSource` returns true, so that we can test the scenario of receiving IBC denom tokens for which the receiver chain is source, but no total escrow amount is stored in state.

I hope this makes sense now. I know it's complicated sometimes to get one's head around with all these :) so please keep asking questions if something is not clear. :)

stackman27: aah so the whole flow is gonna looks something like this; 

```
ChainA (ch-0) -----> (ch-1) ChainB 
Source = chainA 
ChainA sends [10stake from transfer/ch-0 to transfer/ch-1] to chainB

ChainB (ch-1) ------> (ch-0) ChainA
Source = chainA
ChainB sends [10stake from transfer/ch-1/stake to transfer/ch-0] to chainA
```

my question here is, wouldn't chainA be the source when we make transfer from  chainB --> chainA?  Is chainB sending tokens from a different channel than the one it received it from, therefore chainA is not the source chain here?

modules/apps/transfer/keeper/relay.go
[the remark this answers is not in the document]
stackman27: Thank you! this makes sense

modules/apps/transfer/keeper/relay_test.go
stackman27: i think my assumption about sending tokens over different channel than it was received and therefore the chain will be native is right after reading this description you posted (https://github.com/cosmos/ibc-go/pull/3019#discussion_r1115342675)

modules/apps/transfer/keeper/relay.go
womensrights: Yes I think that this would be useful for all ibc assets, not just native assets. I followed up with Cosmostation and it seems like it could simplify generating [data like this](https://www.mintscan.io/osmosis/assets)

crodriguezvega: Great, thanks everyone for weighing in. We'll make the necessary changes to support this then.

stackman27: @crodriguezvega what do you think about making a follow up PR that supports ibc assets and merge this one? I can lead the course of the the follow up PR

crodriguezvega: @stackman27 I think I prefer to continue working on this PR, so that the feature is complete when it gets merged. I started doing some changes this morning. I can push those now and you can pick up from there, if you want. I was starting to adjust the tests that I added.

crodriguezvega: Ok, just pushed my changes. The test should be failing now.

modules/apps/transfer/keeper/relay_test.go
crodriguezvega: I shortened it a bit now to `TestSendTransferSetsTotalEscrowForSourceIBCToken` :)

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -119,3 +123,79 @@ func (suite *KeeperTestSuite) TestMigratorMigrateTracesCorruptionDetection() {
 		migrator.MigrateTraces(suite.chainA.GetContext()) //nolint:errcheck // we shouldn't check the error here because we want to ensure that a panic occurs.
 	})
 }
+
+func (suite *KeeperTestSuite) TestMigrateTotalEscrowForDenom() {
```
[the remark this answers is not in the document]
crodriguezvega: I think testing with one native denom should be fine.

modules/apps/transfer/keeper/relay.go
stackman27: perfect! i will pick it up from here and ask you questions if i get stuck

stackman27:
Hi @crodriguezvega i'm a little confused as to what exactly i should be testing for, currently we expect these functions 
```
TestSendTransferSetsTotalEscrowAmountForSourceIBCToken, TestOnRecvPacketSetsTotalEscrowAmountForSourceIBCToken, TestOnAcknowledgementPacketSetsTotalEscrowAmountForSourceIBCToken,
TestOnTimeoutPacketSetsTotalEscrowAmountForSourceIBCToken
```
to panic because escrowAmount becomes negative, should i be resolving the panic and updating escrowAmount in the test so that there is some defaultEscrowAmount to begin with?

crodriguezvega:
> Hi @crodriguezvega i'm a little confused as to what exactly i should be testing for, currently we expect these functions
> 
> ```
> TestSendTransferSetsTotalEscrowAmountForSourceIBCToken, TestOnRecvPacketSetsTotalEscrowAmountForSourceIBCToken, TestOnAcknowledgementPacketSetsTotalEscrowAmountForSourceIBCToken,
> TestOnTimeoutPacketSetsTotalEscrowAmountForSourceIBCToken
> ```
> 
> to panic because escrowAmount becomes negative, should i be resolving the panic and updating escrowAmount in the test so that there is some defaultEscrowAmount to begin with?
> 
> Also i'm assuming we dont really need any logic changes right?

Hi @stackman27. Sorry for the late reply... You're right: these tests should not panic anymore. The tests should have an initial escrow amount (because we now track total escrow for IBC denoms on chains acting as source) and thus the escrow amount should be decreased when the tokens are received back.

Thanks for updating the PR. I will review this week!

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -35,6 +35,11 @@ service Query {
   rpc EscrowAddress(QueryEscrowAddressRequest) returns (QueryEscrowAddressResponse) {
     option (google.api.http).get = "/ibc/apps/transfer/v1/channels/{channel_id}/ports/{port_id}/escrow_address";
   }
+
+  // TotalEscrowForDenom returns the total amount of tokens of a native denom in escrow.
```
colin-axner, line 39:
```suggestion
  // TotalEscrowForDenom returns the total amount of tokens in escrow based on the denom.
```

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -35,6 +35,11 @@ service Query {
...
+  rpc TotalEscrowForDenom(QueryTotalEscrowFormDenomRequest) returns (QueryTotalEscrowForDenomResponse) {
```
colin-axner, line 40:
```suggestion
  rpc TotalEscrowForDenom(QueryTotalEscrowForDenomRequest) returns (QueryTotalEscrowForDenomResponse) {
```

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -102,4 +107,14 @@ message QueryEscrowAddressRequest {
 message QueryEscrowAddressResponse {
   // the escrow account address
   string escrow_address = 1;
+}
+
+// QueryTotalNativeIBCOut is the request type for TotalNativeIBCOut RPC method.
```
colin-axner, line 112:
```suggestion
// QueryTotalEscrowForDenom is the request type for TotalEscrowForDenom RPC method.
```

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -102,4 +107,14 @@ message QueryEscrowAddressRequest {
...
+message QueryTotalEscrowFormDenomRequest {
+  string denom = 1;
+}
+
+// TotalNativeIBCOutResponse is the response type for TotalNativeIBCOut RPC method.
```
colin-axner, line 117:
```suggestion
// QueryTotalEscrowForDenomResponse is the response type for TotalEscrowForDenom RPC method.
```

modules/apps/transfer/types/trace.go
```diff
@@ -75,6 +75,11 @@ func (dt DenomTrace) GetFullDenomPath() string {
 	return dt.GetPrefix() + dt.BaseDenom
 }
 
+// IsNativeDenom returns true if the denomination is native, thus containing no trace history.
+func (dt DenomTrace) IsNativeDenom() bool {
```
colin-axner, line 79: nice!

modules/apps/transfer/types/keys.go
```diff
@@ -29,6 +29,9 @@ const (
 
 	// DenomPrefix is the prefix used for internal SDK coin representation.
 	DenomPrefix = "ibc"
+
+	KeyTotalEscrowPrefix = "totalEscrow"
+	KeyDenomsPrefix      = "denoms"
```
colin-axner, line 34: It might make sense for this to be a single key? Is there a reason to have 2?

modules/apps/transfer/module.go
```diff
@@ -107,6 +107,10 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	if err := cfg.RegisterMigration(types.ModuleName, 1, m.MigrateTraces); err != nil {
 		panic(fmt.Sprintf("failed to migrate transfer app from version 1 to 2: %v", err))
 	}
+
+	if err := cfg.RegisterMigration(types.ModuleName, 2, m.MigrateTotalEscrowForDenom); err != nil {
```
colin-axner, line 111: I think you need to bump the version as well in this file

modules/apps/transfer/client/cli/query.go
```diff
@@ -167,3 +167,36 @@ func GetCmdQueryDenomHash() *cobra.Command {
 	flags.AddQueryFlagsToCmd(cmd)
 	return cmd
 }
+
+// GetCmdQueryTotalEscrowForDenom defines the command to query the total amount of tokens in escrow for a native denom
+func GetCmdQueryTotalEscrowForDenom() *cobra.Command {
```
colin-axner, line 172: We will need to register the cmd on the cli, see [here](https://github.com/stackman27/ibc-go/blob/stack/total-ibc-out/modules/apps/transfer/client/cli/cli.go#L17)

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -121,3 +121,22 @@ func (q Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRe
...
+func (q Keeper) TotalEscrowForDenom(c context.Context, req *types.QueryTotalEscrowFormDenomRequest) (*types.QueryTotalEscrowForDenomResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(c)
+
+	if q.IsIBCDenom(ctx, req.Denom) {
+		return nil, status.Error(codes.InvalidArgument, "denom is not a native token denomination")
+	}
```
colin-axner, line 135: I think this can be removed now? The associated test case should be expected to pass

modules/apps/transfer/keeper/grpc_query_test.go
```diff
@@ -261,3 +261,80 @@ func (suite *KeeperTestSuite) TestEscrowAddress() {
...
+			"success",
+			func() {
+				req = &types.QueryTotalEscrowFormDenomRequest{
+					Denom: sdk.DefaultBondDenom,
+				}
+			},
+			true,
+		},
+		{
+			"not found denom trace",
```
colin-axner, line 283: maybe add:
```suggestion
			"not found denom trace, valid native denom",
```

modules/apps/transfer/keeper/grpc_query_test.go
```diff
@@ -261,3 +261,80 @@ func (suite *KeeperTestSuite) TestEscrowAddress() {
...
+			suite.SetupTest() // reset
+
+			tc.malleate()
+			ctx := sdk.WrapSDKContext(suite.chainA.GetContext())
+
+			res, err := suite.queryClient.TotalEscrowForDenom(ctx, req)
+
+			if tc.expPass {
+				suite.Require().NoError(err)
+				suite.Require().Equal(int64(0), res.Amount)
```
colin-axner, line 334: Hmm, so if the denom doesn't exist, instead of returning an error, we return 0. Would it be possible to use the bank keeper to check if it is a valid denom on the chain?

colin-axner (commented):
Thanks @stackman27! Great work! I've only reviewed half the files, I will continue my review next week. Let me know if you have any questions

modules/apps/transfer/types/keys.go
crodriguezvega: The reason for 2 keys is to follow the pattern that we use for other paths, e.g. `channelEnds/ports/{identifier}/channels/{identifier}`.

modules/apps/transfer/keeper/grpc_query_test.go
crodriguezvega: Not sure if there exists a way to check that in the SDK? Isn't it possible to just mint tokens with any denomination?

crodriguezvega: Ah, I think there's [a way](https://github.com/cosmos/cosmos-sdk/blob/main/types/coin.go#L855).

crodriguezvega: Ok, I found a way to at least check if the denim is valid using `ValidateDenom` from SDK.

crodriguezvega (commented):
@stackman27 I made some changes (most notably adding checks to e2e test, genesis migration) so this PR is ready for another review now. There are [a few items](https://github.com/cosmos/ibc-go/issues/3368) pending that I will work on in a follow up PR.

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -121,3 +121,22 @@ func (q Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRe
 		EscrowAddress: addr.String(),
 	}, nil
 }
+
+// TotalEscrowForDenom implements the TotalEscrowForDenom gRPC method.
+func (q Keeper) TotalEscrowForDenom(c context.Context, req *types.QueryTotalEscrowFormDenomRequest) (*types.QueryTotalEscrowForDenomResponse, error) {
```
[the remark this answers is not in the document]
colin-axner: I remember there being a note to change these from `q` to `k`, but I can't find the comment or issue. It isn't consistent in the code base, see [ics27](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/keeper/grpc_query.go#L17). My general preference would be to use `k` as that makes logical sense to me, but I'm okay to open a separate issue

.github/workflows/e2e-upgrade.yaml
```diff
@@ -50,10 +50,10 @@ jobs:
     with:
       chain-image: ghcr.io/cosmos/ibc-go-simd
       chain-binary: simd
-      chain-a-tag: pr-3136 # TODO: needs v7.0.0 (with simapp fixes) when cut
-      chain-b-tag: pr-3136 # TODO: needs v7.0.0 (with simapp fixes) when cut
+      chain-a-tag: v7.0.0
+      chain-b-tag: v7.0.0
       chain-upgrade-tag: pr-3164 # TODO: needs v7.1.0 when cut
```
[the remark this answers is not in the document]
colin-axner: Should this change be made outside of this pr?

modules/apps/transfer/types/keys.go
colin-axner: I think I'd prefer it to be one key, I don't find the pattern used for channel/port id path's to be equivalent. In referenced case, `channelEnds` is a key for the 04-channel keeper. Here the `totalEscrow` and `denoms` are referencing the same key, is there a scenario when we use `totalEscrow` as a prefix for a different key? Otherwise I'd prefer it to just be `totalEscrow/{denom}`

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -35,6 +35,11 @@ service Query {
   rpc EscrowAddress(QueryEscrowAddressRequest) returns (QueryEscrowAddressResponse) {
     option (google.api.http).get = "/ibc/apps/transfer/v1/channels/{channel_id}/ports/{port_id}/escrow_address";
   }
+
+  // TotalEscrowForDenom returns the total amount of tokens in escrow based on the denom.
+  rpc TotalEscrowForDenom(QueryTotalEscrowForDenomRequest) returns (QueryTotalEscrowForDenomResponse) {
+    option (google.api.http).get = "/ibc/apps/transfer/v1/total_escrow/{denom}";
```
colin-axner, line 41: Once upon a time I remember being told that the constant before a user provided argument needed to be plural, so in this case `total_escrows/{denom}`, this is why it is `channels`, `ports`. I was never sure why this would be the case, but does anyone have any idea if this has changed?

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -102,4 +107,14 @@ message QueryEscrowAddressRequest {
...
+}
+
+// QueryTotalEscrowForDenomRequest is the request type for TotalEscrowForDenom RPC method.
+message QueryTotalEscrowForDenomRequest {
+  string denom = 1;
+}
+
+// QueryTotalEscrowForDenomResponse is the response type for TotalEscrowForDenom RPC method.
+message QueryTotalEscrowForDenomResponse {
+  string amount = 1;
```
colin-axner, line 119: Should we cast into [`sdk.Int`](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/base/v1beta1/coin.proto#L20)?

proto/ibc/applications/transfer/v1/genesis.proto
```diff
@@ -5,11 +5,16 @@ package ibc.applications.transfer.v1;
...
+import "cosmos/base/v1beta1/coin.proto";
 import "gogoproto/gogo.proto";
 
 // GenesisState defines the ibc-transfer genesis state
 message GenesisState {
   string              port_id      = 1;
   repeated DenomTrace denom_traces = 2 [(gogoproto.castrepeated) = "Traces", (gogoproto.nullable) = false];
   Params              params       = 3 [(gogoproto.nullable) = false];
+  // denom_escrows contains the denomination and the total amount of tokens of
+  // the denomination in escrow in the transfer channels accounts
```
colin-axner, line 17:
```suggestion
  // denom_escrows contains the total amount of tokens escrowed
  // by the transfer module 
```

modules/apps/transfer/keeper/genesis.go
```diff
@@ -28,13 +28,18 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state types.GenesisState) {
 	}
 
 	k.SetParams(ctx, state.Params)
+
+	for _, denomEscrow := range state.DenomEscrows {
+		k.SetTotalEscrowForDenom(ctx, denomEscrow.Denom, denomEscrow.Amount)
+	}
```
colin-axner, line 34: Can we add a comment that the `DenomEscrows` must be valid as duplicate entries will cause invalid state?

modules/apps/transfer/keeper/genesis_test.go
```diff
@@ -3,35 +3,51 @@ package keeper_test
...
 func (suite *KeeperTestSuite) TestGenesis() {
-	var (
-		path   string
-		traces types.Traces
-	)
+	getTrace := func(index uint) string {
+		return fmt.Sprintf("transfer/channelToChain%d", index)
+	}
 
-	for i := 0; i < 5; i++ {
```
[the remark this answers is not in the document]
colin-axner: Can we pull these changes out into a separate pr to main?

proto/ibc/applications/transfer/v1/genesis.proto
```diff
@@ -5,11 +5,16 @@ package ibc.applications.transfer.v1;
...
+  repeated cosmos.base.v1beta1.Coin denom_escrows = 4
```
colin-axner, line 18: Should we call this `total_escrowed`?

modules/apps/transfer/keeper/relay.go
```diff
@@ -230,6 +236,13 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 			return errorsmod.Wrap(err, "unable to unescrow tokens, this may be caused by a malicious counterparty module or a bug: please open an issue on counterparty module")
 		}
 
+		// get the existing total amount in escrow
+		currentTotalEscrow := k.GetTotalEscrowForDenom(ctx, token.GetDenom())
+		newTotalEscrow := currentTotalEscrow.Sub(token.Amount)
+
+		// store the new total amount in escrow
+		k.SetTotalEscrowForDenom(ctx, token.GetDenom(), newTotalEscrow)
```
colin-axner, line 244: In a followup pr, we could consider bundling this with the bank send into two helper functions:
```go
    // escrow token case
    
    escrowAddress := types.GetEscrowAddress(...)
    if err := escrowToken(ctx, sender, escrowAddress, token); err != nil {
        return err
    }
    
    // unescrow case
    
    escrowAddress := types.GetEscrowAddress(...)
    if err := unescrowToken(ctx, escrowAddress, receiver, token); err != nil {
        return err
    }
    
    
    ... 
    
func escrowToken(ctx sdk.Context, sender, escrowAddress sdk.AccAddress, token sdk.Coin) {
        // escrow source tokens. It fails if balance insufficient.
	if err := k.bankKeeper.SendCoins(
		ctx, sender, escrowAddress, sdk.NewCoins(token),
	); err != nil {
		return err
	}

	// get the existing total amount in escrow
	currentTotalEscrow := k.GetTotalEscrowForDenom(ctx, token.GetDenom())
	newTotalEscrow := currentTotalEscrow.Add(token.Amount)

	// store the new total amount in escrow
	k.SetTotalEscrowForDenom(ctx, token.GetDenom(), newTotalEscrow)
		
	return nil
}
    

```

modules/apps/transfer/keeper/migrations.go
```diff
@@ -53,6 +54,33 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
...
+			} else {
+				tokenDenomToTotalEscrowMap[escrowBalance.Denom] = escrowBalance.Amount
+			}
+		}
+	}
+
+	if len(tokenDenomToTotalEscrowMap) != 0 {
+		for denom, amount := range tokenDenomToTotalEscrowMap {
+			m.keeper.SetTotalEscrowForDenom(ctx, denom, amount)
+		}
```
colin-axner, line 78: Do migrations have gas limits?

colin-axner (commented):
Thanks for all the followup fixes! First pass, will continue reviewing next week

proto/ibc/applications/transfer/v1/query.proto
crodriguezvega: Yes, the plural naming is the [convention for resources in REST apis](https://restfulapi.net/resource-naming/). But in this case, it would be more appropriate to make it than `denoms/{denom}/total_escrow`. I will change it to that.

modules/apps/transfer/keeper/relay.go
crodriguezvega: Good idea! I will add it to the list of follow-ups in https://github.com/cosmos/ibc-go/issues/3368.

modules/apps/transfer/keeper/genesis.go
crodriguezvega: I was hoping that we could avoid ending up in invalid state if [`Validate` of `GenesisState`](https://github.com/stackman27/ibc-go/blob/stack/total-ibc-out/modules/apps/transfer/types/genesis.go#L31) is called before `InitGenesis`, and that would also call `Validate` on the collection of coins, which should error if there are duplicates. But if that is not the case, then yes, I agree we can add a comment here.

modules/apps/transfer/types/keys.go
crodriguezvega: I see your point, thanks! And what about if the keys are swapped, so `denoms/total_escrow/{denom}`? In case in the future other information for denoms is stored we can still iterate over the `denoms/total_escrow` to filter any other information. Would that be useful?

proto/ibc/applications/transfer/v1/genesis.proto
crodriguezvega: And `denom_total_escrow`? I was trying to keep it similar to `denom_traces` above...

modules/apps/transfer/keeper/genesis.go
colin-axner: Only a valid genesis state should be provided to `InitGenesis`. I would like a comment to be added regardless so that a future reader can understand the correctness of the code. Since `InitGenesis` is implicitly relying on a check in `genesisState.Validate()`, I think we should make a note of this, maybe in both locations

proto/ibc/applications/transfer/v1/genesis.proto
colin-axner: I personally don't think `denom_traces` is a good name. I don't think we need to keep to that pattern, but happy to abide by what folks think is best

modules/apps/transfer/keeper/keeper_test.go
```diff
@@ -44,3 +48,139 @@ func NewTransferPath(chainA, chainB *ibctesting.TestChain) *ibctesting.Path {
...
+					suite.chainA.GetSimApp().TransferKeeper.SetTotalEscrowForDenom(ctx, "atom", expAmount)
+				})
+				total := suite.chainA.GetSimApp().TransferKeeper.GetTotalEscrowForDenom(ctx, "atom")
+				suite.Require().Equal(math.ZeroInt(), total)
+			}
+		})
+	}
+}
+
+func (suite *KeeperTestSuite) TestGetAllDenomEscrows() {
```
colin-axner, line 101: Could we add a test for many kv mappings since this function performs iteration?

modules/apps/transfer/keeper/keeper.go
```diff
@@ -133,14 +137,83 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
 	for ; iterator.Valid(); iterator.Next() {
-
 		denomTrace := k.MustUnmarshalDenomTrace(iterator.Value())
 		if cb(denomTrace) {
 			break
 		}
 	}
 }
 
+// GetTotalEscrowForDenom gets the total amount of source chain tokens that are in escrow.
```
colin-axner, line 147: Should we mention that they are keyed by denomination?

modules/apps/transfer/keeper/keeper.go
```diff
@@ -133,14 +137,83 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
+	if err != nil {
+		panic(err)
+	}
+
+	store.Set(types.TotalEscrowForDenomKey(denom), bz)
+}
+
+// GetAllDenomEscrows returns the escrow information for all the denominations.
+func (k Keeper) GetAllDenomEscrows(ctx sdk.Context) sdk.Coins {
+	escrows := sdk.Coins{}
```
colin-axner, line 179: should this be a var declaration? I think that is our convention:
```go
    var escrows sdk.Coins
```

modules/apps/transfer/keeper/keeper.go
```diff
@@ -133,14 +137,83 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
+	k.IterateDenomEscrows(ctx, func(denomEscrow sdk.Coin) bool {
+		escrows = append(escrows, denomEscrow)
+		return false
+	})
+
+	return escrows.Sort()
```
colin-axner, line 185: Iteration is lexicographical? Is the sorting necessary?

modules/apps/transfer/keeper/keeper.go
```diff
@@ -133,14 +137,83 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
+}
+
+// IterateDenomEscrows iterates over the denomination escrows in the store
+// and performs a callback function.
+func (k Keeper) IterateDenomEscrows(ctx sdk.Context, cb func(denomEscrow sdk.Coin) bool) {
```
colin-axner, line 190: A new convention we want to do is have every iteration function take in a [`prefix`](https://github.com/cosmos/ibc-go/issues/2867), this allows the caller to decide the iteration starting point without creating a secondary function

modules/apps/transfer/keeper/keeper.go
```diff
@@ -133,14 +137,83 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
+	re := regexp.MustCompile(fmt.Sprintf(`%s\/%s\/(.*[^\s])`, types.KeyTotalEscrowPrefix, types.KeyDenomsPrefix))
```
colin-axner, line 191: what is this?

modules/apps/transfer/keeper/keeper.go
```diff
@@ -133,14 +137,83 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
+
+	store := ctx.KVStore(k.storeKey)
+	iterator := sdk.KVStorePrefixIterator(store, []byte(types.KeyTotalEscrowPrefix))
+
+	defer sdk.LogDeferred(ctx.Logger(), func() error { return iterator.Close() })
+	for ; iterator.Valid(); iterator.Next() {
+		matches := re.FindStringSubmatch(string(iterator.Key()))
+		if len(matches) != 2 { // there should be two matches: 1st one for the whole string, 2nd for the denomination
+			continue
+		}
```
colin-axner, line 201: An alternative is to use [`strings.Contains`](https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/keeper/keeper.go#L376). My main gripe with regexp is the `MustCompile` statement is hard to read, I wonder if we should keep things simple?

modules/apps/transfer/keeper/keeper.go
```diff
@@ -133,14 +137,83 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
+
+		denom := matches[1]
+
+		var amount math.Int
+		if err := amount.Unmarshal(iterator.Value()); err != nil {
+			continue
+		}
```
colin-axner, line 208: Can we indicate why we would continue?

e2e/tests/transfer/base_test.go
```diff
@@ -130,6 +139,12 @@ func (s *TransferTestSuite) TestMsgTransfer_Succeeds_Nonincentivized() {
 		expected := testvalues.StartingTokenAmount
 		s.Require().Equal(expected, actualBalance)
 	})
+
+	t.Run("escrow amount for native denom is updated", func(t *testing.T) {
+		actualTotalEscrow, err := s.QueryTotalEscrowForDenom(ctx, chainA, chainADenom)
+		s.Require().NoError(err)
+		s.Require().Equal(math.ZeroInt(), actualTotalEscrow)
```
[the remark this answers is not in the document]
colin-axner: should this be added as an in code comment?

e2e/tests/transfer/base_test.go
```diff
@@ -106,6 +107,14 @@ func (s *TransferTestSuite) TestMsgTransfer_Succeeds_Nonincentivized() {
 		s.Require().Equal(expected, actualBalance)
 	})
 
+	t.Run("escrow amount for native denom is set", func(t *testing.T) {
+		actualTotalEscrow, err := s.QueryTotalEscrowForDenom(ctx, chainA, chainADenom)
+		s.Require().NoError(err)
+
+		expectedTotalEscrow := math.NewInt(testvalues.IBCTransferAmount)
+		s.Require().Equal(expectedTotalEscrow, actualTotalEscrow)
+	})
```
colin-axner, line 116: Should this be in the `tokens are escrowed` `t.Run`?

modules/apps/transfer/keeper/grpc_query_test.go
```diff
@@ -261,3 +262,99 @@ func (suite *KeeperTestSuite) TestEscrowAddress() {
...
+			true, // denom trace is not found, so the denom is considered a native token
+		},
+		{
+			"invalid ibc denom treated as valid native denom",
+			func() {
+				req = &types.QueryTotalEscrowForDenomRequest{
+					Denom: "ibc/123",
+				}
+			},
+			true, // the denom is considered a native token
```
colin-axner, line 329: should we indicate why it is considered a native token? ie because the denom trace isn't set?

colin-axner (commented):
Almost done! Thanks for all the work! I only have a few files left to review (migrations). I left many suggestions/comments. Let me know if you have any questions

modules/apps/transfer/keeper/keeper.go
crodriguezvega: True, guess we can remove it.

modules/apps/transfer/keeper/keeper.go
crodriguezvega: It's a regular expression that checks that will check that the key has the expected prefixes and will also extract the denomination out of it. :)

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -167,42 +254,83 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
...
-		}, true, false},
+		{
+			"success receive on source chain",
+			func() {}, true, true,
+		},
+		{
+			"success receive on source chain of half the amount",
+			func() {
+				amount = math.NewInt(50)
+				expEscrowAmount = math.NewInt(50)
```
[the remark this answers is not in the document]
crodriguezvega: If I send back the full amount I wouldn't be sure of whether the subtraction happened correctly when the tokens are sent back or something else is wrong in the code, since there could be a situation where we have a bug in setting the escrow amount both on send and recv and the result would be the same as if it worked (the total amount in escrow at the end if 0).

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -149,16 +157,95 @@ func (suite *KeeperTestSuite) TestSendTransfer() {
 	}
 }
 
+func (suite *KeeperTestSuite) TestSendTransferSetsTotalEscrowAmountForSourceIBCToken() {
```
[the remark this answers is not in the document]
crodriguezvega: Yes, I could look into that, but to be honest in my opinion the `TestSendTransfer` is a bit overloaded already (I was actually thinking of opening a PR with a suggestion to split it, since every time I have to come back to it I have a hard time understanding the tests cases)...

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -167,42 +254,83 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
...
 
 		// - coin being sent back to original chain (chainB)
-		{"tries to unescrow more tokens than allowed", func() {
-			amount = sdk.NewInt(1000000)
-		}, true, false},
+		{
+			"tries to unescrow more tokens than allowed",
+			func() {
+				amount = sdk.NewInt(1000000)
+				expEscrowAmount = math.NewInt(100)
```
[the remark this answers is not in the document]
crodriguezvega: If it tries to unescrow more that the actual amount in escrow, the amount stored will not be changed.

proto/ibc/applications/transfer/v1/query.proto
crodriguezvega: Turns out that it seems to work with a URI like `denoms/{denom}/total_escrow`!

modules/apps/transfer/keeper/keeper.go
crodriguezvega: I share a similar feeling about regular expression :) but I think they also simplify the code quite a bit.

modules/apps/transfer/keeper/genesis_test.go
```diff
@@ -3,35 +3,51 @@ package keeper_test
 import (
 	"fmt"
 
+	"cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
```
damiannolan, line 8: nit: imports

```suggestion
	"cosmossdk.io/math"
	sdk "github.com/cosmos/cosmos-sdk/types"
	
	"github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
```

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -3,8 +3,12 @@ package keeper_test
 import (
 	"fmt"
 
+	"cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	banktestutil "github.com/cosmos/cosmos-sdk/x/bank/testutil"
 	transferkeeper "github.com/cosmos/ibc-go/v7/modules/apps/transfer/keeper"
 	transfertypes "github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
+	ibctesting "github.com/cosmos/ibc-go/v7/testing"
```
damiannolan, line 11: ditto 

```suggestion
	"cosmossdk.io/math"
	sdk "github.com/cosmos/cosmos-sdk/types"
	banktestutil "github.com/cosmos/cosmos-sdk/x/bank/testutil"
	
	transferkeeper "github.com/cosmos/ibc-go/v7/modules/apps/transfer/keeper"
	transfertypes "github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
	ibctesting "github.com/cosmos/ibc-go/v7/testing"
```

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -250,7 +383,22 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
...
+
+			// check total amount in escrow of received token denom on receiving chain
+			var denom string
+			var totalEscrow math.Int
+			if tc.recvIsSource {
+				denom = sdk.DefaultBondDenom
+			} else {
+				denom = types.ParseDenomTrace(
+					types.GetPrefixedDenom(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, sdk.DefaultBondDenom),
+				).IBCDenom()
```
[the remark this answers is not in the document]
crodriguezvega: Because the test exercises `OnRecvPacket` and in that callback it will update the total escrow amount for the received denom in case the receiver chain is the source. Since the received tokens will have an IBC denom and the receiving chain is not the source, then we expect that no update will happen.

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -250,7 +383,22 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
 			data := types.NewFungibleTokenPacketData(trace.GetFullDenomPath(), amount.String(), suite.chainA.SenderAccount.GetAddress().String(), receiver, memo)
 			packet := channeltypes.NewPacket(data.GetBytes(), seq, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, clienttypes.NewHeight(1, 100), 0)
 
-			err = suite.chainB.GetSimApp().TransferKeeper.OnRecvPacket(suite.chainB.GetContext(), packet, data)
+			suite.Require().NotPanics(func() {
+				err = suite.chainB.GetSimApp().TransferKeeper.OnRecvPacket(suite.chainB.GetContext(), packet, data)
+			})
```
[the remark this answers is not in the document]
crodriguezvega: I removed then; they were used before but not anymore.

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -232,6 +361,9 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
 				err = path.RelayPacket(packet)
 				suite.Require().NoError(err) // relay committed
 
+				// set escrow amount that would have been stored after successful execution of MsgTransfer
+				suite.chainB.GetSimApp().TransferKeeper.SetTotalEscrowForDenom(suite.chainB.GetContext(), coinFromBToA.GetDenom(), coinFromBToA.Amount)
+
```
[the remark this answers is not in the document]
crodriguezvega: Yeah, good catch. It's not needed.

crodriguezvega (approved):
Thanks a lot for all your work, @stackman27. We will continue with https://github.com/cosmos/ibc-go/issues/3368 to wrap up the feature.

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -35,6 +36,11 @@ service Query {
   rpc EscrowAddress(QueryEscrowAddressRequest) returns (QueryEscrowAddressResponse) {
     option (google.api.http).get = "/ibc/apps/transfer/v1/channels/{channel_id}/ports/{port_id}/escrow_address";
   }
+
+  // TotalEscrowForDenom returns the total amount of tokens in escrow based on the denom.
+  rpc TotalEscrowForDenom(QueryTotalEscrowForDenomRequest) returns (QueryTotalEscrowForDenomResponse) {
+    option (google.api.http).get = "/ibc/apps/transfer/v1/denoms/{denom=**}/total_escrow";
```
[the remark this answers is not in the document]
crodriguezvega: Parse parameters of the URL template that contain slashes. See [this](https://github.com/cosmos/ibc-go/pull/3045).

PR #3022, opened 2023-01-17
fix: correctly close iterator in `07-tendermint` 

damiannolan:
## Description

- An iterator is reassigned in `07-tendermint` while `defer iterator.Close()` is called on both. The reassignment would mean that the first iterator is not closed cleanly using `defer`.

closes: #XXXX

```bash
fix: correctly close iterator in `07-tendermint` 
```

---

write a little note why.

### DIFF
--- a/modules/light-clients/07-tendermint/store.go
+++ b/modules/light-clients/07-tendermint/store.go
@@ -80,7 +80,6 @@ func IterateConsensusMetadata(store sdk.KVStore, cb func(key, val []byte) bool) 
 	iterator := sdk.KVStorePrefixIterator(store, []byte(host.KeyConsensusStatePrefix))
 
 	// iterate over processed time and processed height
-	defer iterator.Close()
 	for ; iterator.Valid(); iterator.Next() {
 		keySplit := strings.Split(string(iterator.Key()), "/")
 		// processed time key in prefix store has format: "consensusState/<height>/processedTime"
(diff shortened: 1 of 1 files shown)


charleenfei (approved):

modules/light-clients/07-tendermint/store.go
```diff
@@ -80,7 +80,6 @@ func IterateConsensusMetadata(store sdk.KVStore, cb func(key, val []byte) bool)
 	iterator := sdk.KVStorePrefixIterator(store, []byte(host.KeyConsensusStatePrefix))
 
 	// iterate over processed time and processed height
-	defer iterator.Close()
```
chatton, line 83: I think the defer pattern is still better to use, I worry that calling iterator.Close() could later be accidentally re-converted to a defer statement as it is typically more idiomatic. We can make defer version work in a few ways, either create a new variable to defer 

e.g.
```golang
iter := iterator
defer iter.Close()
```

or use a new variable name below rather than re-assign. (I don't see any reason why we would need to re-assign below.)

WDYT?

damiannolan: Nice suggestion, I agree sticking to the defer pattern is probably best. I've pushed an update using a new variable instead of reassignment.

chatton (approved):
LGTM!

AdityaSripal (approved):

Issue #3032, opened 2023-01-20
Add migration handler to enable `09-localhost` client and store sentinel `connectionEnd`


damiannolan:
## Summary

Add a migration handler for chain upgrades to enable the `09-localhost` client and store a sentinel localhost `connectionEnd` in state.

1. Create the `09-localhost` client and set in state using the `02-client` keeper `SetClientState`
2. Create the sentinal localhost `connectionEnd` using the `CreateLocalhost` method added to `03-connection` in https://github.com/cosmos/ibc-go/issues/3031

This can be run as an automatic migration handler by bumping the core IBC module consensus version and registering a new migration in `module.go`.

```go
err := cfg.RegisterMigration(exported.ModuleName, 3, m.Migrate3to4)
if err != nil {
	panic(err)
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

Issue #3031, opened 2023-01-20
Add sentinel localhost `connectionEnd` to genesis state


damiannolan:
## Summary

Similarly to how we create the `09-localhost` client on `InitGenesis` using a simple boolean flag, add a sentinel `connectionEnd` to the ibc store. We can create simple function to store the sentinel localhost connection in the `03-connection` keeper. This can be reused in a migration handler to enable localhost. 

Example:

```go
func (k Keeper) CreateLocalhost(ctx sdk.Context) {
        counterparty := types.NewCounterparty(exported.Localhost, types.LocalhostID, commitmenttypes.NewMerklePrefix(k.GetCommitmentPrefix().Bytes()))
	connectionEnd := types.ConnectionEnd{
		State:        types.OPEN,
		ClientId:     exported.Localhost,
		Versions:     types.ExportedVersionsToProto(types.GetCompatibleVersions()),
		Counterparty: counterparty,
	}

	store := ctx.KVStore(k.storeKey)
	bz := k.cdc.MustMarshal(&connectionEnd)
	store.Set(host.ConnectionKey(types.LocalhostID), bz)
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3040, opened 2023-01-23
Add sentinel localhost connection end to genesis state

chatton:
## Description

closes: #3031
closes: #3032

N/A

entry will be added as part of feature branch

---

write a little note why.


AdityaSripal (approved):
Nice and clean!

chatton:
We discussed making this migration happen automatically, and also removing the boolean field and always creating this connection end in state in `InitGenesis`, let me know if you have any thoughts on these changes @AdityaSripal

modules/core/03-connection/keeper/keeper.go
```diff
@@ -208,6 +208,25 @@ func (k Keeper) GetAllConnections(ctx sdk.Context) (connections []types.Identifi
...
+	return types.ConnectionEnd{
+		State:        types.OPEN,
+		ClientId:     exported.Localhost,
+		Versions:     types.ExportedVersionsToProto(types.GetCompatibleVersions()),
+		Counterparty: counterparty,
+	}
+}
+
+// CreateLocalhostConnectionEnd writes a sentinel localhost ConnectionEnd to state.
+func (k Keeper) CreateLocalhostConnectionEnd(ctx sdk.Context) {
```
crodriguezvega, line 223: Do we really need this function? Cannot we use the existing `SetConnection` and pass as argument the sentinel localhost connection end created by `GetSentinelLocalhostConnectionEnd`?

modules/core/03-connection/keeper/keeper.go
```diff
@@ -208,6 +208,25 @@ func (k Keeper) GetAllConnections(ctx sdk.Context) (connections []types.Identifi
 	return connections
 }
 
+// GetSentinelLocalhostConnectionEnd returns the sentinel localhost connection end.
+func (k Keeper) GetSentinelLocalhostConnectionEnd() types.ConnectionEnd {
```
crodriguezvega, line 212: Maybe also a bit of a nit, but should we rename this to `GenerateSentinelLocalhostConnection` or `CreateSentinelLocalhostConnection` because we usually use the verb `Get` for methods that retrieve data from the store. I would also remove the suffix `End` to keep it consistent with the naming in the rest of the keeper.

modules/core/03-connection/keeper/keeper.go
```diff
@@ -208,6 +208,25 @@ func (k Keeper) GetAllConnections(ctx sdk.Context) (connections []types.Identifi
...
+	counterparty := types.NewCounterparty(exported.Localhost, types.LocalhostID, commitmenttypes.NewMerklePrefix(k.GetCommitmentPrefix().Bytes()))
+	return types.ConnectionEnd{
```
crodriguezvega, line 214: Should we use the constructor function `NewConnectionEnd`?

crodriguezvega (commented):
Thank you, @chatton. I have left some ideas; curious to read your thoughts about them.

modules/core/03-connection/keeper/keeper.go
chatton: Yes this sounds like a good change!

modules/core/03-connection/genesis.go
```diff
@@ -19,6 +19,7 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, gs types.GenesisState) {
 	}
 	k.SetNextConnectionSequence(ctx, gs.NextConnectionSequence)
 	k.SetParams(ctx, gs.Params)
+	k.CreateLocalhostConnectionEnd(ctx)
```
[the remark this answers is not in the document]
chatton: I think this is a good idea! A dedicated function is maybe not required.

modules/core/03-connection/migrations/v7/localhost.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	connectiontypes "github.com/cosmos/ibc-go/v7/modules/core/03-connection/types"
+)
+
+// MigrateLocalhostConnectionEnd creates the sentinel localhost connection end to enable
+// localhost ibc functionality.
+func MigrateLocalhostConnectionEnd(ctx sdk.Context, connectionKeeper ConnectionKeeper) {
```
crodriguezvega, line 11: Sorry, super nit:

```suggestion
func MigrateLocalhostConnection(ctx sdk.Context, connectionKeeper ConnectionKeeper) {
```

modules/core/module.go
```diff
@@ -124,8 +124,8 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	channeltypes.RegisterMsgServer(cfg.MsgServer(), am.keeper)
 	types.RegisterQueryService(cfg.QueryServer(), am.keeper)
 
-	m := clientkeeper.NewMigrator(am.keeper.ClientKeeper)
```
crodriguezvega, line 127: Maybe this is due to my lack of knowledge of how migrations work, but if we replace `Migrate2to3` with `Migration3to4` and a chain upgrades from v6.0.0 to v7.1.0, wouldn't it miss the `Migrate2to3` migration?

chatton: right, my understanding was that chains needed to update to each version one by one in order to run all the migrations. I feel like it should be possible, to just register all migrations at once (as they should be idempotent) however I'm not sure if this is the case. Maybe @AdityaSripal  has an answer to this?

crodriguezvega: Yeah, I am also not sure if the chains really need to upgrade to each version of ibc-go or if they can jump versions as long as they run the migrations of the versions in between.

chatton: if this is the case then we would preferably just add it as a migration instead of replace

crodriguezvega: Yeah, it should be possible [to register multiple migrations](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/12-upgrade.md#registering-migrations).

damiannolan (approved):

PR #3085, opened 2023-01-30
chore: adding initial localhost client testing

damiannolan:
## Description

- Adds unit testing for 09-localhost client
- Fix verify membership/non-membership key paths to omit prefix

closes: #XXXX

```bash
chore: adding initial localhost client testing
```

---

write a little note why.

### DIFF
--- a/modules/light-clients/09-localhost/client_state.go
+++ b/modules/light-clients/09-localhost/client_state.go
@@ -100,6 +100,7 @@ func (cs ClientState) VerifyMembership(
 		return sdkerrors.Wrapf(clienttypes.ErrFailedChannelStateVerification, "todo: update error -- not found for path %s", path)
 	}
 
+	// The commitment prefix (eg: "ibc") is omitted when operating on the core IBC store
 	bz := store.Get([]byte(merklePath.KeyPath[1]))
 	if bz == nil {
 		return sdkerrors.Wrapf(clienttypes.ErrFailedChannelStateVerification, "todo: update error -- not found for path %s", path)
@@ -136,6 +137,7 @@ func (cs ClientState) VerifyNonMembership(
 		return sdkerrors.Wrapf(clienttypes.ErrFailedChannelStateVerification, "todo: update error -- not found for path %s", path)
 	}
 
+	// The commitment prefix (eg: "ibc") is omitted when operating on the core IBC store
 	bz := store.Get([]byte(merklePath.KeyPath[1]))
 	if bz != nil {
 		return sdkerrors.Wrapf(clienttypes.ErrFailedChannelStateVerification, "todo: update error -- found for path %s", path)
--- a/modules/light-clients/09-localhost/client_state.go
+++ b/modules/light-clients/09-localhost/client_state.go
@@ -62,10 +62,17 @@ func (cs ClientState) ZeroCustomFields() exported.ClientState {
 }
 
 // Initialize ensures that initial consensus state for localhost is nil.
-func (cs ClientState) Initialize(_ sdk.Context, _ codec.BinaryCodec, _ sdk.KVStore, consState exported.ConsensusState) error {
+func (cs ClientState) Initialize(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore, consState exported.ConsensusState) error {
 	if consState != nil {
 		return sdkerrors.Wrap(clienttypes.ErrInvalidConsensus, "initial consensus state for localhost must be nil.")
 	}
+
+	cs = ClientState{
+		ChainId:      ctx.ChainID(),
+		LatestHeight: clienttypes.NewHeight(clienttypes.ParseChainID(ctx.ChainID()), uint64(ctx.BlockHeight())),
+	}
+
+	clientStore.Set([]byte(host.KeyClientState), clienttypes.MustMarshalClientState(cdc, &cs))
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


modules/light-clients/09-localhost/client_state.go
```diff
@@ -95,7 +91,17 @@ func (cs ClientState) VerifyMembership(
...
+	if !ok {
+		return sdkerrors.Wrapf(clienttypes.ErrFailedChannelStateVerification, "todo: update error -- not found for path %s", path)
+	}
+
+	// todo: revisit this
+	if len(merklePath.GetKeyPath()) != 2 && merklePath.KeyPath[0] != exported.StoreKey {
+		return sdkerrors.Wrapf(clienttypes.ErrFailedChannelStateVerification, "todo: update error -- not found for path %s", path)
+	}
+
+	bz := store.Get([]byte(merklePath.KeyPath[1]))
```
chatton, line 104: could be a good candidate for an unexported function.
```go
func readPathBytes(path exported.Path,) ([]byte, error) {...}
```

modules/light-clients/09-localhost/client_state.go
```diff
@@ -122,7 +128,17 @@ func (cs ClientState) VerifyNonMembership(
...
```
chatton, line 141: could use the function I suggested above here too

modules/light-clients/09-localhost/client_state.go
```diff
@@ -153,8 +169,10 @@ func (cs ClientState) UpdateStateOnMisbehaviour(_ sdk.Context, _ codec.BinaryCod
 func (cs ClientState) UpdateState(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore, clientMsg exported.ClientMessage) []exported.Height {
 	height := clienttypes.GetSelfHeight(ctx)
 
-	clientState := NewClientState(ctx.ChainID(), height)
-	clientStore.Set([]byte(host.KeyClientState), clienttypes.MustMarshalClientState(cdc, clientState))
+	cs.ChainId = ctx.ChainID()
+	cs.LatestHeight = height
+
+	clientStore.Set([]byte(host.KeyClientState), clienttypes.MustMarshalClientState(cdc, &cs))
```
chatton, line 175: just to make sure I understand the intention correctly, here we modify `cs` the method receiver, but it is not a pointer, so the CientState in the store will be updated, but the instance we act on i.e. 

```go
clientState.UpdateState(...)
```

will remain unchanged.

chatton (commented):
Looking good so far, I'll take a more in depth look at the test cases once the PR is marked ready for review!

modules/light-clients/09-localhost/client_state.go
damiannolan: yeah, correct, this is essentially the same as how the solomachine client state is doing updates also.

modules/light-clients/09-localhost/client_state.go
damiannolan: I think its more or less ok as is, given its a relatively small method compared to solomachine or tm

damiannolan:
Created https://github.com/cosmos/ibc-go/issues/3099 to address error returns in a follow-up PR.

modules/light-clients/09-localhost/client_state.go
```diff
@@ -95,7 +91,16 @@ func (cs ClientState) VerifyMembership(
...
+	merklePath, ok := path.(commitmenttypes.MerklePath)
+	if !ok {
+		return sdkerrors.Wrapf(clienttypes.ErrFailedChannelStateVerification, "todo: update error -- not found for path %s", path)
+	}
+
+	if len(merklePath.GetKeyPath()) != 2 {
+		return sdkerrors.Wrapf(clienttypes.ErrFailedChannelStateVerification, "todo: update error -- not found for path %s", path)
+	}
+
+	bz := store.Get([]byte(merklePath.KeyPath[1]))
```
damiannolan, line 103: `merklePath.KeyPath` is a `[]string` where the first element of the key path `"/ibc/"` (the prefix that is applied [here](https://github.com/cosmos/ibc-go/blob/main/modules/core/03-connection/keeper/verify.go#L177)), but here we pass the core IBC store for verification, so we can trim the prefix or use `keyPath[1]` (i.e. the key path without the prefix)

chatton: I feel like we should include this reasoning in a comment alongside these checks.

chatton (approved):
Really nice work with these tests, and nice job catching the ibc prefix issue! 👍

modules/light-clients/09-localhost/client_state.go
damiannolan: good call, done!

chatton (approved):
ack changes 👍

modules/light-clients/09-localhost/client_state.go
AdityaSripal: Should we verify that the store is prefixed by KeyPath[0]? I don't know if there's an easy way to do that

modules/light-clients/09-localhost/client_state.go
```diff
@@ -153,8 +169,10 @@ func (cs ClientState) UpdateStateOnMisbehaviour(_ sdk.Context, _ codec.BinaryCod
 func (cs ClientState) UpdateState(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore, clientMsg exported.ClientMessage) []exported.Height {
 	height := clienttypes.GetSelfHeight(ctx)
 
-	clientState := NewClientState(ctx.ChainID(), height)
-	clientStore.Set([]byte(host.KeyClientState), clienttypes.MustMarshalClientState(cdc, clientState))
+	cs.ChainId = ctx.ChainID()
```
AdityaSripal, line 172: Maybe should be a separate PR. But do we need the chainID if the function got removed above

AdityaSripal (approved):

PR #3104, opened 2023-02-03
feat: add metadata for IBC tokens

0xmuralik:
## Description

Fixes: [#14181](https://github.com/cosmos/cosmos-sdk/issues/14181) 

See: https://github.com/cosmos/cosmos-sdk/pull/14894

```bash
type: commit message
```

---

write a little note why.


crodriguezvega (commented):
Thanks a lot for this, @0xmuralik! This will be a great UX improvement!

I am wondering if we should have also a migration to add the metadata for the existing denom traces. We can fetch them using [`GetAllDenomTraces`](https://github.com/cosmos/ibc-go/blob/b69b8276d77e7b37633ad99643ef1912ec8f8175/modules/apps/transfer/keeper/keeper.go#L118). That way we would have metadata available already without the need for some tokens to be transferred. What do you think?

modules/apps/transfer/keeper/relay.go
```diff
@@ -282,6 +283,24 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 		return err
 	}
 
+	if !k.bankKeeper.HasDenomMetaData(ctx, voucherDenom) {
+		metadata := banktypes.Metadata{
+			Description: getMetaDataDescription(denomTrace),
+			DenomUnits: []*banktypes.DenomUnit{
+				{
+					Denom:    denomTrace.BaseDenom,
```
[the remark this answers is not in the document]
0xmuralik: VoucherDenom is the IBCHash, so wouldn't it better to have a readable for DenomUnits? Maybe we can use the full path here if needed. Instead of having a seperate hash for each unit we will have paths like...
`
{
    Denom: transfer/channel-10/foobar
    Exponent: 6
},
{
    Denom: transfer/channel-10/ufoobar
    Exponent: 0
}
`

crodriguezvega:
> Sounds good. We can set MetaData in the initgenesis [here](https://github.com/cosmos/ibc-go/blob/ceb9464ced75d4c64b3a71abf6ceab4a50e908be/modules/apps/transfer/keeper/genesis.go#L16) after `SetDenomTrace`.

Yes, true, we should also do it in `InitGenesis`!

But wouldn't we also need a migration handler for chain upgrades? We could run this as an automatic migration handler by bumping the consensus version of transfer.

modules/apps/transfer/keeper/migrations.go
```diff
@@ -43,6 +43,11 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
 			if !equalTraces(newTrace, dt) {
 				newTraces = append(newTraces, newTrace)
 			}
+
+			if !m.keeper.bankKeeper.HasDenomMetaData(ctx, newTrace.IBCDenom()) {
+				m.keeper.SetDenomMetaData(ctx, newTrace)
+			}
+
```
colin-axner, line 50: This migration may have already been run and might not be run again. I think it is safer to make a separate migration only setting the denom metadata. Please let me know if you have any questions on how to do this. I'm also happy to receive that change in a followup pr, it would make reviews faster/easier

modules/apps/transfer/keeper/relay.go
```diff
@@ -310,6 +315,38 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
...
+				Denom:    denomTrace.BaseDenom,
+				Exponent: 0,
+			},
+		},
+		// Setting base as IBChash Denom as SetDenomMetaData uses Base as storeKey
+		// and the bank keeper will only have the IBCHash to get the denom metadata
+		Base:    denomTrace.IBCDenom(),
+		Display: denomTrace.GetFullDenomPath(),
+		Name:    getMetaDataName(denomTrace),
+		Symbol:  getMetaDataSymbol(denomTrace),
```
colin-axner, line 332: Can we have an example print out of how this might look? Would be great if we could get product input on the desired formatting? cc @womensrights @adiraviraj @tmsdkeys

colin-axner (commented):
Thanks @0xmuralik! :)

modules/apps/transfer/keeper/relay.go
colin-axner: For context, this pr would add metadata about IBC tokens into the state machine. This means when clients query a bank balance, instead of only seeing `ibc/{hash}` they could also provide a flag to obtain back the metadata information stored here, such as the base denomination or the full history of this tokens hops (via channel/portIDs)

modules/apps/transfer/keeper/migrations.go
0xmuralik: I have created a new migration and registered it with as version 3 migration. Feel free to take up the task if required.

modules/apps/transfer/keeper/relay.go
0xmuralik: I added a unit test for the new migration. The example metadata is available [here](https://github.com/0xmuralik/ibc-go/blob/dafd5f1cbbfaf77b3326ba7bdabffd480ccb80e7/modules/apps/transfer/keeper/migrations_test.go#L137).

0xmuralik:
@colin-axner @crodriguezvega If there are any additional changes required please let me know. Else fell free to take up this task in a different PR and close this one if necesssary.

modules/apps/transfer/keeper/relay.go
tmsdkeys: I've known people asking for this on Discord, to have the metadata about IBC tokens as you would regular sdk tokens so think it's a good idea. Wrt the formatting, looks okay for me. The `Base` field should normally equal `DenomUnits.Denom` with `Exponent` according to the bank module's metadata, but the reasoning is provided why it needs to deviate. 
Rest looks okay for me. Maybe @womensrights has insights from user research?

crodriguezvega:
> @colin-axner @crodriguezvega If there are any additional changes required please let me know. Else fell free to take up this task in a different PR and close this one if necesssary.

Thank you for all the work on this one, @0xmuralik. Yes, if there's any more changes we need, we can let you know or take over. We should also be able to push changes directly to your branch.

modules/apps/transfer/keeper/migrations.go
```diff
@@ -53,6 +54,19 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
 	return nil
 }
 
+// MigrateMetaData sets token metadata for all the IBC denom traces
```
damiannolan, line 57:
```suggestion
// MigrateMetadata sets token metadata for all the IBC denom traces
```

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -3,6 +3,7 @@ package keeper_test
 import (
 	"fmt"
 
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	transferkeeper "github.com/cosmos/ibc-go/v7/modules/apps/transfer/keeper"
 	transfertypes "github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
```
damiannolan, line 8: nit: import separation

```suggestion
	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
	
	transferkeeper "github.com/cosmos/ibc-go/v7/modules/apps/transfer/keeper"
	transfertypes "github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
```

modules/apps/transfer/types/expected_keepers.go
```diff
@@ -25,6 +26,8 @@ type BankKeeper interface {
 	SendCoinsFromAccountToModule(ctx sdk.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error
 	BlockedAddr(addr sdk.AccAddress) bool
 	IsSendEnabledCoin(ctx sdk.Context, coin sdk.Coin) bool
+	HasDenomMetaData(ctx sdk.Context, denom string) bool
+	SetDenomMetaData(ctx sdk.Context, denomMetaData banktypes.Metadata)
```
damiannolan, line 30: I think it would good to change all instances of `MetaData` to `Metadata`

modules/apps/transfer/module.go
```diff
@@ -107,6 +107,10 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	if err := cfg.RegisterMigration(types.ModuleName, 1, m.MigrateTraces); err != nil {
 		panic(fmt.Sprintf("failed to migrate transfer app from version 1 to 2: %v", err))
 	}
+
+	if err := cfg.RegisterMigration(types.ModuleName, 2, m.MigrateMetadata); err != nil {
+		panic(fmt.Sprintf("failed to migrate transfer app from version 2 to 3: %v", err))
+	}
```
damiannolan, line 113: we'll be running some migrations for total escrow denoms maybe we could include denom metadata migrations within the same consensus version bump. cc @crodriguezvega

modules/apps/transfer/types/expected_keepers.go
0xmuralik: I changed the ones decalred within IBC. But didn't change the methods from bankkeeper.

modules/apps/transfer/keeper/relay.go
womensrights: I think this is a great first step, but also adding in a human readable name to the current output would precisely cover what the metadata is used for. Front end developers are currently maintaining their own registries of assets to be displayed for their application. An example of such a registry is [for Injective](https://github.com/InjectiveLabs/utils-go/blob/master/pkg/token/token_meta.json#L25-L36). Front ends need:

- token symbol, this is covered
- the location of the decimal place for the denom, I think this is covered from the exponent, from my understanding, e.g. the precision of ATOM is 6 dp, ATOM exponent is 6, uATOM is 0. In the example output given its not clear though as you have used an exponent of 0 for ubar and foo, foo should have exponent 6 assuming it is an sdk coin? 
- human readable name, having the path is good, but then you would need to check which chain the token came from additionally to make this human readable

So I think this is already an improvement, just if there was also a way to add an additional human readable name somehow then this would be great but I'm not sure this is easily done. Also as a sidenote, I know there is a constraint on the metadata from the sdk types so this is not easy to change, but for the ibc denoms, you don't get any new information from looking at the `Display` compared to `Name` so in the long term I would question the need to have both

crodriguezvega: Thank you for the input, @womensrights!

- Regarding the exponent, if I am not mistaken I think at the moment it will be difficult to set it to the correct value, since this metadata is set on the receiving chain and the [`FungibleTokenPacketData` does not contain that information at the moment](https://github.com/cosmos/ibc-go/blob/cf519bc4d4285aa0b69511ba08d69b7456744f5d/proto/ibc/applications/transfer/v2/packet.proto#L10). However, [with channel upgradability it would be possible to extend the packet data and add extra information](https://github.com/cosmos/ibc/discussions/719).
- Regarding the lack of chain ID. Maybe the sending chain's chain ID could be added in the memo field? Not that I particularly want to continue adding more information in the memo file :), but maybe it's a solution for now. Your comment `if there was also a way to add an additional human readable name somehow` refers to adding the chain ID, right?

womensrights: For the exponent, as you linked the `FungibleTokenPacketData` there isn't explicitly an exponent field but where does the logic to know the decimal place for the amount in the packet for a specific denom come from in this case? 

The chainID would also not completely work as sometimes the chainID is not the same as what is considered the human branded name for the chain or you would need to remove the additional numbers. e.g. I think bandchain has chainID "laozi-mainnet" and the human readable name would be Bandchain

crodriguezvega: > For the exponent, as you linked the FungibleTokenPacketData there isn't explicitly an exponent field but where does the logic to know the decimal place for the amount in the packet for a specific denom come from in this case?

The amount comes from the [sdk Coin that is used in `MsgTransfer`](https://github.com/cosmos/ibc-go/blob/main/proto/ibc/applications/transfer/v1/tx.proto#L38). [Whatever that amount is, we put it in the `FungibleTokenPacketData`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/keeper/relay.go#LL132C1-L132C1), but we don't have any information about the exponent.

> The chainID would also not completely work as sometimes the chainID is not the same as what is considered the human branded name for the chain or you would need to remove the additional numbers. e.g. I think bandchain has chainID "laozi-mainnet" and the human readable name would be Bandchain

True, good point! Then I guess we cannot do anything about this at the moment?

womensrights: ok I see, I guess [this adr](https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-024-coin-metadata.md) has not yet been implemented and right now its assumed that sdk coins all have 6 decimal places.

Yes I don't think there can be anything done about the chainIDs

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -235,3 +236,82 @@ func (suite *KeeperTestSuite) TestMigrateTotalEscrowForDenom() {
...
+		denomTrace       transfertypes.DenomTrace
+		expectedMetadata banktypes.Metadata
+	)
+
+	testCases := []struct {
+		msg      string
+		malleate func()
+	}{
+		{
+			"success with denom trace with one hop",
```
colin-axner, line 251: needs test cases for:
- multiple denom traces
- 1 already set denom trace with metadata and 1 without metadata set

modules/apps/transfer/keeper/relay.go
```diff
@@ -277,6 +278,10 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 		return errorsmod.Wrap(err, "failed to mint IBC tokens")
 	}
 
+	if !k.bankKeeper.HasDenomMetaData(ctx, voucherDenom) {
```
colin-axner, line 281: if we are running migrations such that all traces have metadata set, shouldn't we just make setting metadata a functionality of `SetDenomTrace`? This avoids accidental incidents where the trace is set but the metadata isn't. There should never be an instance when you want to set the denom trace without setting the metadata as well?

modules/apps/transfer/keeper/relay.go
```diff
@@ -423,3 +428,37 @@ func (k Keeper) DenomPathFromHash(ctx sdk.Context, denom string) (string, error)
 	fullDenomPath := denomTrace.GetFullDenomPath()
 	return fullDenomPath, nil
 }
+
+// SetDenomMetadata sets an IBC token's denomination metadata
+func (k Keeper) SetDenomMetadata(ctx sdk.Context, denomTrace types.DenomTrace) {
```
colin-axner, line 433: I would recommend moving to keeper.go and making the function private

modules/apps/transfer/keeper/relay.go
```diff
@@ -423,3 +428,37 @@ func (k Keeper) DenomPathFromHash(ctx sdk.Context, denom string) (string, error)
...
+		Description: getMetadataDescription(denomTrace),
+		DenomUnits: []*banktypes.DenomUnit{
+			{
+				Denom:    denomTrace.BaseDenom,
+				Exponent: 0,
+			},
+		},
+		// Setting base as IBC hash denom since bank keepers's SetDenomMetaData uses
+		// Base as storeKey and the bank keeper will only have the IBC hash to get
+		// the denom metadata
```
colin-axner, line 444:
```suggestion
		// Base as storeKey and the IBC hash is what gives this token uniqueness 
		// on the executing chain
```

modules/apps/transfer/keeper/migrations.go
colin-axner, line 1: Requires e2e test as well. Is there an issue for this already? (I'd recommend writing the test in a followup pr)

modules/apps/transfer/keeper/genesis.go
```diff
@@ -14,6 +14,7 @@ func (k Keeper) InitGenesis(ctx sdk.Context, state types.GenesisState) {
 
 	for _, trace := range state.DenomTraces {
 		k.SetDenomTrace(ctx, trace)
+		k.SetDenomMetadata(ctx, trace)
```
colin-axner, line 17: tests should be modified to assert that metadata is set

modules/apps/transfer/keeper/relay.go
```diff
@@ -277,6 +278,10 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 		return errorsmod.Wrap(err, "failed to mint IBC tokens")
 	}
 
+	if !k.bankKeeper.HasDenomMetaData(ctx, voucherDenom) {
+		k.SetDenomMetadata(ctx, denomTrace)
```
colin-axner, line 282: tests should be modified to assert metadata is set

colin-axner (commented):
LGTM but see comments. More test assertions are needed and I'd recommend blending the added functionality into the existing `SetDenomTrace` function to avoid accidental bugs in the future due to not calling both `SetDenomTrace` and `SetMetadata`

modules/apps/transfer/keeper/migrations.go
crodriguezvega: Yeap, there is [this one](https://github.com/cosmos/ibc-go/issues/3768).

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -385,22 +404,25 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
...
-			var (
-				denom       string
-				totalEscrow sdk.Coin
-			)
-			if tc.recvIsSource {
-				denom = sdk.DefaultBondDenom
-			} else {
-				denom = trace.IBCDenom()
-			}
-			totalEscrow = suite.chainB.GetSimApp().TransferKeeper.GetTotalEscrowForDenom(suite.chainB.GetContext(), denom)
```
crodriguezvega, line 397: I realised that this could be simplified. It was actually testing the wrong thing, I think, but looks like by chance the tests passed...

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -321,7 +323,7 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
 		{
 			"failure: receive on module account",
 			func() {
-				receiver = suite.chainA.GetSimApp().AccountKeeper.GetModuleAddress(types.ModuleName).String()
+				receiver = suite.chainB.GetSimApp().AccountKeeper.GetModuleAddress(types.ModuleName).String()
```
crodriguezvega, line 326: This was also wrong, but it worked because the module address of transfer for chain A and chain B is the same.

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -385,22 +404,25 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
...
+					suite.Require().False(found)
+				} else {
+					denomMetadata, found := suite.chainB.GetSimApp().BankKeeper.GetDenomMetaData(suite.chainB.GetContext(), denomTraceOnB.IBCDenom())
+					suite.Require().True(found)
+					suite.Require().Equal(expDenomMetadataOnB, denomMetadata)
+				}
 			} else {
 				suite.Require().Error(err)
+
+				_, found := suite.chainB.GetSimApp().BankKeeper.GetDenomMetaData(suite.chainB.GetContext(), denomTraceOnB.IBCDenom())
```
crodriguezvega, line 424: There is one test failing for this check. But I need some help to understand why.

modules/apps/transfer/keeper/relay.go
crodriguezvega: Added the assertions.

modules/apps/transfer/keeper/migrations_test.go
crodriguezvega: Added the tests.

modules/apps/transfer/keeper/relay.go
```diff
@@ -277,6 +277,10 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 		return errorsmod.Wrap(err, "failed to mint IBC tokens")
 	}
 
+	if !k.bankKeeper.HasDenomMetaData(ctx, voucherDenom) {
```
colin-axner, line 280: Not strictly related to my proposal of setting denom trace and metadata simultaneously, I realize this code placement is quite odd. It is in between two actions that must be atomic (mint and send minted tokens to receiver).  I would prefer it to go right after the `HasDenomTrace` function

modules/apps/transfer/keeper/relay_test.go
colin-axner: Checks on errors are unnecessary as state reversals will occur by baseapp. I removed the assertion, I don't think this situation justifies the extra hassle

modules/apps/transfer/keeper/relay.go
crodriguezvega: I moved the code up. If it's ok with you, maybe we can address in a different issue/PR performing `SetDenomTrace` and `setDenomMetadata` actions in the same function call.

colin-axner: sure I don't feel strongly about it

modules/apps/transfer/keeper/genesis_test.go
```diff
@@ -48,10 +48,16 @@ func (suite *KeeperTestSuite) TestGenesis() {
...
+	suite.Require().Equal(denomTraces.Sort(), genesis.DenomTraces)
 	suite.Require().Equal(escrows.Sort(), genesis.TotalEscrowed)
 
 	suite.Require().NotPanics(func() {
 		suite.chainA.GetSimApp().TransferKeeper.InitGenesis(suite.chainA.GetContext(), *genesis)
 	})
+
+	ctx := suite.chainA.GetContext()
+	for _, denomTrace := range denomTraces {
+		_, found := suite.chainA.GetSimApp().BankKeeper.GetDenomMetaData(ctx, denomTrace.IBCDenom())
```
damiannolan, line 60: `ctx` var feels redundant given that we pass `suite.chainA.GetContext()` directly above in many other places

```suggestion
	for _, denomTrace := range denomTraces {
		_, found := suite.chainA.GetSimApp().BankKeeper.GetDenomMetaData(suite.chainA.GetContext(), denomTrace.IBCDenom())
```

modules/apps/transfer/keeper/keeper.go
```diff
@@ -190,6 +191,40 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
+	metadata := banktypes.Metadata{
+		Description: getMetadataDescription(denomTrace),
+		DenomUnits: []*banktypes.DenomUnit{
+			{
+				Denom:    denomTrace.BaseDenom,
+				Exponent: 0,
+			},
+		},
+		// Setting base as IBC hash denom since bank keepers's SetDenomMetadata uses
+		// Base as storeKey and the IBC hash is what gives this token uniqueness
```
damiannolan, line 205: nit: might be nice to differentiate between store key and key path. 

```suggestion
		// Base as key path and the IBC hash is what gives this token uniqueness
```

modules/apps/transfer/module.go
```diff
@@ -126,6 +126,10 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	if err := cfg.RegisterMigration(types.ModuleName, 3, m.MigrateParams); err != nil {
 		panic(fmt.Sprintf("failed to migrate transfer app version 3 to 4 (self-managed params migration): %v", err))
 	}
+
+	if err := cfg.RegisterMigration(types.ModuleName, 4, m.MigrateDenomMetadata); err != nil {
+		panic(fmt.Sprintf("failed to migrate transfer app from version 4 to 5 (set denom metadata migration): %v", err))
+	}
```
damiannolan, line 132: We could wrap both params migration and denom metadata in an anonymous func here, but having two separate also works. 

Happy to keep as is if everyone is fine with bumping consensus version twice in the one release version.

damiannolan (approved):
LGTM! Looks good to me, left some suggestions which were only nits

modules/apps/transfer/module.go
colin-axner: I'm fine keeping as is. One benefit is that it makes it quite clear which migration is doing what

modules/apps/transfer/keeper/migrations.go
```diff
@@ -65,6 +66,21 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
 	return nil
 }
 
+// MigrateMigrateDenomMetadataMetadata sets token metadata for all the IBC denom traces
+func (m Migrator) MigrateDenomMetadata(ctx sdk.Context) error {
+	m.keeper.IterateDenomTraces(ctx,
+		func(dt types.DenomTrace) (stop bool) {
+			// check if the metadata for the given denom trace already exists
```
chatton, line 73:
```suggestion
			// check if the metadata for the given denom trace does not already exist.
```

modules/apps/transfer/module.go
damiannolan: Agreed! Will resolve this comment :)

modules/apps/transfer/keeper/migrations.go
```diff
@@ -65,6 +66,21 @@ func (m Migrator) MigrateTraces(ctx sdk.Context) error {
 	return nil
 }
 
+// MigrateMigrateDenomMetadataMetadata sets token metadata for all the IBC denom traces
```
chatton, line 69:
```suggestion
// MigrateDenomMetadata sets token metadata for all the IBC denom traces.
```

modules/apps/transfer/keeper/keeper.go
```diff
@@ -190,6 +191,40 @@ func (k Keeper) IterateDenomTraces(ctx sdk.Context, cb func(denomTrace types.Den
...
+	return fmt.Sprintf("IBC token from %s", denomTrace.GetFullDenomPath())
+}
+
+func getMetadataName(denomTrace types.DenomTrace) string {
+	return fmt.Sprintf("%s IBC token", denomTrace.GetFullDenomPath())
+}
+
+func getMetadataSymbol(denomTrace types.DenomTrace) string {
+	return strings.ToUpper(denomTrace.BaseDenom)
+}
```
chatton, line 226: [nit] I don't really think these fns are adding much value. They are only used here, I think we could just inline this.

chatton (approved):
LGTM thanks @0xmuralik, just left a few small comments but it looks great 👍

crodriguezvega:
Thanks for the reviews, @colin-axner, @damiannolan and @chatton.

Issue #3034, opened 2023-01-20
Add `09-localhost` client state implementation


damiannolan:
## Summary

Based on the work in https://github.com/cosmos/ibc-go/pull/2700, add a new client state implementation for `09-localhost`. 
The localhost client should be accompanied by a sentinel localhost `connectionEnd` which should be stored in state upon `InitGenesis` or enabed via a chain upgrade migration handler. 

Update the `09-localhost` client in https://github.com/cosmos/ibc-go/pull/2700 to conform to the latest `ClientState` implementation in ibc-go/v7. A feature branch has been created [here](https://github.com/cosmos/ibc-go/compare/09-localhost?expand=1). 

The body of work is detailed in the following issues:

- [x] https://github.com/cosmos/ibc-go/issues/3056
- [x] https://github.com/cosmos/ibc-go/issues/3058
- [x] https://github.com/cosmos/ibc-go/issues/3031
- [x] https://github.com/cosmos/ibc-go/issues/3032
- [x] https://github.com/cosmos/ibc-go/issues/3099
- [x] https://github.com/cosmos/ibc-go/issues/3033
- [x] https://github.com/cosmos/ibc-go/issues/3055
- [x] https://github.com/cosmos/ibc-go/issues/2537
- [x] https://github.com/cosmos/ibc-go/issues/3167
- [x] https://github.com/cosmos/ibc-go/issues/3166
- [x] #3193
- [x] https://github.com/cosmos/ibc-go/issues/3194
- [x] https://github.com/cosmos/ibc-go/issues/3206
- [x] https://github.com/cosmos/ibc-go/issues/3208

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3114, opened 2023-02-07
chore: disallow localhost client creation in 02-client

damiannolan:
## Description

- This PR prevents creation of `09-localhost` clients in `02-client`

ref: #3034 

```bash
chore: disallow localhost client creation in 02-client
```

---

write a little note why.

### DIFF
--- a/modules/core/02-client/keeper/client.go
+++ b/modules/core/02-client/keeper/client.go
@@ -17,7 +17,7 @@ func (k Keeper) CreateClient(
 	ctx sdk.Context, clientState exported.ClientState, consensusState exported.ConsensusState,
 ) (string, error) {
 	params := k.GetParams(ctx)
-	if !params.IsAllowedClient(clientState.ClientType()) {
+	if !params.IsAllowedClient(clientState.ClientType()) || clientState.ClientType() == exported.Localhost {
 		return "", sdkerrors.Wrapf(
 			types.ErrInvalidClientType,
 			"client state type %s is not registered in the allowlist", clientState.ClientType(),
--- a/modules/core/02-client/keeper/client.go
+++ b/modules/core/02-client/keeper/client.go
@@ -16,8 +16,12 @@ import (
 func (k Keeper) CreateClient(
 	ctx sdk.Context, clientState exported.ClientState, consensusState exported.ConsensusState,
 ) (string, error) {
+	if clientState.ClientType() == exported.Localhost {
+		return "", sdkerrors.Wrapf(types.ErrInvalidClientType, "cannot create client of type: %s", clientState.ClientType())
+	}
+
 	params := k.GetParams(ctx)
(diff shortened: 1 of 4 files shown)


modules/core/02-client/keeper/client_test.go
```diff
@@ -13,22 +13,39 @@ import (
...
+			"success: 07-tendermint client type supported",
+			ibctm.NewClientState(testChainID, ibctm.DefaultTrustLevel, trustingPeriod, ubdPeriod, maxClockDrift, testClientHeight, commitmenttypes.GetSDKSpecs(), ibctesting.UpgradePath),
+			suite.consensusState,
+			true,
+		},
+		{
+			"success: 06-solomachine client type supported",
+			solomachine.NewClientState(0, &solomachine.ConsensusState{PublicKey: suite.solomachine.ConsensusState().PublicKey, Diversifier: suite.solomachine.Diversifier, Timestamp: suite.solomachine.Time}),
+			&solomachine.ConsensusState{PublicKey: suite.solomachine.ConsensusState().PublicKey, Diversifier: suite.solomachine.Diversifier, Timestamp: suite.solomachine.Time},
+			true,
```
damiannolan, line 37: I've adapted the tests here slightly. This one was actually setup incorrectly and the case was failing on `clientState.Initialise()` due to non-matching consensus state, rather than it being an unsupported client. Solomachine is now in the list of allowed clients so this test case should pass.

damiannolan:
Initially I removed the localhost client type from `AllowedClients` but noticed some failing tests as this is used in `InitGenesis`. I decided to keep localhost as an allowed client so genesis can continue to work as normal but added an additional check explicitly for client creation.

modules/core/02-client/keeper/client.go
```diff
@@ -17,7 +17,7 @@ func (k Keeper) CreateClient(
 	ctx sdk.Context, clientState exported.ClientState, consensusState exported.ConsensusState,
 ) (string, error) {
 	params := k.GetParams(ctx)
-	if !params.IsAllowedClient(clientState.ClientType()) {
+	if !params.IsAllowedClient(clientState.ClientType()) || clientState.ClientType() == exported.Localhost {
 		return "", sdkerrors.Wrapf(
 			types.ErrInvalidClientType,
 			"client state type %s is not registered in the allowlist", clientState.ClientType(),
```
crodriguezvega, line 23: Maybe a nit, but wouldn't this error be a bit misleading for localhost? I mean, the reader of the error would see that localhost is in the allow list...

damiannolan: Yes, I suppose it would! 

We can either change the error msg or put the localhost check in its own conditional, what do you think?

AdityaSripal (approved):

modules/core/02-client/keeper/client.go
damiannolan: Updated!

crodriguezvega (approved):
Thank you, @damiannolan!

PR #3161, opened 2023-02-16
imp: improve validateHeight function naming and code documentation

colin-axner:
## Description

While removing the localhost addition diffs, I came across this nit. 

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/core/03-connection/keeper/handshake.go
+++ b/modules/core/03-connection/keeper/handshake.go
@@ -74,7 +74,7 @@ func (k Keeper) ConnOpenTry(
 	// generate a new connection
 	connectionID := k.GenerateConnectionIdentifier(ctx)
 
-	if err := k.validateHeight(ctx, clientState.ClientType(), consensusHeight); err != nil {
+	if err := k.validateConsensusHeight(ctx, clientState.ClientType(), consensusHeight); err != nil {
 		return "", err
 	}
 
@@ -160,7 +160,7 @@ func (k Keeper) ConnOpenAck(
 	consensusHeight exported.Height, // latest height of chainA that chainB has stored on its chainA client
 ) error {
 	// Check that chainB client hasn't stored invalid height
-	if err := k.validateHeight(ctx, clientState.ClientType(), consensusHeight); err != nil {
+	if err := k.validateConsensusHeight(ctx, clientState.ClientType(), consensusHeight); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


modules/core/03-connection/keeper/handshake.go
```diff
@@ -288,11 +288,16 @@ func (k Keeper) ConnOpenConfirm(
...
+// of this chain's consensus state. For non-localhost clients, the consensus height must be a past height.
+// The localhost consensus height must be less than or equal to the current height.
+func (k Keeper) validateConsensusHeight(
 	ctx sdk.Context,
 	clientType string,
 	consensusHeight exported.Height,
 ) error {
+	// if the consensusHeight is greater than the current height,
+	// the self consensus state will fail to be retrieved.
 	if clientType == exported.Localhost {
```
colin-axner, line 301: I think I'd actually prefer to check the height still in this case. It isn't necessary, but I think it is good to avoid skipping checks altogether for the localhost, it makes more sense to apply the logic of the check even if it is adjust just slightly (in this case, the current height is a valid consensus height). How would people feel about me adjust this code?

modules/core/03-connection/keeper/handshake.go
```diff
@@ -74,7 +74,7 @@ func (k Keeper) ConnOpenTry(
 	// generate a new connection
 	connectionID := k.GenerateConnectionIdentifier(ctx)
 
-	if err := k.validateHeight(ctx, clientState.ClientType(), consensusHeight); err != nil {
```
colin-axner, line 77: I found this function name to vague. This function only makes sense when validating the consensus height stored on the counterparty chain's light client

modules/core/03-connection/keeper/handshake.go
damiannolan: Sounds good to me

colin-axner: If we update the connection handshake to disallow working for the localhost client, we can remove this entirely?

damiannolan: Yeah. We can revert the addition of this `validateHeight` function back to the original code if you like!

colin-axner: Yea I think I'll revert the changes just because this check is a bit naunced and not reusable after these two checks

damiannolan (approved):

chatton (approved):
LGTM

Issue #2537, opened 2022-10-12
Allow localhost enabling to be a parameter


AdityaSripal:
Perhaps we should add a parameter that allows chains to enable or disable localhost on their chain

This would then be checked in the localhost verification case to ensure it is enabled. Localhost verification should fail if the parameter is set to false

EDIT from @chatton 

Additional tasks to be completed after the internal audit:

- [x] Use allowed clients list to dynamically enable/disable and client type.
- [x] Add Unauthorized status for light clients. This status should be returned if the client type is not in the allowed_list
- [x] Add light client status checks in ChanOpenInit and ConnOpenInit.

colin-axner:
There already is a parameter, clsoing

colin-axner:
One possible approach: 
- add a enable/disable field to client state
- add msg to change this value (gov v1 msg, authority is only gov module) 
- update localhost client state upon execution
- use client state field in `Status`

damiannolan:
I think in order to do gov v1 message execution it will incur an API breaking change to the ibc core `NewKeeper()` function. 

SDK modules which have migrated to use self managed params now take the additional arg to `NewKeeper()` functions: `authority` and store it as a field on the keeper.

[See here](https://github.com/cosmos/cosmos-sdk/blob/main/x/auth/keeper/keeper.go#L69) and below where the additional `authority` arg is added

colin-axner:
Could we preemptively get the API breaking change in v7? Or does it feel too late in the cycle for that?

damiannolan:
Sounds good to me. I think if its just changing the core `NewKeeper` function to accept an additional arg and store it as a field on the keeper then its totally fine

chatton:
closing this as #3196 has been merged into the feature branch.

Issue #2010, opened 2022-08-15
Migrate params to be self managed


colin-axner:
## Summary

Migrate all our on-chain params to be self managed, the SDK is removing x/param

See example [pr](https://github.com/cosmos/cosmos-sdk/pull/12363)

See upgrading [docs](https://github.com/cosmos/cosmos-sdk/blob/main/UPGRADING.md#modules)

Ensure unset module params don't cause a panic on genesis export

The module will be removed after April 18th, 2023

### Issues:

- [x] #3500
- [x] #3501
- [x] #3502
- [x] #3503
- [x] #3504
- [ ] ~~#3621~~ Since there is no timeline for this yet, I'll remove it as requirement to close the epic issue.
- [x] #3704
- [x] #3705
- [ ] Add param e2es to compatibility tests.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Looking at the [PR for migrating `x/mint`](https://github.com/cosmos/cosmos-sdk/pull/12363) it looks like we need to:

In `tx.proto`:

- Add an [`UpdateParams` RPC endpoint](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/mint/v1beta1/tx.proto#L20) to update params that receives a [`MsgUpdateParams`](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/mint/v1beta1/tx.proto#L26) and returns a  [`MsgUpdateParamsResponse`](https://github.com/cosmos/cosmos-sdk/blob/main/proto/cosmos/mint/v1beta1/tx.proto#L43). Add the `Msg` service if needed.

In `types/msgs.go`:
- [Implement `sdk.Msg` for `MsgUpdateParams`](https://github.com/cosmos/cosmos-sdk/blob/main/x/mint/types/msgs.go).

In `keeper/msg_server.go`:
- Implement [RPC endpoint for `UpdateParams`](https://github.com/cosmos/cosmos-sdk/blob/main/x/mint/keeper/msg_server.go#L27).

In `types/keys.go`:
- Add key [`ParamsKey`](https://github.com/cosmos/cosmos-sdk/blob/main/x/mint/types/keys.go#L6). This is the key that will be used to store the params.

In `keeper/keeper.go`:
[...]

PR #3178, opened 2023-02-21
Enable or disable localhost client via allowed_clients

chatton:
## Description

closes: #2537

The issue referenced mentions adding a new `enabled` field for localhost client, however this didn't fit super cleanly with the existing fields as we already have a list of allowed clients. This PR instead uses the presence/absence of the localhost client ID in this array to return the `Status` of the client as `Active`/`Frozen` respectively. 

N/A (included in feature release notes)

---

write a little note why.

- [x] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/10-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).


modules/core/02-client/keeper/keeper.go
```diff
@@ -31,10 +34,11 @@ type Keeper struct {
 	paramSpace    paramtypes.Subspace
 	stakingKeeper types.StakingKeeper
 	upgradeKeeper types.UpgradeKeeper
+	authority     string
 }
 
 // NewKeeper creates a new NewKeeper instance
-func NewKeeper(cdc codec.BinaryCodec, key storetypes.StoreKey, paramSpace paramtypes.Subspace, sk types.StakingKeeper, uk types.UpgradeKeeper) Keeper {
+func NewKeeper(cdc codec.BinaryCodec, key storetypes.StoreKey, paramSpace paramtypes.Subspace, sk types.StakingKeeper, uk types.UpgradeKeeper, authority string) Keeper {
```
damiannolan, line 41: If we're going to go ahead with this approach we should cherry pick this API breaking change out to a separate PR that we can backport to v7

chatton: do you just mean the method signature change? Not the `MsgUpdateParams` ?

damiannolan: Yeah, just the method signature change and addition of the `authority` field on the keeper struct

modules/core/02-client/keeper/keeper_test.go
```diff
@@ -461,3 +463,73 @@ func (suite KeeperTestSuite) TestIterateClientStates() { //nolint:govet // this
...
+			msgUpdateParams := &types.MsgUpdateParams{}
+			tc.malleate(msgUpdateParams)
+
+			_, err := ibcKeeper.ClientKeeper.UpdateParams(ctxA, msgUpdateParams)
+
+			clientParams = ibcKeeper.ClientKeeper.GetParams(ctxA)
+			cs = suite.chainA.GetClientState(exported.Localhost)
+			localhostClientState = cs.(*localhost.ClientState)
+
+			suite.Require().Equal(clientParams.IsAllowedClient(exported.Localhost), localhostClientState.Enabled)
```
crodriguezvega, line 526: Wouldn't it be more safe to check both `clientParams.IsAllowedClient(exported.Localhost)` and `localhostClientState.Enabled` against an expected value? Otherwise we could have the situation were `UpdateParams` doesn't do what it's supposed to do when disabling the params (neither params get updated nor client state gets updated) and the test would still pass.

proto/ibc/lightclients/localhost/v2/localhost.proto
```diff
@@ -13,4 +13,7 @@ message ClientState {
 
   // the latest block height
   ibc.core.client.v1.Height latest_height = 1 [(gogoproto.nullable) = false];
+
+  // whether or not the localhost client should be enabled
```
crodriguezvega, line 17:
```suggestion
  // whether or not the localhost client is enabled
```

modules/core/02-client/keeper/keeper.go
```diff
@@ -403,3 +408,25 @@ func (k Keeper) ClientStore(ctx sdk.Context, clientID string) sdk.KVStore {
...
 }
+
+// UpdateParams implements MsgServer.UpdateParams method.
+// It defines a method to update the 02-client module parameters.
+func (k Keeper) UpdateParams(goCtx context.Context, req *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
+	if k.authority != req.Authority {
+		return nil, errors.Wrapf(govtypes.ErrInvalidSigner, "invalid authority; expected %s, got %s", k.authority, req.Authority)
+	}
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	k.SetParams(ctx, req.Params)
```
crodriguezvega, line 419: This is a nit, but should we move this call to the end once all the checks have passed?

modules/light-clients/09-localhost/client_state.go
```diff
@@ -57,11 +61,7 @@ func (cs ClientState) Initialize(ctx sdk.Context, cdc codec.BinaryCodec, clientS
 		return sdkerrors.Wrap(clienttypes.ErrInvalidConsensus, "initial consensus state for localhost must be nil.")
 	}
 
-	clientState := ClientState{
```
crodriguezvega, line 60: Why was this removed?

crodriguezvega (commented):
Nice work, @chatton! Some of this work can also partially count for https://github.com/cosmos/ibc-go/issues/2010 :)

I think we should add in this PR or in the one @damiannolan is working on documentation for enabling/disabling the localhost.

modules/light-clients/09-localhost/client_state.go
chatton: Because `Initialize` is an interface function, and previously we were not actually using the method receiver. The usage looks like

```go
func (k Keeper) CreateLocalhostClient(ctx sdk.Context) error {
	clientState := localhost.NewClientState(types.GetSelfHeight(ctx), k.GetParams(ctx).IsAllowedClient(exported.Localhost))
	return clientState.Initialize(ctx, k.cdc, k.ClientStore(ctx, exported.Localhost), nil)
}
```

This enables us to configure the clientState.Enabled field based on the `allowed_clients` list.

modules/core/02-client/keeper/keeper_test.go
chatton: yes I think that's a good idea!

modules/core/02-client/keeper/keeper.go
chatton: Sure I think that makes sense!

damiannolan:
I think it would be good to update the localhost client update logic [here](https://github.com/cosmos/ibc-go/blob/09-localhost/modules/core/02-client/keeper/keeper.go#L64-L66) to check the status. This would prevent unnecessary writes to state when the localhost client is disabled.

```go
func (k Keeper) UpdateLocalhostClient(ctx sdk.Context, clientState exported.ClientState) []exported.Height {
	if status := clientState.Status(ctx, k.ClientStore(ctx, exported.Localhost), k.cdc); status == exported.Active {
		clientState.UpdateState(ctx, k.cdc, k.ClientStore(ctx, exported.Localhost), nil)
	}

	return nil
}
```

chatton:
closing in favour of #3196

Issue #2896, opened 2022-12-07
migrate error packages


tac0turtle:
## Summary

Cosmos SDK has spun out the errors package into its own go.mod in light of the recent conversation on reducing dependency on the sdk, ibc-go should be migrated to cosmossdk.io/errors to take advantage of this. 

We did not move the default error types over cause the idea is that users/app devs will create their own with their custom codespace. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
> We did not move the default error types over cause the idea is that users/app devs will create their own with their custom codespace.

Should ibc-go not be using the default types from the sdk root codespace then, such as `ErrInvalidCoins`, `ErrInvalidAddress` etc?

PR #3184, opened 2023-02-21
Add IBC module level errors

chatton:
## Description

part of: #2896

This PR duplicates the errors found [here](https://github.com/cosmos/cosmos-sdk/blob/main/types/errors/errors.go) and uses the `ibc` as the module name. They will be used in a follow up PR.

N/A

---

write a little note why.

### DIFF
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -6,137 +6,139 @@ import (
 	"github.com/cosmos/ibc-go/v7/modules/core/exported"
 )
 
+const codespace = exported.ModuleName
+
 var (
 	// ErrTxDecode is returned if we cannot parse a transaction.
-	ErrTxDecode = errorsmod.Register(exported.ModuleName, 2, "tx parse error")
+	ErrTxDecode = errorsmod.Register(codespace, 2, "tx parse error")
 
 	// ErrInvalidSequence is used the sequence number (nonce) is incorrect
 	// for the signature.
-	ErrInvalidSequence = errorsmod.Register(exported.ModuleName, 3, "invalid sequence")
+	ErrInvalidSequence = errorsmod.Register(codespace, 3, "invalid sequence")
 
 	// ErrUnauthorized is used whenever a request without sufficient
 	// authorization is handled.
-	ErrUnauthorized = errorsmod.Register(exported.ModuleName, 4, "unauthorized")
+	ErrUnauthorized = errorsmod.Register(codespace, 4, "unauthorized")
 
 	// ErrInsufficientFunds is used when the account cannot pay requested amount.
-	ErrInsufficientFunds = errorsmod.Register(exported.ModuleName, 5, "insufficient funds")
+	ErrInsufficientFunds = errorsmod.Register(codespace, 5, "insufficient funds")
 
 	// ErrUnknownRequest is used when the request body.
-	ErrUnknownRequest = errorsmod.Register(exported.ModuleName, 6, "unknown request")
+	ErrUnknownRequest = errorsmod.Register(codespace, 6, "unknown request")
 
 	// ErrInvalidAddress is used when an address is found to be invalid.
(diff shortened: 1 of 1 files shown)


damiannolan:
What do you think about assigning the ibc module name to a `codespace` constant and using that in the calls to `Register`?

```go
const codespace = exported.ModuleName
```

chatton:
> What do you think about assigning the ibc module name to a `codespace` constant and using that in the calls to `Register`?
> 
> ```go
> const codespace = exported.ModuleName
> ```

don't feel strongly either way! Happy to make this change if you think it makes things clearer

internal/errors/errors.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// ErrNotFound defines an error when requested entity doesn't exist in the state.
+	ErrNotFound = errorsmod.Register(codespace, 38, "not found")
+
+	// ErrIO should be used to wrap internal errors caused by external operation.
+	// Examples: not DB domain error, file writing etc...
+	ErrIO = errorsmod.Register(codespace, 39, "Internal IO error")
+
+	// ErrAppConfig defines an error occurred if min-gas-prices field in BaseConfig is empty.
+	ErrAppConfig = errorsmod.Register(codespace, 40, "error in app.toml")
```
colin-axner, line 137: Should we consider pruning down the errors that aren't applicable to us? I don't think we need to copy over all of them, but I don't have a strong opinion on the matter. I think the idea is that these are now our errors

chatton: yeah I was thinking we could do that. I'll try and remove all the ones we don't use anywhere in the codebase.

chatton: question: Will the error code need to remain the same or can it change? Will changing the error code potentially be state machine breaking?

chatton: I'm thinking specifically about [this](https://github.com/cosmos/ibc-go/blob/b40dbc646c7191aaee12d3a70243f9b6f3f80919/modules/core/04-channel/types/acknowledgement.go#L35-L41) function.

colin-axner: It can change. Error codes are state machine breaking, but we don't plan to backport this to a minor release? If included in a major release, it won't be an issue. It does not need to maintain parity with the SDK errors. I believe just changing the codespace will change the ABCI error code anyways

internal/errors/errors.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/cosmos/ibc-go/v7/modules/core/exported"
+)
+
+const codespace = exported.ModuleName
+
+var (
+	// ErrInvalidSequence is used the sequence number (nonce) is incorrect
+	// for the signature.
+	ErrInvalidSequence = errorsmod.Register(codespace, 1, "invalid sequence")
```
colin-axner, line 14: A couple years ago, it'd be the case that if you used error code 1, the error message would get redacted and replaced by "internal error". I'm unsure if this has changed, but it would be good to double check

chatton: Thanks for the heads up I will look into it

colin-axner: Looks like it just got changed https://github.com/cosmos/cosmos-sdk/pull/11960/files#diff-cbd3ee4d9e83821aab653ac4a36825cf310ecc8fc8d225024e3f74a305621ba3L124

colin-axner: looks like it got fixed in beta7 release https://github.com/cosmos/cosmos-sdk/blob/ce25dd2cc101737959e38fd888bbd4765c14836e/errors/abci.go#L46

colin-axner: So I think this should be just fine :+1:

colin-axner (approved):
Looks great to me. Thanks @chatton :slightly_smiling_face:

damiannolan (approved):
Will be great to have deprecation warnings gone after this migration! Thanks ❤️

PR #3196, opened 2023-02-23
Allow disabling of client type via allowed_clients param

chatton:
## Description

This PR directly replaces #3194

closes: #2537

N/A included in feature branch

---

write a little note why.

### DIFF
--- a/docs/ibc/light-clients/client-state.md
+++ b/docs/ibc/light-clients/client-state.md
@@ -29,7 +29,7 @@ implementer is in charge of determining which checks are required. See the [Tend
 - An `Expired` status indicates that a client is not allowed to be used because it was not updated for longer than the trusting period.
 - An `Unknown` status indicates that there was an error in determining the status of a client.
 - An `Unauthorized` status indicates that usage of the client is not allowed.
-  - > Note: Removing a client type from the [`AllowedClients`](https://github.com/cosmos/ibc-go/blob/v6.0.0/modules/core/02-client/types/client.pb.go#L345) array will prevent all usage of that client type.
+  - > Removing a client type from the [`AllowedClients`](https://github.com/cosmos/ibc-go/blob/v6.0.0/modules/core/02-client/types/client.pb.go#L345) array will prevent all usage of that client type.
 
 All possible `Status` types can be found [here](https://github.com/cosmos/ibc-go/blob/v6.0.0/modules/core/exported/client.go#L26-L36).
 
--- a/docs/ibc/light-clients/client-state.md
+++ b/docs/ibc/light-clients/client-state.md
@@ -28,8 +28,6 @@ implementer is in charge of determining which checks are required. See the [Tend
 - A `Frozen` status indicates that misbehaviour was detected in the counterparty chain and the client is not allowed to be used.
 - An `Expired` status indicates that a client is not allowed to be used because it was not updated for longer than the trusting period.
 - An `Unknown` status indicates that there was an error in determining the status of a client.
-- An `Unauthorized` status indicates that usage of the client is not allowed.
-  - > Removing a client type from the [`AllowedClients`](https://github.com/cosmos/ibc-go/blob/v6.0.0/modules/core/02-client/types/client.pb.go#L345) array will prevent all usage of that client type.
 
 All possible `Status` types can be found [here](https://github.com/cosmos/ibc-go/blob/v6.0.0/modules/core/exported/client.go#L26-L36).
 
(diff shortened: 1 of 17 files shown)


modules/core/02-client/keeper/keeper.go
```diff
@@ -403,3 +403,12 @@ func (k Keeper) ClientStore(ctx sdk.Context, clientID string) sdk.KVStore {
 	clientPrefix := []byte(fmt.Sprintf("%s/%s/", host.KeyClientStorePrefix, clientID))
 	return prefix.NewStore(ctx.KVStore(k.storeKey), clientPrefix)
 }
+
+// GetClientStatus returns the status for a given clientState. If the client type is not in the allowed
+// clients param field, Unauthorized is returned, otherwise the client state status is returned.
+func (k Keeper) GetClientStatus(ctx sdk.Context, clientState exported.ClientState, clientID string) exported.Status {
```
colin-axner, line 409: Should we add documentation somewhere that removing clients from the `AllowedClients` parameter will disable all usage of that client type?

colin-axner (approved):
This is wonderful! So fast! We should probably add a changelog entry for this and I think it would be good to add a little docs on this functionality. It might be unclear when to use `clientState.Status()` vs `clientKeeper.GetClientStatus`

modules/core/02-client/keeper/keeper.go
chatton: Yes I think we can improve this docstring a bit!

damiannolan (approved):
Beautiful, super clean PR! LGTM 🙌

docs/ibc/light-clients/client-state.md
```diff
@@ -28,6 +28,8 @@ implementer is in charge of determining which checks are required. See the [Tend
 - A `Frozen` status indicates that misbehaviour was detected in the counterparty chain and the client is not allowed to be used.
 - An `Expired` status indicates that a client is not allowed to be used because it was not updated for longer than the trusting period.
 - An `Unknown` status indicates that there was an error in determining the status of a client.
+- An `Unauthorized` status indicates that usage of the client is not allowed.
+  - > Note: Removing a client type from the [`AllowedClients`](https://github.com/cosmos/ibc-go/blob/v6.0.0/modules/core/02-client/types/client.pb.go#L345) array will prevent all usage of that client type.
```
chatton, line 32: Added a note in the client-state.md doc. 

cc @colin-axner

colin-axner: Hmm, the only issue with the doc here is that this is for the `clientState.Status` message. Light client developers don't need to use the `Unauthorized` status

chatton: That is a very good point! Let me see if I can find a better place for this.

docs/ibc/light-clients/client-state.md
```diff
@@ -28,6 +28,8 @@ implementer is in charge of determining which checks are required. See the [Tend
 - A `Frozen` status indicates that misbehaviour was detected in the counterparty chain and the client is not allowed to be used.
 - An `Expired` status indicates that a client is not allowed to be used because it was not updated for longer than the trusting period.
 - An `Unknown` status indicates that there was an error in determining the status of a client.
+- An `Unauthorized` status indicates that usage of the client is not allowed.
```
crodriguezvega, line 31: The formatting of the rendered text looks a bit weird with the `- >`. Was its use intentional?

crodriguezvega (commented):
Nice work, @chatton!

Should we add some tests to exercise the new conditional checks added to `ChanOpenInit` and `ConnOpenInit`?

damiannolan:
Could we add the status check before updating the localhost client in `BeginBlocker` in this PR?

docs/ibc/light-clients/client-state.md
chatton: this was intentional to add a `note` section, but I think Colin brought up a good point that this is not the best place for these docs.

modules/core/04-channel/keeper/handshake.go
```diff
@@ -50,6 +51,15 @@ func (k Keeper) ChanOpenInit(
 		)
 	}
 
+	clientState, found := k.clientKeeper.GetClientState(ctx, connectionEnd.ClientId)
+	if !found {
+		return "", nil, sdkerrors.Wrapf(clienttypes.ErrClientNotFound, "clientID (%s)", connectionEnd.ClientId)
+	}
+
+	if status := k.clientKeeper.GetClientStatus(ctx, clientState, connectionEnd.ClientId); status != exported.Active {
```
AdityaSripal, line 59: Do we need this for ChanCloseInit as well?

AdityaSripal (commented):
I think we need a check on ChanCloseInit as well correct?

Also does the current test coverage cover the new k.ClientStatus function completely? Would be good to ensure we get an error if we try to call a method on a client that has a valid internal status but is not part of the AllowedClients params

chatton:
Based on PR feedback, I've 
* added a new Troubleshooting section to the docs
* added test cases which explicitly remove the tendermint client from the allow list
* added the check to `ChanCloseInit`
* added the check in `BeginBlocker`

AdityaSripal (approved):
Thanks for the quick fixes @chatton !

colin-axner (approved):
Excellent!

damiannolan (approved):

PR #3214, opened 2023-02-27
[POC] Use a hooks interface to support executing custom logic after client operations

chatton:
## Description

This PR enables hooks to allow for arbitrary code execution upon client events (upgrade/create/update)

ref: https://github.com/cosmos/ibc-go/issues/2880

```bash
type: commit message
```

---

write a little note why.


modules/core/02-client/types/exp/hooks.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// Hooks specifies methods to enable writing custom logic which executes
+// after specific client lifecycle methods.
+// Note: this interface is experimental and could be removed in favour of a different solution, requiring code changes.
+type Hooks interface {
+	// OnClientCreated is executed upon client creation.
+	OnClientCreated(ctx sdk.Context, clientID string) error
+	// OnClientUpdated is executed when a client is updated.
+	OnClientUpdated(ctx sdk.Context, clientID string, consensusHeights []exported.Height) error
```
AdityaSripal, line 16:
```suggestion
	OnClientUpdated(ctx sdk.Context, clientID string, clientMsg exported.ClientMessage, consensusHeights []exported.Height) error
```

modules/core/02-client/types/exp/hooks.go
```diff
@@ -0,0 +1,56 @@
...
+	// OnClientUpgraded is executed when a client is upgraded.
+	OnClientUpgraded(ctx sdk.Context, clientID string) error
+}
```
AdityaSripal, line 19: I think a hook for Misbehaviour is also necessary. CCV has expressed interest in something like this

`OnClientMisbehaviour(ctx sdk.Context, clientID string, clientMsg exported.ClientMessage)`

AdityaSripal (commented):
Very in favor of this direction!!!

modules/core/02-client/types/exp/hooks.go
AdityaSripal: This is so the hook also has access to the header which in the tendermint case has the unhashed validator set as well as other potentially useful information like `DataHash`

modules/core/02-client/types/exp/hooks.go
AdityaSripal, line 1: I think `exported` is a better name than `exp`

chatton: this was `experimental` rather than `exported`

AdityaSripal: Is the goal to eventually have a release where this is under the experimental directory?

chatton: We can name the directory to be more explicit. Although this PR as it currently exists is not sufficient to achieve what is being achieved by using an `expected keepers` interface. Hooks functionality ensures core ibc logic executes, but in the interface implementation, that logic is being changed i.e. clients on *not* frozen when misbehaviour is detected. We still need to figure out the best way of enabling this.

AdityaSripal: Theoretically, if the hook has access to the client keeper; they can simply unfreeze the client again. Though perhaps that's a bit hacky

chatton: [this](https://github.com/cosmos/ibc-go/pull/2967#issuecomment-1458949771) comment gives more details about possible issues.

crodriguezvega:
Thanks for drafting this alternative solution, @chatton. Closing the PR for now for the same reason as [here](https://github.com/cosmos/ibc-go/pull/2967#issuecomment-1582036087).

Issue #2827, opened 2022-11-24
Investigate non-determinstic failures in e2e


colin-axner:
## Summary

Occasionally the e2e's will fail [unexpectedly](https://github.com/cosmos/ibc-go/actions/runs/3530533337/jobs/5922676431). The behaviour appears to be non-determinstic and only occur around 5% of the time. It is unpredictable which test will fail. As we increase our reliance on the e2e's, we should also ensure we understand why this instances are failing and fix the issue if possible. [Here](https://github.com/cosmos/ibc-go/actions/runs/3530940914/jobs/5923576191) is another example of a different test failing which typically succeeds

If we make our e2e's [importable via a workflow](https://github.com/cosmos/ibc-go/issues/2501), it would be less than ideal if our tests randomly failed

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chatton:
I think the first link was an issue with the specific test [which  has been resolved](https://github.com/cosmos/ibc-go/pull/2858), however the second link is the real problem. The periodic network failures during set up. These are relatively consistent.

colin-axner:
Looking at this [failure](https://github.com/cosmos/ibc-go/actions/runs/4383152405/jobs/7673162930) with tm logs on debug as well as this [failure](https://github.com/cosmos/ibc-go/actions/runs/4378231603/jobs/7663140896), I noticed something really intersting. 

The failing node contains 38 or 39 instances of:
```
Inbound Peer rejected                        module=p2p err="auth failure: secret conn failed: proto: illegal wireType 7" numPeers=4
```

All other nodes have instances of:
```
D[2023-03-10|09:42:20.344] served RPC HTTP response                     module=rpc-server method=POST url=/ status=200 duration=0 remoteAddr=172.18.0.1:34828
```
while the failing node does not. 


It appears to me that the HTTP request to query the node status is somehow ending up on the p2p port, which tendermint then attempts to perform the secret handshake used for encrypted communication. Obviously the response expected back is an HTTP response and not a encoded public key, thus causing a failure. What I'm really not certain of is how this network routing failure is occurring? Is interchain test returning the p2p port on accident? Is docker to blame? Why do we see this issue so randomly? The depth of this sudoku puzzle continues...

colin-axner:
Opened an issue on [cometbft](https://github.com/cometbft/cometbft/issues/516)

[...]

PR #3234, opened 2023-03-02
fix: address non-deterministic e2e failures

colin-axner:
## Description

this is an attempt to fix the non-deterministic failures we are seeing on our e2e's. With the help of @chatton, we believe we solved an issue I was hitting locally where interchain test would accidentally retry executing a docker command on a container via a race condition. My interchain test fork just increases the timeout to make the race condition less likely. We think this may then be downstream affecting chain startup by a chain having a malformed config setup by accidentally having executed a command twice, but we aren't certain, so this just gives us a shot to see if this is the case or not

see https://github.com/colin-axner/interchaintest/commit/8516da44071163a620ede42096583834f834e846

From running CI first commit
**Passes: 4**
**Failures: 1**

From running CI second commit
**Passes: 1**
**Failures: 1**

Looking at the logs on the second failure, it seems 2 of the nodes just begin to stop and shut down. Maybe this is a resource issue, but I'm a bit confused why each node is able to do a clean shutdown without any abrupt errors?

ref: #2827 

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -234,3 +234,5 @@ replace (
 replace github.com/cosmos/ibc-go/v7 => ../
 
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+
+replace github.com/strangelove-ventures/interchaintest/v7 => github.com/colin-axner/interchaintest/v7 v7.0.0-20230302165704-8516da440711
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -235,4 +235,4 @@ replace github.com/cosmos/ibc-go/v7 => ../
 
 replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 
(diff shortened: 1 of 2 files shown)


colin-axner:
Looks like there is still a [failure](https://github.com/cosmos/ibc-go/actions/runs/4316420981/jobs/7533777415). We should get the logs from the failing chain to see why the chain stopped running. 

As far as this fix goes https://github.com/colin-axner/interchaintest/commit/8516da44071163a620ede42096583834f834e846, I think probably we should try to reuse the existing context deadline rather than creating a new one with an arbitrary value. My understanding of why a timeout is used is in the case you never get a response from the server, in which case you'd want to timeout. This [case](https://github.com/colin-axner/interchaintest/blob/main/internal/dockerutil/startcontainer.go#L25) is handling the scenario when the request timed out, but the caller still has provided more time for the request attempt to go through. It's unclear to me if there is a scenario in which retrying a request should succeed? (which is why maybe we should change it to reuse the caller context deadline until we hit a scenario where it fails)

Issue #3241, opened 2023-03-06
add import/export genesis e2e


colin-axner:
## Summary

Add an e2e test for init and export genesis. 

## Proposal

- startup chains
- create channel/send packets
- halt chain, export genesis, write file
- restart chain
- send packet over existing channel

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

ThanhNhann:
can I handle this ser?

colin-axner:
Hi! Yes you can! That would be wonderful!

I've started looking into this so here's some more information:

I think we should add the test into the `e2e/tests/upgrades` directory under a `genesis_test.go` file. We should create a new `TestGenesisTestSuite` as the upgrades one is skipped on regular pr's to ibc-go. 
```go
func TestGenesisTestSuite(t *testing.T) {
    suite.Run(t, new(GenesisTestSuite))
}

type GenesisTestSuite struct {
    testsuite.E2ETestSuite
}
```

We will need to make modifications to interchain test to complete this issue. I've opened this issue https://github.com/strangelove-ventures/interchaintest/issues/418. You are welcome to take that on if you'd like, but I had planned to give it a shot tomorrow on my fork as I know the changes I'd like to do so I don't think it'll take me very long. 

The changes will add a function to the `CosmosChain` struct which will allow us to set the `halt-height` parameter in the `app.toml`. This would be the function we set on the cosmos chains we create. I think we could set the function by adding it as a field to `testconfig.ChainOptionConfiguration` and then calling `GetChains(opts)` before `SetupChainsRelayersAndChannel`
```go
func setHaltHeight(node *cosmos.ChainNode) error {
    appToml := make(testutil.Toml)

    appToml["halt-height"] = haltHeight

    return testutil.ModifyTomlConfigFile(
        ctx, 
        node.logger(),
        node.DockerClient,
        node.TestName,
        node.VolumeName,
        "config/app.toml",
        appToml,   
    )    
}
```
see similar [code](https://github.com/strangelove-ventures/interchaintest/blob/main/chain/cosmos/chain_node.go#L255)

For the genesis import/export test we would set the halt height to X so all nodes would halt at X. Once all the nodes are halted, we should export state from each validator. We can use exec cmd `<bin> export`. I'm not sure if there's an rpc we could call instead?

[...]

PR #3282, opened 2023-03-11
e2e: add export genesis e2e

ThanhNhann:
## Description
- This pr add export genesis e2e

closes: #3241 

---

write a little note why.


e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,159 @@
...
+	err = test.WaitForBlocks(timeoutCtx, int(blocksAfterUpgrade), chain)
+	s.Require().NoError(err, "chain did not produce blocks after halt")
+
+	height, err = chain.Height(ctx)
+	s.Require().NoError(err, "error fetching height after halt")
+
+	s.Require().Greater(height, haltHeight, "height did not increment after halt")
+}
+
+func WriteFile(path string, body []byte) error {
```
chatton, line 153: Hi @ThanhNhann, this function is going to write the contents of the genesis file to the host running `go test` not to the container where the chain is running, so the chains will not see this genesis file.

It looks like we don't have a function to copy a file **to** a container but [here](https://github.com/cosmos/ibc-go/blob/b4503f4b8f910d46e4c69fff8ea13e91812d8a75/e2e/dockerutil/dockerutil.go#L51) is a function we use to copy a file from a container.

You will likely need to write the contents of the file into all of the containers using the docker client in a manner similar to this.

ThanhNhann: thank @chatton, it helps me a lot

chatton: I will also add we want to make sure we don't re-import `tendermint` as we are now using `comeft`. In order to copy the the contents, it might make sense to just load the file as a `[]byte` using the function I linked and then write the file using a new function without needing to use `tmjson` at all.

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,159 @@
...
+import (
+	"context"
+	"os"
+	"time"
+	"encoding/json"
+	"testing"
+
+	"github.com/cosmos/ibc-go/e2e/testconfig"
+	"github.com/cosmos/ibc-go/e2e/testsuite"
+	tmjson "github.com/tendermint/tendermint/libs/json"
```
colin-axner, line 12: I think we can use [cometbft](https://github.com/cometbft/cometbft/tree/main/libs/json)?

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	configFileOverrides := make(map[string]any)
+	appTomlOverrides := make(test.Toml)
+	configFileOverrides["halt-height"] = haltHeight
+
+	testconfig.DefaultChainOptions().ChainAConfig.ConfigFileOverrides = appTomlOverrides
+
+	ctx := context.Background()
+	relayer, channelA := s.SetupChainsRelayerAndChannel(ctx)
+	chainA, chainB := s.GetChains()
```
colin-axner, line 41: I believe this isn't working as intended since the chain config instance returned by a call to `testconfig.DefaultChainOptions()` will be different than the instance later called by `SetupChainsRelayerAndChannel`. The fix is fairly simple though:
```suggestion
	configFileOverrides := make(map[string]any)
	appTomlOverrides := make(test.Toml)
	
	appTomlOverrides["halt-height"] = haltHeight
	configFileOverrides["config/app.toml"] = appTomlOverrides

	chainOpts := func(options *ChainOptions) {
	    options.ChainAConfig.ConfigFileOverrides = configFileOverrides
	}
	
	// create chains with specified chain configuration options
	chainA, chainB := s.GetChains(chainOpts)
	
	ctx := context.Background()
	relayer, channelA := s.SetupChainsRelayerAndChannel(ctx)
```

I would recommend committing these changes locally as I haven't tried to format/compile

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 5, chainA, chainB), "failed to wait for blocks")
+
+	t.Run("Halt chain and export genesis", func(t *testing.T) {
+		s.HaltChainAndExportGenesis(ctx, chainA, int64(haltHeight))
+	})
+
+	t.Run("restart relayer", func(t *testing.T) {
+		s.StopRelayer(ctx, relayer)
+		s.StartRelayer(relayer)
```
colin-axner, line 95: Sorry, I have forgotten why we need to stop/start the relayer?

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,159 @@
...
+	})
+
+	t.Run("native IBC token transfer from chainA to chainB, sender is source of tokens", func(t *testing.T) {
+		transferTxResp, err := s.Transfer(ctx, chainA, chainAWallet, channelA.PortID, channelA.ChannelID, testvalues.DefaultTransferAmount(chainADenom), chainAAddress, chainBAddress, s.GetTimeoutHeight(ctx, chainB), 0, "")
+		s.Require().NoError(err)
+		s.AssertValidTxResponse(transferTxResp)
+	})
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 5, chainA, chainB), "failed to wait for blocks")
```
colin-axner, line 104: I think we can added some commented out code with a TODO linking to an issue tracking the functionality of testing relaying after genesis restart, might be a nice way to encourage relayers to support the functionality by showing we have a test to see if it works as expected. I can open a pr to your branch with what I'm thinking if that's okay?

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,159 @@
...
+func (s *GenesisTestSuite) HaltChainAndExportGenesis(ctx context.Context, chain *cosmos.CosmosChain, haltHeight int64) {
+	var genesisState GenesisState
+	
+	timeoutCtx, timeoutCtxCancel := context.WithTimeout(ctx, time.Minute*2)
+	defer timeoutCtxCancel()
+
+	height, err := chain.Height(ctx)
+	s.Require().NoError(err, "error fetching height before halt")
+	
+	err = test.WaitForBlocks(timeoutCtx, int(haltHeight - int64(height)) +1, chain)
```
colin-axner, line 116: What's the +1 for? The chain should halt at the halt height I believe?

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,159 @@
...
+	s.Require().NoError(err)
+	err = chain.StopAllNodes(ctx)
+	s.Require().NoError(err, "error stopping node(s)")
+
+	state, err := chain.ExportState(ctx, int64(haltHeight-int64(height))+1)
```
colin-axner, line 121: I think we can export state at `haltHeight` (this would be the height we continue from when the chain restarts)

colin-axner (commented):
Great work so far @ThanhNhann :pray: The test flow was really easy to follow. I left a few suggestions that should hopefully get the test closer to passing. Let us know if you run into any issues

e2e/tests/upgrades/genesis_test.go
ThanhNhann: Yes sure, go ahead sir

e2e/tests/upgrades/genesis_test.go
ThanhNhann: that's my mistake, updated

e2e/tests/upgrades/genesis_test.go
```diff
@@ -32,20 +32,23 @@ func (s *GenesisTestSuite) TestIBCGenesis() {
...
-
+	
+	appTomlOverrides["halt-height"] = haltHeight
+	configFileOverrides["config/app.toml"] = appTomlOverrides
+	chainOpts := func(options * testconfig.ChainOptions) {
+	    options.ChainAConfig.ConfigFileOverrides = configFileOverrides
+	}
+	
+	// create chains with specified chain configuration options
+	chainA, chainB := s.GetChains(chainOpts)
```
chatton, line 43: I think this is the first time we'll have a test adding additional configuration to chain options, this code should work, if the options aren't being applied as expected I can help look into why 👍

ThanhNhann: thank @chatton

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,152 @@
...
+	timeoutCtx, timeoutCtxCancel := context.WithTimeout(ctx, time.Minute*2)
+	defer timeoutCtxCancel()
+
+	err := test.WaitForBlocks(timeoutCtx, int(haltHeight), chain)
+	s.Require().Error(err, "chain did not halt at halt height")
+
+	err = chain.StopAllNodes(ctx)
+	s.Require().NoError(err, "error stopping node(s)")
+
+	state, err := chain.ExportState(ctx, int64(haltHeight))
```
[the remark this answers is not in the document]
chatton: Sorry @ThanhNhann , I completely missed this comment.

I haven't personally used that function before, but briefly looking at the implementation

```go
func (tn *ChainNode) ExportState(ctx context.Context, height int64) (string, error) {
	tn.lock.Lock()
	defer tn.lock.Unlock()

	_, stderr, err := tn.ExecBin(ctx, "export", "--height", fmt.Sprint(height))
	if err != nil {
		return "", err
	}
	// output comes to stderr for some reason
	return string(stderr), nil
}
```

It looks like the function unexpectedly was displaying to stderr, it might be possible that it is now correctly being displayed to stdout and is being discarded here.

🤔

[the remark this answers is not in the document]
chatton: I believe this is [what is happening](https://github.com/cosmos/cosmos-sdk/blob/e7097b146830edb541f47382e54d9ba1fd60d085/server/export.go#L99-L103)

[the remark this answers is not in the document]
ThanhNhann: Thank @chatton, I think your comment above is total true, maybe we should make a pr to interchain test for this

e2e/dockerutil/dockerutil.go
```diff
@@ -70,3 +71,37 @@ func GetFileContentsFromContainer(ctx context.Context, dc *dockerclient.Client,
...
+	}
+	return nil
+}
+
+//chainAbsoluteGenesisFilePaths get absolute path of Genesis file of a chain
+func chainAbsoluteGenesisFilePaths(cfg ibc.ChainConfig) string {
+	return fmt.Sprintf("/var/cosmos-chain/%s/config/genesis.json", cfg.Name)
+}
+
+//getDockerContainerID get docker container id
```
colin-axner, line 94:
```suggestion
// getDockerContainerID get docker container id
```

e2e/tests/upgrades/genesis_test.go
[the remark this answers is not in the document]
colin-axner: Has this occurred yet? Are changes required for interchain test?

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	s.Require().NoError(err)
+	err = tmjson.Unmarshal([]byte(state), &genesisState)
+	s.Require().NoError(err)
+	genesisJson, err := tmjson.MarshalIndent(genesisState, "", "  ")
```
colin-axner, line 123: genesisJSON?

colin-axner (commented):
Wow, fantastic work @ThanhNhann! The code is looking really good! Apologies for taking so long to get back

e2e/dockerutil/dockerutil.go
```diff
@@ -70,3 +71,37 @@ func GetFileContentsFromContainer(ctx context.Context, dc *dockerclient.Client,
...
+	err = dc.CopyToContainer(ctx, containerID, genesisFilePath, content, options)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+//chainAbsoluteGenesisFilePaths get absolute path of Genesis file of a chain
+func chainAbsoluteGenesisFilePaths(cfg ibc.ChainConfig) string {
+	return fmt.Sprintf("/var/cosmos-chain/%s/config/genesis.json", cfg.Name)
```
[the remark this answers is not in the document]
chatton: there are some [internal functions](https://github.com/strangelove-ventures/interchaintest/blob/330268c42d5ce6a245fef7254e0fbd6bf172adc6/chain/internal/tendermint/tendermint_node.go#L101) that actually do what we are trying to do. I would be happy to leave this as is, and maybe we could open an issue to export some of these internal utility functions that would be quite useful for some of these more complex tests.

e2e/tests/upgrades/genesis_test.go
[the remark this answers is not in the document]
ThanhNhann: hi @colin-axner, I made a pr for the issue above and merged https://github.com/strangelove-ventures/interchaintest/pull/475, and it's ok now sir

e2e/dockerutil/dockerutil.go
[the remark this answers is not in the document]
ThanhNhann: Yeah, I checked and saw the relevant func above at [here](https://github.com/strangelove-ventures/interchaintest/blob/330268c42d5ce6a245fef7254e0fbd6bf172adc6/chain/cosmos/chain_node.go#L153), but it's the internal func so I can't use it, I'm asking about the change now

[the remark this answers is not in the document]
ThanhNhann: I'm just making a [pull request](https://github.com/strangelove-ventures/interchaintest/pull/553) for this

anhductn2001:
Hi all, I am testing with this E2E test and I realize that after halt-chain I restart the node again, the node will generate one more block then continue halt. I tried on different chains and same result.
The solution I am thinking of now is that after halt-chain we reconfigure halt-height = 0. 
Anyone have any ideas? Thanks!! ( and I'm having a little trouble configuring the param in app.toml)

colin-axner:
Apologies for the delayed reply @anhductn2001. That sounds correct to me! Where does the node get started again in this e2e? We could overwrite the app.toml file like we do with genesis before restarting or we could potentially pass the halt height as a flag when restarting the node

e2e/dockerutil/dockerutil.go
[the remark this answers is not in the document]
DimitrisJim: just a quick ping, seems like that PR ^ was merged right?

anhductn2001:
Can you approve CI again @chatton ? Tks

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,155 @@
...
+	s.Require().Error(err, "chain did not halt at halt height")
+
+	err = chain.StopAllNodes(ctx)
+	s.Require().NoError(err, "error stopping node(s)")
+
+	state, err := chain.ExportState(ctx, int64(haltHeight))
+	s.Require().NoError(err)
+
+	str := strings.SplitAfter(state, "server")
+	state = str[1]
```
colin-axner, line 125: Could we add some comments on what this is doing?

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	newGenesisJson := strings.ReplaceAll(state, fmt.Sprintf("\"initial_height\":%d", 0), fmt.Sprintf("\"initial_height\":%d", haltHeight+2))
```
colin-axner, line 127: Will the test fail if we start from the halt height?

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+type GenesisState map[string]json.RawMessage
+
+func TestGenesisTestSuite(t *testing.T) {
+	suite.Run(t, new(GenesisTestSuite))
+}
+
+type GenesisTestSuite struct {
+	testsuite.E2ETestSuite
+	cosmos.ChainNode
```
chatton, line 28: what's the reason for this embedding? This seems a bit odd to have a test case itself be composed of a chain node?

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,155 @@
...
+		s.Require().NoError(err)
+
+		expected := testvalues.StartingTokenAmount - 2*testvalues.IBCTransferAmount
+		s.Require().Equal(expected, actualBalance)
+	})
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 5, chainA, chainB), "failed to wait for blocks")
+}
+
+func (s *GenesisTestSuite) HaltChainAndExportGenesis(ctx context.Context, chain *cosmos.CosmosChain, relayer ibc.Relayer, haltHeight int64) {
```
chatton, line 111: I think it would make sense for this function to accept the relative path the the app.toml file which will be copied into the chain nodes. WDYT?

e2e/tests/upgrades/genesis_test.go
colin-axner, line 127:
```suggestion
        // perform a non-zero height upgrade which avoids requiring the counterparty light clients to be upgraded as well
	newGenesisJson := strings.ReplaceAll(state, fmt.Sprintf("\"initial_height\":%d", 0), fmt.Sprintf("\"initial_height\":%d", haltHeight+2))
```

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,155 @@
...
+	state = str[1]
+
+	newGenesisJson := strings.ReplaceAll(state, fmt.Sprintf("\"initial_height\":%d", 0), fmt.Sprintf("\"initial_height\":%d", haltHeight+2))
+
+	for _, node := range chain.Nodes() {
+		err := node.OverwriteGenesisFile(ctx, []byte(newGenesisJson))
+		s.Require().NoError(err)
+	}
+	for _, node := range chain.Nodes() {
+		err := node.CopyFile(ctx, "app.toml", "config/app.toml")
```
chatton, line 134: there is no function that we can use which would act on something like `ConfigFileOverrides` for this is there? Copying the file over is not idea but I think it might be the best approach we have for now!

e2e/tests/upgrades/app.toml
colin-axner, line 1: how is this used? Is there a way we could use a generated version from the binary instead?

e2e/tests/upgrades/genesis_test.go
chatton, line 127: minor preference for using the backtick syntax

```golang
fmt.Sprintf(`"initial_height":%d`, 0), fmt.Sprintf(`"initial_height":%d`, haltHeight+2))
```

to avoid the quote escaping.

chatton (approved):
Thank you for your work on this @ThanhNhann it looks great! I just had a few small questions/suggestions but nothing major. Leaving my approval here, happy to get this merged 👍

colin-axner (commented):
Overall LGTM! Thank you @ThanhNhann! I left a few questions in my review

e2e/tests/upgrades/genesis_test.go
colin-axner: What about reading the existing `config/app.toml`, unmarshaling, setting the halt height to 0, marshal, and write the file. Main concern about referencing the `app.toml` here is having to maintain a config file across versions

colin-axner: I'm unsure if it is possible to use [this function](https://github.com/strangelove-ventures/interchaintest/blob/main/testutil/toml.go#L46) directly like interchain test does [here](https://github.com/strangelove-ventures/interchaintest/blob/main/chain/cosmos/chain_node.go#L296), but I'm okay adding our own version of that function

anhductn2001: Yess, The way I copy file here is the easiest way to do it but it may be affected by version upgrade.

anhductn2001: Great, I was also reading [this function](https://github.com/strangelove-ventures/interchaintest/blob/main/chain/cosmos/chain_node.go#L296) while looking for ideas but it seems to be quite complicated. I tried copying the file first to make sure everything works. Now I will try to add this function and use it.

anhductn2001: Should we create another PR for it or do it right away?

e2e/tests/upgrades/genesis_test.go
anhductn2001: Yes, I will change it

e2e/tests/upgrades/genesis_test.go
ThanhNhann: I think we should up in this pr with the function you mentioned above, look great @anhductn2001

e2e/tests/upgrades/genesis_test.go
anhductn2001: I think there is no need in case of file copy. Now i will try changing halt-height directly in node, things might change!

e2e/tests/upgrades/genesis_test.go
anhductn2001: Yep, it's superfluous, I'll delete it.

e2e/tests/upgrades/app.toml
anhductn2001: now it's been retrieved and fixed manually, i will try to fix it directly as discussed above

e2e/tests/upgrades/genesis_test.go
ThanhNhann: it's like when we export the genesis state from `state, err := chain.ExportState(ctx, int64(haltHeight))`, the variable `state` will include of some thing before the real state we need so we split it by 2 lines above
Screenshot 2023-08-29 at 10 35 33

e2e/tests/upgrades/genesis_test.go
ThanhNhann: Ah it's not, I forgot to update it after testing, thank @colin-axner

anhductn2001:
OK, I just add func to change halt-height directly in node. Let's approve CI and review again. TY!!

e2e/tests/upgrades/genesis_test.go
colin-axner: Looks like the export function is simply reading from [stdout](https://github.com/strangelove-ventures/interchaintest/blob/main/chain/cosmos/chain_node.go#L1072) which has log entries, I wonder if it should be using the `--output-document`

I opened https://github.com/strangelove-ventures/interchaintest/issues/721. Want to link that issue in the code comment for now? A bit fragile to leave this handling, but unsure of a way to make it more robust without an update in interchaintest cc @chatton

e2e/tests/upgrades/genesis_test.go
colin-axner:
```diff
diff --git a/e2e/tests/upgrades/genesis_test.go b/e2e/tests/upgrades/genesis_test.go
index 78c4ab0b7..e03a1a19d 100644
--- a/e2e/tests/upgrades/genesis_test.go
+++ b/e2e/tests/upgrades/genesis_test.go
@@ -121,7 +121,7 @@ func (s *GenesisTestSuite) HaltChainAndExportGenesis(ctx context.Context, chain
        str := strings.SplitAfter(state, "server")
        state = str[1]
 
-       appTomlOverrides := make(testsuite.Toml)
+       appTomlOverrides := make(test.Toml)
 
        appTomlOverrides["halt-height"] = 0
 
@@ -131,7 +131,7 @@ func (s *GenesisTestSuite) HaltChainAndExportGenesis(ctx context.Context, chain
        }
 
        for _, node := range chain.Nodes() {
-               err := testsuite.ModifyTomlConfigFile(
+               err := test.ModifyTomlConfigFile(
                        ctx,
                        zap.NewExample(),
                        node.DockerClient,
```

This passes for me locally

colin-axner: I decided to push the fix, figured we could always revert the commit

e2e/tests/upgrades/genesis_test.go
```diff
@@ -0,0 +1,163 @@
...
+	err := test.WaitForBlocks(timeoutCtx, int(haltHeight), chain)
+	s.Require().Error(err, "chain did not halt at halt height")
+
+	err = chain.StopAllNodes(ctx)
+	s.Require().NoError(err, "error stopping node(s)")
+
+	state, err := chain.ExportState(ctx, int64(haltHeight))
+	s.Require().NoError(err)
+
+	// I dont know why but the state be exported have 2 line notice about IAVL fast node, so we need delete it.
```
colin-axner, line 120:
```suggestion
	// state exports currently read from stdout which includes log entries, this discards the log entry by the server module
	// this may need to be updated if log entries are inserted in between exporting of state and the server module logging
	// see issue: https://github.com/strangelove-ventures/interchaintest/issues/721
```

e2e/tests/upgrades/genesis_test.go
anhductn2001: Thank you @colin-axner. LGTM! I thought we have to add this function ourselves, but no.

colin-axner (approved):
Thank you @ThanhNhann and @anhductn2001, code looks amazing! I appreciate y'all continuing with this work! :rocket:

PR #3287, opened 2023-03-13
ADR 8 Interface and Packet Data Implementations

AdityaSripal:
## Description

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/types/packet.go
+++ b/modules/apps/27-interchain-accounts/types/packet.go
@@ -7,6 +7,7 @@ import (
 	errorsmod "cosmossdk.io/errors"
 	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/ibc-go/v7/modules/core/exported"
 )
 
 // MaxMemoCharLength defines the maximum length for the InterchainAccountPacketData memo field
@@ -24,6 +25,8 @@ var (
 	// timeout.
 	DefaultRelativePacketTimeoutTimestamp = uint64((time.Duration(10) * time.Minute).Nanoseconds())
 )
+
+var _ exported.CallbackPacketDataI = (*InterchainAccountPacketData)(nil)
 
 // ValidateBasic performs basic validation of the interchain account packet data.
 // The memo may be empty.
--- a/modules/apps/27-interchain-accounts/types/packet.go
+++ b/modules/apps/27-interchain-accounts/types/packet.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"encoding/json"
 	"time"
 
 	errorsmod "cosmossdk.io/errors"
@@ -45,6 +46,57 @@ func (iapd InterchainAccountPacketData) ValidateBasic() error {
 // GetBytes returns the JSON marshalled interchain account packet data.
 func (iapd InterchainAccountPacketData) GetBytes() []byte {
 	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&iapd))
+}
+
+// ADR-8 middleware should callback on the sender address on the source chain
+// if the sender address is an IBC Actor (i.e. smart contract that accepts IBC callbacks)
+func (iapd InterchainAccountPacketData) GetSrcCallbackAddress() (addr string) {
+	if len(iapd.Memo) == 0 {
+		return
+	}
+
+	jsonObject := make(map[string]interface{})
+	// the jsonObject must be a valid JSON object
+	err := json.Unmarshal([]byte(iapd.Memo), &jsonObject)
+	if err != nil {
+		return
+	}
+
+	callbackData, ok := jsonObject["callbacks"].(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	callbackAddr := callbackData["src_callback_address"].(string)
+	return callbackAddr
+}
+
+// ADR-8 middleware should callback on the receiver address on the destination chain
+// if the receiver address is an IBC Actor (i.e. smart contract that accepts IBC callbacks)
+func (iapd InterchainAccountPacketData) GetDestCallbackAddress() (addr string) {
+	if len(iapd.Memo) == 0 {
+		return
+	}
+
+	jsonObject := make(map[string]interface{})
+	// the jsonObject must be a valid JSON object
+	err := json.Unmarshal([]byte(iapd.Memo), &jsonObject)
+	if err != nil {
+		return
+	}
+
+	callbackData, ok := jsonObject["callbacks"].(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	callbackAddr := callbackData["dest_callback_address"].(string)
+	return callbackAddr
+}
+
+// no-op on this method to use relayer passed in gas
+func (fptd InterchainAccountPacketData) UserDefinedGasLimit() uint64 {
+	return 0
 }
 
 // GetBytes returns the JSON marshalled interchain account CosmosTx.
--- a/modules/apps/transfer/types/packet.go
+++ b/modules/apps/transfer/types/packet.go
@@ -9,6 +9,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	ibcerrors "github.com/cosmos/ibc-go/v7/internal/errors"
+	"github.com/cosmos/ibc-go/v7/modules/core/exported"
 )
 
 var (
@@ -23,6 +24,8 @@ var (
 	// timeout.
 	DefaultRelativePacketTimeoutTimestamp = uint64((time.Duration(10) * time.Minute).Nanoseconds())
 )
+
+var _ exported.CallbackPacketDataI = (*FungibleTokenPacketData)(nil)
 
 // NewFungibleTokenPacketData contructs a new FungibleTokenPacketData instance
 func NewFungibleTokenPacketData(
--- a/modules/apps/transfer/types/packet.go
+++ b/modules/apps/transfer/types/packet.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"encoding/json"
 	"strings"
 	"time"
 
@@ -61,4 +62,59 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
 // GetBytes is a helper for serialising
 func (ftpd FungibleTokenPacketData) GetBytes() []byte {
 	return sdk.MustSortJSON(mustProtoMarshalJSON(&ftpd))
+}
+
+// ADR-8 middleware should callback on the sender address on the source chain
+// if the sender address is an IBC Actor (i.e. smart contract that accepts IBC callbacks)
+func (ftpd FungibleTokenPacketData) GetSrcCallbackAddress() (addr string) {
+	if len(ftpd.Memo) == 0 {
+		return
+	}
+
+	jsonObject := make(map[string]interface{})
+	// the jsonObject must be a valid JSON object
+	err := json.Unmarshal([]byte(ftpd.Memo), &jsonObject)
+	if err != nil {
+		return
+	}
+
+	callbackData, ok := jsonObject["callbacks"].(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	if callbackData["src_callback_address"] == ftpd.Sender {
+		return ftpd.Sender
+	}
+	return
+}
+
+// ADR-8 middleware should callback on the receiver address on the destination chain
+// if the receiver address is an IBC Actor (i.e. smart contract that accepts IBC callbacks)
+func (ftpd FungibleTokenPacketData) GetDestCallbackAddress() (addr string) {
+	if len(ftpd.Memo) == 0 {
+		return
+	}
+
+	jsonObject := make(map[string]interface{})
+	// the jsonObject must be a valid JSON object
+	err := json.Unmarshal([]byte(ftpd.Memo), &jsonObject)
+	if err != nil {
+		return
+	}
+
+	callbackData, ok := jsonObject["callbacks"].(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	if callbackData["dest_callback_address"] == ftpd.Receiver {
+		return ftpd.Receiver
+	}
+	return
+}
+
+// no-op on this method to use relayer passed in gas
+func (fptd FungibleTokenPacketData) UserDefinedGasLimit() uint64 {
+	return 0
 }
--- a/modules/core/exported/callbacks.go
+++ b/modules/core/exported/callbacks.go
@@ -1,0 +1,19 @@
+package exported
+
+type CallbackPacketData interface {
+	// may return the empty string
+	GetSrcCallbackAddress() string
+
+	// may return the empty string
+	GetDestCallbackAddress() string
+
+	// UserDefinedGasLimit allows the sender of the packet to define inside the packet data
+	// a gas limit for how much the ADR-8 callbacks can consume. If defined, this will be passed
+	// in as the gas limit so that the callback is guaranteed to complete within a specific limit.
+	// On recvPacket, a gas-overflow will just fail the transaction allowing it to timeout on the sender side.
+	// On ackPacket and timeoutPacket, a gas-overflow will reject state changes made during callback but still
+	// commit the transaction. This ensures the packet lifecycle can always complete.
+	// If the packet data returns 0, the remaining gas limit will be passed in (modulo any chain-defined limit)
+	// Otherwise, we will set the gas limit passed into the callback to the `min(ctx.GasLimit, UserDefinedGasLimit())`
+	UserDefinedGasLimit() uint64
+}
--- a/modules/core/exported/callbacks.go
+++ b/modules/core/exported/callbacks.go
@@ -1,6 +1,6 @@
 package exported
 
-type CallbackPacketData interface {
+type CallbackPacketDataI interface {
 	// may return the empty string
 	GetSrcCallbackAddress() string
 
--- a/modules/core/exported/codec.go
+++ b/modules/core/exported/codec.go
@@ -1,0 +1,12 @@
+package exported
+
+import (
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+)
+
+func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
+	registry.RegisterInterface(
+		"ibc.core.exported.v1.CallbackPacketDataI",
+		(*CallbackPacketDataI)(nil),
+	)
+}
--- a/modules/apps/27-interchain-accounts/types/codec.go
+++ b/modules/apps/27-interchain-accounts/types/codec.go
@@ -7,6 +7,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 	"github.com/cosmos/gogoproto/proto"
+	"github.com/cosmos/ibc-go/v7/modules/core/exported"
 )
 
 // ModuleCdc references the global interchain accounts module codec. Note, the codec
@@ -21,6 +22,11 @@ var ModuleCdc = codec.NewProtoCodec(codectypes.NewInterfaceRegistry())
 func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
 	registry.RegisterImplementations((*authtypes.AccountI)(nil), &InterchainAccount{})
 	registry.RegisterImplementations((*authtypes.GenesisAccount)(nil), &InterchainAccount{})
+
+	registry.RegisterImplementations(
+		(*exported.CallbackPacketDataI)(nil),
+		&InterchainAccountPacketData{},
+	)
 }
 
 // SerializeCosmosTx serializes a slice of sdk.Msg's using the CosmosTx type. The sdk.Msg's are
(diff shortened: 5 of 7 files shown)


modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
 	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&iapd))
 }
 
+/**
+
+ADR-8 CallbackPacketData implementation
```
AdityaSripal, line 56: Is this the right place for this documentation? Should I instead have it in a README in ICS27 directory?

modules/apps/transfer/types/packet.go
```diff
@@ -62,3 +66,95 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
 func (ftpd FungibleTokenPacketData) GetBytes() []byte {
 	return sdk.MustSortJSON(mustProtoMarshalJSON(&ftpd))
 }
+
+/**
+
+ADR-8 CallbackPacketData implementation
```
AdityaSripal, line 72: ditto

modules/apps/transfer/types/packet.go
```diff
@@ -62,3 +66,95 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
...
+However, we may remove this restriction at a later date if it proves useful.
+
+**/
+
+// ADR-8 middleware should callback on the sender address on the source chain
+// if the sender address is an IBC Actor (i.e. smart contract that accepts IBC callbacks)
+// The desired callback address must be confirmed in the memo under the "callbacks" key.
+// This ensures that the callback is explicitly desired by the user and not just called
+// automatically.
+func (ftpd FungibleTokenPacketData) GetSrcCallbackAddress() (addr string) {
```
AdityaSripal, line 106: Do we need to reiterate in the memo here?

I think it makes sense on dest chain so we don't have automatic triggering of undefined smart contract logic. Plenty of bugs in Ethereum because of this behaviour.

But why wouldn't we want the automatic callback on ack and timeout if the sender itself is a smart contract. I think it makes sense to remove this logic and just return the sender directly here.

The middleware is responsible for determining if the address is a smart contract before executing

modules/apps/27-interchain-accounts/types/packet.go
charleenfei: it might make sense to have it in the ICS27 directory _as well_, but I think it would be nice to have here as well so there is an example of how to make use of this ADR-8

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
...
+		return
+	}
+
+	callbackAddr := callbackData["src_callback_address"].(string)
+	return callbackAddr
+}
+
+// ADR-8 middleware should callback on the receiver address on the destination chain
+// if the receiver address is an IBC Actor (i.e. smart contract that accepts IBC callbacks)
+func (iapd InterchainAccountPacketData) GetDestCallbackAddress() (addr string) {
```
AdityaSripal, line 105: I have the feeling this should just be disabled honestly. If you want custom logic to run on DEST, why not have it be a message encoded in the ICA tx itself?

I think for ICA, only sender callbacks make sense and for that I don't think we need a double confirmation in the memo

nicolaslara: I think it makes sense to disable this. I'm can't remember if ICA allows for multiple messages in one ICA tx but even if it doesn't, a callback would just extend this to two messages instead of multiple messages (which would be more useful)

modules/apps/transfer/types/packet.go
nicolaslara: I think it makes sense to have this documentation here as long as it's referenced from the ICS27 readme as two specific examples of how to implement the callbacks.

modules/apps/transfer/types/packet.go
```diff
@@ -62,3 +66,95 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
...
+		"dest_callback_address": {contractAddrOnDestChain},
+		"src_callback_msg": {jsonObjectForSrcChainCallback},
+		"dest_callback_msg": {jsonObjectForDestChainCallback},
+	}
+
+}
+```
+
+For transfer, we will enforce that the src_callback_address is the same as sender and dest_callback_address is the same as receiver.
+However, we may remove this restriction at a later date if it proves useful.
```
nicolaslara, line 97: Does it make sense to make that restriction optional? Maybe have an optional field `allow_non_receiving_callback` (or a better name) that defaults to false if non-existing?

modules/apps/transfer/types/packet.go
```diff
@@ -62,3 +66,95 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
...
+		return
+	}
+
+	callbackData, ok := jsonObject["callbacks"].(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	if callbackData["src_callback_address"] == ftpd.Sender {
+		return ftpd.Sender
```
nicolaslara, line 124: This is prevents us from specifying a callback as anything other than the sender. I think there's value in calling a different address (as long as it's explicitly specified by the sender). 

As an example use case, a FE may want to build packets that trigger a contract on Ack but without having to write/deploy a contract themselves and have all transactions be sent through that contract.

modules/apps/27-interchain-accounts/types/packet.go
AdityaSripal: It does allow multiple messages

modules/apps/transfer/types/packet.go
AdityaSripal: Yea, i just put this as an initial behaviour. Isn't this the behaviour of your hooks at the moment?

We could change it

nicolaslara: yeah, that's what we're currently doing (mostly to prevent user error). Just thought since callbacks were a generalization that they could allow for more flexibility here (see my other comment about making it optional based on a key in the memo)

modules/apps/27-interchain-accounts/types/packet.go
crodriguezvega: I believe we don't really make use of README files in our different modules, so maybe it's better to add this to our regular docs site?

crodriguezvega (commented):
Thank you @AdityaSripal!

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
...
+// ADR-8 middleware should callback on the sender address on the source chain
+// if the sender address is an IBC Actor (i.e. smart contract that accepts IBC callbacks)
+func (iapd InterchainAccountPacketData) GetSrcCallbackAddress() (addr string) {
+	if len(iapd.Memo) == 0 {
+		return
+	}
+
+	jsonObject := make(map[string]interface{})
+	// the jsonObject must be a valid JSON object
+	err := json.Unmarshal([]byte(iapd.Memo), &jsonObject)
```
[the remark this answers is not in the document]
damiannolan: I'd prefer not adding a dependency on a 3rd party module. Would there be any benefits? stdlib seems fine to me

modules/apps/transfer/types/packet.go
```diff
@@ -62,3 +66,95 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
...
+
+```json
+{
+	// ... other memo fields we don't care about
+	"callbacks": {
+		"src_callback_address": {contractAddrOnSrcChain},
+		"dest_callback_address": {contractAddrOnDestChain},
+		"src_callback_msg": {jsonObjectForSrcChainCallback},
+		"dest_callback_msg": {jsonObjectForDestChainCallback},
+	}
```
damiannolan, line 91: Is it intentional to omit methods for retrieving the callback msgs from the interface?

damiannolan: As I understand it these are the raw json msgs the contract would accept. Furthermore, if we add additional methods for retrieving them then we're parsing JSON twice to extract values. In this case I'd imagine having a single method `GetPacketCallbacks` which returned an object/strongly typed value rather than a string, because a middleware implementing `OnRecvPacket` for example, needs not only the contract address, but also the json msg to perform some action e.g. swap and stake.

modules/apps/27-interchain-accounts/types/packet.go
colin-axner: I think in-code and external facing documentation (information under `docs/`) are great. The in-code documentation should help explain the hard-coded values while the external facing documentation should explain how to interact with it

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
 	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&iapd))
 }
 
+/**
```
colin-axner, line 54:
```suggestion
/*
```

modules/core/exported/callbacks.go
```diff
@@ -0,0 +1,19 @@
+package exported
+
+type CallbackPacketDataI interface {
```
colin-axner, line 3:
```suggestion
type CallbackPacketData interface {
```
I think `ibcexported.CallbackPacketData` should make it apparent it is an interface?

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
...
+The Memo format is defined like so:
+
+```json
+{
+	// ... other memo fields we don't care about
+	"callbacks": {
+		"src_callback_address": {contractAddrOnSrcChain},
+		"dest_callback_address": {contractAddrOnDestChain},
+		"src_callback_msg": {jsonObjectForSrcChainCallback},
+		"dest_callback_msg": {jsonObjectForDestChainCallback},
```
colin-axner, line 72: Should we note this as additional information? This implementation doesn't actually require `src_callback_msg`/`dest_callback_msg`

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
...
+	}
+
+}
+```
+
+**/
```
colin-axner, line 78:
```suggestion
*/
```

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
...
+
+// ADR-8 middleware should callback on the sender address on the source chain
+// if the sender address is an IBC Actor (i.e. smart contract that accepts IBC callbacks)
```
colin-axner, line 81: godoc needs to start with the type name. 

Maybe we could have something like:
```go
// GetSrcCallbackAddress returns the source callback address provided in the packet data memo. 
// The memo is expected to specify the callback address in the following format:
// { "callbacks": { "src_callback_address": {contractAddrOnSrcChain}}
// 
// If no callback address is specified, an empty string is returned.

File: modules/apps/27-interchain-accounts/types/packet.go [hunk 14]
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
...
+func (iapd InterchainAccountPacketData) GetSrcCallbackAddress() (addr string) {
```
colin-axner, line 82: see comment below on named return value

modules/apps/27-interchain-accounts/types/packet.go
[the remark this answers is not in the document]
colin-axner: We could also add a helper function? I agree with @damiannolan on not relying on third party modules

modules/apps/transfer/types/packet.go
```diff
@@ -62,3 +66,95 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
 func (ftpd FungibleTokenPacketData) GetBytes() []byte {
 	return sdk.MustSortJSON(mustProtoMarshalJSON(&ftpd))
 }
+
+/**
+
+ADR-8 CallbackPacketData implementation
+
+FungibleTokenPacketData implements CallbackPacketDataI interface. This will allow middlewares targetting specific VMs
```
colin-axner, line 74:
```suggestion
FungibleTokenPacketData implements CallbackPacketDataI interface. This will allow middlewares targeting specific VMs
```

modules/apps/transfer/types/packet.go
colin-axner: I'm ok with an additional field, but it'd be nice to know this is a feature that would be requested fairly soon. Is there a use case we have in mind?

modules/apps/transfer/types/packet.go
colin-axner: I don't have a preference. Both seem like valid options to me. I guess I just wouldn't want to change the behaviour in the future (once contracts are written to expect certain behaviour, it shouldn't be adjusted)

modules/apps/transfer/types/packet.go
```diff
@@ -62,3 +66,95 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
...
+// The desired callback address must be confirmed in the memo under the "callbacks" key.
+// This ensures that the callback is explicitly desired by the user and not just called
+// automatically.
+func (ftpd FungibleTokenPacketData) GetSrcCallbackAddress() (addr string) {
+	if len(ftpd.Memo) == 0 {
+		return
+	}
+
+	jsonObject := make(map[string]interface{})
+	// the jsonObject must be a valid JSON object
```
colin-axner, line 112: Is this comment stating that the unmarshal should fail if the memo is not a valid jsonObject?

modules/apps/transfer/types/packet.go
colin-axner: > As an example use case, a FE may want to build packets that trigger a contract on Ack but without having to write/deploy a contract themselves and have all transactions be sent through that contract.

This sounds like a pretty good example. I guess the idea is being able to specify on the source side a contract call to continue execution on the packet flow that does not relate to the original sender (external user, or smart contract with a separate purpose)

modules/core/exported/codec.go
```diff
@@ -0,0 +1,12 @@
+package exported
+
+import (
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+)
+
+func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
```
colin-axner, line 7: godoc :)

colin-axner (commented):
More fabulous work from the protocol mage :mage_man: :magic_wand: :)

Conceptually looks great to me. I left many nits. 

The main component missing is tests. Each function added should have a unit test (helps us detect regressions in logic change, thinking of accidental changes in merge conflict fixing). In addition, it might be nice to add tests which mirror the expected ADR 8 flow, this might give implementers some concrete compiled code to look at as well. It can be very basic ( unmarshaling packet data into the interface and getting the src/dest addresses)

0xekez:
Hi folks, thanks for the work here. It would be great if we could use this work with DAO DAO. I do have some feedback on the callbacks here.

image

From my read of the docs, it seems like the callback is called only with the original packet sent. This is fine, though it would be great if data from the initiator of the transaction, and the receiver on the host chain could also add data there. For example, here is the callback form I came up with on my own ICA implementation:

https://git.sr.ht/~ekez/ica-secret/tree/main/item/packages/polytone/src/callback.rs#L10-28

I wonder if there is any flexibility from you folks about including more information in these callbacks?

modules/apps/27-interchain-accounts/types/packet.go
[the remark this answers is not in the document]
crodriguezvega: I am fine using standard library, but might be worth considering using a third party library if we see ourselves doing more and more JSON parsing on the memo field. I found JSON parsing pretty error prone and it's such a common activity that sounds logical to rely on a well tested third party library to take care of that.

0xekez:
> @AdityaSripal: The host chain can return data in the ack.

Thanks for the response and clarification. It seems like the module will [return the response information](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/keeper/relay.go#L127) from the message executed in the ACK success variant which seems reasonable. Thanks!

modules/apps/transfer/types/packet.go
```diff
@@ -62,3 +66,95 @@ func (ftpd FungibleTokenPacketData) ValidateBasic() error {
...
+	err := json.Unmarshal([]byte(ftpd.Memo), &jsonObject)
+	if err != nil {
+		return
+	}
+
+	callbackData, ok := jsonObject["callbacks"].(map[string]interface{})
```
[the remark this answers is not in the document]
colin-axner: > since non-strings would still fail as they can't be compared to ftpd.Receiver

Is it possible to have additional fields in the callbacks section that will not be strings? Maybe something like:
```json
{
    "callbacks": {
        "src_callback_address": "contractAddrOnSrcChain",
        "dest_callback_address": "contractAddrOnDestChain",
        "delegation_amount": 100, 
    }
}
```
I guess the callbacks section isn't intended to hold contract specific arguments? Do we have it specified somewhere what fields are okay for the callbacks section?

modules/core/exported/callbacks.go
crodriguezvega, line 3: There was a user on Discord asking about the flow of token transfer + ICA call (i.e. on acknowledgement of a token transfer, execute ICA call to send a transaction to the interchain account). It seems like the user was not using a smart contract, so I suggested to use the message router to send a `MsgSendTx` to the interchain account sub-controller. After discussing with @womensrights, for such flow we will probably need also an ADR-8 interface so that when processing the token transfer acknowledgement, the necessary information to send a TX to an interchain account is provided (e.g connection ID, owner address, maybe interchain account packet data?). If a new interface is required (maybe is good to have separate interfaces for different use cases), should we rename this interface to something like `ContractCallbackPacketDataI` to make it specific to the use case of smart contracts?

modules/apps/transfer/types/packet.go
colin-axner: opened https://github.com/cosmos/ibc-go/issues/3351 to track this discussion

modules/core/exported/callbacks.go
AdityaSripal: They should still be able to use the interfaces defined here. They would have to make their own custom middleware to do the go logic though.

damiannolan (approved):
ACK changes! Nice work

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,77 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
...
+	if err != nil {
+		return ""
+	}
+
+	callbackData, ok := jsonObject["callbacks"].(map[string]interface{})
+	if !ok {
+		return ""
+	}
+
+	callbackAddr, ok := callbackData["src_callback_address"].(string)
```
colin-axner, line 104: should we keep as `src_callback_address`? Or change to `source_callback_address`?

modules/core/exported/packet.go
```diff
@@ -0,0 +1,46 @@
...
+// Acknowledgement defines the interface used to return
+// acknowledgements in the OnRecvPacket callback.
+type Acknowledgement interface {
+	Success() bool
+	Acknowledgement() []byte
+}
+
+// CallbackPacketData defines the interface used by ADR 8 implementations
+// to obtain callback addresses associated with a specific packet data type.
+// This is an optional interface which indicates support for ADR 8 implementations.
```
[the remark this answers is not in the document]
colin-axner: added https://github.com/cosmos/ibc-go/pull/3287/commits/523f40b017821c13b750c507e7a361afe66328fb, I just used a link to main. I think that's fine since I don't expect our `docs/architecture` layout to change for a very long time

modules/apps/27-interchain-accounts/types/packet.go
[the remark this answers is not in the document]
chatton: I would also lean towards not pulling in a third party library for json processing.

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -47,6 +51,83 @@ func (iapd InterchainAccountPacketData) GetBytes() []byte {
...
+		return
+	}
+
+	callbackData, ok := jsonObject["callbacks"].(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	callbackAddr := callbackData["dest_callback_address"].(string)
+	return callbackAddr
```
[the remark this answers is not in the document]
chatton: +1

chatton (approved):
Nice job LGTM, thanks for all the work you've put into this 🙏

modules/core/exported/packet.go
```diff
@@ -0,0 +1,48 @@
...
+type CallbackPacketData interface {
+	// GetSourceCallbackAddress should return the callback address of a packet data on the source chain.
+	// This may or may not be the sender of the packet. If no source callback address exists for the packet,
+	// an empty string may be returned.
+	GetSourceCallbackAddress() string
+
+	// GetDestCallbackAddress should return the callback address of a packet data on the destination chain.
+	// This may or may not be the receiver of the packet. If no dest callback address exists for the packet,
+	// an empty string may be returned.
+	GetDestCallbackAddress() string
```
fedekunze, line 37: curious why these are `string` as opposed to addresses

PR #3309, opened 2023-03-17
e2e: support icad binary using new genesis commands

crodriguezvega:
## Description

Our simd still uses genesis commands that are at the top level, but icad has those commands now under the `genesis` command, so when I run the compatibility tests for I got [an error because the `add-genesis-account` command was not found](https://github.com/cosmos/interchain-accounts-demo/actions/runs/4445109899/jobs/7803951509). This is a bit of a wacky way of solving the immediate problem, but open for better alternatives.

I guess we should also change our commands in simd to move the genesis commands under `genesis`, to keep it consistent with SDK's simd, but this can be done in a separate PR.

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.


e2e/semverutil/semver.go
```diff
@@ -34,7 +34,7 @@ func (fr FeatureReleases) IsSupported(versionStr string) bool {
 		return true
 	}
 
-	if semverGTE(versionStr, fr.MajorVersion) {
+	if fr.MajorVersion != "" && semverGTE(versionStr, fr.MajorVersion) {
```
crodriguezvega, line 37: If the major version was empty, then it would return as supported.

chatton: does the check on line 33 above not already account for this?

e2e/testconfig/testconfig.go
```diff
@@ -214,25 +226,27 @@ func newDefaultSimappConfig(cc ChainConfig, name, chainID, denom string) ibc.Cha
...
 }
 
 // getGenesisModificationFunction returns a genesis modification function that handles the GenesisState type
 // correctly depending on if the govv1beta1 gov module is used or if govv1 is being used.
 func getGenesisModificationFunction(cc ChainConfig) func(ibc.ChainConfig, []byte) ([]byte, error) {
+	binary := cc.Binary
 	version := cc.Tag
 
-	if govGenesisFeatureReleases.IsSupported(version) {
+	if (binary == "defaultBinary" && govGenesisFeatureReleases.IsSupported(version)) || (binary == icadBinary && icadGovGenesisFeatureReleases.IsSupported(version)) {
```
chatton, line 249: This condition is getting quick complex. Can we maybe split this out into separate named booleans?

I believe you also meant to do `binary == defaultBinary` (not the string "defaultBinary")

It looks like the tests are not failing as we are using the wrong genesis modification function for simd.

.github/workflows/e2e-compatibility-workflow-call.yaml
```diff
@@ -40,6 +40,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: cosmos/ibc-go
+          ref: carlos/e2e-support-binary-new-genesis-commands
```
chatton, line 43: we shouldn't need to change this at all 🤔 

Just realising now, you are just temporarily using this to run icad tests right? You can do this in an easier way via

`Actions` > `Manual E2E (Icad)` > 
image

e2e/semverutil/semver.go
crodriguezvega: I think not... The problem I had was when `MajorVersion` is `""`, then `semverGTE(versionStr, fr.MajorVersion)` will pass and I think the right behaviour would be the opposite (i.e. if no major version is specified, then we should assume that the feature has not been introduced in a new major version).

.github/workflows/e2e-compatibility-workflow-call.yaml
crodriguezvega: > Just realising now, you are just temporarily using this to run icad tests right? 

Yes, that's right. And I needed this to be triggered from the interchain-accounts-demo repository.

e2e/testconfig/testconfig.go
```diff
@@ -370,25 +382,30 @@ func newDefaultSimappConfig(cc ChainConfig, name, chainID, denom string, cometCf
...
 
 // getGenesisModificationFunction returns a genesis modification function that handles the GenesisState type
 // correctly depending on if the govv1beta1 gov module is used or if govv1 is being used.
 func getGenesisModificationFunction(cc ChainConfig) func(ibc.ChainConfig, []byte) ([]byte, error) {
+	binary := cc.Binary
 	version := cc.Tag
 
-	if govGenesisFeatureReleases.IsSupported(version) {
+	doesSimdSupportGovv1Genesis := binary == defaultBinary && govGenesisFeatureReleases.IsSupported(version)
+	doesIcadSupportGovv1Genesis := icadGovGenesisFeatureReleases.IsSupported(version)
```
chatton, line 406: love it! Much more readable!

e2e/semverutil/semver.go
chatton: My assumption was that 

```go
if !semver.IsValid(versionStr) {
		return true
}
```

would be executed if `fr.MajorVersion` was an empty string, if this is not the case then I'm okay with this change 👍

chatton (approved):
LGTM! Thanks for updating the tests to work with icad 👍

e2e/semverutil/semver.go
colin-axner: Could we add a test case?

e2e/testconfig/testconfig.go
colin-axner, line 406: should this have `binary == icadBinary`? I wonder if there's a way we could programmatically tell if we should use gov v1 genesis?

e2e/testconfig/testconfig.go
```diff
@@ -370,25 +382,30 @@ func newDefaultSimappConfig(cc ChainConfig, name, chainID, denom string, cometCf
...
+		Bech32Prefix:           "cosmos",
+		CoinType:               fmt.Sprint(sdk.GetConfig().GetCoinType()),
+		Denom:                  denom,
+		GasPrices:              fmt.Sprintf("0.00%s", denom),
+		GasAdjustment:          1.3,
+		TrustingPeriod:         "508h",
+		NoHostMount:            false,
+		ModifyGenesis:          getGenesisModificationFunction(cc),
+		ConfigFileOverrides:    configFileOverrides,
+		UsingNewGenesisCommand: useNewGenesisCommand,
```
colin-axner, line 395: This is a field specified by interchain test? Seems a bit odd in my opinion since this seems to be very cosmos specific?

e2e/testconfig/testconfig.go
```diff
@@ -360,6 +370,8 @@ func newDefaultSimappConfig(cc ChainConfig, name, chainID, denom string, cometCf
 	tmTomlOverrides["log_level"] = cometCfg.LogLevel // change to debug in ~/.ibc-go-e2e-config.json to increase cometbft logging.
 	configFileOverrides["config/config.toml"] = tmTomlOverrides
 
+	useNewGenesisCommand := cc.Binary == icadBinary && icadNewGenesisCommandsFeatureReleases.IsSupported(cc.Tag)
```
colin-axner, line 373: It would be nice if it was possible to check if the command exists by using the specified binary. But it seems this is likely interchain test could have handled?

colin-axner (approved):
LGTM, thanks for fixing this! Hopefully we can figure out ways to better handle breaking changes to the cli interface across versions as I don't like referencing specific binary names in our test config setup, otherwise it seems contradictory to our support for any binary name/input

e2e/testconfig/testconfig.go
chatton: yes it is, my understanding of how the cobra fields are defined is that any chain could specify whatever args they want, which means the command interchain test ends up using might not actually be correct. It is however in this case. I think there is probably a more generic solution that could be used somewhere though.

e2e/testconfig/testconfig.go
chatton: I think this should be possible using `<bin> help --json` and parsing it as appropriate. I agree that this should be happening in interchaintest.

PR #3311, opened 2023-03-20
chore: fix linter warnings

crodriguezvega:
## Description

There were mainly 3 warning types:
- variables names `cap` redefined a built-in function.
- redundant `if ...; err != nil` check when it would be possible to just return the function called.
- Unused named parameters.

closes: #XXXX

N/A

---

write a little note why.


.golangci.yml
```diff
@@ -29,6 +29,9 @@ linters:
 
 issues:
   exclude-rules:
+    - text: "unused-parameter"
```
crodriguezvega, line 32: This linter rule complains when a named argument in a functions is not used

chatton (commented):
Thanks for sorting out these issues, maybe we can pin the linter to the current latest version in the [workflow](https://github.com/cosmos/ibc-go/blob/6be595a66e891ec10dbc7afcf36dfc509c685da9/.github/workflows/golangci.yml#L25) so we don't run into this issue unexpectedly again? WDYT?

modules/apps/27-interchain-accounts/controller/ibc_middleware_test.go
```diff
@@ -107,11 +107,7 @@ func SetupICAPath(path *ibctesting.Path, owner string) error {
 		return err
 	}
 
-	if err := path.EndpointB.ChanOpenConfirm(); err != nil {
-		return err
-	}
-
-	return nil
+	return path.EndpointB.ChanOpenConfirm()
```
damiannolan, line 110: so an update to the linter starting trigger all these?
sometimes I prefer an explicit function call, and there may be scenarios where we may want to log something. 

Happy to honour the linter here, just playing devils advocate. I think sometimes linters can be overly aggressive. We can always tweak the config later I guess.

damiannolan (approved):
Thanks @crodriguezvega ❤️

modules/apps/27-interchain-accounts/controller/ibc_middleware_test.go
DimitrisJim: It might be beneficial to add `//nolint` directives in that case with an additional comment for context or is that too noisy? I can definitely remember cases where a linter was being silly.

crodriguezvega: Yes, looks like golangci-lint was bumped to v1.52 and hell broke loose...

I had second thoughts when I was changing these as well, but thought the improvement was justified. I don't think the link would have complained if there was some extra logic inside the if statement.

crodriguezvega:
> maybe we can pin the linter to the current latest version in the [workflow](https://github.com/cosmos/ibc-go/blob/6be595a66e891ec10dbc7afcf36dfc509c685da9/.github/workflows/golangci.yml#L25) so we don't run into this issue unexpectedly again? WDYT?

I can do that, if you guys prefer that. To me fixing these warnings when they pop up is not a big deal, but if it's disturbing you the rest, I am fine fixing the version.

chatton:
> > maybe we can pin the linter to the current latest version in the [workflow](https://github.com/cosmos/ibc-go/blob/6be595a66e891ec10dbc7afcf36dfc509c685da9/.github/workflows/golangci.yml#L25) so we don't run into this issue unexpectedly again? WDYT?
> 
> I can do that, if you guys prefer that. To me fixing these warnings when they pop up is not a big deal, but if it's disturbing for the rest, I am fine fixing the version.

My main concern is that periodically linting that works locally will cause failures on CI, I. would prefer to intentionally upgrade the linter periodically so CI stays consistent until we change it ourselves.

crodriguezvega:
> My main concern is that periodically linting that works locally will cause failures on CI, I. would prefer to intentionally upgrade the linter periodically so CI stays consistent until we change it ourselves.

All right, [I fixed the version](https://github.com/cosmos/ibc-go/blob/carlos/fix-linter-warnings/.github/workflows/golangci.yml#L25).

chatton (approved):
Thanks for fixing these issues 🥇

modules/apps/27-interchain-accounts/controller/ibc_middleware_test.go
colin-axner: I agree with @damiannolan, I think it'd be nice to be able to turn off this specific linting check. I don't see a lot of value in enforcing it, but I do see value in not enforcing it

PR #3097, opened 2023-02-01
proposal: capability addition to ibc-go structure

tac0turtle:
## Description

This is a proposal on how to adopt capability. In the near future we will be able to remove the dependency of the sdk from this module making it standalone. The reason for the separate go.mod is that this module will most likely hit v1 and not iterate further with the plans of the sdk to introduce internal message routers. Tying it to an ibc-go release could be done as well, but didnt see the point since there wont be iteration on this work for a while. 

before going further wanted to double check if this is acceptable

closes: #3076

### Commit Message / Changelog Entry

```bash
feat: migrate capability module from cosmos-sdk to ibc-go
```

see the [guidelines](https://github.com/cosmos/ibc-go/blob/main/CONTRIBUTING.md#commit-messages) for commit messages. (view raw markdown for examples)

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#pr-targeting)).
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/10-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/master/CONTRIBUTING.md#testing).
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [commit message](https://github.com/cosmos/ibc-go/blob/main/CONTRIBUTING.md#commit-messages) to be used for the changelog entry in the PR description for review.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.
- [ ] Review `Codecov Report` in the comment section below once CI passes.

PR #3325, opened 2023-03-22
adr 8 callback packet data impl followups 

colin-axner:
## Description

applies review suggestions in #3287 

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.


modules/apps/27-interchain-accounts/types/codec_test.go
```diff
@@ -1,16 +1,41 @@
...
+
+	packetData := types.InterchainAccountPacketData{
+		Type: types.EXECUTE_TX,
+		Data: []byte("data"),
+		Memo: fmt.Sprintf("{\"callbacks\": {\"src_callback_address\": \"%s\"}}", sourceCallbackAddress),
+	}
+
+	types.RegisterInterfaces(types.ModuleCdc.InterfaceRegistry())
+	data, err := types.ModuleCdc.MarshalInterfaceJSON(&packetData)
+	suite.Require().NoError(err)
```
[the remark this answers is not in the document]
AdityaSripal: In this case, I think we should just remove all the codec related code here

colin-axner:
Tagging as r4r. The codec related code should still be removed. I will do that tomorrow, or it can be done in a followup or on the upstream pr if we want this merged upstream before then

damiannolan (approved):
Looks good to me! Nice work on the tests, super clean.

damiannolan:
Is there some linter fixes that are needed in this PR? 

Is the idea to target this feature branch in the middleware implementation to flesh things out before merging to main?

colin-axner:
> Is there some linter fixes that are needed in this PR?

No, the upstream pr isn't up to date with main, so I think bumping to the tip of main should hopefully fix the linter

> Is the idea to target this feature branch in the middleware implementation to flesh things out before merging to main?

Not the current intention. This pr is just to for ease of review

colin-axner:
Tried merging to main but it causes crazy diffs and a bad commit log. Reverted that commit, we can bump on the upstream branch which compares to main

modules/apps/27-interchain-accounts/types/packet_test.go
```diff
@@ -132,21 +136,71 @@ func (suite *TypesTestSuite) TestGetCallbackAddresses() {
...
+			if tc.expPass {
+				suite.Require().Equal(expSrcCbAddr, srcCbAddr)
+			} else {
+				suite.Require().Equal("", srcCbAddr)
+			}
+		})
+	}
+}
+
+func (suite *TypesTestSuite) TestGetDestCallbackAddress() {
```
AdityaSripal, line 166: Seems like an odd test to have for a no-op function

modules/core/exported/packet.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Acknowledgement defines the interface used to return
+// acknowledgements in the OnRecvPacket callback.
+type Acknowledgement interface {
+	Success() bool
+	Acknowledgement() []byte
+}
+
+// CallbackPacketData defines the interface used by ADR 8 implementations
```
AdityaSripal, line 23: Shouldn't we clarify that this is an optional interface?

AdityaSripal (approved):
Some minor nits

modules/apps/27-interchain-accounts/types/packet_test.go
colin-axner: If unresolved merge conflicts cause this to return a non empty string, I'd prefer the tests to catch the issue. It's quite easy for [merge conflicts to slip through](https://github.com/cosmos/ibc-go/pull/3097#discussion_r1146205788)

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -26,7 +26,7 @@ var (
 	DefaultRelativePacketTimeoutTimestamp = uint64((time.Duration(10) * time.Minute).Nanoseconds())
 )
 
-var _ exported.CallbackPacketDataI = (*InterchainAccountPacketData)(nil)
+var _ exported.CallbackPacketData = (*InterchainAccountPacketData)(nil)
```
chatton, line 29: I like this change 👍

modules/apps/27-interchain-accounts/types/codec.go
```diff
@@ -22,11 +21,6 @@ var ModuleCdc = codec.NewProtoCodec(codectypes.NewInterfaceRegistry())
 func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
 	registry.RegisterImplementations((*authtypes.AccountI)(nil), &InterchainAccount{})
 	registry.RegisterImplementations((*authtypes.GenesisAccount)(nil), &InterchainAccount{})
-
-	registry.RegisterImplementations(
-		(*exported.CallbackPacketDataI)(nil),
-		&InterchainAccountPacketData{},
-	)
```
chatton, line 29: why do we not need to register this type against `CallbackPacketData` anymore?

modules/apps/transfer/types/packet_test.go
```diff
@@ -43,3 +44,213 @@ func TestFungibleTokenPacketDataValidateBasic(t *testing.T) {
...
+			false,
+		},
+		{
+			"memo has callbacks in json struct but does not have string value for src_callback_address key",
+			types.FungibleTokenPacketData{
+				Denom:    denom,
+				Amount:   amount,
+				Sender:   addr1,
+				Receiver: addr2,
+				Memo:     "{\"callbacks\": {\"src_callback_address\": 10}}",
```
chatton, line 105: don't think this is a big enough deal to change it, but I find using backticks can be more readable than requiring to escape quotes 

e.g.

```go
Memo: `{"callbacks": {"src_callback_address": 10}}`,
```

chatton (commented):
Love the test coverage, just had a question about the interface registration, otherwise it all looks great nice job 👍

modules/apps/27-interchain-accounts/types/codec.go
colin-axner: So, in order to unmarshal into an interface we need to marshal into an interface. That is, the packet data needs to be marshaled into an any (including the @type field in the encoding) so that it can be unmarshaled into an any where the type registration lookup can occur. The issue is that this requires changing the encoding of the packet data which is a breaking change for over the wire encoding. We would need channel upgradability to do so

The latest discussion on this topic was to expect ADR 8 implementations in the short term to hard code the packet data types they use and reference the CallbackPacketData as a compliance check. That is, the interface won't be used for encoding/decoding (hence removing the codec references), and will only be used to check that the packet data types implement the associated interface

modules/apps/transfer/types/packet_test.go
colin-axner: Oh that's a lot cleaner. I will do this change when I get a chance (might not be until after this merge or monday)

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -66,27 +66,32 @@ The Memo format is defined like so:
 {
 	// ... other memo fields we don't care about
 	"callbacks": {
-		"src_callback_address": {contractAddrOnSrcChain},
+		"src_callback_address": {contractAddrOnSourceChain},
 		"dest_callback_address": {contractAddrOnDestChain},
```
crodriguezvega, line 70: This is nit, but if we spell out `Src` as `Source`, should we do the same with `Dst` and written `Destination` instead?

modules/apps/27-interchain-accounts/types/packet_test.go
```diff
@@ -83,11 +85,13 @@ func (suite *TypesTestSuite) TestValidateBasic() {
 	}
 }
 
-func (suite *TypesTestSuite) TestGetCallbackAddresses() {
+func (suite *TypesTestSuite) TestGetSourceCallbackAddress() {
+	expSrcCbAddr := "srcCbAddr"
```
crodriguezvega, line 89: I think this could be a constant.
```suggestion
	const expSrcCbAddr = "srcCbAddr"
```

modules/apps/27-interchain-accounts/types/packet_test.go
```diff
@@ -105,7 +109,7 @@ func (suite *TypesTestSuite) TestGetCallbackAddresses() {
 				Data: []byte("data"),
 				Memo: "memo",
 			},
-			"",
+			false,
 		},
 		{
 			"memo is does not have callbacks in json struct",
```
crodriguezvega, line 115:
```suggestion
			"memo does not have callbacks in json struct",
```

modules/apps/27-interchain-accounts/types/packet.go
```diff
@@ -66,27 +66,32 @@ The Memo format is defined like so:
...
-func (iapd InterchainAccountPacketData) GetSrcCallbackAddress() string {
+// GetSourceCallbackAddress returns the source callback address provided in the packet data memo.
+// If no callback address is specified, an empty string is returned.
+//
+// The memo is expected to specify the callback address in the following format:
+// { "callbacks": { "src_callback_address": {contractAddrOnSourceChain}}
+//
+// ADR-8 middleware should callback on the returned address if it is a PacketActor
+// (i.e. smart contract that accepts IBC callbacks).
+func (iapd InterchainAccountPacketData) GetSourceCallbackAddress() string {
```
crodriguezvega, line 89: What would be returned if the json struct has 2 `callbacks` entries or 2 `src_callback_address` fields inside the same `callbacks` object?

crodriguezvega (commented):
Nice!

modules/apps/transfer/types/packet_test.go
```diff
@@ -43,3 +44,213 @@ func TestFungibleTokenPacketDataValidateBasic(t *testing.T) {
...
+				Denom:    denom,
+				Amount:   amount,
+				Sender:   addr1,
+				Receiver: addr2,
+				Memo:     "memo",
+			},
+			false,
+		},
+		{
+			"memo is does not have callbacks in json struct",
```
crodriguezvega, line 77:
```suggestion
			"memo does not have callbacks in json struct",
```

modules/apps/transfer/types/packet_test.go
```diff
@@ -43,3 +44,213 @@ func TestFungibleTokenPacketDataValidateBasic(t *testing.T) {
...
```
crodriguezvega, line 176:
```suggestion
			"memo does not have callbacks in json struct",
```

modules/apps/27-interchain-accounts/types/codec.go
chatton: I see thanks for the detailed explanation! 👍

modules/apps/27-interchain-accounts/types/packet.go
colin-axner: I don't thinkso. We changed this to be consistent with existing [terminology](https://github.com/cosmos/ibc-go/blob/main/modules/core/exported/channel.go#L28). If we didn't have an existing precedence, it likely would have remained `src`

modules/apps/27-interchain-accounts/types/packet.go
damiannolan: The second entry will likely override the first if there is two identical keys in the stringified JSON object.

https://go.dev/play/p/SQQBSfr2gDq

Issue #1618, opened 2022-06-29
Implement `ChannelUpgradeInit` functionality for `04-channel`


damiannolan:
## Summary

Implement the `ChannelUpgradeInit` functionality as defined in [ICS004 - Upgrade Handshake](https://github.com/cosmos/ibc/blob/master/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#Upgrade-Handshake):

- Add `ChannelUpgradeInit` functionality to [IBC core `Msg` server implementation](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/msg_server.go)
- Add `ChanUpgradeInit` to `04-channel` keeper
- Add `WriteUpgradeInitChannel` to `04-channel` keeper

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3331, opened 2023-03-23
adding `ChanUpgradeInit` implementation

damiannolan:
## Description

closes: #1618 

**NA**

---

write a little note why.


modules/core/04-channel/types/channel.go
```diff
@@ -126,3 +126,23 @@ func (ic IdentifiedChannel) ValidateBasic() error {
 	channel := NewChannel(ic.State, ic.Ordering, ic.Counterparty, ic.ConnectionHops, ic.Version)
 	return channel.ValidateBasic()
 }
+
+// SubsetOf returns true if Order is a valid subset of the provided Order.
+func (o Order) SubsetOf(order Order) bool {
```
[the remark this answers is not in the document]
damiannolan: I'm onboard if there's consensus on this approach

modules/core/04-channel/keeper/events.go
```diff
@@ -271,3 +271,22 @@ func emitChannelClosedEvent(ctx sdk.Context, packet exported.PacketI, channel ty
 		),
 	})
 }
+
+// emitChannelUpgradeInitEvent emits a channel upgrade init event
+func emitChannelUpgradeInitEvent(ctx sdk.Context, portID string, channelID string, channel types.Channel) {
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
```
chatton, line 278: we should include the upgrade sequence too

modules/core/04-channel/keeper/events.go
```diff
@@ -271,3 +271,22 @@ func emitChannelClosedEvent(ctx sdk.Context, packet exported.PacketI, channel ty
...
+			types.EventTypeChannelUpgradeInit,
+			sdk.NewAttribute(types.AttributeKeyPortID, portID),
+			sdk.NewAttribute(types.AttributeKeyChannelID, channelID),
+			sdk.NewAttribute(types.AttributeCounterpartyPortID, channel.Counterparty.PortId),
+			sdk.NewAttribute(types.AttributeCounterpartyChannelID, channel.Counterparty.ChannelId),
+			sdk.NewAttribute(types.AttributeKeyConnectionID, channel.ConnectionHops[0]),
+			sdk.NewAttribute(types.AttributeVersion, channel.Version),
```
chatton, line 285: should this be `AttributeNewVersion`?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	if propsedUpgradeChannel.State != types.INITUPGRADE || propsedUpgradeChannel.Counterparty.PortId != channel.Counterparty.PortId ||
+		propsedUpgradeChannel.Counterparty.ChannelId != channel.Counterparty.ChannelId {
+		return 0, errorsmod.Wrap(types.ErrInvalidChannel, "proposed channel upgrade is invalid")
+	}
+
+	if strings.TrimSpace(propsedUpgradeChannel.Version) == "" {
+		return 0, errorsmod.Wrap(types.ErrInvalidChannelVersion, "channel version must be not be empty")
+	}
```
chatton, line 47: should this be caught in validate basic?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if !channel.Ordering.SubsetOf(propsedUpgradeChannel.Ordering) {
+		return 0, errorsmod.Wrap(types.ErrInvalidChannelOrdering, "channel ordering must be a subset of the new ordering")
+	}
+
+	upgradeSequence, found := k.GetUpgradeSequence(ctx, portID, channelID)
+	if !found {
```
chatton, line 54: technically this could all be replaced with

```go
         upgradeSequence++
         k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
```

but maybe the more explicit version is better

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,87 @@
...
+		k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
+	} else {
+		upgradeSequence++
+		k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
+	}
+
+	upgradeTimeout := types.UpgradeTimeout{
+		TimeoutHeight:    counterpartyTimeoutHeight,
+		TimeoutTimestamp: counterpartyTimeoutTimestamp,
+	}
```
chatton, line 65: it looks like we are missing validation for these

`abortTransactionUnless(counterpartyTimeoutHeight != 0 || counterpartyTimeoutTimestamp != 0)`

EDIT: this is happening in validate basic!

chatton (approved):
LGTM, the only thing I would like to know is if we should version validation into the ValidateBasic of MsgChanUpgradeInit

It might be nice to refactor the IsSubsetOf function too.

modules/core/keeper/msg_server.go
```diff
@@ -697,7 +697,45 @@ func (k Keeper) Acknowledgement(goCtx context.Context, msg *channeltypes.MsgAckn
...
+	msg.ProposedUpgradeChannel.Version = version
+
+	k.ChannelKeeper.WriteUpgradeInitChannel(ctx, msg.PortId, msg.ChannelId, msg.ProposedUpgradeChannel)
+
+	ctx.Logger().Info("channel upgrade init callback succeeded", "channel-id", msg.ChannelId, "version", version)
+
+	return &channeltypes.MsgChannelUpgradeInitResponse{
+		ChannelId: msg.ChannelId,
+		Version:   version,
+	}, nil
```
damiannolan, line 738: upgrade sequence may also be desirable in the response

modules/core/04-channel/types/channel.go
```diff
@@ -126,3 +126,23 @@ func (ic IdentifiedChannel) ValidateBasic() error {
...
+
+	switch o {
+	case ORDERED:
+		if order == UNORDERED {
+			return true
+		}
+	case UNORDERED:
+		if order == ORDERED {
+			return false
+		}
```
DimitrisJim, line 144: Another equivalent though likely less explicit:

```suggestion
	case ORDERED:
            order == UNORDERED
	case UNORDERED:
            return false
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// ChanUpgradeInit is called by a module to initiate a channel upgrade handshake with
+// a module on another chain.
+func (k Keeper) ChanUpgradeInit(
+	ctx sdk.Context,
+	portID string,
+	channelID string,
+	chanCap *capabilitytypes.Capability,
+	propsedUpgradeChannel types.Channel,
```
charleenfei, line 23: nit: `proposedUpgradeChannel`

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if channel.State != types.OPEN {
+		return 0, errorsmod.Wrapf(types.ErrInvalidChannelState, "expected %s, got %s", types.OPEN, channel.State)
+	}
+
+	if !k.scopedKeeper.AuthenticateCapability(ctx, chanCap, host.ChannelCapabilityPath(portID, channelID)) {
+		return 0, errorsmod.Wrapf(types.ErrChannelCapabilityNotFound, "caller does not own capability for channel, port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	if propsedUpgradeChannel.State != types.INITUPGRADE || propsedUpgradeChannel.Counterparty.PortId != channel.Counterparty.PortId ||
```
charleenfei, line 40: isn't this first check `if propsedUpgradeChannel.State != types.INITUPGRADE ||` already in `ValidateBasic`?

damiannolan: you are correct! :D

modules/core/04-channel/types/channel.go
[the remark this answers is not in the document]
damiannolan: I'm just gonna pull this in cause I like it! Thanks for the free code

charleenfei (approved):
two nits otherwise lgtm!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,87 @@
...
+func (k Keeper) WriteUpgradeInitChannel(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+	channelUpgrade types.Channel,
+) {
+	defer telemetry.IncrCounter(1, "ibc", "channel", "upgrade-init")
+
+	k.SetChannel(ctx, portID, channelID, channelUpgrade)
+	k.Logger(ctx).Info("channel state updated", "port-id", portID, "channel-id", channelID, "previous-state", "OPEN", "new-state", "INITUPGRADE")
```
crodriguezvega, line 84:
```suggestion
	k.Logger(ctx).Info("channel state updated", "port-id", portID, "channel-id", channelID, "previous-state", types.OPEN.String(), "new-state", types.INITUPGRADE.String())
```

modules/core/04-channel/types/channel.go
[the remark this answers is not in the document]
damiannolan: Added this plus the generic contains. Thanks @chatton

modules/core/04-channel/keeper/upgrade.go
damiannolan: I noticed this too but I left it as is for consistency 😛 

I've updated it as per your suggestion and opened https://github.com/cosmos/ibc-go/issues/3335

modules/core/04-channel/keeper/events.go
damiannolan: Aye me too!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	upgradeSequence, found := k.GetUpgradeSequence(ctx, portID, channelID)
+	if !found {
+		upgradeSequence = 1
+		k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
+	} else {
+		upgradeSequence = upgradeSequence + 1
+		k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
+	}
```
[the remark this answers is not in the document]
damiannolan: Maybe this is more idiomatic Go...


```go
upgradeSequence := uint64(1)
if seq, found := k.GetUpgradeSequence(ctx, portID, channelID); found {
	upgradeSequence = seq + 1
}
```

modules/core/04-channel/keeper/upgrade.go
damiannolan: Moved this to `ValidateBasic` 👍

modules/core/04-channel/keeper/upgrade.go
damiannolan: https://github.com/cosmos/ibc-go/pull/3331#discussion_r1147334242

damiannolan:
Since we can upgrade the connection, we should probably verify that the new connection exists:

```
if !k.HasConnection(ctx, proposedUpgradeChannel.ConnectionHops[0]) {
    return err
}
```

I will push an update for this. Perhaps this also needs to be added to the spec

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+func (k Keeper) ChanUpgradeInit(
+	ctx sdk.Context,
+	portID string,
+	channelID string,
+	chanCap *capabilitytypes.Capability,
+	proposedUpgradeChannel types.Channel,
+	counterpartyTimeoutHeight clienttypes.Height,
+	counterpartyTimeoutTimestamp uint64,
+) (uint64, error) {
+	channel, found := k.GetChannel(ctx, portID, channelID)
```
crodriguezvega, line 25: Would it make sense to fail fast if the parameters of the current channel (ordering, version, connection hops) and the proposed new channel are exactly the same? Is it worth initiating the handshake if they are exactly the same? @AdityaSripal

damiannolan: Good point! There may be a use case for updating one channel end's connection maybe? In that scenario we may want to allow one side to proceed even if they are identical.

crodriguezvega: Yeah, I thought about that as well, but then would it still make sense that on init side the channel parameters are the same but on the try side they are different? Like would it work if the try side decides to change ordering unilaterally, for example? And for the version, for example try side upgrading to use fee middleware?

I see in the spec that on ack the ordering should be the same for both sides:

```
// both channel ends must be mutually compatible.
// this means that the ordering must be the same and 
// any future introduced fields that must be compatible
// should also be checked
if counterpartyChannel.ordering != proposedUpgradeChannel.ordering {
  restoreChannel(portIdentifier, channelIdentifier)
  return
}
```

damiannolan: > Yeah, I thought about that as well, but then would it still make sense that on init side the channel parameters are the same as the current channel but on the try side they are different?

I agree with you that if only one side is changing (connection only I guess) then they are/should be the initiating party. Happy to add a `reflect.DeepEquals()` check here if there's consensus on that! 

> In the case of changing ordering, if the try side decides to change ordering unilaterally, I think the handshake will already fail on try.

Yep, you cannot change order unilaterally. Both channel ends must use the same ordering. 

> In the case of changing versions, if init side keeps the current channel version (let's say ics20-1 if we talk about transfer), but try side decides to use fee middleware

Yeah if you want to use ics29, both channel ends need to use it.

modules/core/keeper/msg_server.go
```diff
@@ -697,7 +697,46 @@ func (k Keeper) Acknowledgement(goCtx context.Context, msg *channeltypes.MsgAckn
...
+
+	upgradeSequence, err := k.ChannelKeeper.ChanUpgradeInit(ctx, msg.PortId, msg.ChannelId, chanCap, msg.ProposedUpgradeChannel, msg.TimeoutHeight, msg.TimeoutTimestamp)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade init callback failed", "error", errorsmod.Wrap(err, "channel handshake upgrade init failed"))
+		return nil, errorsmod.Wrap(err, "channel handshake upgrade init failed")
+	}
+
+	version, err := cbs.OnChanUpgradeInit(ctx, msg.ProposedUpgradeChannel.Ordering, msg.ProposedUpgradeChannel.ConnectionHops,
+		msg.PortId, msg.ChannelId, upgradeSequence, msg.ProposedUpgradeChannel.Counterparty, msg.ProposedUpgradeChannel.Version,
+		"previous-version-can-be-looked-up-from-within-cb??",
```
AdityaSripal, line 722: The full channel version should be passed in

modules/core/keeper/msg_server.go
```diff
@@ -697,7 +697,45 @@ func (k Keeper) Acknowledgement(goCtx context.Context, msg *channeltypes.MsgAckn
...
+	sequence, err := k.ChannelKeeper.ChanUpgradeInit(ctx, msg.PortId, msg.ChannelId, cap, msg.ProposedUpgradeChannel, msg.TimeoutHeight, msg.TimeoutTimestamp)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade init callback failed", "error", errorsmod.Wrap(err, "channel handshake upgrade init failed"))
+		return nil, errorsmod.Wrap(err, "channel handshake upgrade init failed")
+	}
+
+	version, err := cbs.OnChanUpgradeInit(ctx, msg.ProposedUpgradeChannel.Ordering, msg.ProposedUpgradeChannel.ConnectionHops,
+		msg.PortId, msg.ChannelId, sequence, msg.ProposedUpgradeChannel.Counterparty, msg.ProposedUpgradeChannel.Version,
+		"previous-version-can-be-looked-up-from-within-cb??",
+	)
```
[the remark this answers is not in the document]
AdityaSripal: We don't want the baseApp Version, but the entire channel version to be passed into our callback.

I don't mind the extra channel lookup over changing the return args of the previous call

modules/core/04-channel/keeper/upgrade.go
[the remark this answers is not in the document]
AdityaSripal: > This is as per spec. As an alternative we could run a migration handler to set the initial upgrade sequence for all OPEN channels to 1... but that might be overkill...

Yes this feels like overkill imv

modules/core/04-channel/keeper/upgrade.go
chatton: ack'ing this comment thread, the additional check on connection hops looks good to me. 👍

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,91 @@
...
+	if seq, found := k.GetUpgradeSequence(ctx, portID, channelID); found {
+		upgradeSequence = seq + 1
+	}
+
+	upgradeTimeout := types.UpgradeTimeout{
+		TimeoutHeight:    counterpartyTimeoutHeight,
+		TimeoutTimestamp: counterpartyTimeoutTimestamp,
+	}
+
+	k.SetUpgradeRestoreChannel(ctx, portID, channelID, channel)
```
chatton, line 69: channel will have a state of `types.INITUPGRADE` (above on line 41), the channel stored here should be the channel before any modifications were made right?

damiannolan: I've changed it to the following: 

```go
restoreChannel := channel
channel.State = types.INITUPGRADE
if reflect.DeepEqual(channel, proposedUpgradeChannel) {
	return 0, "", errorsmod.Wrap(types.ErrChannelExists, "existing channel end is identical to proposed upgrade channel end")
}
```

I don't really like changing the state of the channel just to do a deep equals comparison. Easy to lead to bugs like what you just caught. But we can worry about code vanity later on this one I guess.

chatton: LGTM thanks for the quick fix

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,92 @@
...
+// a module on another chain.
+func (k Keeper) ChanUpgradeInit(
+	ctx sdk.Context,
+	portID string,
+	channelID string,
+	chanCap *capabilitytypes.Capability,
+	proposedUpgradeChannel types.Channel,
+	counterpartyTimeoutHeight clienttypes.Height,
+	counterpartyTimeoutTimestamp uint64,
+) (uint64, string, error) {
```
AdityaSripal, line 27: nit: can you name the return args here so i know the string here is supposed to be the previous version?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if !k.scopedKeeper.AuthenticateCapability(ctx, chanCap, host.ChannelCapabilityPath(portID, channelID)) {
+		return 0, "", errorsmod.Wrapf(types.ErrChannelCapabilityNotFound, "caller does not own capability for channel, port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	restoreChannel := channel
+	channel.State = types.INITUPGRADE
+	if reflect.DeepEqual(channel, proposedUpgradeChannel) {
+		return 0, "", errorsmod.Wrap(types.ErrChannelExists, "existing channel end is identical to proposed upgrade channel end")
+	}
```
AdityaSripal, line 45: nit: godoc here to explain what this is doing

modules/core/keeper/msg_server.go
AdityaSripal, line 1: Tests of this file in a future PR?

AdityaSripal (approved):
Nice work! Some minor nits and a more important point about the test assertions.

Approving since the logic is right and don't want to block on addressing my comments

modules/core/keeper/msg_server.go
damiannolan: Yeah, so this file is gaining coverage by using the test helpers added to the `Endpoint` testing struct. That is how its currently being covered for the opening handshake handlers also. So once they are put to use it will be covered. 

At the moment the tests in msg_server_test.go are testing packet lifecycle while using for example, `endpoint.ChanOpenInit(...)` etc

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -0,0 +1,123 @@
...
+			)
+
+			if tc.expPass {
+				suite.Require().NoError(err)
+
+				expSequence, found := suite.chainA.GetSimApp().GetIBCKeeper().ChannelKeeper.GetUpgradeSequence(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().True(found)
+				suite.Require().Equal(expSequence, sequence)
+
+				expVersion, found := suite.chainA.GetSimApp().GetIBCKeeper().ChannelKeeper.GetAppVersion(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
```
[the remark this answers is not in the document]
damiannolan: Good call!

damiannolan:
Thanks @AdityaSripal! Will address nits in a follow up, going to get this merged now to unblock some other PRs.

Issue #1619, opened 2022-06-29
Implement `ChannelUpgradeTry` functionality for `04-channel`


damiannolan:
## Summary

Implement the `ChannelUpgradeTry` functionality as defined in [ICS004 - Upgrade Handshake](https://github.com/cosmos/ibc/blob/master/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#Upgrade-Handshake):

- [x] Add `ChannelUpgradeTry` functionality to [IBC core `Msg` server implementation](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/msg_server.go) (#3740)
- [x] Add `ChanUpgradeTry` to `04-channel` keeper (#3739)
- [x] Add `WriteUpgradeTryChannel` to `04-channel` keeper (#3703)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3333, opened 2023-03-23
Implement Channel Upgrade Try functionality for 04 channel

chatton:
## Description

closes: #1619

This PR adds the ChanUpgradeTry handshake. 

A few smaller items will be addressed in follow up issues.

- [Event Emmision](https://github.com/cosmos/ibc-go/issues/3398)
- [Clean up upgrade channel inputs](https://github.com/cosmos/ibc-go/issues/3412)
- adding a `validateProposedUpgradeChannelFields` function
- create a generic `k.getConnectionForVerification` function
- generic `if !connectionEnd.IsOpen()`
- Add helper function for configuring upgrade channel see [this comment](https://github.com/cosmos/ibc-go/pull/3333#discussion_r1159391597)

N/A  - part of feature branch

---

write a little note why.


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -77,3 +81,112 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+	}
+
+	if counterPartyChannel.Ordering != proposedUpgradeChannel.Ordering {
+		return errorsmod.Wrapf(types.ErrInvalidChannelOrdering, "channel ordering of counterparty channel and proposed channel must be equal")
+	}
+
+	connection, err := k.GetConnection(ctx, channel.ConnectionHops[0])
+	if err != nil {
+		return err
+	}
```
damiannolan, line 134: Connection hops is an upgradable field, I think we should be operating on the `proposedUpgradeChannel` connection hops instead. If we are upgrading/altering the connection of the underlying channel end, then we should be proving counterparty state against the new connection. cc. @AdityaSripal 

```suggestion
	connection, err := k.GetConnection(ctx, proposedUpgradeChannel.ConnectionHops[0])
	if err != nil {
		return err
	}
```

damiannolan: Addressed in https://github.com/cosmos/ibc-go/pull/3333/commits/165e23d0f275be6eafe3c416e5b659b7b4f05543

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -77,3 +80,107 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+	if counterpartyChannel.Ordering != proposedUpgradeChannel.Ordering {
+		return errorsmod.Wrapf(types.ErrInvalidChannelOrdering, "channel ordering of counterparty channel and proposed channel must be equal")
+	}
+
+	connection, err := k.GetConnection(ctx, proposedUpgradeChannel.ConnectionHops[0])
+	if err != nil {
+		return err
+	}
+
+	if err := k.connectionKeeper.VerifyChannelState(ctx, connection, proofHeight, channelProof, portID, channelID, counterpartyChannel); err != nil {
```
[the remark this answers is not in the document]
damiannolan: Addressed in [9ecf67b](https://github.com/cosmos/ibc-go/pull/3333/commits/9ecf67b77d4ca354c8d14245b7fc751ea34afc5c)

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -95,6 +96,188 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+	upgradeTimeout := types.UpgradeTimeout{TimeoutHeight: timeoutHeight, TimeoutTimestamp: timeoutTimestamp}
+	if err := k.connectionKeeper.VerifyChannelUpgradeTimeout(ctx, connectionEnd, proofHeight, proofUpgradeTimeout, proposedUpgradeChannel.Counterparty.PortId,
+		proposedUpgradeChannel.Counterparty.ChannelId, upgradeTimeout); err != nil {
+		return 0, "", err
+	}
+
+	if err := k.connectionKeeper.VerifyChannelUpgradeSequence(ctx, connectionEnd, proofHeight, proofUpgradeSequence, proposedUpgradeChannel.Counterparty.PortId,
+		proposedUpgradeChannel.Counterparty.ChannelId, counterpartyUpgradeSequence); err != nil {
+		return 0, "", err
+	}
```
damiannolan, line 172: We should be doing state verification of counterparty using the original channel connection. This is currently using the connection looked up from the proposed upgrade channel connection hops.

This might look a little ugly as depending on whether or not we are in a crossing hello then we'll need to retrieve the connection from the either current channel (state: OPEN) or the restore channel (crossing hello, state: INITUPGRADE).

ref: https://github.com/cosmos/ibc/pull/950#discussion_r1153264262 

cc. @AdityaSripal

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -95,6 +96,188 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+			k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
+		} else {
+			errorReceipt := types.NewErrorReceipt(upgradeSequence, errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "upgrade sequence %d was not smaller than the counter party chain upgrade sequence %d", upgradeSequence, counterpartyUpgradeSequence))
+			// the upgrade sequence is incremented so both sides start the next upgrade with a fresh sequence.
+			upgradeSequence++
+
+			k.SetUpgradeErrorReceipt(ctx, portID, channelID, errorReceipt)
+			k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
+
+			// TODO: emit error receipt events
```
damiannolan, line 204: I don't mind doing error receipt events in a follow up PR, considering this is already quite large

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -95,6 +96,188 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+			}
+			return 0, "", errorsmod.Wrap(types.ErrUpgradeAborted, "proposed upgrade channel did not equal expected channel")
+		}
+
+		// retrieve restore channel to return previous version
+		restoreChannel, found := k.GetUpgradeRestoreChannel(ctx, portID, channelID)
+		if !found {
+			// TODO(this case should be unreachable): write error receipt for upgrade sequence and abort / cancel upgrade
+			return 0, "", errorsmod.Wrapf(types.ErrChannelNotFound, "upgrade aborted, error receipt written for upgrade sequence: %d", upgradeSequence)
+		}
```
damiannolan, line 254: since this is technically unreachable should we panic? if it is reached then something is horribly wrong and some manual intervention or migration code is probably required.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -95,6 +96,188 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+	ctx sdk.Context,
+	portID,
+	channelID string,
+	upgradeSequence uint64,
+	channelUpgrade types.Channel,
+) {
+	defer telemetry.IncrCounter(1, "ibc", "channel", "upgrade-try")
+
+	k.SetChannel(ctx, portID, channelID, channelUpgrade)
+	k.Logger(ctx).Info("channel state updated", "port-id", portID, "channel-id", channelID, "previous-state", types.OPEN.String(), "new-state", types.TRYUPGRADE.String())
```
damiannolan, line 276: minor nit: in crossing hellos previous state would be `INITUPGRADE`

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -296,7 +296,11 @@ message MsgChannelUpgradeTry {
 }
 
 // MsgChannelUpgradeTryResponse defines the MsgChannelUpgradeTry response type
-message MsgChannelUpgradeTryResponse {}
+message MsgChannelUpgradeTryResponse {
+  string channel_id       = 1;
+  string version          = 2;
+  uint64 upgrade_sequence = 3;
+}
```
damiannolan, line 303: should we include some kind of `success` flag to indicate whether or not the TRY handler succeeded successfully or whether an error receipt was written for sequence `n`.

modules/core/keeper/msg_server.go
```diff
@@ -741,7 +741,75 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		msg.ProposedUpgradeChannel,
+		msg.TimeoutHeight,
+		msg.TimeoutTimestamp,
+		msg.ProofChannel,
+		msg.ProofUpgradeTimeout,
+		msg.ProofUpgradeSequence,
+		msg.ProofHeight,
+	)
+
+	if errorsmod.IsOf(err, channeltypes.ErrUpgradeAborted, channeltypes.ErrInvalidUpgradeSequence, channeltypes.ErrUpgradeTimeout) {
```
damiannolan, line 774: Here we need to handle certain errors which signal that an error receipt has been written and must be committed. If we return an error to sdk baseapp then state changes are reverted

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -95,6 +96,188 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+	// check if upgrade timed out by comparing it with the latest height of the chain
+	selfHeight := clienttypes.GetSelfHeight(ctx)
+	if !timeoutHeight.IsZero() && selfHeight.GTE(timeoutHeight) {
+		return 0, "", k.RestoreChannel(ctx, portID, channelID, upgradeSequence, types.ErrUpgradeTimeout)
+	}
+
+	// check if upgrade timed out by comparing it with the latest timestamp of the chain
+	if timeoutTimestamp != 0 && uint64(ctx.BlockTime().UnixNano()) >= timeoutTimestamp {
+		return 0, "", k.RestoreChannel(ctx, portID, channelID, upgradeSequence, types.ErrUpgradeTimeout)
+	}
```
damiannolan, line 183: Should we just reject the transaction here instead of writing an error receipt? Opened an issue on the spec https://github.com/cosmos/ibc/issues/953

charleenfei: from @AdityaSripal's [reply](https://github.com/cosmos/ibc/issues/953#issuecomment-1494210735) `I write an error receipt here to support the cancellation case, but we could remove it and only rely on timeout.`

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -95,6 +96,188 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+}
+
+// WriteUpgradeTryChannel writes a channel which has successfully passed the UpgradeTry handshake step.
+// An event is emitted for the handshake step.
+func (k Keeper) WriteUpgradeTryChannel(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+	upgradeSequence uint64,
+	channelUpgrade types.Channel,
```
colin-axner, line 271: #3401 bug exists here as well. I know it isn't clean, but I actually think the channel end should be retrieved from state instead of writing the proposed upgrade channel. This comes from a general security principal that you should be recreating the data you want to write to disk rather than writing the user input (as in this case it is unsafe without an additional check). This is partially why we do this for [openAck](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/handshake.go#L275) and openConfirm. I'm open to differing opinions, but it would be nice to be defensive here, if another field is added to the channel end which cannot be modified in an upgrade, we would need to remember to add another defensive check

colin-axner: What if we add `proposedChannelUpgradeArgs` or some similar type which contains all the fields that can be modified. This can be passed as an argument to `MsgChannelUpgradeInit`, `ChanUpgradeInit`, `OnChanUpgradeInit` and `WriteChannelUpgradeInit`. It seems redundant to check that the `proposedUpgradeChannel` hasn't modified certain fields, seems safer to simply exclude them from consideration by adding an extra type

chatton: nice catch, but do we need to create a separate type for this? We can fetch the existing channel and then modify it based on values present in `channelUpgrade`. A `proposedChannelUpgradeInfo` would surely contain the same fields. I do think it's a good idea to not set the channel directly though.

chatton: The benefit of a separate type I can think of is that not all fields are upgradable, and so some might end up being ignored. With a custom type we can provide only fields that are valid to upgrade.

modules/core/keeper/msg_server.go
chatton: One thought is to wrap any individual error that should be caught here in `ErrUpgradeAborted` and do a single check.

modules/core/04-channel/keeper/upgrade.go
chatton: this is now being handled in the above `getUpgradeTryConnectionEnd` function which handles the crossing hellos case.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,17 +88,249 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+		upgradeSequence, found = k.GetUpgradeSequence(ctx, portID, channelID)
+		if !found {
+			errorReceipt := types.NewErrorReceipt(upgradeSequence, errorsmod.Wrapf(types.ErrUpgradeAborted, "upgrade sequence %d was not smaller than the counter party chain upgrade sequence %d", upgradeSequence, counterpartyUpgradeSequence))
+			k.SetUpgradeErrorReceipt(ctx, portID, channelID, errorReceipt)
+
+			return 0, "", errorsmod.Wrapf(types.ErrUpgradeAborted, "upgrade aborted, error receipt written for upgrade sequence: %d", upgradeSequence)
+		}
+
+		if upgradeSequence != counterpartyUpgradeSequence {
+			errorReceipt := types.NewErrorReceipt(upgradeSequence, errorsmod.Wrapf(types.ErrUpgradeAborted, "upgrade sequence %d was not equal to the counter party chain upgrade sequence %d", upgradeSequence, counterpartyUpgradeSequence))
```
crodriguezvega, line 248:
```suggestion
			errorReceipt := types.NewErrorReceipt(upgradeSequence, errorsmod.Wrapf(types.ErrUpgradeAborted, "upgrade sequence ≠ counterparty chain upgrade sequence (%d ≠ %d)", upgradeSequence, counterpartyUpgradeSequence))
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,17 +88,249 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+
+			// set to the max of the two
+			if counterpartyUpgradeSequence > upgradeSequence {
+				upgradeSequence = counterpartyUpgradeSequence
+				k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
+			}
+
+			k.SetUpgradeErrorReceipt(ctx, portID, channelID, errorReceipt)
+			return 0, "", errorsmod.Wrapf(types.ErrUpgradeAborted, "upgrade aborted, error receipt written for upgrade sequence: %d", upgradeSequence)
```
crodriguezvega, line 257: The upgrade sequence mentioned in this error message could be different than the one in the error receipt if it gets updated in the if clause just above.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,17 +88,249 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+		}
+
+		// if there is a crossing hello, i.e an UpgradeInit has been called on both channelEnds,
+		// then we must ensure that the proposedUpgrade by the counterparty is the same as the currentChannel
+		// except for the channel state (upgrade channel will be in TRYUPGRADE and current channel will be in INITUPGRADE)
+		// if the proposed upgrades on either side are incompatible, then we will restore the channel and cancel the upgrade.
+		channel.State = types.TRYUPGRADE
+		k.SetChannel(ctx, portID, channelID, channel)
```
crodriguezvega, line 265: Should this `SetChannel` (maybe for clarity) be done at the end?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,17 +88,249 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+
 	k.SetChannel(ctx, portID, channelID, channelUpgrade)
 	k.Logger(ctx).Info("channel state updated", "port-id", portID, "channel-id", channelID, "previous-state", types.OPEN.String(), "new-state", types.INITUPGRADE.String())
 
 	emitChannelUpgradeInitEvent(ctx, portID, channelID, upgradeSequence, channelUpgrade)
+	return nil
+}
+
+// ChanUpgradeTry is called by a module to accept the first step of a channel upgrade
+// handshake initiated by a module on another chain.
```
crodriguezvega, line 119: Maybe it would be good to add something here in the good about the upgrade sequence that is returned, because if this function returns a not `nil` error, then `upgradeSequence` will be zero, but in fact the upgrade sequence stored could be different.

modules/core/04-channel/types/errors.go
```diff
@@ -37,10 +37,13 @@ var (
...
-	ErrUpgradeErrorNotFound    = errorsmod.Register(SubModuleName, 29, "upgrade error receipt not found")
+	ErrInvalidChannelVersion        = errorsmod.Register(SubModuleName, 24, "invalid channel version")
+	ErrPacketNotSent                = errorsmod.Register(SubModuleName, 25, "packet has not been sent")
+	ErrInvalidTimeout               = errorsmod.Register(SubModuleName, 26, "invalid packet timeout")
+	ErrInvalidUpgradeTimeout        = errorsmod.Register(SubModuleName, 27, "either timeout height or timeout timestamp must be non-zero")
+	ErrUpgradeSequenceNotFound      = errorsmod.Register(SubModuleName, 28, "upgrade sequence not found")
+	ErrChannelUpgradeRestoreFailure = errorsmod.Register(SubModuleName, 29, "failed to restore channel during upgrade")
+	ErrUpgradeAborted               = errorsmod.Register(SubModuleName, 30, "upgrade aborted")
+	ErrUpgradeTimeout               = errorsmod.Register(SubModuleName, 31, "upgrade timeout")
+	ErrUpgradeErrorNotFound         = errorsmod.Register(SubModuleName, 32, "upgrade error receipt not found")
```
crodriguezvega, line 48:
```suggestion
	ErrUpgradeErrorReceiptNotFound         = errorsmod.Register(SubModuleName, 32, "upgrade error receipt not found")
```

crodriguezvega (commented):
Great work so far, @charleenfei and @chatton. This handler is quite a beast! 🐻 I left some comments for now, will take an another look again in a couple of days. :)

modules/core/04-channel/keeper/upgrade.go
colin-axner: Yes exactly. The benefit of that is that there is less room for error if the submitter specifies the non-modifiable fields incorrectly. We could use a similar syntax to client state `ZeroCustomFields()` function which returns a client state with the custom fields zeroed out (except we would do the opposite). I'm happy to long as we aren't setting the `proposedUpgradeChannel` directly into state and are simply modifying the existing one

modules/core/04-channel/keeper/upgrade.go
chatton: The updated channel state is used just a few lines down and I think separating them out makes things a little harder to follow in this case

modules/core/04-channel/keeper/upgrade.go
chatton: I think a helper like that is the best option, this PR is getting a little unwieldy so I think I will create a follow up for this if you're okay with it!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,23 +89,252 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+			connectiontypes.ErrInvalidConnectionState,
+			"connection state is not OPEN (got %s)", connectiontypes.State(connectionEnd.GetState()).String(),
+		)
+	}
+
+	if connectionEnd.GetCounterparty().GetConnectionID() != counterpartyChannel.ConnectionHops[0] {
+		return 0, "", errorsmod.Wrapf(connectiontypes.ErrInvalidConnection, "unexpected counterparty channel connection hops, expected %s but got %s", connectionEnd.GetCounterparty().GetConnectionID(), counterpartyChannel.ConnectionHops[0])
+	}
+
+	if err := k.connectionKeeper.VerifyChannelState(ctx, connectionEnd, proofHeight, proofChannel, proposedUpgradeChannel.Counterparty.PortId,
```
colin-axner, line 176: So we are verifying 3 things:
- channel
- sequence 
- timeout

From my understanding, there is no particular reason that sequence and timeout are separated within the provable store (they could be under the same upgrade type). The channel is under its' own path for no strong reason, but one benefit is that you can be certain the counterparty has paused packet processing. There could be a construction where all 3 are under the same provable upgrade type:
```
interface Upgrade {
    proposedChannel Channel
    sequence uint64
    timeout Timeout
}
```

We could add a `k.connectionKeeper.VerifyUpgrade` which does all 3, but I don't have a strong preference, just mentioning as food for thought. The main downside of such a function is passing 3 proofs/values at once. The argument list would be very large:
```go
(ctx, connectionEnd, channel.Counterparty.PortId, channel.Counterparty.ChannelId, proofHeight, proofChannel, counterpartyChannel, proofUpgradeTimeout, upgradeTimeout, proofUpgradeSequence, upgradeSequence
```

After seeing that argument list, I think I actually prefer doing the split here

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,23 +89,252 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+	if timeoutTimestamp != 0 && uint64(ctx.BlockTime().UnixNano()) >= timeoutTimestamp {
+		if err := k.RestoreChannelAndWriteErrorReceipt(ctx, portID, channelID, upgradeSequence, types.ErrUpgradeTimeout); err != nil {
+			return 0, "", errorsmod.Wrap(types.ErrUpgradeAborted, err.Error())
+		}
+		return 0, "", errorsmod.Wrapf(types.ErrUpgradeAborted, "block timestamp >= upgrade timeout timestamp (%s >= %s)", ctx.BlockTime(), time.Unix(0, int64(timeoutTimestamp)))
+	}
+
+	switch channel.State {
+	case types.OPEN:
+		upgradeSequence = uint64(0)
```
colin-axner, line 211: This should be 1?

colin-axner (commented):
Ay, excellent work y'all!! Looking good :sunglasses: :sun_with_face: 

I left a lot of thoughts/suggestions for improvements. I've only gotten part way through the upgradeTry handler. Will continue my review later

modules/core/04-channel/keeper/upgrade.go
charleenfei: I tend to agree with @chatton that it is clearer to keep the logic together, rather than having the channel state being updated in one place and then actually set further down with the potential restore channel logic happening in between

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,23 +89,252 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+
+	if !channel.Ordering.SubsetOf(proposedUpgradeChannel.Ordering) {
+		return 0, "", errorsmod.Wrap(types.ErrInvalidChannelOrdering, "channel ordering must be a subset of the new ordering")
+	}
+
+	if counterpartyChannel.Ordering != proposedUpgradeChannel.Ordering {
+		return 0, "", errorsmod.Wrapf(types.ErrInvalidChannelOrdering, "channel ordering of counterparty channel and proposed channel must be equal")
+	}
+
+	connectionEnd, err := k.getUpgradeTryConnectionEnd(ctx, portID, channelID, channel)
```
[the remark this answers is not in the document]
chatton: yes already ran into this problem looking into upgrade ack! The only thing that changes is the condition for a crossing hello

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,23 +89,252 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+	}
+
+	if proposedUpgradeChannel.Counterparty.PortId != channel.Counterparty.PortId ||
+		proposedUpgradeChannel.Counterparty.ChannelId != channel.Counterparty.ChannelId {
+		return 0, "", errorsmod.Wrap(types.ErrInvalidChannel, "counterparty port ID and channel ID cannot be upgraded")
+	}
+
+	if !channel.Ordering.SubsetOf(proposedUpgradeChannel.Ordering) {
+		return 0, "", errorsmod.Wrap(types.ErrInvalidChannelOrdering, "channel ordering must be a subset of the new ordering")
+	}
```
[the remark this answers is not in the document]
charleenfei: this is a good idea, we will update in a followup pr along with other util methods/small updates (see PR description)

modules/core/04-channel/keeper/upgrade.go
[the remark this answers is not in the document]
charleenfei: we will make a generic function for this in a followup pr

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -86,23 +89,252 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+
+	if connectionEnd.GetState() != int32(connectiontypes.OPEN) {
+		return 0, "", errorsmod.Wrapf(
+			connectiontypes.ErrInvalidConnectionState,
+			"connection state is not OPEN (got %s)", connectiontypes.State(connectionEnd.GetState()).String(),
+		)
+	}
+
+	if connectionEnd.GetCounterparty().GetConnectionID() != counterpartyChannel.ConnectionHops[0] {
```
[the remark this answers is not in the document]
charleenfei: which spec changes are you referring to?

modules/core/04-channel/keeper/upgrade.go
charleenfei: shouldn't it actually still be 0 because we are in channel state `OPEN`? we update the sequence further down the code block

modules/core/keeper/msg_server.go
```diff
@@ -726,22 +726,92 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		return nil, errorsmod.Wrap(err, "could not retrieve module from port-id")
+	}
+
+	cbs, ok := k.Router.GetRoute(module)
+	if !ok {
+		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
+		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+	}
+
+	upgradeSequence, previousVersion, err := k.ChannelKeeper.ChanUpgradeTry(
```
AdityaSripal, line 755: I still prefer getting the previousVersion separately in a new store call.

I know its more gas, but just sat for a couple minutes wondering why this would return previous version.

AdityaSripal (commented):
Looking good. Have a couple comments for discussion

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -85,24 +82,245 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+	if !found {
+		panic(fmt.Sprintf("failed to retrieve channel %s on port %s", channelID, portID))
+	}
+
+	// assign directly the fields that are modifiable.
+	// counterparty fields may not be changed.
+	channel.State = types.TRYUPGRADE
+	channel.Version = proposedUpgradeVersion
+	channel.Ordering = channelUpgrade.Ordering
+	channel.ConnectionHops = channelUpgrade.ConnectionHops
```
crodriguezvega, line 302: Nit: this code updating the channel fields [is duplicated in `WriteUpgradeInitChannel`](https://github.com/cosmos/ibc-go/blob/cian/issue%231619-implement-channelupgradetry-functionality-for-04-channel/modules/core/04-channel/keeper/upgrade.go#L94-L99). Maybe we can have a method on the channel end struct to update all upgradable fields?

modules/core/keeper/msg_server.go
```diff
@@ -726,22 +726,92 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
 		return nil, errorsmod.Wrapf(err, "channel upgrade init callback failed for port ID: %s, channel ID: %s", msg.PortId, msg.ChannelId)
 	}
 
-	msg.ProposedUpgradeChannel.Version = version
```
crodriguezvega, line 729: Why was this removed? In `WriteUpgradeInitChannel` the version of the channel [is updated with the one in the proposed upgrade channel](https://github.com/cosmos/ibc-go/blob/cian/issue%231619-implement-channelupgradetry-functionality-for-04-channel/modules/core/04-channel/keeper/upgrade.go#L97), but without this, we wouldn't take into account the version returned by the application callback. Maybe instead of modifying the proposed upgrade channel, the version can be passed as a separate argument, as it is done in this PR for `WriteUpgradeTryChannel`.

crodriguezvega (commented):
You guys are rockin' it with this one. 🪨 🎸 
Left a couple of extra comments. :)

modules/core/keeper/msg_server.go
chatton: It looks like this is an oversight from a refactor that we did to [`WriteUpgradeTryChannel`](https://github.com/cosmos/ibc-go/blob/00a313677d15572d3677802088ee885e4ed410c6/modules/core/keeper/msg_server.go#L807)

that refactor was intended to be more explicit by not modifying the input message and passing the version directly. I think we should still not modify the message, but we should pass the version into WriteUpgradeInit in the same way we do in the linked code.

modules/core/keeper/msg_server.go
chatton: Actually technically it's not more gas, the current code was doing a call to the restore channel in order to return it anyway, I'm happy enough to pull this out to the msg server level.

modules/core/04-channel/keeper/upgrade.go
chatton: Are you happy if we introduce a helper function in a follow up?

crodriguezvega: Yes, more than happy! :)

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -75,34 +65,226 @@ func (k Keeper) ChanUpgradeInit(
...
+
+	if err := k.connectionKeeper.VerifyChannelUpgradeSequence(ctx, connectionEnd, proofHeight, proofUpgradeSequence, channel.Counterparty.PortId,
+		channel.Counterparty.ChannelId, counterpartyUpgradeSequence); err != nil {
+		return 0, err
+	}
+
+	// check if upgrade timed out by comparing it with the latest height of the chain
+	selfHeight := clienttypes.GetSelfHeight(ctx)
+	if !timeoutHeight.IsZero() && selfHeight.GTE(timeoutHeight) {
+		if err := k.RestoreChannelAndWriteErrorReceipt(ctx, portID, channelID, upgradeSequence, types.ErrUpgradeTimeout); err != nil {
```
crodriguezvega, line 157: Restoring the channel and writing a receipt here would only be needed if the channel state is in `INITUPGRADE`, right? (crossing hello). Otherwise, just aborting the transaction and calling timeout on the counterparty would be enough, right? I think the same applies for the timestamp check below.

chatton: yes you're right, this actually ended up happening in #3418 which is probably not where that change should have been introduced 😓

charleenfei:
closing in favour of https://github.com/cosmos/ibc-go/pull/3478

Issue #1612, opened 2022-06-29
Implement `RestoreChannel` functionality for upgrade cancellation


damiannolan:
## Summary

`RestoreChannel` is called upon aborting a channel upgrade handshake in-progress.

```go
func (k Keeper) RestoreChannel(ctx sdk.Context, portID, channelID string, upgradeSequence uint64) error {
    return nil    
}
```

`RestoreChannel` should reinstate the `channelEnd` to its original pre-upgrade state while also setting the `errorReceipt` in state. 

A relayer can then send a `ChanUpgradeCancelMsg` to the counterparty so that it can restore its `channelEnd` to its pre-upgrade state as well. Once both channel ends are back to the pre-upgrade state, packet processing will resume with the original channel and application parameters.

[See ICS004 - Upgrades specification for details](https://github.com/cosmos/ibc/blob/master/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#Utility-Functions)
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

Issue #2471, opened 2022-10-05
Abstract IBC exported interface types to separate go module


damiannolan:
## Summary

Opening this issue to discuss and explore the idea of potentially abstracting out the interface types maintained under `modules/core/exported` to a separate Go module. 

## Problem Definition

One issue we face in the cosmos ecosystem with respect to IBC is that IBC applications maintained in their own repositories must stay in lock step with `ibc-go` releases. That is, in order for an IBC application to be included in a chain distribution's `app.go` it must depend on the same version of `ibc-go` that the chain itself depends on. 
The same would be true for applications maintained inside of `ibc-go` if #614 was to be implemented. Applications _must_ follow the `ibc-go` release cycle and release in tandem an appropriate application version that is compatible. 

The problem arises when wiring the IBC [`Router`](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/types/router.go#L11) in `app.go`, specifically, compatibility issues arise between `ibc-go` versions for the [`IBCModule`](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/types/module.go#L13) interface when invoking [`AddRoute()`](https://github.com/cosmos/ibc-go/blob/main/testing/simapp/app.go#L456). The Go compiler breaks if an `IBCModule` depends on a different version of `ibc-go` due to the parameters defined in the various handshake callback methods and packet APIs.
This includes both concrete types _and_ interface types defined within `ibc-go`.

## Proposal

A potential solution could be to extract the existing interfaces defined in `modules/core/exported` to a separate Go module. As well as the various parameters defined in the handshake callbacks and packet APIs of the `IBCModule` interface.
This would allow IBC applications that live in their own repositories to freely maintain their own release cycle and only update to newer versions of `ibc-go` when explicitly forced to do so.

A spike or investigation into this as a potential solution could be carried out. This would help determine the pros and cons of the changes to be made and if it would be worth the effort in doing so.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chatton:
Some additional thoughts here, I think depending on an interface is always a good idea. In this particular case, if the main problem is the method signature on the interface functions, it should also be possible to accept a single argument in the interface functions, some sort of options type (this would of course be a breaking change for all existing versions of ibc-go)

```golang
type OnAckOpts struct {
   ...
}
```

[...]

PR #3352, opened 2023-03-27
Implement Restore Channel Functionality

chatton:
## Description

closes: #1612

This PR adds a keeper function to restore an upgrade channel.

Note: there are some issues with routing the application callback. See comments in the review for more details, I don't want to merge until we figure out the best resolution.

~Will wait for #3331 to be merged before marking as R4R~

---

write a little note why.


modules/core/keeper/keeper.go
```diff
@@ -64,14 +64,14 @@ func NewKeeper(
...
-	channelKeeper := channelkeeper.NewKeeper(cdc, key, clientKeeper, connectionKeeper, portKeeper, scopedKeeper)
+	channelKeeper := channelkeeper.NewKeeper(cdc, key, clientKeeper, connectionKeeper, &portKeeper, scopedKeeper)
 
 	return &Keeper{
 		cdc:              cdc,
 		ClientKeeper:     clientKeeper,
 		ConnectionKeeper: connectionKeeper,
 		ChannelKeeper:    channelKeeper,
-		PortKeeper:       portKeeper,
+		PortKeeper:       &portKeeper,
```
chatton, line 74: the port keeper's router is assigned [here](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/core/keeper/keeper.go#L90) however k.portKeeper is a value type instead of a reference type, so the assignment is not preserved across function calls. This means that if `k.portKeeper.Router` is ever accessed, it will have been an empty value.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -89,3 +98,41 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+
+	module, _, err := k.LookupModuleByChannel(ctx, portID, channelID)
+	if err != nil {
+		return errorsmod.Wrap(err, "could not retrieve module from port-id")
+	}
+
+	portKeeper, ok := k.portKeeper.(*portkeeper.Keeper)
+	if !ok {
+		panic("todo: handle this situation")
+	}
```
chatton, line 130: would love some input about this. A few things were attempted.

1. Add an entry to expected keepers. - this created a circular import between 04-channel and 05-port
2. Add IBCModule to exported package. - this created a circular import between exported and 04-channel as the interface imports types from channel.

Casting directly to `portkeeper.Keeper` defeats the purpose of an expected keepers interface. Let me know if there is a nice way to work around this without introducing more packages.

modules/core/keeper/keeper.go
charleenfei: nice catch!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -90,3 +99,41 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+
+	cbs, found := portKeeper.Router.GetRoute(module)
+	if !found {
+		return errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+	}
+
+	return cbs.OnChanUpgradeRestore(ctx, portID, channelID)
```
crodriguezvega, line 138: I know this is the logic that the spec specifies at the moment, but I am wondering if we could change the implementation (maybe the spec as well) to do the application callback in the `ChannelUpgradeTry` function in `core/keeper/msg_server.go` so that either we execute there the callback for success or the callback for restore depending on the error returned by `ChanUpgradeTry` in `04-channel/keeper/upgrade.go`. I think that would make the code cleaner, since we can keep it consistent at which layer we execute application callbacks.

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -121,3 +121,81 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+			suite.coordinator.Setup(path)
+
+			originalChannel := path.EndpointA.GetChannel()
+
+			chanCap, _ := suite.chainA.GetSimApp().GetScopedIBCKeeper().GetCapability(suite.chainA.GetContext(), host.ChannelCapabilityPath(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID))
+			channelUpgrade = types.NewChannel(types.INITUPGRADE, types.UNORDERED, types.NewCounterparty(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID), []string{path.EndpointA.ConnectionID}, fmt.Sprintf("%s-v2", mock.Version))
+
+			tc.malleate()
```
crodriguezvega, line 164: Do we need to call `malleate` here if it's called later on as well?

modules/core/keeper/keeper.go
chatton: We only ever accessed the router via `channelKeeper.Router` and never through `portKeeper.Router`

modules/core/04-channel/keeper/upgrade.go
crodriguezvega: Maybe if we can [execute the callback at the message server layer](https://github.com/cosmos/ibc-go/pull/3352/files#r1151544321), then we don't have this problem? :)

modules/core/04-channel/keeper/upgrade_test.go
chatton: good catch, this looks like a mistake 👍

modules/core/04-channel/types/errors.go
```diff
@@ -37,8 +37,10 @@ var (
...
-	ErrInvalidChannelVersion = errorsmod.Register(SubModuleName, 24, "invalid channel version")
-	ErrPacketNotSent         = errorsmod.Register(SubModuleName, 25, "packet has not been sent")
-	ErrInvalidTimeout        = errorsmod.Register(SubModuleName, 26, "invalid packet timeout")
-	ErrInvalidUpgradeTimeout = errorsmod.Register(SubModuleName, 27, "either timeout height or timeout timestamp must be non-zero")
+	ErrInvalidChannelVersion  = errorsmod.Register(SubModuleName, 24, "invalid channel version")
+	ErrPacketNotSent          = errorsmod.Register(SubModuleName, 25, "packet has not been sent")
+	ErrInvalidTimeout         = errorsmod.Register(SubModuleName, 26, "invalid packet timeout")
+	ErrInvalidUpgradeTimeout  = errorsmod.Register(SubModuleName, 27, "either timeout height or timeout timestamp must be non-zero")
+	ErrRestoreChannelNotFound = errorsmod.Register(SubModuleName, 28, "restore channel not found")
+	ErrChannelRestored        = errorsmod.Register(SubModuleName, 29, "channel restored")
```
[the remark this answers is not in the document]
chatton: Yeah it feels a little off, what do you think makes the most sense here for which error type we should be writing in the ErrorReceipt?

modules/core/04-channel/keeper/upgrade.go
AdityaSripal: Should IBCModule be pulled out into its own separate package?

Maybe with one day giving it a separate go.mod

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -121,3 +121,81 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+				chanCap, channelUpgrade, path.EndpointB.Chain.GetTimeoutHeight(), 0,
+			)
+
+			suite.Require().NoError(err)
+
+			// update the channel to have some other state.
+			modifiedChannel := originalChannel
+			modifiedChannel.Version = "different-version"
+			modifiedChannel.ConnectionHops = []string{"different-connection-id"}
+			path.EndpointA.Chain.GetSimApp().IBCKeeper.ChannelKeeper.SetChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, modifiedChannel)
```
AdityaSripal, line 177: Why do you have this code block?

AdityaSripal (changes requested):
Nice work!!

modules/core/04-channel/keeper/upgrade.go
damiannolan: Definitely worth considering, but I think its likely part of a much larger body of work and related to https://github.com/cosmos/ibc-go/issues/2471

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -90,3 +99,41 @@ func (k Keeper) WriteUpgradeInitChannel(
 
 	emitChannelUpgradeInitEvent(ctx, portID, channelID, upgradeSequence, channelUpgrade)
 }
+
+// RestoreChannel restores the given channel to the state prior to upgrade.
+func (k Keeper) RestoreChannel(ctx sdk.Context, portID, channelID string, upgradeSequence uint64) error {
+	_, code, _ := errorsmod.ABCIInfo(types.ErrChannelRestored, false) // discard non-determinstic codespace and log values
```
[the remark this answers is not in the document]
chatton: Are you thinking something like

`RestoreChannel(ctx sdk.Context, portID, channelID string, upgradeSequence uint64, err error) error` ?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -90,3 +99,41 @@ func (k Keeper) WriteUpgradeInitChannel(
...
+
+	errorReceipt := types.ErrorReceipt{
+		Sequence: upgradeSequence,
+		Error:    fmt.Sprintf("ABCI code: %d: %s", code, restoreErrorString),
```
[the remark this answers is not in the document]
chatton: I was using [this](https://github.com/cosmos/ibc-go/blob/5ae77b2a2fc35328505ee753d7ff524d3b66b657/modules/core/04-channel/types/acknowledgement.go#L32-L42) as a reference. It seems we always include the same error string, but emit details through events. If we have many different constants for strings, simply switching the constant use could cause an apphash mismatch.

What do you think about including more details in events for this use case too?

modules/core/04-channel/keeper/upgrade_test.go
chatton: after discussion on slack: It would make more sense to actually test `ibckeeper. ChanUpgradeInit` as we would be better able to test the full flow.

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -121,3 +121,72 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+
+			upgradeSequence := uint64(1)
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+			suite.coordinator.Setup(path)
+
+			path.EndpointA.ChannelConfig.Version = fmt.Sprintf("%s-v2", mock.Version)
+
+			originalChannel := path.EndpointA.GetChannel()
+
+			err := path.EndpointA.ChanUpgradeInit(path.EndpointB.Chain.GetTimeoutHeight(), 0)
```
chatton, line 163: updated the test to broadcast the message directly as the updated channel gets written in `ibckeeper.OnChanUpgradeInit` instead of `channelkeeper.OnChanUpgradeInit`.

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -484,6 +501,28 @@ func (suite *KeeperTestSuite) TestSetRestoreChannel() {
 	suite.Require().Equal(expChannel, restoreChannel)
 }
 
+func (suite *KeeperTestSuite) TestDeleteUpgradeRestoreChannel() {
```
AdityaSripal, line 504: nit: But I think you can test all store functions, `Set/Get/Delete` in a single test function. The prior test is just a subset of this test

AdityaSripal (approved):
Looks great! Just a final nit on testing, doesn't need to block this PR

modules/core/04-channel/keeper/upgrade.go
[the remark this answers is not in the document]
AdityaSripal: Yes

modules/core/04-channel/keeper/keeper_test.go
damiannolan: Agree!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -11,9 +11,12 @@ import (
 	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
 	connectiontypes "github.com/cosmos/ibc-go/v7/modules/core/03-connection/types"
 	"github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+	portkeeper "github.com/cosmos/ibc-go/v7/modules/core/05-port/keeper"
+	porttypes "github.com/cosmos/ibc-go/v7/modules/core/05-port/types"
 	host "github.com/cosmos/ibc-go/v7/modules/core/24-host"
 )
 
+
```
damiannolan, line 19: formatter says no

```suggestion
```

damiannolan (approved):
Nice work!

Happy to revisit the app callback routing for `OnRestoreChannel` later. If you are going to close the issue maybe we can open another one for that.

🚀

modules/core/04-channel/keeper/upgrade.go
chatton: how dare the formatter question my whitespace.

modules/core/04-channel/keeper/keeper_test.go
chatton: I refactored the tests to use subtests for each field 👍

Issue #3349, opened 2023-03-27
Add a `UnmarshalPacketData` optional interface function to support ADR 8 callbacks


colin-axner:
## Summary

In order for ADR 8 to work generically with the given middleware construction, it needs to ask the underlying app to decode the packet data bytes into an interface function. 

## Problem Definition

Middleware has no way of generically decoding packet data bytes without making assertions on the underlying app

## Proposal

Add an optional interface which IBC apps can implement. It should take in bytes and return an interface. For ICS 20 the `FungibleTokenPacketData` should be returned. For ICS 27 the `InterchainAccountsPacketData` should be returned. ADR 8 can cast these interfaces to the `CallbacksPacketData` if supported. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3353, opened 2023-03-27
imp: add UnmarshalPacketData interface function

colin-axner:
## Description

Added a quick implementation. I just chose the first names that came to mind, please feel free to suggest an alternative

closes: #3349 

```bash
imp: add optional PacketDataUnmarshaler interface to allow middlewares to decode packet data into interfaces. 
```

---

write a little note why.


modules/apps/29-fee/ibc_middleware.go
```diff
@@ -347,3 +350,14 @@ func (im IBCMiddleware) WriteAcknowledgement(
 func (im IBCMiddleware) GetAppVersion(ctx sdk.Context, portID, channelID string) (string, bool) {
 	return im.keeper.GetAppVersion(ctx, portID, channelID)
 }
+
+// UnmarshalPacketData attempts to use the underlying app to unmarshal the packet data.
+// If the underlying app does not support the PacketDataUnmarshaler interface, an error is returned.
+func (im IBCMiddleware) UnmarshalPacketData(bz []byte) (interface{}, error) {
```
colin-axner, line 356: this is only necessary if the ADR 8 implementation is higher than ICS 29, which should not be the case. I added the function anyways so we could discuss here whether to keep it or not

colin-axner:
Will rebase onto the feature branch after the first pr is merged

modules/apps/transfer/ibc_module.go
```diff
@@ -300,3 +305,15 @@ func (im IBCModule) OnTimeoutPacket(
 
 	return nil
 }
+
+// UnmarshalPacketData attempts to unmarshal the provided packet data bytes
+// into a FungibleTokenPacketData. This function implements the optional
+// PacketDataUnmarshaler interface required for ADR 008 support.
+func (im IBCModule) UnmarshalPacketData(bz []byte) (interface{}, error) {
```
damiannolan, line 312: do we have a `PacketData` interface, and would it be helpful or have any benefit?

e.g. something like

```go
type PacketData interface {
    GetBytes() []byte
    ValidateBasic() error
}
```

modules/apps/29-fee/ibc_middleware.go
damiannolan: hmm, I guess its best to keep it?

modules/apps/29-fee/types/errors.go
```diff
@@ -16,4 +16,5 @@ var (
 	ErrFeeNotEnabled                 = errorsmod.Register(ModuleName, 9, "fee module is not enabled for this channel. If this error occurs after channel setup, fee module may not be enabled")
 	ErrRelayerNotFoundForAsyncAck    = errorsmod.Register(ModuleName, 10, "relayer address must be stored for async WriteAcknowledgement")
 	ErrFeeModuleLocked               = errorsmod.Register(ModuleName, 11, "the fee module is currently locked, a severe bug has been detected")
+	ErrUnsupportedAction             = errorsmod.Register(ModuleName, 12, "unsupported action")
```
damiannolan, line 19: nit: maybe this go into `internal/errors` pkg? Seems pretty generic

feel free to ignore

damiannolan (approved):
Super clean as always. Left a few minor suggestions/questions, otherwise LGTM 🚀

modules/core/05-port/types/module.go
```diff
@@ -138,3 +138,9 @@ type Middleware interface {
 	IBCModule
 	ICS4Wrapper
 }
+
+// PacketDataUnmarshaler defines an optional interface which allows a middleware to
+// request the packet data to be unmarshaled by the base application.
+type PacketDataUnmarshaler interface {
+	UnmarshalPacketData([]byte) (interface{}, error)
+}
```
chatton, line 146: I looked very briefly into what this could look like with generics as I think it would be nice to be able to specify the data type that will be unmarshalled, but things get a little tricky with middleware!

chatton (approved):
LGTM! Nice job, I was hoping this would be a perfect use case for generics but I think it would end up adding a lot of complexity. (which seems to usually be the case 😓 )

e.g.

```golang


type PacketDataUnmarshaler [T any] interface {
	UnmarshalPacketData([]byte) (T, error)
}

var _ porttypes.PacketDataUnmarshaler[types.FungibleTokenPacketData] = IBCModule{}


// UnmarshalPacketData attempts to unmarshal the provided packet data bytes
// into a FungibleTokenPacketData. This function implements the optional
// PacketDataUnmarshaler interface required for ADR 008 support.
func (im IBCModule) UnmarshalPacketData(bz []byte) (types.FungibleTokenPacketData, error) {
	var packetData types.FungibleTokenPacketData
	if err := types.ModuleCdc.UnmarshalJSON(bz, &packetData); err != nil {
		return types.FungibleTokenPacketData{}, err
	}

	return packetData, nil
}

```

modules/apps/29-fee/ibc_middleware.go
AdityaSripal: Agree that it should be kept. Let's remove ordering requirements from our stack whenever possible

modules/core/05-port/types/module.go
AdityaSripal: @alpe I think wrote up a quick POC on generics here: https://github.com/cosmos/ibc-go/pull/3359

Planning to compare the two solutions this week

AdityaSripal (approved):
Thank you @colin-axner !!

modules/apps/transfer/ibc_module.go
colin-axner: We could do this, but it wouldn't provide any functional benefits, would just clean up the API slightly. It would be an API breaking change to add which is why we have avoided it for now. I believe the interface would likely become unnecessary in the existence of a flat map packet data

modules/core/05-port/types/module.go
colin-axner: I think the solution with generics that could potentially work here is to change the interface function from `UnmarshalPacketData` to `GetPacketDataType() T` which returns the packet data type to unmarshal into. The only benefit is that the apps wouldn't have to call `cdc.Unmarshal`, but I'm not super convinced this is a significant benefit, especially given that this interface may eventually be removed with the existence of a flat map packet data

Happy to review a follow pr if someone wanted to explore that solution

colin-axner:
merging! Thanks for the reviews everyone :)

Issue #1783, opened 2022-07-26
Confusion regarding next sequence recv


adizere:
## Summary

A relayer operator reported this confusion:

> what is the expected behaviour of the /ibc/core/channel/v1/channels/{channel_id}/ports/{port_id}/next_sequence endpoint? it seems to always return “next_sequence_receive”: “1" (https://rest-cosmoshub.ecostake.com/ibc/core/channel/v1/channels/channel-141/ports/transfer/next_sequence)

The answer, thanks for @crodriguezvega, was:

> I see that this endpoint [retrieves the next sequence recv](https://github.com/cosmos/ibc-go/blob/48a6ae512b4ea42c29fdf6c6f5363f50645591a2/modules/core/04-channel/keeper/grpc_query.go#L482) and I think that value [is set on ](https://github.com/cosmos/ibc-go/blob/48a6ae512b4ea42c29fdf6c6f5363f50645591a2/modules/core/04-channel/keeper/packet.go#L281)RecvPacket[ only for ordered channels](https://github.com/cosmos/ibc-go/blob/48a6ae512b4ea42c29fdf6c6f5363f50645591a2/modules/core/04-channel/keeper/packet.go#L281). So I guess it returns 1 because the transfer channel is unordered and the next sequence recv has never been incremented after being initialized to 1 either in ChanOpenInit ([here](https://github.com/cosmos/ibc-go/blob/7370a8b438b5a7d9e3982673bcfbfc550145b1f2/modules/core/04-channel/keeper/handshake.go#L83)) or ChanOpenTry ([here](https://github.com/cosmos/ibc-go/blob/7370a8b438b5a7d9e3982673bcfbfc550145b1f2/modules/core/04-channel/keeper/handshake.go#L192)).

## Expected Behaviour

No confusion regarding this undocumented feature.

Possible suggestion for a fix:

> Just an idea: could it be set to `null` for unordered channels to avoid the confusion?

## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Thanks for opening the issue @adizere! I agree with the suggested fix of using null as a return on unordered channels

crodriguezvega:
[...]

PR #3357, opened 2023-03-28
Handle NextSequenceReceive for unordered channels.

DimitrisJim:
## Description

closes: #1783 

```bash
fix: handle unordered channels in `NextSequenceReceive` query
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/grpc_query.go
+++ b/modules/core/04-channel/keeper/grpc_query.go
@@ -480,14 +480,25 @@ func (q Keeper) NextSequenceReceive(c context.Context, req *types.QueryNextSeque
 	}
 
 	ctx := sdk.UnwrapSDKContext(c)
-	sequence, found := q.GetNextSequenceRecv(ctx, req.PortId, req.ChannelId)
+	channel, found := q.GetChannel(ctx, req.PortId, req.ChannelId)
 	if !found {
 		return nil, status.Error(
 			codes.NotFound,
-			errorsmod.Wrapf(types.ErrSequenceReceiveNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
+			errorsmod.Wrapf(types.ErrChannelNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
 		)
 	}
 
+	// Return the next sequence received for ordered channels and 0 for unordered channels.
+	var sequence uint64
+	if channel.Ordering == types.ORDERED {
+		sequence, found = q.GetNextSequenceRecv(ctx, req.PortId, req.ChannelId)
+		if !found {
+			return nil, status.Error(
+				codes.NotFound,
+				errorsmod.Wrapf(types.ErrSequenceReceiveNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
+			)
+		}
+	}
 	selfHeight := clienttypes.GetSelfHeight(ctx)
 	return types.NewQueryNextSequenceReceiveResponse(sequence, nil, selfHeight), nil
 }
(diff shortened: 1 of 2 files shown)


charleenfei (approved):
nice work! lgtm

AdityaSripal (approved):
LGTM!

DimitrisJim:
Thanks for the reviews! :raised_hands:

modules/core/04-channel/keeper/grpc_query.go
```diff
@@ -480,14 +480,25 @@ func (q Keeper) NextSequenceReceive(c context.Context, req *types.QueryNextSeque
...
 			codes.NotFound,
-			errorsmod.Wrapf(types.ErrSequenceReceiveNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
+			errorsmod.Wrapf(types.ErrChannelNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
 		)
 	}
 
+	// Return the next sequence received for ordered channels and 0 for unordered channels.
+	var sequence uint64
+	if channel.Ordering == types.ORDERED {
+		sequence, found = q.GetNextSequenceRecv(ctx, req.PortId, req.ChannelId)
```
colin-axner, line 494: Thanks @DimitrisJim! Code looks great! 

Just want to note that if we add another channel ordering, like `ORDERED_ALLOW_TIMEOUT` we may need to adjust this handling. It's not incredibly clear whether `ORDERED` channels or `UNORDERED` channels are the exception here, but since we are referring to the next sequence receive, it probably makes sense to add an if statement for the channel types which do not use that value (UNORDERED). I see the referenced issue suggested this structure of code.

An alternative solution:
```go
if channel.Ordering == types.UNORDERED {
    // unordered channels do not make use of the next sequence receive
    return 0, nil
}

sequence, found := q.GetNextSequenceRecv(ctx, req.PortId, req.ChannelId)

// etc
```

In this scenario, since we perform the query by default, `ORDERED_ALLOW_TIMEOUT` would function properly without additional changes. I believe we will need to modify this code with the addition of ordered allow timeout channels (not yet implemented)

DimitrisJim: Yep, can totally see that point, wasn't aware of the possibility of another ordered variant. Does it make sense to fix it pronto? (Note that some grep-ing found one other case where we special case on `ORDERED` during [packet acknowledgement](https://github.com/cosmos/ibc-go/blob/ae27fa5960d523e4476776308ecd0b4cd3377671/modules/core/04-channel/keeper/packet.go#L442) but I'm unsure if the same semantics would apply).

colin-axner: Yea, the new ordered variant is a bit of a new concept for us as well, which is why this issue wasn't so obvious. I only barely made the connection thinking about explicit return values (I prefer the code to be explicit with what it returns rather than being implicit by not executing some conditional)

I think it could make sense to fix since we know it will be an issue? Might save someone some time debugging later. I'm also perfectly happy having an issue opened. 

> (Note that some grep-ing found one other case where we special case on ORDERED during [packet acknowledgement](https://github.com/cosmos/ibc-go/blob/ae27fa5960d523e4476776308ecd0b4cd3377671/modules/core/04-channel/keeper/packet.go#L442) but I'm unsure if the same semantics would apply).

When implementing the new channel ordering, we will need to modify this code (already specified in the spec). We will likely need to do some grep-ing to look into all switches/conditionals on channel ordering. Core IBC isn't too well setup abstraction wise for new channel types. It's somewhat unclear how many different ordering types will ever exist

DimitrisJim: I'll fix this up today. Better to keep things future proof as much as possible.

Issue #3264, opened 2023-03-08
Avoid sequence mismatch when using relayer wallet to sign ibc tx msgs


colin-axner:
## Summary

An indeterminstic failure can be seen [here](https://github.com/cosmos/ibc-go/actions/runs/4366212377/jobs/7636050384)
```
  testsuite.go:431: 
        	Error Trace:	/home/runner/work/ibc-go/ibc-go/e2e/testsuite/testsuite.go:431
        	            				/home/runner/work/ibc-go/ibc-go/e2e/tests/interchain_accounts/incentivized_test.go:292
        	Error:      	Should not be: 0
        	Test:       	TestIncentivizedInterchainAccountsTestSuite/TestMsgSendTx_FailedBankSend_Incentivized
        	Messages:   	code: 32
        	            	codespace: sdk
        	            	data: ""
        	            	events: []
        	            	gas_used: "0"
        	            	gas_wanted: "0"
        	            	height: "0"
        	            	info: ""
        	            	logs: []
        	            	raw_log: 'account sequence mismatch, expected 9, got 8: incorrect account sequence'
        	            	timestamp: ""
        	            	tx: null
        	            	txhash: B915E76EC510455F9A27ECA78521CFC0E97F6FE3059BEC1CF4A2DC5D96C0F528
```

   	            	

## Problem Definition

We use the [relayer wallet](https://github.com/cosmos/ibc-go/blob/main/e2e/tests/interchain_accounts/incentivized_test.go#L119) to sign the `MsgRegisterCounterpartyPayee` which makes sense, but this causes a sequence mismatch since the relayer wallet may be submitting a tx at the same time.

## Proposal

Not sure since the `MsgRegisterCounterpartyPayee` needs to be signed by the relayer wallet. Maybe we can stop the relayer/restart while sending this tx?

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
I think we can actually look for a substring match of `account sequence mismatch` and then retry the tx?

chatton:
I think for now we can specifically retry on the mismatch error, but it might be useful to have a mechanism to retry on one of a list of errors if more pop up.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #3402, opened 2023-04-03
fix: allow retries for messages signed by relayer.

DimitrisJim:
## Description

closes: #3264

```bash
fix: allow retries for messages signed by relayer.
```

---

write a little note why.


e2e/testsuite/testsuite.go
```diff
@@ -611,6 +628,43 @@ func (s *E2ETestSuite) QueryGranterGrants(ctx context.Context, chain *cosmos.Cos
...
+	if s.relayers == nil {
+		return false
+	}
+	return s.relayers[user]
+}
+
+// retryNtimes retries the provided function up to the provided number of attempts.
+func (s *E2ETestSuite) retryNtimes(f func() (sdk.TxResponse, error), attempts int) (sdk.TxResponse, error) {
+	// Ignore account sequence mismatch errors.
+	allowedMessages := []string{"account sequence mismatch"}
```
crodriguezvega, line 642: Since this is a configuration parameter, maybe we can move it to the top [together with the `const`](https://github.com/cosmos/ibc-go/blob/main/e2e/testsuite/testsuite.go#L48-L55)? We can make it an unexpected `var`.

e2e/testsuite/testsuite.go
```diff
@@ -611,6 +628,43 @@ func (s *E2ETestSuite) QueryGranterGrants(ctx context.Context, chain *cosmos.Cos
...
+		resp, err = f()
+		if !containsMessage(resp.RawLog, allowedMessages) {
+			return resp, err
+		}
+		s.T().Logf("retrying tx due to ignored raw log: %s", resp.RawLog)
+	}
+	return resp, err
+}
+
+// containsPrefix returns true if the string s contains any of the prefixes in the slice.
```
crodriguezvega, line 656:
```suggestion
// containsMessage returns true if the string s contains any of the prefixes in the slice.
```

e2e/testsuite/testsuite.go
```diff
@@ -611,6 +628,43 @@ func (s *E2ETestSuite) QueryGranterGrants(ctx context.Context, chain *cosmos.Cos
...
+func containsMessage(s string, messages []string) bool {
+	var hasPrefix bool
```
crodriguezvega, line 658:
```suggestion
	var containsText bool
```

e2e/testsuite/testsuite.go
crodriguezvega, line 642:
```suggestion
	retryMessages := []string{"account sequence mismatch"}
```

e2e/testsuite/testsuite.go
```diff
@@ -281,7 +288,17 @@ func (s *E2ETestSuite) BroadcastMessages(ctx context.Context, chain *cosmos.Cosm
...
 
-	resp, err := cosmos.BroadcastTx(ctx, broadcaster, user, msgs...)
+	// Retry the operation a few times if the user signing the transaction is a relayer. (See issue #3264)
+	var resp sdk.TxResponse
+	var err error
+	broadcastFunc := func() (sdk.TxResponse, error) {
+		return cosmos.BroadcastTx(ctx, broadcaster, user, msgs...)
+	}
+	if s.IsRelayer(user) {
+		resp, err = s.retryNtimes(broadcastFunc, 5)
```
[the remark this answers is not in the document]
crodriguezvega: Maybe we can make it a const and put it in the top of the file as well?

e2e/testsuite/testsuite.go
```diff
@@ -611,6 +628,43 @@ func (s *E2ETestSuite) QueryGranterGrants(ctx context.Context, chain *cosmos.Cos
...
+
+// retryNtimes retries the provided function up to the provided number of attempts.
+func (s *E2ETestSuite) retryNtimes(f func() (sdk.TxResponse, error), attempts int) (sdk.TxResponse, error) {
+	// Ignore account sequence mismatch errors.
+	allowedMessages := []string{"account sequence mismatch"}
+	var err error
+	var resp sdk.TxResponse
+	for i := 0; i < attempts; i++ {
+		// If the response's raw log doesn't contain any of the allowed prefixes we return, else, we retry.
+		resp, err = f()
```
crodriguezvega, line 647: Small optimization: should we also check that `err` is not nil? Otherwise, if it is `nil` then the transaction should be succeeded, right? And we don't need to check the response log or retry.

e2e/testsuite/testsuite.go
```diff
@@ -611,6 +628,43 @@ func (s *E2ETestSuite) QueryGranterGrants(ctx context.Context, chain *cosmos.Cos
...
+		if !containsMessage(resp.RawLog, allowedMessages) {
+			return resp, err
+		}
+		s.T().Logf("retrying tx due to ignored raw log: %s", resp.RawLog)
```
crodriguezvega, line 651: Not sure what you mean with `ignored raw log`...

crodriguezvega (commented):
Thank you, @DimitrisJim! It's looking good; I just left a bunch for nits.

e2e/testsuite/testsuite.go
DimitrisJim: oh `resp.RawLog` is what we look into to check if we should retry. Open to suggestions for a better log message here.

e2e/testsuite/testsuite.go
```diff
@@ -60,6 +60,7 @@ type E2ETestSuite struct {
 
 	grpcClients    map[string]GRPCClients
 	paths          map[string]path
+	relayers       map[ibc.Wallet]bool
```
chatton, line 63: while we do currently just run a single test in each run, we should try and make sure that if we change that we need to make as few changes as possible. With relayers being at the top level, this means that these relayer wallets would be shared across all tests. What might make sense to do is to create a mapping of testName -> relayer pair. Similar to how we do [here](https://github.com/cosmos/ibc-go/blob/9bca644453430b5daab860cc6e3bb1fd3b45245d/e2e/testsuite/testsuite.go#L250).

This could even be a `map[string]map[ibc.Wallet]bool` which isn't super pretty, but maybe we could work with an alias of some sort here.

e2e/testsuite/testsuite.go
DimitrisJim: As far as I've understood the issue, `err` is `nil` when this happens. The test was failing when we tried to check that the `resp` was a valid tx response [here](https://github.com/cosmos/ibc-go/blob/d840c699d988e32cb55f2c289d69f8ef2e5cea50/e2e/testsuite/testsuite.go#L431) and it failed when `gasUsed == 0`.

Happy to change if I misunderstood something.

e2e/testsuite/testsuite.go
chatton: one thing to watch out for is that sometimes the rawlog doesn't actually container a whole lot of information. It might be useful to dump the full tx so it is more clearly obvious it is an empty response. We do that [here](https://github.com/cosmos/ibc-go/blob/9bca644453430b5daab860cc6e3bb1fd3b45245d/e2e/testsuite/testsuite.go#L439)

e2e/testsuite/testsuite.go
```diff
@@ -611,6 +628,43 @@ func (s *E2ETestSuite) QueryGranterGrants(ctx context.Context, chain *cosmos.Cos
...
+	}
+	return resp, err
+}
+
+// containsPrefix returns true if the string s contains any of the prefixes in the slice.
+func containsMessage(s string, messages []string) bool {
+	var hasPrefix bool
+	for _, message := range messages {
+		if strings.Contains(s, message) {
+			hasPrefix = true
```
chatton: [nit] we can just return here instead of using a boolean and breaking

chatton (commented):
Looking great! Left a few comments/suggestions

e2e/testsuite/testsuite.go
DimitrisJim: does it make sense to add an additional check for the error here? Should we retry the message if an error occurred but `RawLog` contains one of the retry messages?

e2e/testsuite/testsuite.go
DimitrisJim: nice catch, we don't actually need a separate variable to keep track of it.

e2e/testsuite/testsuite.go
chatton: hm, maybe we can do

```go
resp, err = f()

if !s.IsValidTxResponse(resp) {
    continue
}

if !containsMessage(resp.RawLog, retryMessages) {
    return resp, err
}
// continue loop 
```

This would involve adding a function `IsValidTxResponse` (which would just do what the assertion function does but return a bool I guess)

DimitrisJim: I'm probably missing something here :smile: I was under the impression we'd like to skip invalid tx's that contain those messages in their `RawLog`, any other invalid `tx`s are just returned back to broadcast message which will eventually pass them over to its callee and have `AssertValidTxResponse` fail on us.

If we add a call to check if its a valid response beforehand that would `continue` without leaving us the chance to check for the `RawLog` (basically a retry again but for cases where we might not want that). 

I do think I'm not seeing something obvious though.

chatton: you are definitely correct! I was getting mixed up!   We definitely would need to do the contains check first. 

Thinking about it some more, I think we may be able to fail in the case of an error, and exit entirely then, but retry if the raw logs contain one of these strings.

The failures we're seeing because of the sequence mismatch are returning a nil error, so an error in the broadcast itself can still be considered a failure.

colin-axner (commented):
Excellent work @DimitrisJim! Left a few suggestions, but overall looks great to me! Have we seen this solution work in the logs?

e2e/testsuite/testsuite.go
DimitrisJim: I did go with an alias here and basically just added a couple of functions on it. Lmk what you think.

e2e/testsuite/testsuite.go
DimitrisJim: I've added the check for err as suggested. I'll open the issue if there's at least a 2/3 consensus.

e2e/testsuite/testsuite.go
```diff
@@ -611,6 +628,43 @@ func (s *E2ETestSuite) QueryGranterGrants(ctx context.Context, chain *cosmos.Cos
...
+	return resp, err
+}
+
+// containsMessages returns true if the string s contains any of the messages in the slice.
+func containsMessage(s string, messages []string) bool {
+	var containsText bool
+	for _, message := range messages {
+		if strings.Contains(s, message) {
+			containsText = true
+			break
```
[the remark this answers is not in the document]
colin-axner: Was there a reason for using the `containsText` var?

e2e/testsuite/testsuite.go
```diff
@@ -105,6 +105,25 @@ func newPath(chainA, chainB *cosmos.CosmosChain) path {
...
+	r[testName][relayer] = true
+}
+
+// containsRelayer returns true if the given relayer is in the relayer set for the given test name.
+func (r relayerMap) containsRelayer(testName string, wallet ibc.Wallet) bool {
+	if relayerSet, ok := r[testName]; ok {
+		return relayerSet[wallet]
+	}
+	return false
+}
```
colin-axner, line 125: should this go in `relayer/relayer.go`? Seems like a convenience struct we can hide in the background

e2e/testsuite/testsuite.go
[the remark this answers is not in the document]
DimitrisJim: oh no I've removed that in the most recent push to directly return here, no reason for a variable all together as you say

colin-axner (approved):
Thanks @DimitrisJim :slightly_smiling_face:

e2e/testsuite/testsuite.go
colin-axner: I can open an issue and we can discuss it on there

e2e/testsuite/testsuite.go
DimitrisJim: Sound good to me, done in fb515f5

DimitrisJim:
@colin-axner re: have we seen it work in the logs. Not so far as far as I can tell, I've searched the logs for the added log `"retrying tx due to ..."` i.e searching the logs for the log stating the retry) and haven't come up with anything. Unfortunate but can't really do much better when dealing with non-determinism.

chatton (approved):
LGTM! Thanks @DimitrisJim

Issue #3373, opened 2023-03-30
Correctly Handle RestoreChannel application callbacks


chatton:
## Summary

We ran into issues with circular dependencies when adding the RestoreChannel function. We are currently
[directly casting](https://github.com/cosmos/ibc-go/blob/3e34171819ae3975f89f02cc3ebb31d6c724e773/modules/core/04-channel/keeper/upgrade.go#L117-L120) to a concrete type which means routing to an application callback for OnRestoreChannel won't work with a custom portKeeper implementation.

See [this comment](https://github.com/cosmos/ibc-go/pull/3352#discussion_r1150197282) for more details and suggested solutions.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chatton:
yep we can 👍

PR #3407, opened 2023-04-04
Correctly handle restore channel application callbacks

chatton:
## Description

closes: #3373

See [this comment](https://github.com/cosmos/ibc-go/pull/3352#discussion_r1150197282) for information about the problem this is solving.

This PR introduces a new interface local to the keeper package. The interface is a subset of the application callbacks (only containing the function requiring for performing a restore operation).

It would be possible to pass the entire interface i.e. `porttypes.IBCModule`, but I think this mechanism has a few benefits.

-  Limits the scope of what can be done to what is required to perform a restore.
-  Limits dependencies on additional `05-port` to make moving the IBCModule to a separate go mod in the future easier.

I think this pattern is a nice way of handling issues of circular dependencies. 

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -11,8 +11,6 @@ import (
 	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
 	connectiontypes "github.com/cosmos/ibc-go/v7/modules/core/03-connection/types"
 	"github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
-	portkeeper "github.com/cosmos/ibc-go/v7/modules/core/05-port/keeper"
-	porttypes "github.com/cosmos/ibc-go/v7/modules/core/05-port/types"
 	host "github.com/cosmos/ibc-go/v7/modules/core/24-host"
 )
 
@@ -95,8 +93,18 @@ func (k Keeper) WriteUpgradeInitChannel(
 	emitChannelUpgradeInitEvent(ctx, portID, channelID, upgradeSequence, channelUpgrade)
 }
 
+// Restorable defines an interface for an ibc module which can be restored.
+// All upgradable modules should implement this interface.
+type Restorable interface {
+	OnChanUpgradeRestore(
+		ctx sdk.Context,
+		portID,
+		channelID string,
+	) error
+}
+
 // RestoreChannel restores the given channel to the state prior to upgrade.
-func (k Keeper) RestoreChannel(ctx sdk.Context, portID, channelID string, upgradeSequence uint64, err error) error {
+func (k Keeper) RestoreChannel(ctx sdk.Context, restorable Restorable, portID, channelID string, upgradeSequence uint64, err error) error {
 	errorReceipt := types.NewErrorReceipt(upgradeSequence, err)
 	k.SetUpgradeErrorReceipt(ctx, portID, channelID, errorReceipt)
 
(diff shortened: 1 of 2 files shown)


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -109,20 +117,5 @@ func (k Keeper) RestoreChannel(ctx sdk.Context, portID, channelID string, upgrad
...
-
-	portKeeper, ok := k.portKeeper.(*portkeeper.Keeper)
-	if !ok {
-		panic("todo: handle this situation")
-	}
-
-	cbs, found := portKeeper.Router.GetRoute(module)
-	if !found {
-		return errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
-	}
```
crodriguezvega, line 125: All this code here to retrieve the application to which we need to callback, would still be needed, right? But one level up? That's it, in the functions where `RestoreChannel` is called, like `ChanUpgradeTimeout`.

crodriguezvega (commented):
Thank you, @chatton. This is definitely a nice construction. I still think that maybe it would be nice to execute the callback at the message server layer (the same place where other callbacks are executed), but I am open to go forward with the solution if the rest are happy with it and makes the code cleaner.

chatton:
@crodriguezvega are you thinking that all calls to restore should be at the message server layer? We would need to then remove all calls to restore in the current try PR. I would actually like to have the restore call happen in a single place, at the moment it happens in several places.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -95,8 +93,18 @@ func (k Keeper) WriteUpgradeInitChannel(
...
 }
 
+// Restorable defines an interface for an ibc module which can be restored.
+// All upgradable modules should implement this interface.
+type Restorable interface {
+	OnChanUpgradeRestore(
+		ctx sdk.Context,
+		portID,
+		channelID string,
+	) error
```
AdityaSripal, line 103: I thought we agreed not to return an error

AdityaSripal (changes requested):

modules/core/04-channel/keeper/upgrade.go
chatton: yes! This PR was opened before that change was made.

crodriguezvega:
> @crodriguezvega are you thinking that all calls to restore should be at the message server layer? We would need to then remove all calls to restore in the current try PR. I would actually like to have the restore call happen in a single place, at the moment it happens in several places.

If we could have the call to restore channel in one place (e.g. the message server layer), I think that would nice. And if we could further take out the callback call from the restore channel function, I think that would also be good. Would it work if for all situations that require restore channel we return the same error (e.g. `ErrUpgradeAborted`) and then we do a check like `errorsmod.IsOf(err, channeltypes.ErrUpgradeAborted)` in the message server handler? If the check passes, then we know we need to restore channel/write error receipt + call callback. Let me know what you think, @chatton!

chatton:
@crodriguezvega  I would love to be able to move the restore to a single place, the only thing that might be a problem here is that different restore calls end up needing a different upgrade sequence. This would mean either returning additional fields for the restore, or returning a function which makes the method signature very messy.

For example [here](https://github.com/cosmos/ibc-go/blob/00a313677d15572d3677802088ee885e4ed410c6/modules/core/04-channel/keeper/upgrade.go#L193-L195) we restore to the upgrade sequence directly, but later on we [increment it](https://github.com/cosmos/ibc-go/blob/00a313677d15572d3677802088ee885e4ed410c6/modules/core/04-channel/keeper/upgrade.go#L213).

It might be tricker than it seems to do a generic restore, but I think that would be the best outcome. I think once we have an initial version of the try handler merged we can look into a refactor. WDYT?

crodriguezvega:
>It might be tricker than it seems to do a generic restore, but I think that would be the best outcome. I think once we have an initial version of the try handler merged we can look into a refactor. WDYT?

Sounds good to me, @chatton. Thank you!

crodriguezvega:
@chatton Will this still be needed?

chatton:
@crodriguezvega  it will not, we can close

PR #3418, opened 2023-04-06
Use ErrorReceipt as Error type

chatton:
## Description

This PR attempts to clean up the chan upgrade try implementation in a few ways.

1. Make the `ErrorReceipt` type itself a valid error type to be able to return that directly instead of writing it to state and special casing the ErrUpgradeAborted in several places.
2. Handle the restore at a single level (msg server layer)
3. Simplify error receipt logic.

---

write a little note why.


proto/ibc/core/channel/v1/upgrade.proto
```diff
@@ -23,5 +23,5 @@ message ErrorReceipt {
   // the channel upgrade sequence
   uint64 sequence = 1;
   // the error message detailing the cause of failure
-  string error = 2;
+  string error_message = 2;
```
chatton, line 26: this needed to be changed as it clashes with the `Error()` function name.

modules/core/keeper/msg_server.go
```diff
@@ -807,7 +810,7 @@ func (k Keeper) ChannelUpgradeTry(goCtx context.Context, msg *channeltypes.MsgCh
 		ctx.Logger().Error("channel upgrade try callback failed", "port-id", msg.PortId, "channel-id", msg.ChannelId, "error", err.Error())
 
 		// we are not returning the error here because we want to commit the error receipt to state
-		if err := k.ChannelKeeper.RestoreChannelAndWriteErrorReceipt(ctx, msg.PortId, msg.ChannelId, upgradeSequence, err); err != nil {
+		if err := k.ChannelKeeper.RestoreChannel(ctx, msg.PortId, msg.ChannelId); err != nil {
```
chatton, line 813: [q] do we want to write an error receipt in the case of an application call back error? This new code as it currently is will just do a restore but not write an error receipt to state.

crodriguezvega: I believe [according to the spec](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#utility-functions) also an error receipt would be written in this scenario. cc: @AdityaSripal

proto/ibc/core/channel/v1/upgrade.proto
crodriguezvega, line 26: Nit: since the type name already indicates this is an error, maybe the field can be just `message`?

```suggestion
  string message = 2;
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -275,16 +267,14 @@ func (k Keeper) WriteUpgradeTryChannel(
...
 
-// TODO: should we pull out the error receipt logic from this function? They seem like two discrete operations.
-
-// RestoreChannelAndWriteErrorReceipt restores the given channel to the state prior to upgrade.
-func (k Keeper) RestoreChannelAndWriteErrorReceipt(ctx sdk.Context, portID, channelID string, upgradeSequence uint64, err error) error {
-	errorReceipt := types.NewErrorReceipt(upgradeSequence, err)
-	k.SetUpgradeErrorReceipt(ctx, portID, channelID, errorReceipt)
-
+// RestoreChannel restores the given channel to the state prior to upgrade.
+func (k Keeper) RestoreChannel(ctx sdk.Context, portID, channelID string) error {
```
crodriguezvega, line 271: If this function is now called only at the message server layer maybe we can either:
- Move the call to `OnChanUpgradeRestore` also to the message server layer;
- or pass `cbs` to this function and call the callback here (and thus not needing to retrieve the callback here), since `cbs` should already be known at the message server layer, right?

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -431,7 +429,14 @@ func (suite *KeeperTestSuite) TestChanUpgradeTry() {
 				proofChannel, proofUpgradeTimeout, proofUpgradeSequence, proofHeight,
 			)
 
-			suite.chainB.GetSimApp().GetIBCKeeper().ChannelKeeper.WriteUpgradeTryChannel(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, upgradeChannel.Version, upgradeSequence, upgradeChannel)
+			// mock what is happening in the msg server handler.
```
crodriguezvega, line 432: Why is this needed?

crodriguezvega (commented):
Thank you, @chatton. I think this is a very nice improvement!

modules/core/04-channel/keeper/upgrade_test.go
chatton: this is needed as the current tests are testing the channel keeper function and not the ibckeeper function. The error receipt is written in the ibc keeper. An alternative is to test the ibckeeper instead, and pass in a MsgChannelUpgradeTry and this would not be needed.

proto/ibc/core/channel/v1/upgrade.proto
chatton: yeah I'm happy enough to make this change

crodriguezvega:
@chatton Can we close this one? Is this now included in the https://github.com/cosmos/ibc-go/pull/3478?

colin-axner:
Can we apply this idea to the newest changes? I really liked the idea of returning the error receipt all the way to the msg server and then aborting the handshake there

Issue #3422, opened 2023-04-06
Code consistency nits in grpc handler


colin-axner:
## Summary

- Use `k` instead of `q, see [comment](https://github.com/cosmos/ibc-go/pull/3019#discussion_r1138913048)
- Unwrap the sdk context first, see [comment](https://github.com/cosmos/ibc-go/pull/2193#discussion_r964639500)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

ruthishvitwit:
I have changed the q's to k but I am not able to undertand on how to unwrap the sdk context first .

PR #3424, opened 2023-04-07
Fix:Changed q's to k and Repositioned UnWrapSDKContext

ruthishvitwit:
## Description

As per the issue Code consistency nits in grpc handler https://github.com/cosmos/ibc-go/issues/3422 ,
I have changed the q's to k and changed the UnWrapSDKContext to first line in all required files.
 

closes: #3422

```bash
fix: Changed q's to k and Repositioned UnWrapSDKContext
```

---

write a little note why.


crodriguezvega (approved):
Thank you, @ruthishvitwit!

modules/core/04-channel/keeper/grpc_query.go
```diff
@@ -398,7 +398,7 @@ func (q Keeper) UnreceivedPackets(c context.Context, req *types.QueryUnreceivedP
 
 	ctx := sdk.UnwrapSDKContext(c)
 
-	channel, found := q.GetChannel(sdk.UnwrapSDKContext(c), req.PortId, req.ChannelId)
+	channel, found := k.GetChannel(sdk.UnwrapSDKContext(c), req.PortId, req.ChannelId)
```
DimitrisJim, line 401: Seems like this existed from before but I guess it wouldn't be a bad idea to fix it now:

```suggestion
	channel, found := k.GetChannel(ctx, req.PortId, req.ChannelId)
```

modules/apps/29-fee/keeper/grpc_query.go
```diff
@@ -17,12 +17,11 @@ var _ types.QueryServer = Keeper{}
 
 // IncentivizedPackets implements the Query/IncentivizedPackets gRPC method
 func (k Keeper) IncentivizedPackets(goCtx context.Context, req *types.QueryIncentivizedPacketsRequest) (*types.QueryIncentivizedPacketsResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx).WithBlockHeight(int64(req.QueryHeight))
```
damiannolan, line 20: I don't think it makes sense to move this context unwrapping to above the `nil` check. This uses `WithBlockHeight()` and passes the `req.QueryHeight` as an arg. 

As you can see the linter is complaining with - SA5011: possible nil pointer dereference (staticcheck)

damiannolan (commented):
Thanks for the PR!

modules/apps/29-fee/keeper/grpc_query.go
colin-axner: Would you recommend we leave the context wrapping always after the req == nil check or separate the two pieces of logic here (wrapping the context, setting the query height)

damiannolan: I think either is fine, but technically the fail fast approach if `req == nil` is probably more efficient. No need to unwrap the context if the request is empty because we'll fail anyways.

damiannolan (approved):
LGTM! Thanks @ruthishvitwit @crodriguezvega 🙏

colin-axner (approved):

modules/core/04-channel/keeper/grpc_query.go
DimitrisJim: bump don't think this was addressed, no? @crodriguezvega

Issue #3445, opened 2023-04-13
Add verifiable upgrade type and validate basic functions


chatton:
## Summary 

Part of #3438

Reference [this commit](https://github.com/cosmos/ibc-go/pull/3438/commits/7c4a32ccd19d98ad0b8ffb9ace470947ab4a9076) and [this one](https://github.com/cosmos/ibc-go/pull/3438/commits/c4857a0810e4304ce8f330e839c7e263934d15d6).
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
I believe this has been completed!

PR #3451, opened 2023-04-13
Add verifiable upgrade type and validate basic functions

chatton:
## Description

closes: #3445

This PR also introduces a helper function to check if a connection is open. Other checks can be refactored in a follow up.

Follow up Items after this PR is merged:

- Add ValidateBasic tests for Upgrade Types
- Determine the correct usage of `latest_sequence_send` field.
- Implement [this suggestion](https://github.com/cosmos/ibc-go/pull/3451#discussion_r1168870480)

N/A part of feature branch

---

write a little note why.


modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -557,3 +560,88 @@ func (suite *KeeperTestSuite) TestUpgradeTimeoutAccessors() {
...
+		path            *ibctesting.Path
+	)
+
+	tests := []struct {
+		name     string
+		malleate func()
+		expPass  bool
+	}{
+		{
+			name: "modify ordering to invalid value",
```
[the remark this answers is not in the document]
chatton: I hadn't really considered the ordering here, but I think it makes sense to move the success cases to the top.

modules/core/04-channel/types/upgrade.go
```diff
@@ -0,0 +1,43 @@
...
+func (u Upgrade) ValidateBasic() error {
+	if err := u.UpgradeFields.ValidateBasic(); err != nil {
+		return errorsmod.Wrap(err, "proposed upgrade fields are invalid")
+	}
+
+	if !u.Timeout.IsValid() {
+		return errorsmod.Wrap(ErrInvalidUpgrade, "upgrade timeout cannot be empty")
+	}
+
+	// TODO: determine if last packet sequence sent can be 0?
```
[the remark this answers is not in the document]
chatton: Yes I think 0 should technically be a valid value (no packets sent yet), maybe @colin-axner / @AdityaSripal could confirm if we should let 0 be an accepted value here?

modules/core/04-channel/types/upgrade.go
```diff
@@ -0,0 +1,43 @@
...
+		return errorsmod.Wrap(ErrInvalidChannelOrdering, muf.Ordering.String())
+	}
+
+	if len(muf.ConnectionHops) != 1 {
+		return errorsmod.Wrap(ErrTooManyConnectionHops, "current IBC version only supports one connection hop")
+	}
+
+	if muf.Version == "" {
+		return errorsmod.Wrap(ErrInvalidUpgrade, "proposed upgrade version cannot be empty")
+	}
```
damiannolan, line 35: we should add `strings.TrimSpace(version)`

modules/core/04-channel/types/upgrade.go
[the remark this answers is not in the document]
damiannolan: [Packet sequences start at 1](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/handshake.go#L92). But iirc from yesterday this value can be determined within the state machine in `WriteChanUpgradeInit()` as opposed to expecting a user provided value

proto/ibc/core/channel/v1/upgrade.proto
```diff
@@ -6,6 +6,25 @@ option go_package = "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types";
...
+// to allow the counterparty to block sends after the upgrade has started.
+message Upgrade {
+  ModifiableUpgradeFields upgrade_fields   = 1 [(gogoproto.nullable) = false];
+  UpgradeTimeout          timeout          = 2 [(gogoproto.nullable) = false];
+  uint64                  last_packet_sent = 3;
+}
+
+// ModifiableUpgradeFields are the fields in a channel end which may be changed
+// during a channel upgrade.
+message ModifiableUpgradeFields {
```
damiannolan, line 23: what do you think about calling this just `UpgradeFields`?

modules/core/04-channel/keeper/keeper.go
```diff
@@ -571,6 +572,30 @@ func (k Keeper) DeleteUpgradeTimeout(ctx sdk.Context, portID, channelID string)
...
+	}
+	if reflect.DeepEqual(proposedUpgrade, currentFields) {
+		return errorsmod.Wrap(types.ErrChannelExists, "existing channel end is identical to proposed upgrade channel end")
+	}
+
+	if !currentFields.Ordering.SubsetOf(proposedUpgrade.Ordering) {
+		return errorsmod.Wrap(types.ErrInvalidChannelOrdering, "channel ordering must be a subset of the new ordering")
+	}
+
+	return k.connectionKeeper.CheckIsOpen(ctx, proposedUpgrade.ConnectionHops[0])
```
damiannolan, line 596: Personally I think I prefer just doing the check explicitly in here rather than adding another expected keeper interface method but happy to keep this as is if there's consensus on adding this instead. 
My vote would be for doing the check directly here.

modules/core/04-channel/types/expected_keepers.go
```diff
@@ -18,6 +18,8 @@ type ClientKeeper interface {
 
 // ConnectionKeeper expected account IBC connection keeper
 type ConnectionKeeper interface {
+	CheckIsOpen(ctx sdk.Context, connectionID string) error
+	HasConnection(ctx sdk.Context, connectionID string) bool
```
damiannolan, line 22: Is `HasConnection` being used?

proto/ibc/core/channel/v1/upgrade.proto
```diff
@@ -6,6 +6,25 @@ option go_package = "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types";
...
 import "ibc/core/client/v1/client.proto";
+import "ibc/core/channel/v1/channel.proto";
+
+// Upgrade is a verifiable type which contains the relevant information
+// for an attempted upgrade. It provides the proposed changes to the channel
+// end, the timeout for this upgrade attempt and the last packet sequence sent
+// to allow the counterparty to block sends after the upgrade has started.
+message Upgrade {
+  ModifiableUpgradeFields upgrade_fields   = 1 [(gogoproto.nullable) = false];
+  UpgradeTimeout          timeout          = 2 [(gogoproto.nullable) = false];
```
damiannolan, line 17: wondering should we do something like:

```protobuf
UpgradeFields upgrade_fields            = 1 [(gogoproto.nullable) = false];
UpgradeTimeout upgrade_timeout          = 2 [(gogoproto.nullable) = false];
```

damiannolan (commented):
Looks great. I left some suggestions, mostly around naming

charleenfei (approved):
so much easier to review 🚀

modules/core/04-channel/types/expected_keepers.go
chatton: it's actually not anymore, this was being used before the CheckIsOpen was introduced (there was a HasConnection check as well as a connection.GetSate() == OPEN) so if we keep CheckIsOpen we can remove HasConnection here

modules/core/04-channel/keeper/keeper.go
chatton: The idea is that this could be used in many other places, it also covers the not found case. Quickly looking it looks like we have around 7 other duplicate checks.

My preference would be to re-use this helper

proto/ibc/core/channel/v1/upgrade.proto
chatton: yeah I think either name could works, the `Modifiable` very explicit, but I don't feel super strongly about it.

modules/core/04-channel/keeper/keeper.go
```diff
@@ -571,6 +572,30 @@ func (k Keeper) DeleteUpgradeTimeout(ctx sdk.Context, portID, channelID string)
...
 
+// ValidateUpgradeFields validates the proposed upgrade fields against the existing channel.
+// It returns an error if the following constraints are not met:
+// - there exists at least one valid proposed change to the existing channel fields
+// - the proposed order is a subset of the existing order
+// - the proposed connection hops do not exist
+// - the proposed version is non-empty (checked in ModifiableUpgradeFields.ValidateBasic())
+// - the proposed connection hops are not open
+func (k Keeper) ValidateUpgradeFields(ctx sdk.Context, proposedUpgrade types.UpgradeFields, existingChannel types.Channel) error {
+	currentFields := types.UpgradeFields{
```
crodriguezvega, line 583: Should we have maybe a helper function that creates an `UpgradeFields` object from a channel? Something like `func From(channel types.Channel): UpgradeFields`

crodriguezvega (commented):
Thank you, @chatton. I have left a bunch of suggestions.

modules/core/04-channel/keeper/keeper.go
DimitrisJim: yeah there was a bit of bike-shedding over the name/scope of this function between us. Initially we had it as `IsOpen` with a single boolean return but that didn't make it as self contained as we'd like. Returning the error allowed it to basically also check for any required error conditions and return them, simplifying the validation, hence the rename to `CheckIsOpen`.

Cian did say this function might not fly with everyone though :smile:

chatton: I'm happy to just remove the function entirely and do the explicit checking where needed! 👍

proto/ibc/core/channel/v1/upgrade.proto
```diff
@@ -6,6 +6,25 @@ option go_package = "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types";
...
+import "ibc/core/channel/v1/channel.proto";
+
+// Upgrade is a verifiable type which contains the relevant information
+// for an attempted upgrade. It provides the proposed changes to the channel
+// end, the timeout for this upgrade attempt and the last packet sequence sent
+// to allow the counterparty to block sends after the upgrade has started.
+message Upgrade {
+  UpgradeFields  upgrade_fields   = 1 [(gogoproto.nullable) = false];
+  UpgradeTimeout timeout          = 2 [(gogoproto.nullable) = false];
+  uint64         last_packet_sent = 3;
```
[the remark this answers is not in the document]
damiannolan: We could probably make this work as `next_sequence_send` also, which may be more conformant to the existing `04-channel` primitives

proto/ibc/core/channel/v1/upgrade.proto
DimitrisJim: Alternatively, to cut down a bit on the repetition of `upgrade`, it could be:

```
UpgradeFields  fields           = 1 [(gogoproto.nullable) = false];
UpgradeTimeout timeout          = 2 [(gogoproto.nullable) = false];
```

chatton: definitely in favour of renaming these to be consistent.

I think I like `fields` and `timeout` since the type makes it very clear what it is already

modules/core/03-connection/keeper/keeper.go
```diff
@@ -83,6 +83,18 @@ func (k Keeper) HasConnection(ctx sdk.Context, connectionID string) bool {
 	return store.Has(host.ConnectionKey(connectionID))
 }
 
+// CheckIsOpen returns nil if the connection with the given id is OPEN. Otherwise an error is returned.
+func (k Keeper) CheckIsOpen(ctx sdk.Context, connectionID string) error {
```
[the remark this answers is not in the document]
chatton: happy to remove this 👍

proto/ibc/core/channel/v1/upgrade.proto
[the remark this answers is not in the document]
chatton: I like `latest_sequence_send`

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -557,3 +560,88 @@ func (suite *KeeperTestSuite) TestUpgradeTimeoutAccessors() {
...
+		},
+		{
+			name: "attempt upgrade when connection is not open",
+			malleate: func() {
+				connection := path.EndpointA.GetConnection()
+				connection.State = connectiontypes.UNINITIALIZED
+				path.EndpointA.SetConnection(connection)
+			},
+			expPass: false,
+		},
```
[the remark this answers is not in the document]
chatton: I think these should be tested in a separate test that tests validate basic. This PR is already quite large, are you happy if we create a follow up PR to add tests for ValidateBasic of `UpgradeFields` and `Upgrade`?

proto/ibc/core/channel/v1/upgrade.proto
```diff
@@ -6,6 +6,25 @@ option go_package = "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types";
...
+import "ibc/core/channel/v1/channel.proto";
+
+// Upgrade is a verifiable type which contains the relevant information
+// for an attempted upgrade. It provides the proposed changes to the channel
+// end, the timeout for this upgrade attempt and the latest packet sequence sent
+// to allow the counterparty to block sends after the upgrade has started.
+message Upgrade {
+  UpgradeFields  fields               = 1 [(gogoproto.nullable) = false];
+  UpgradeTimeout timeout              = 2 [(gogoproto.nullable) = false];
+  uint64         latest_sequence_send = 3;
```
crodriguezvega, line 18: Just trying to understand how the flushing of in-flight packets will work, so excuses if I am missing something. 

If we have 2 chains (A and B) and A initiates the upgrade (i.e. executes `MsgChannelUpgradeInit`) this `latest_sequence_send` in the message can be use for chain A to know that it needs to block further sending of packets. But how can it be used by chain B to block sends? Doesn't chain B keep track of its own `next_sequence_send` that does not necessarily match the `next_sequence_send` on chain A?

e2e/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/cosmos/cosmos-sdk v0.47.0
 	github.com/cosmos/gogoproto v1.4.6
 	github.com/cosmos/ibc-go/v7 v7.0.0
-	github.com/cosmos/interchain-accounts v0.5.0
```
chatton, line 10: there was an error I'm not quite sure I can explain happening suddenly

```
On branch cian/issue#3445-add-verifiable-upgrade-type-and-validate-basic-functions
        downloaded: h1:b9P3j2ZPziqYUYzb7zHVucZg84EfiGY+T+fPgszVxfI=                                                         [135/1987]
        sum.golang.org: h1:0Pzv5xHq+TVY7ExRIsrzJ33+t22TUoJ1k8UgSxvv1X4=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.

github.com/cosmos/ibc-go/e2e/testsuite imports
        github.com/strangelove-ventures/interchaintest/v7/chain/cosmos imports
        github.com/cosmos/cosmos-sdk/x/upgrade imports
        github.com/cosmos/cosmos-sdk/x/upgrade/client/cli imports
        github.com/cosmos/cosmos-sdk/x/upgrade/plan imports
        github.com/hashicorp/go-getter imports
        cloud.google.com/go/storage tested by
        cloud.google.com/go/storage.test imports
        cloud.google.com/go/httpreplay imports
        cloud.google.com/go/httpreplay/internal/proxy imports
        github.com/google/martian/v3/mitm: github.com/cosmos/interchain-accounts@v0.5.0: verifying go.mod: checksum mismatch
        downloaded: h1:b9P3j2ZPziqYUYzb7zHVucZg84EfiGY+T+fPgszVxfI=
        sum.golang.org: h1:0Pzv5xHq+TVY7ExRIsrzJ33+t22TUoJ1k8UgSxvv1X4=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.

github.com/cosmos/ibc-go/e2e/testsuite imports
        github.com/strangelove-ventures/interchaintest/v7/chain/cosmos imports
        github.com/cosmos/cosmos-sdk/x/upgrade imports
        github.com/cosmos/cosmos-sdk/x/upgrade/client/cli imports
        github.com/cosmos/cosmos-sdk/x/upgrade/plan imports
        github.com/hashicorp/go-getter imports
        github.com/aws/aws-sdk-go/service/s3 imports
        github.com/aws/aws-sdk-go/aws/awsutil imports
        github.com/jmespath/go-jmespath tested by
        github.com/jmespath/go-jmespath.test imports
        github.com/jmespath/go-jmespath/internal/testify/assert: github.com/cosmos/interchain-accounts@v0.5.0: verifying go.mod: check
sum mismatch
        downloaded: h1:b9P3j2ZPziqYUYzb7zHVucZg84EfiGY+T+fPgszVxfI=
        sum.golang.org: h1:0Pzv5xHq+TVY7ExRIsrzJ33+t22TUoJ1k8UgSxvv1X4=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.

```

It looks like some sort of mismatch in v0.5.0 in the one we've imported and the one that is imported in interchaintest.

modules/core/03-connection/keeper/keeper.go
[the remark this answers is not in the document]
damiannolan: This isn't being used anymore right? Can we remove it now?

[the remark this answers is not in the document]
chatton: removed all usages but not the function itself 😅

modules/core/04-channel/keeper/keeper.go
```diff
@@ -571,6 +572,49 @@ func (k Keeper) DeleteUpgradeTimeout(ctx sdk.Context, portID, channelID string)
 	store.Delete(host.ChannelUpgradeTimeoutKey(portID, channelID))
 }
 
+// ValidateUpgradeFields validates the proposed upgrade fields against the existing channel.
+// It returns an error if the following constraints are not met:
+// - there exists at least one valid proposed change to the existing channel fields
+// - the proposed order is a subset of the existing order
+// - the proposed connection hops do not exist
+// - the proposed version is non-empty (checked in ModifiableUpgradeFields.ValidateBasic())
```
damiannolan, line 580: I guess this godoc is outdated? ModifiableUpgradeFields is now UpgradeFields?

modules/core/04-channel/types/upgrade.go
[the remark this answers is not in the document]
colin-axner: Maybe we should remove this function? I think we should have `MsgUpgradeInit` and `MsgUpgradeTry` use the `UpgradeFields` and `UpgradeTimeout` directly rather than wrapping with `Upgrade`. I suggest this because it seems unnecessary to expose a field populated by the state machine to the relayer. This would render this `ValidateBasic` unnecessary as the msg should do basic validation on the fields and timeout and then the state machine would construct the upgrade type with the fields, timeout and retrieved packet sequence send. I guess we could keep if for the situation where the entire counterparty upgrade type is provided in the TRY and ACK, in this situation the sequence should not equal 0. 

My vote is to modify the init/try field for self upgrade to use the fields/timeout directly in the msg type and add a if sequence != 0 check to be used to verify the fully formed counterparty upgrade

modules/core/04-channel/types/upgrade.go
```diff
@@ -0,0 +1,45 @@
...
+		return errorsmod.Wrap(ErrInvalidUpgrade, "upgrade timeout cannot be empty")
+	}
+
+	// TODO: determine if last packet sequence sent can be 0?
+	return nil
+}
+
+// ValidateBasic performs a basic validation of the proposed upgrade fields
+func (uf UpgradeFields) ValidateBasic() error {
+	if !collections.Contains(uf.Ordering, []Order{ORDERED, UNORDERED}) {
```
colin-axner, line 27: I wonder if there is a variable we can use for `[]Order{ORDERED, UNORDERED}`? Such as `SupportedOrderings`?

colin-axner (commented):
LGTM! See relevant [comments](https://github.com/cosmos/ibc-go/pull/3451#discussion_r1168870480). Thanks for the quick changes! :)

modules/core/04-channel/keeper/keeper.go
```diff
@@ -571,6 +572,49 @@ func (k Keeper) DeleteUpgradeTimeout(ctx sdk.Context, portID, channelID string)
...
+// - the proposed version is non-empty (checked in UpgradeFields.ValidateBasic())
+// - the proposed connection hops are not open
+func (k Keeper) ValidateUpgradeFields(ctx sdk.Context, proposedUpgrade types.UpgradeFields, existingChannel types.Channel) error {
+	currentFields := extractUpgradeFields(existingChannel)
+
+	if reflect.DeepEqual(proposedUpgrade, currentFields) {
+		return errorsmod.Wrap(types.ErrChannelExists, "existing channel end is identical to proposed upgrade channel end")
+	}
+
+	if !currentFields.Ordering.SubsetOf(proposedUpgrade.Ordering) {
```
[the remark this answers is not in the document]
damiannolan: Yeah I think this reads a little off. The proposed upgrade ordering should be a subset of the current channel ordering. I think this should read as:

```
if !proposedUpgrade.Ordering.SubsetOf(currentFields.Ordering) {}
```

Which means we should modify `SubsetOf` like so:

```diff
func (o Order) SubsetOf(order Order) bool {
-	if supported, ok := orderSubsets[o]; ok {
-		return collections.Contains(order, supported)
+	if supported, ok := orderSubsets[order]; ok {
+		return collections.Contains(o, supported)
	}

	return false
}
```

ref: https://github.com/cosmos/ibc-go/blob/04-channel-upgrades/modules/core/04-channel/types/channel.go#L149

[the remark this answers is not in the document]
colin-axner: Looks good! This can be fixed in a followup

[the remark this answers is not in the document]
damiannolan: I opened https://github.com/cosmos/ibc-go/issues/3469, I'll pick it up as soon as this is merged

damiannolan (approved):
LGTM, lets remove `Upgrade.ValidateBasic()` and update msgs to include `UpgradeFields` and `UpgradeTimeout` in follow ups.

Issue #3452, opened 2023-04-13
Represent unlimited ICS20 authorization amount as MaxInt256


Vvaradinov:
## Summary
ICS20 authz spending limit always subtracts even if the value represents the maximum amount.

## Problem Definition
The new ICS20 authorization included in v7 takes in a spending limit as a list of `sdk.Coin` and type `int64`. At Evmos we use ERC20 token allowance approvals that support an unlimited amount (which are represented as the max `uint256` value). If we pass in the max `uint256` amount effectively making it unlimited to the ICS20 authorization we assume that the amount will not be subtracted on each transfer.

## Proposal

We propose the unlimited value be represented as the `MaxUint256`. This will require an additional  check and skip substracting from the `SpendLimit` if the amount is `MaxUint256`.
```go
var MaxUint256 = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1))
if allocation.SpendLimit.AmountOf(msgTransfer.Token.Denom).Equal(sdk.NewIntFromBigInt(MaxUint256)) {
	return authz.AcceptResponse{Accept: true, Delete: false, Updated: &a}, nil
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3454, opened 2023-04-13
imp: represent unlimited approvals with MaxUint256 value

Vvaradinov:
## Description
This PR implements the changes proposed in https://github.com/cosmos/ibc-go/issues/3452

closes: #3452

```bash
imp(statemachine)!: represent unlimited approvals with a the MaxInt256 value
```

---

write a little note why.


chatton (commented):
Thanks for the PR @Vvaradinov!

My concern about giving a special value to nil here means that it because easier to accidentally configure unlimited spending if the field is omitted.

I think I would lean towards creating a named sentinel value to pass in explicitly as the SpendLimit.

e.g. 

```go
allocation.SpendLimit = UnlimitedSpending()
```

What do you think?

crodriguezvega:
I agree with the risk described above by @chatton.

And besides that, and after discussing also with @womensrights, we think that it is risky to not specify the unlimited spending for a particular denomination. With the approach of using `nil` (or even the sentinel value) the authorisation would grant the grantee the possibility to spend without any limits _any_ denomination. We think that it would be safer to follow an approach closer to what Ethereum does and use the maximum `Int` value for `Amount` in `sdk.Coin` to indicate unlimited spending for a particular denomination. What do you think?

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -4,10 +4,10 @@ import (
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/authz"
-
 	ibcerrors "github.com/cosmos/ibc-go/v7/internal/errors"
 	channeltypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
 	host "github.com/cosmos/ibc-go/v7/modules/core/24-host"
+	"math"
```
fedekunze, line 10: @Vvaradinov could you fix the linter

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -31,39 +31,46 @@ func (a TransferAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authz.Accep
...
+					if limitLeft.IsZero() {
+						a.Allocations = append(a.Allocations[:index], a.Allocations[index+1:]...)
+						if len(a.Allocations) == 0 {
+							return authz.AcceptResponse{Accept: true, Delete: true}, nil
+						}
+						return authz.AcceptResponse{Accept: true, Delete: false, Updated: &TransferAuthorization{
+							Allocations: a.Allocations,
+						}}, nil
+					}
+				}
```
fedekunze, line 61:
```suggestion
				if coin.Amount.Int64() == math.MaxInt64 {
					continue
				}
				
				limitLeft, isNegative = allocation.SpendLimit.SafeSub(msgTransfer.Token)
				if isNegative {
					return authz.AcceptResponse{}, errorsmod.Wrapf(ibcerrors.ErrInsufficientFunds, "requested amount is more than spend limit")
				}

				if !isAllowedAddress(ctx, msgTransfer.Receiver, allocation.AllowList) {
					return authz.AcceptResponse{}, errorsmod.Wrap(ibcerrors.ErrInvalidAddress, "not allowed address for transfer")
				}

				if limitLeft.IsZero() {
					a.Allocations = append(a.Allocations[:index], a.Allocations[index+1:]...)
					if len(a.Allocations) == 0 {
						return authz.AcceptResponse{Accept: true, Delete: true}, nil
					}
					return authz.AcceptResponse{Accept: true, Delete: false, Updated: &TransferAuthorization{
						Allocations: a.Allocations,
					}}, nil
				}
			}
```

CHANGELOG.md
```diff
@@ -43,8 +43,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### State Machine Breaking
 
 ### Improvements
-
 * (tests) [\#3138](https://github.com/cosmos/ibc-go/pull/3138) Support benchmarks and fuzz tests through `testing.TB`.
+* (ics20) [\#3454](https://github.com/cosmos/ibc-go/pull/3454) Represent authz unlimited spending limit as `nil` value.
```
fedekunze, line 47:
```suggestion
* (ics20) [\#3454](https://github.com/cosmos/ibc-go/pull/3454) Represent authz unlimited spending limit as `math.MaxInt64` value.
```

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -31,28 +32,37 @@ func (a TransferAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authz.Accep
...
-		}
-
-		if limitLeft.IsZero() {
-			a.Allocations = append(a.Allocations[:index], a.Allocations[index+1:]...)
-			if len(a.Allocations) == 0 {
-				return authz.AcceptResponse{Accept: true, Delete: true}, nil
+	for index, allocation := range a.Allocations {
+		if allocation.SourceChannel == msgTransfer.SourceChannel && allocation.SourcePort == msgTransfer.SourcePort {
+			for _, coin := range allocation.SpendLimit {
+				if coin.Amount.Int64() == math.MaxInt64 {
```
fedekunze, line 43:
```suggestion
				// skip spending limit if the amount is set to max int64
				if coin.Amount.Int64() == math.MaxInt64 {
```

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -31,28 +32,38 @@ func (a TransferAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authz.Accep
...
-			return authz.AcceptResponse{}, errorsmod.Wrap(ibcerrors.ErrInvalidAddress, "not allowed address for transfer")
-		}
-
-		if limitLeft.IsZero() {
-			a.Allocations = append(a.Allocations[:index], a.Allocations[index+1:]...)
-			if len(a.Allocations) == 0 {
-				return authz.AcceptResponse{Accept: true, Delete: true}, nil
+	for index, allocation := range a.Allocations {
+		if allocation.SourceChannel == msgTransfer.SourceChannel && allocation.SourcePort == msgTransfer.SourcePort {
+			for _, coin := range allocation.SpendLimit {
```
damiannolan, line 42: SpendLimit is defined as `sdk.Coins` so it can contain multiple coins of various denomination. 

If we are iterating on the SpendLimit coins here and doing `SafeSub(msgTransfer.Token)` below then we are deducting more than we should be, right? Essentially over spending.

It seems like this is not a bug at the moment due to the fact that `allocation.SpendLimit` is not being updated / reassigned with the new `limitLeft` until after the loop exits. This behaviour does seem a little off tho.

I'd like to add a test to verify these thoughts. I can check it out later today.

damiannolan: Rough testcase could be something like:

```
{
	"test multiple coins does not overspend",
	func() {
		transferAuthz.Allocations[0].SpendLimit = sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(100)), sdk.NewCoin("test-denom", sdk.NewInt(100)), sdk.NewCoin("test-denom2", sdk.NewInt(100)))
		msgTransfer.Token = sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(50))
	},
	func(res authz.AcceptResponse, err error) {
		suite.Require().NoError(err)

		updatedTransferAuthz, ok := res.Updated.(*types.TransferAuthorization)
		suite.Require().True(ok)

		remainder := updatedTransferAuthz.Allocations[0].SpendLimit.AmountOf(sdk.DefaultBondDenom)
		suite.Require().True(sdk.NewInt(50).Equal(remainder))
	},
},
```

damiannolan:
Looks like some unit test is failing 👀 Will take a closer look later today

modules/apps/transfer/types/transfer_authorization.go
colin-axner: Nice catch @damiannolan! That logic holds up to me:
```go
for index, allocation := range a.Allocations {
	if allocation.SourceChannel == msgTransfer.SourceChannel && allocation.SourcePort == msgTransfer.SourcePort {
		// imagine: sdk.Coins{token1, token2, token3}
		for _, coin := range allocation.SpendLimit { 
		    limitLeft = SafeSub(token1.Amount) // 3 times
		}
	}
}
    return
```

Vvaradinov: Good point @damiannolan would a simple check for the message transfer Token suffice before subtracting? Something like this 
```go
if coin.Denom == msgTransfer.Token.Denom {
     limitLeft, isNegative = allocation.SpendLimit.SafeSub(msgTransfer.Token)
}
```

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -31,29 +32,42 @@ func (a TransferAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authz.Accep
...
 
-		limitLeft, isNegative := allocation.SpendLimit.SafeSub(msgTransfer.Token)
-		if isNegative {
-			return authz.AcceptResponse{}, errorsmod.Wrapf(ibcerrors.ErrInsufficientFunds, "requested amount is more than spend limit")
-		}
+		for _, coin := range allocation.SpendLimit {
+			// skip spending limit if the amount is set to max int64
+			if coin.Amount.Int64() == math.MaxInt64 {
+				continue
+			}
```
damiannolan, line 49: Now that we `continue` here we should move the `isAllowedAddress()` check to before this statement. Perhaps even outside of the scope of this coin iteration

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -190,6 +230,13 @@ func (suite *TypesTestSuite) TestTransferAuthorizationValidateBasic() {
 			},
 			true,
 		},
+		{
+			"success: with max Int allocation spend limit",
+			func() {
+				transferAuthz.Allocations[0].SpendLimit = sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(math.MaxInt64)))
+			},
+			true,
+		},
```
fedekunze, line 239: can you add a failure test case where the amount is greater than `math.Int64`? You can set it using 

```golang
sdk.NewIntFromBigInt(new(big.Int).SetUint64(math.MaxUint64))
```

fedekunze (approved):
LGTM. Minor comment for additional test case

modules/apps/transfer/types/transfer_authorization_test.go
damiannolan: should this fail or should it behave as a normal spend limit and have the transfer token amount deducted? 

At the moment, the code panic's due to calling `Int64()` on the amount [here](https://github.com/cosmos/ibc-go/pull/3454/files#diff-44955ffabce2c0f3471ea97e335ca2a5e05b70a4349dca81e707b097292b7b07R51) when feeding in `sdk.NewIntFromBigInt(new(big.Int).SetUint64(math.MaxUint64))` in a testcase. This is expected according to `sdk.Int` godoc for `Int64()`: `panics if out of range`

Adapting the code to use the below can avoid panics, but then treats values greater than MaxInt64 as valid cases, which imo doesn't make sense as you can end up deducting a transfer amount amount from greater than MaxInt64 and landing on the unlimited spend limit sentinel value accidentally

```diff
+ if coin.Amount.Equal(sdkmath.NewInt(math.MaxInt64)) {
- if coin.Amount.Int64() == math.MaxInt64 {
	continue
}
```

Vvaradinov: @damiannolan it seems like the `SpendLimit` can accept a value of size `uint256` using `sdk.NewIntFromBigInt(MaxUint256)`. Do you see this as problematic? On our end this works well since Solidity developers can directly pass `max(uint256)`. I've updated the PR to reflect this change.

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -1,17 +1,23 @@
...
 	host "github.com/cosmos/ibc-go/v7/modules/core/24-host"
 )
 
 var _ authz.Authorization = (*TransferAuthorization)(nil)
 
+// MaxUint256 is the maximum value for a 256 bit unsigned integer used in EVM compatible chains.
+var (
+	MaxUint256 = new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), 256), big.NewInt(1))
+)
+
```
fedekunze, line 20: remove var, it's not necessary for anything but the check

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -36,35 +42,44 @@ func (a TransferAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authz.Accep
...
 			return authz.AcceptResponse{}, errorsmod.Wrap(ibcerrors.ErrInvalidAddress, "not allowed address for transfer")
 		}
 
-		if limitLeft.IsZero() {
-			a.Allocations = append(a.Allocations[:index], a.Allocations[index+1:]...)
-			if len(a.Allocations) == 0 {
-				return authz.AcceptResponse{Accept: true, Delete: true}, nil
+		for _, coin := range allocation.SpendLimit {
+			// skip spending limit if the amount is set to max int64
+			if coin.Amount.Equal(sdk.NewIntFromBigInt(MaxUint256)) {
```
fedekunze, line 51:
```suggestion
			if coin.Amount.Equal(sdk.NewIntFromBigInt(new(big.Int).SetUint64(math.MaxUint64)))) {
```

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -190,6 +226,13 @@ func (suite *TypesTestSuite) TestTransferAuthorizationValidateBasic() {
 			},
 			true,
 		},
+		{
+			"success: with max Int allocation spend limit",
+			func() {
+				transferAuthz.Allocations[0].SpendLimit = sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(math.MaxInt64)))
```
fedekunze, line 232: update test with maxUint64

CHANGELOG.md
```diff
@@ -43,8 +43,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### State Machine Breaking
 
 ### Improvements
-
 * (tests) [\#3138](https://github.com/cosmos/ibc-go/pull/3138) Support benchmarks and fuzz tests through `testing.TB`.
+* (ics20) [\#3454](https://github.com/cosmos/ibc-go/pull/3454) Represent authz unlimited spending limit as `math.MaxInt64` value.
```
fedekunze, line 47:
```suggestion
* (ics20) [\#3454](https://github.com/cosmos/ibc-go/pull/3454) Represent authz unlimited spending limit as `math.MaxUint64` value.
```

modules/apps/transfer/types/transfer_authorization.go
damiannolan: We could make it unexported, but atm its leveraged in tests. I like the idea of having this as a package level variable so it can be evaluated once on pkg init.

damiannolan (approved):
Sorry for the delay on this, was travelling yesterday. Leaving my approval, would also be good to get your 👍 on the latest changes @fedekunze @Vvaradinov 

Thanks guys! ❤️

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -36,6 +40,10 @@ func (a TransferAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authz.Accep
 			continue
 		}
 
+		if allocation.SpendLimit.AmountOf(msgTransfer.Token.Denom).Equal(sdk.NewIntFromBigInt(MaxUint256)) {
+			return authz.AcceptResponse{Accept: true, Delete: false, Updated: &a}, nil
```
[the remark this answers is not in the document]
Vvaradinov: Actually let's keep it since the tests are checking for the `Updated`.

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -36,15 +40,19 @@ func (a TransferAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authz.Accep
 			continue
 		}
 
+		if !isAllowedAddress(ctx, msgTransfer.Receiver, allocation.AllowList) {
+			return authz.AcceptResponse{}, errorsmod.Wrap(ibcerrors.ErrInvalidAddress, "not allowed address for transfer")
+		}
+
+		if allocation.SpendLimit.AmountOf(msgTransfer.Token.Denom).Equal(sdk.NewIntFromBigInt(MaxUint256)) {
```
fedekunze, line 47: nit, add a comment here and a godoc comment on the `Accept` function.

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -36,15 +40,19 @@ func (a TransferAuthorization) Accept(ctx sdk.Context, msg sdk.Msg) (authz.Accep
...
+			return authz.AcceptResponse{Accept: true, Delete: false, Updated: &a}, nil
+		}
+
 		limitLeft, isNegative := allocation.SpendLimit.SafeSub(msgTransfer.Token)
```
fedekunze, line 51: ditto comment

CHANGELOG.md
fedekunze, line 47:
```suggestion
* (ics20) [\#3454](https://github.com/cosmos/ibc-go/pull/3454) Support transfer authorization unlimited spending limit when the max `uint256` value is provided.
```

fedekunze (approved):
ACK

crodriguezvega:
Thanks for all the work on this one, @Vvaradinov! I opened a [PR](https://github.com/Vvaradinov/ibc-go/pull/1) with an idea suggested by @damiannolan and @colin-axner (making `maxUint256` un-exported and adding an exported function to retrieve it) and updating the documentation.

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -128,3 +139,12 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
 	}
 	return false
 }
+
+// GetUnboundedSpendLimitSentinelValue returns the sentinel value that can be used
+// as the amount for a denomination's spend limit for which spend limit updating
+// should be disabled. Please note that using this sentinel value means that a grantee
+// will be granted the privilege to do ICS20 token transfers for the total amount
+// of the denomination available at the granter's account.
+func GetUnboundedSpendLimitSentinelValue() sdkmath.Int {
```
damiannolan, line 148: nit; I'd favour something a little less long-winded

```suggestion
// UnboundedSpendLimit returns the sentinel value that can be used
// as the amount for a denomination's spend limit for which spend limit updating
// should be disabled. Please note that using this sentinel value means that a grantee
// will be granted the privilege to do ICS20 token transfers for the total amount
// of the denomination available at the granter's account.
func UnboundedSpendLimit() sdkmath.Int {
```

damiannolan (approved):
ACK, thanks @crodriguezvega

docs/apps/transfer/authorizations.md
```diff
@@ -12,7 +12,7 @@ It takes:
 
 - a `SourcePort` and a `SourceChannel` which together comprise the unique transfer channel identifier over which authorized funds can be transferred.
 
-- a `SpendLimit` that specifies the maximum amount of tokens the grantee can spend. The `SpendLimit` is updated as the tokens are spent. This `SpendLimit` may also be updated to increase or decrease the limit as the granter wishes.
+- a `SpendLimit` that specifies the maximum amount of tokens the grantee can transfer. The `SpendLimit` is updated as the tokens are transfered, unless the sentinel value of the maximum value for a 256-bit unsigned integer (i.e. 2^256 - 1) is used for the amount, in which case the `SpendLimit` will not be updated (please be aware that using this sentinel value will grant the grantee the privilege to transfer **all** the tokens of a given denomination available at the granter's account). The helper function `GetUnboundedSpendLimitSentinelValue` in the `types` package of the `transfer` module provides the sentinel value that can be used. This `SpendLimit` may also be updated to increase or decrease the limit as the granter wishes.
```
damiannolan, line 15:
```suggestion
- a `SpendLimit` that specifies the maximum amount of tokens the grantee can transfer. The `SpendLimit` is updated as the tokens are transfered, unless the sentinel value of the maximum value for a 256-bit unsigned integer (i.e. 2^256 - 1) is used for the amount, in which case the `SpendLimit` will not be updated (please be aware that using this sentinel value will grant the grantee the privilege to transfer **all** the tokens of a given denomination available at the granter's account). The helper function `UnboundedSpendLimit` in the `types` package of the `transfer` module provides the sentinel value that can be used. This `SpendLimit` may also be updated to increase or decrease the limit as the granter wishes.
```

crodriguezvega (approved):
ACK. Thanks a lot for all this work, @Vvaradinov!

modules/apps/transfer/types/transfer_authorization.go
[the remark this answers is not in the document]
colin-axner: I think updated being nil is useful for two reasons:
* readability, in this situation no update should occur to the allocations (no undesired sideaffects, removes possibility of buggy code from causing issues)
* less gas costs (returning an updated value causes kv store read/writes)

Is it not possible to update the tests?

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -3,7 +3,6 @@ package types_test
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/authz"
-
```
colin-axner, line 6: nit: import formatting

colin-axner (commented):
Nice work :) just left a small note on the updated return value

modules/apps/transfer/types/transfer_authorization.go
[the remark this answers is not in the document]
damiannolan: I can make quick PR to adapt it. Your reasoning seems good enough for me!

Issue #3447, opened 2023-04-13
Modify UpgradeInit to use new upgrade type


chatton:
## Summary

Part of #3438

See [this commit](https://github.com/cosmos/ibc-go/pull/3438/commits/e9b06ad461457a1f5d1a0289a0bbe56db9ccee99).
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

charleenfei:
closed with https://github.com/cosmos/ibc-go/pull/3456

PR #3456, opened 2023-04-13
Refactor ChanUpgradeInit to use new upgrade type

chatton:
## Description

closes: #3447
closes: #3456

---

write a little note why.


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -36,71 +29,34 @@ func (k Keeper) ChanUpgradeInit(
...
-
-	upgradeTimeout := types.UpgradeTimeout{
-		TimeoutHeight:    counterpartyTimeoutHeight,
-		TimeoutTimestamp: counterpartyTimeoutTimestamp,
-	}
-
-	k.SetUpgradeRestoreChannel(ctx, portID, channelID, restoreChannel)
-	k.SetUpgradeSequence(ctx, portID, channelID, upgradeSequence)
-	k.SetUpgradeTimeout(ctx, portID, channelID, upgradeTimeout)
+	channel.UpgradeSequence++
```
charleenfei, line 36: should we set the upgrade sequence always to start on 0 and increment to 1 on UpgradeInit?

modules/core/04-channel/keeper/events.go
```diff
@@ -273,17 +273,18 @@ func emitChannelClosedEvent(ctx sdk.Context, packet exported.PacketI, channel ty
...
-			sdk.NewAttribute(types.AttributeCounterpartyChannelID, channel.Counterparty.ChannelId),
-			sdk.NewAttribute(types.AttributeKeyConnectionID, channel.ConnectionHops[0]),
-			sdk.NewAttribute(types.AttributeVersion, channel.Version),
-			sdk.NewAttribute(types.AttributeKeyUpgradeSequence, fmt.Sprintf("%d", upgradeSequence)),
+			sdk.NewAttribute(types.AttributeCounterpartyPortID, currentChannel.Counterparty.PortId),
+			sdk.NewAttribute(types.AttributeCounterpartyChannelID, currentChannel.Counterparty.ChannelId),
+			sdk.NewAttribute(types.AttributeKeyUpgradeConnectionHops, upgrade.Fields.ConnectionHops[0]),
+			sdk.NewAttribute(types.AttributeKeyUpgradeVersion, upgrade.Fields.Version),
+			sdk.NewAttribute(types.AttributeKeyUpgradeSequence, fmt.Sprintf("%d", currentChannel.UpgradeSequence)),
+			sdk.NewAttribute(types.AttributeKeyUpgradeOrder, upgrade.Fields.Ordering.String()),
```
chatton, line 287: I've added new keys to be explicit about the upgrade fields.

modules/core/04-channel/types/upgrade.go
```diff
@@ -35,20 +35,6 @@ func NewUpgradeTimeout(height clienttypes.Height, timestamp uint64) UpgradeTimeo
 	}
 }
 
-// ValidateBasic performs a basic validation of the upgrade fields
-func (u Upgrade) ValidateBasic() error {
```
chatton, line 39: follow up to [this discussion](https://github.com/cosmos/ibc-go/pull/3451#discussion_r1168870480)

modules/core/keeper/msg_server.go
```diff
@@ -763,3 +774,23 @@ func (k Keeper) ChannelUpgradeTimeout(goCtx context.Context, msg *channeltypes.M
 func (k Keeper) ChannelUpgradeCancel(goCtx context.Context, msg *channeltypes.MsgChannelUpgradeCancel) (*channeltypes.MsgChannelUpgradeCancelResponse, error) {
 	return nil, nil
 }
+
+// extractUpgradeFromMessage returns the proposed upgrade from the provided MsgChannelUpgradeInit message.
+func (k Keeper) extractUpgradeFromMessage(ctx sdk.Context, msg *channeltypes.MsgChannelUpgradeInit) (channeltypes.Upgrade, error) {
+	seq, found := k.ChannelKeeper.GetNextSequenceSend(ctx, msg.PortId, msg.ChannelId)
+	if !found {
+		return channeltypes.Upgrade{}, channeltypes.ErrSequenceSendNotFound
```
chatton, line 782: I'm unsure if this is the desired behaviour. If we don't want to error here, what value should be set for `LatestSequenceSend`, 0?

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -261,12 +261,11 @@ message MsgChannelUpgradeInit {
...
-  string                    port_id                  = 1;
-  string                    channel_id               = 2;
-  Channel                   proposed_upgrade_channel = 3 [(gogoproto.nullable) = false];
-  ibc.core.client.v1.Height timeout_height           = 4 [(gogoproto.nullable) = false];
-  uint64                    timeout_timestamp        = 5;
-  string                    signer                   = 6;
+  string         port_id         = 1;
+  string         channel_id      = 2;
+  UpgradeTimeout upgrade_timeout = 3 [(gogoproto.nullable) = false];
+  UpgradeFields  upgrade_fields  = 4 [(gogoproto.nullable) = false];
```
chatton, line 267: maybe `timeout` and `fields` are sufficient as names?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -36,101 +27,32 @@ func (k Keeper) ChanUpgradeInit(
 		return 0, "", errorsmod.Wrapf(types.ErrInvalidChannelState, "expected %s, got %s", types.OPEN, channel.State)
 	}
 
-	if !k.scopedKeeper.AuthenticateCapability(ctx, chanCap, host.ChannelCapabilityPath(portID, channelID)) {
-		return 0, "", errorsmod.Wrapf(types.ErrChannelCapabilityNotFound, "caller does not own capability for channel, port ID (%s) channel ID (%s)", portID, channelID)
-	}
```
damiannolan, line 41: I know its kind of pseudo auth, but this was consistent with the opening handshake handlers. Are we good with removing this capability authentication entirely?

proto/ibc/core/channel/v1/tx.proto
damiannolan: mega nit here: can we put `upgrade_fields` before `upgrade_timeout`. e.g. swap 3 and 4

modules/core/keeper/msg_server.go
```diff
@@ -711,30 +711,41 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
 		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
 	}
 
-	upgradeSequence, previousVersion, err := k.ChannelKeeper.ChanUpgradeInit(ctx, msg.PortId, msg.ChannelId, chanCap, msg.ProposedUpgradeChannel, msg.TimeoutHeight, msg.TimeoutTimestamp)
+	proposedUpgrade, err := k.extractUpgradeFromMessage(ctx, msg)
+	if err != nil {
+		return nil, errorsmod.Wrap(err, "failed to get upgrade")
+	}
+
+	upgradeSequence, previousVersion, err := k.ChannelKeeper.ChanUpgradeInit(ctx, msg.PortId, msg.ChannelId, proposedUpgrade)
```
damiannolan, line 719: I feel like this logic may be better encapsulated within 04-channel than at the core msg server. 

E.g. pass fields and timeout and construct the upgrade within there.

We could maybe return the `Upgrade` and reassign the proposed/app negotiated version before passing to `WriteUpgradeInit`

```
channelUpgrade, previousVersion, err := k.Channelkeeper.ChanUpgradeInit(ctx, msg.PortId, msg.ChannelId, msg.UpgradeFields, msg.UpgradeTimeout)
```

modules/core/keeper/msg_server.go
damiannolan: I think if next sequence send is not found then we should error

modules/core/04-channel/types/upgrade.go
damiannolan: >  I guess we could keep if for the situation where the entire counterparty upgrade type is provided in the TRY and ACK, in this situation the sequence should not equal 0.

Was chatting with @colin-axner about this and it may still be useful to keep this around for validating the counterparty upgrade.

proto/ibc/core/channel/v1/upgrade.proto
```diff
@@ -30,9 +30,9 @@ message UpgradeFields {
 // must no longer proceed with the upgrade handshake.
 message UpgradeTimeout {
   // block height after which the upgrade times out
-  ibc.core.client.v1.Height timeout_height = 1 [(gogoproto.nullable) = false];
+  ibc.core.client.v1.Height height = 1 [(gogoproto.nullable) = false];
   // block timestamp (in nanoseconds) after which the upgrade times out
-  uint64 timeout_timestamp = 2;
+  uint64 timestamp = 2;
```
damiannolan, line 35: Nice! 👍

modules/core/04-channel/types/events.go
```diff
@@ -39,7 +39,11 @@ const (
 	AttributeKeyChannelOrdering  = "packet_channel_ordering"
 	AttributeKeyConnection       = "packet_connection"
 
-	AttributeKeyUpgradeSequence = "upgrade_sequence"
+	// upgrade specific keys
+	AttributeKeyUpgradeSequence       = "upgrade_sequence"
+	AttributeKeyUpgradeVersion        = "upgrade_version"
+	AttributeKeyUpgradeConnectionHops = "upgrade_connection_hops"
+	AttributeKeyUpgradeOrder          = "upgrade_order"
```
damiannolan, line 46: nit but aligns more so with `AttributeKeyChannelOrdering` 

```suggestion
	AttributeKeyUpgradeOrder          = "upgrade_ordering"
```

damiannolan (commented):
Really great work pumping this out so quickly with all the various changes ❤️ 

I left some nits and suggestion/open discussion about the `ChanUpgradeInit` method sig and return values.

modules/core/keeper/msg_server.go
chatton: Yeah I think this is probably a little nicer

proto/ibc/core/channel/v1/upgrade.proto
colin-axner: Nice haha, this was bothering me, https://github.com/cosmos/ibc/issues/957

modules/core/keeper/msg_server.go
```diff
@@ -711,31 +711,41 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+	k.ChannelKeeper.WriteUpgradeInitChannel(ctx, msg.PortId, msg.ChannelId, channel, proposedUpgrade)
 
-	ctx.Logger().Info("channel upgrade init callback succeeded", "channel-id", msg.ChannelId, "version", version)
+	ctx.Logger().Info("channel upgrade init callback succeeded", "channel-id", msg.ChannelId, "version", proposedVersion)
 
 	return &channeltypes.MsgChannelUpgradeInitResponse{
 		ChannelId:       msg.ChannelId,
-		Version:         version,
-		UpgradeSequence: upgradeSequence,
+		Version:         proposedVersion,
```
colin-axner, line 747: should we return the proposed upgrade instead?

modules/core/04-channel/types/upgrade.go
```diff
@@ -45,7 +46,12 @@ func (u Upgrade) ValidateBasic() error {
 		return errorsmod.Wrap(ErrInvalidUpgrade, "upgrade timeout cannot be empty")
 	}
 
-	// TODO: determine if last packet sequence sent can be 0?
+	// check LatestSequenceSend to show intent. As this is a uint64 a negative value is not possible.
+	//nolint
+	if u.LatestSequenceSend < 0 {
```
colin-axner, line 51:
```suggestion
	if u.LatestSequenceSend == 0 {
```

modules/core/04-channel/types/upgrade.go
```diff
@@ -45,7 +46,12 @@ func (u Upgrade) ValidateBasic() error {
 		return errorsmod.Wrap(ErrInvalidUpgrade, "upgrade timeout cannot be empty")
 	}
 
-	// TODO: determine if last packet sequence sent can be 0?
+	// check LatestSequenceSend to show intent. As this is a uint64 a negative value is not possible.
+	//nolint
```
colin-axner, line 50: should this be removed?

modules/core/04-channel/types/upgrade.go
```diff
@@ -45,7 +46,12 @@ func (u Upgrade) ValidateBasic() error {
 		return errorsmod.Wrap(ErrInvalidUpgrade, "upgrade timeout cannot be empty")
 	}
 
-	// TODO: determine if last packet sequence sent can be 0?
+	// check LatestSequenceSend to show intent. As this is a uint64 a negative value is not possible.
```
colin-axner, line 49:
```suggestion
	// a fully formed upgrade must have a non zero LatestSequenceSend. As this is a uint64 a negative value is not possible.
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -22,115 +14,64 @@ func (k Keeper) ChanUpgradeInit(
...
 	portID string,
 	channelID string,
-	chanCap *capabilitytypes.Capability,
-	proposedUpgradeChannel types.Channel,
-	counterpartyTimeoutHeight clienttypes.Height,
-	counterpartyTimeoutTimestamp uint64,
-) (upgradeSequence uint64, previousVersion string, err error) {
+	upgradeFields types.UpgradeFields,
+	upgradeTimeout types.UpgradeTimeout,
+) (proposedUpgrade types.Upgrade, err error) {
```
colin-axner, line 19: I wonder if the named return values are still necessary?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -22,115 +14,64 @@ func (k Keeper) ChanUpgradeInit(
...
-	restoreChannel := channel
-	channel.State = types.INITUPGRADE
-	if reflect.DeepEqual(channel, proposedUpgradeChannel) {
-		return 0, "", errorsmod.Wrap(types.ErrChannelExists, "existing channel end is identical to proposed upgrade channel end")
+	if err := k.ValidateUpgradeFields(ctx, upgradeFields, channel); err != nil {
+		return proposedUpgrade, err
 	}
 
-	connectionEnd, err := k.GetConnection(ctx, proposedUpgradeChannel.ConnectionHops[0])
+	proposedUpgrade, err = k.constructProposedUpgrade(ctx, portID, channelID, upgradeFields, upgradeTimeout)
```
colin-axner, line 33: nice naming :)

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -22,115 +14,64 @@ func (k Keeper) ChanUpgradeInit(
...
+	return types.Upgrade{
+		Fields: types.UpgradeFields{
+			Ordering:       fields.Ordering,
+			ConnectionHops: fields.ConnectionHops,
+			Version:        fields.Version,
+		},
+		Timeout: types.UpgradeTimeout{
+			Height:    timeout.Height,
+			Timestamp: timeout.Timestamp,
+		},
```
colin-axner, line 74: What is the reasoning for not doing this? 
```suggestion
		Fields: fields,
		Timeout: timeout,
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -22,115 +14,64 @@ func (k Keeper) ChanUpgradeInit(
...
+		LatestSequenceSend: seq - 1,
```
colin-axner, line 75: should we add a comment for why we do the minus 1?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -22,115 +14,64 @@ func (k Keeper) ChanUpgradeInit(
...
 // WriteUpgradeInitChannel writes a channel which has successfully passed the UpgradeInit handshake step.
 // An event is emitted for the handshake step.
-func (k Keeper) WriteUpgradeInitChannel(
-	ctx sdk.Context,
-	portID,
-	channelID string,
-	upgradeSequence uint64,
-	channelUpgrade types.Channel,
-) {
+func (k Keeper) WriteUpgradeInitChannel(ctx sdk.Context, portID, channelID string, currentChannel types.Channel, upgrade types.Upgrade) {
```
colin-axner, line 46: The only thing we have to be careful with about passing in the channel is to ensure it is obtained and not modified in state before calling this function

modules/core/04-channel/keeper/upgrade.go
colin-axner: context: https://github.com/cosmos/ibc/issues/886

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -94,21 +91,14 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
-		{
-			"invalid proposed channel counterparty",
-			func() {
-				channelUpgrade.Counterparty = types.NewCounterparty(mock.PortID, "channel-100")
+				upgrade.Fields.ConnectionHops = []string{"connection-100"}
 			},
 			false,
 		},
 		{
 			"invalid proposed channel upgrade ordering",
```
crodriguezvega, line 99:
```suggestion
			"stricter proposed channel upgrade ordering",
```

modules/core/04-channel/keeper/events.go
```diff
@@ -273,17 +273,18 @@ func emitChannelClosedEvent(ctx sdk.Context, packet exported.PacketI, channel ty
...
-			sdk.NewAttribute(types.AttributeCounterpartyChannelID, channel.Counterparty.ChannelId),
-			sdk.NewAttribute(types.AttributeKeyConnectionID, channel.ConnectionHops[0]),
-			sdk.NewAttribute(types.AttributeVersion, channel.Version),
-			sdk.NewAttribute(types.AttributeKeyUpgradeSequence, fmt.Sprintf("%d", upgradeSequence)),
+			sdk.NewAttribute(types.AttributeCounterpartyPortID, currentChannel.Counterparty.PortId),
+			sdk.NewAttribute(types.AttributeCounterpartyChannelID, currentChannel.Counterparty.ChannelId),
+			sdk.NewAttribute(types.AttributeKeyUpgradeConnectionHops, upgrade.Fields.ConnectionHops[0]),
+			sdk.NewAttribute(types.AttributeKeyUpgradeVersion, upgrade.Fields.Version),
+			sdk.NewAttribute(types.AttributeKeyUpgradeSequence, fmt.Sprintf("%d", currentChannel.UpgradeSequence)),
+			sdk.NewAttribute(types.AttributeKeyUpgradeOrdering, upgrade.Fields.Ordering.String()),
```
crodriguezvega, line 287: Would be good to add a test for this event, so that if in the future new upgrade fields are added, then the test should fail, so that we don't forget to add the field here. Alternatively, instead of adding each upgrade field with its own key, would it work to have one key for all the upgrade fields and the value is a JSON-encoded string of all of them?

modules/core/04-channel/types/msgs.go
```diff
@@ -508,21 +505,17 @@ func (msg MsgChannelUpgradeInit) ValidateBasic() error {
...
-	}
+
 	_, err := sdk.AccAddressFromBech32(msg.Signer)
 	if err != nil {
 		return errorsmod.Wrapf(ibcerrors.ErrInvalidAddress, "string could not be parsed as address: %v", err)
 	}
 
-	return nil
+	if !msg.Timeout.IsValid() {
+		return errorsmod.Wrap(ErrInvalidUpgrade, "upgrade timeout cannot be empty")
```
crodriguezvega, line 515: An error message like this would make it similar to [the one for `Packet`](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/packet.go#L107).

```suggestion
		return errorsmod.Wrap(ErrInvalidUpgrade, "upgrade timeout height and upgrade timeout timestamp cannot both be 0")
```

crodriguezvega (commented):
Looks very clean!

modules/core/04-channel/keeper/upgrade.go
chatton: I think there is not actually a good reason now that I look at it 😅 , will update. 👍

modules/core/04-channel/keeper/upgrade.go
chatton: I'm not a huge fan of named return arguments anyway (magic var initialisation ) and now with fewer return args I would agree

modules/core/04-channel/keeper/upgrade.go
chatton: We can add a comment if you think it will help, I think the combination of the names `GetNextSequenceSend` and `LatestSequenceSend` make it pretty clear, but I can add a comment to make it extra explicit if you think it makes it more clear.

modules/core/04-channel/keeper/upgrade.go
chatton: we could do an additional read of the channel directly in the function, I'm not sure it's necessary though, WDYT?

e2e/tests/core/04-channel/channel_test.go
```diff
@@ -30,18 +31,23 @@ func (s *ChannelUpgradeTestSuite) TestChannelUpgrade() {
...
+			channelA.PortID, channelA.ChannelID, upgradeFields, upgradeTimeout, rlyWallet.FormattedAddress(),
 		)
 
 		s.Require().NoError(msgChanUpgradeInit.ValidateBasic())
 
 		txResp, err := s.BroadcastMessages(ctx, chainA, rlyWallet, msgChanUpgradeInit)
 		s.Require().NoError(err)
 		s.AssertValidTxResponse(txResp)
+
+		s.Require().NoError(test.WaitForBlocks(ctx, 5, chainA))
```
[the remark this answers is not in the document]
chatton: I think we can, the test still passes without it 👍

modules/core/keeper/msg_server.go
chatton: yeah that probably makes more sense

modules/core/04-channel/types/msgs_test.go
```diff
@@ -494,9 +487,8 @@ func (suite *TypesTestSuite) TestMsgChannelUpgradeInitValidateBasic() {
 		suite.Run(tc.name, func() {
 			msg = types.NewMsgChannelUpgradeInit(
 				ibctesting.MockPort, ibctesting.FirstChannelID,
-				types.Channel{State: types.INITUPGRADE, Version: mock.Version},
-				clienttypes.NewHeight(0, 10000),
-				0,
+				types.NewUpgradeFields(types.UNORDERED, []string{ibctesting.FirstChannelID}, mock.Version),
```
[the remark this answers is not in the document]
chatton: yes it should be, interesting, we are not validating a valid connection id format in validate basic. Is this something we should be doing?

modules/core/04-channel/keeper/events.go
chatton: Are you happy if we do this in a followup, I think a few changes would need to happen to the tests (as we are testing the channel keeper function but events are emitted at the message server layer)

modules/core/04-channel/keeper/upgrade.go
damiannolan: It is probably safer to do an additional read but I'm also happy to leave as is for now. @colin-axner and I discussed yesterday briefly about refactoring grpc handlers into their associated submodules rather than having the entry point at the core layer. This would mean we could reduce the exported APIs and prevent them from being misused. If this were a private method then it would be much safer passing `currentChannel`.

modules/core/04-channel/keeper/upgrade.go
damiannolan: we could also consider adding a new keeper method to sweep it under the rug a little. E.g. `k.GetLastSequenceSend(ctx, portID, channelID)` which would read the same key/value as `GetNextSequenceSend` but handle the off by one.

I think if we do that it can be done later in a different PR.

damiannolan (approved):
Excellent work ❤️ 

LGTM!

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -122,92 +112,30 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
 
-			err = suite.chainA.GetSimApp().IBCKeeper.ChannelKeeper.RestoreChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, upgradeSequence, types.ErrInvalidChannel)
-
-			actualChannel, ok := path.EndpointA.Chain.GetSimApp().IBCKeeper.ChannelKeeper.GetChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
-			errReceipt, errReceiptPresent := suite.chainA.GetSimApp().IBCKeeper.ChannelKeeper.GetUpgradeErrorReceipt(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+			_, err := suite.chainA.GetSimApp().IBCKeeper.ChannelKeeper.ChanUpgradeInit(
+				suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, upgrade.Fields, upgrade.Timeout,
+			)
 
 			if tc.expPass {
```
charleenfei, line 132: we should check that the channel state is in INITUPGRADE now?

modules/core/04-channel/types/msgs_test.go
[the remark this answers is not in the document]
colin-axner: We could be. Only the first connection ID can be required to be in the "connection-X" format (thinking of multihop non len 1 context). I don't have a strong preference since it will fail on the handler trying to get the connection

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -22,115 +14,57 @@ func (k Keeper) ChanUpgradeInit(
...
+		return types.Upgrade{}, errorsmod.Wrapf(types.ErrInvalidChannelState, "expected %s, got %s", types.OPEN, channel.State)
 	}
 
-	// set the restore channel to the current channel and reassign channel state to INITUPGRADE,
-	// if the channel == proposedUpgradeChannel then fail fast as no upgradable fields have been modified.
-	restoreChannel := channel
-	channel.State = types.INITUPGRADE
-	if reflect.DeepEqual(channel, proposedUpgradeChannel) {
-		return 0, "", errorsmod.Wrap(types.ErrChannelExists, "existing channel end is identical to proposed upgrade channel end")
+	if err := k.ValidateUpgradeFields(ctx, upgradeFields, channel); err != nil {
```
colin-axner, line 29: In a separate pr, how would folks feel about making this function private and placed in `upgrade.go`?

colin-axner (approved):
Yay! :tada: Thanks @chatton for pumping this out! :)

modules/core/04-channel/keeper/upgrade.go
chatton: I think I made the exported as it was used in another package, if it's possible though I'm in favour of making it unexported.

chatton: it looks like [here](https://github.com/cosmos/ibc-go/blob/f29e2f8f0487948a36d820dd61f9542d86de91ce/modules/core/04-channel/keeper/keeper_test.go#L639)

Issue #3435, opened 2023-04-11
e2e: Replace `AssertValidTxResponse` and error check with `AssertSuccessTx` or `AssertFailedTx`


colin-axner:
## Summary

Instead of asserting that a response from broadcasting a tx is not nil, we should immediately fail if it is nil and provide helper functions to see if the tx was successful or not. See [comment](https://github.com/cosmos/ibc-go/pull/3402#discussion_r1159885971):

Instead of:
```go
    transferTxResp, err := s.Transfer(args)
    s.Require().NoError(err)
    s.AssertValidTxResponse(transferTxResp)
```
use
```go
    resp := s.Transfer(args)
    s.AssertTxSuccess(resp)
```
or
```go
    resp := s.Transfer(args)
    s.AssertTxFailure(resp, transfertypes.ErrInsufficientBalance) // expected error code 
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chatton:
Updated link to [BroadcastMessages](https://github.com/cosmos/ibc-go/blob/d3038f8317bbf4a9f43c578f7b6d4a6a5c4ebda5/e2e/testsuite/tx.go#L31) and the [error return](https://github.com/cosmos/ibc-go/blob/d3038f8317bbf4a9f43c578f7b6d4a6a5c4ebda5/e2e/testsuite/tx.go#L58)

PR #3475, opened 2023-04-18
refactor calls based on required error success in broadcast message.

DimitrisJim:
## Description

closes: #3435

```bash
imp: refactor calls based on required error success in broadcast message.
```

---

write a little note why.


e2e/testsuite/tx.go
```diff
@@ -93,9 +91,16 @@ func containsMessage(s string, messages []string) bool {
 	return false
 }
 
-// AssertValidTxResponse verifies that an sdk.TxResponse
+// TODO: Maybe Allow specified error code too, see ensure ics20 transfer fails in base test (and there's one more)
+func (s *E2ETestSuite) AssertTxFailure(resp sdk.TxResponse, expectedMsg string) {
```
chatton, line 95: I can see this argument list growing over time as we need to assert on different fields, what do you think of this construct


```go
type TxAssertion func(t *testing.T, resp sdk.TxResponse)


func AssertTxCode(code int) TxAssertion {
   return func(t *testing.T, tx sdk.TxResponse) {
	  require.Equal(t, code, int(tx.Code), "expected tx code to match")
   }
}

func AssertTxMessage(msg string) TxAssertion {
	return func(t *testing.T, tx sdk.TxResponse) {
		require.Contains(t, msg, string(tx.RawLog), "expected tx msg to match")
	}
}

func AssertTxSuccess(resp sdk.TxResponse) TxAssertion{
	return func(t *testing.T, resp sdk.TxResponse) {
		errorMsg := fmt.Sprintf("%+v", resp)
		s.Require().NotEmpty(resp.TxHash, errorMsg)
		s.Require().NotEqual(int64(0), resp.GasUsed, errorMsg)
		s.Require().NotEqual(int64(0), resp.GasWanted, errorMsg)
		s.Require().NotEmpty(resp.Events, errorMsg)
		s.Require().NotEmpty(resp.Data, errorMsg)
	}

}

func (s E2ETestSuite) AssertTx(tx sdk.TxResponse, assertions... TxAssertion) {
   for _, fn := range assertions {
      fn(s.T(), tx)
   }
}


func main(){
	s.AssertTx(tx, AssertTxCode(0), AssertTxMessage("foo"))
}
```

We could provide default assertion functions, to use in most tests, but also allows granular control for any assertion we want to make.

Just a thought!

DimitrisJim: Yeah, I'm really liking this approach. Better to put them in a file of their own.

chatton: cool! I think this sort of thing is actually a good usecase for `.` imports to help in readability so that we don't need to fully qualify the function each time. 

e.g. 

`assertions.AssertTxMessage("foo"), assertions.AssertTxMessage("foo"))`

is harder to read than

`AssertTxMessage("foo"), AssertTxMessage("foo"))`

unless we want to include the package name in the name.

i.e.

`assert.TxMessage("foo"), assert.TxMessage("foo"))`

colin-axner, line 95: Commenting on the original api, we could take in a [sdkerror](https://github.com/cosmos/cosmos-sdk/blob/main/errors/errors.go#L90), then we could check the error msg and code/codespace?

e2e/testsuite/tx.go
```diff
@@ -93,9 +91,16 @@ func containsMessage(s string, messages []string) bool {
...
+	errorMsg := fmt.Sprintf("expected tx to fail with error: %s, got: %s", expectedMsg, resp.RawLog)
+	s.Require().NotEqual(resp.Code, 0)
+	s.Require().Contains(resp.RawLog, expectedMsg, errorMsg)
+}
+
+// AssertTxSuccess verifies that an sdk.TxResponse
 // has non-empty values.
-func (s *E2ETestSuite) AssertValidTxResponse(resp sdk.TxResponse) {
+func (s *E2ETestSuite) AssertTxSuccess(resp sdk.TxResponse) {
```
colin-axner, line 103: This should assert the error code is 0

e2e/tests/transfer/localhost_test.go
```diff
@@ -118,12 +114,11 @@ func (s *LocalhostTransferTestSuite) TestMsgTransfer_Localhost() {
...
 	t.Run("send packet localhost ibc transfer", func(t *testing.T) {
-		txResp, err := s.Transfer(ctx, chainA, userAWallet, transfertypes.PortID, msgChanOpenInitRes.GetChannelId(), testvalues.DefaultTransferAmount(chainADenom), userAWallet.FormattedAddress(), userBWallet.FormattedAddress(), clienttypes.NewHeight(1, 100), 0, "")
-		s.Require().NoError(err)
-		s.AssertValidTxResponse(txResp)
+		txResp := s.Transfer(ctx, chainA, userAWallet, transfertypes.PortID, msgChanOpenInitRes.GetChannelId(), testvalues.DefaultTransferAmount(chainADenom), userAWallet.FormattedAddress(), userBWallet.FormattedAddress(), clienttypes.NewHeight(1, 100), 0, "")
+		s.AssertTxSuccess(txResp)
 
 		events := testsuite.ABCIToSDKEvents(txResp.Events)
-		packet, err = ibctesting.ParsePacketFromEvents(events)
+		packet, err := ibctesting.ParsePacketFromEvents(events)
```
colin-axner, line 121: this should cause tests to fail as the packet needs to be assigned outside of the t.Run?

e2e/tests/transfer/base_test.go
```diff
@@ -418,14 +412,13 @@ func (s *TransferTestSuite) TestMsgTransfer_WithMemo() {
...
-		s.Require().NoError(err)
+		transferTxResp := s.Transfer(ctx, chainA, chainAWallet, channelA.PortID, channelA.ChannelID, testvalues.DefaultTransferAmount(chainADenom), chainAAddress, chainBAddress, s.GetTimeoutHeight(ctx, chainB), 0, "memo")
 
 		if memoFeatureReleases.IsSupported(chainAVersion) {
-			s.AssertValidTxResponse(transferTxResp)
+			s.AssertTxSuccess(transferTxResp)
 		} else {
 			s.Require().Equal(uint32(2), transferTxResp.Code)
-			s.Require().Contains(transferTxResp.RawLog, "errUnknownField")
+			s.Require().Contains("errUnknownField", transferTxResp.RawLog)
```
DimitrisJim, line 421: as far as I could tell this is coming from [`unknownproto`](https://github.com/cosmos/cosmos-sdk/blob/7b10ada768f4524edbbbcb78026903c324bdae5e/codec/unknownproto/unknown_fields.go#L313) in the sdk, couldn't find a registered error for it elsewhere.

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -635,11 +626,8 @@ func (s *UpgradeTestSuite) TestV7ChainUpgradeAddLocalhost() {
 }
 
 // RegisterInterchainAccount will attempt to register an interchain account on the counterparty chain.
-func (s *UpgradeTestSuite) RegisterInterchainAccount(ctx context.Context, chain *cosmos.CosmosChain, user ibc.Wallet, msgRegisterAccount *intertxtypes.MsgRegisterAccount) error {
-	txResp, err := s.BroadcastMessages(ctx, chain, user, msgRegisterAccount)
-	s.Require().NoError(err)
-	s.AssertValidTxResponse(txResp)
-	return err
+func (s *UpgradeTestSuite) RegisterInterchainAccount(ctx context.Context, chain *cosmos.CosmosChain, user ibc.Wallet, msgRegisterAccount *intertxtypes.MsgRegisterAccount) sdk.TxResponse {
```
DimitrisJim, line 629: I added this back in since it seemed like an odd change. The functions that currently wrap message creation currently sometimes create the message themselves (when the type of the message is consistent) or they take it as an arg when it isn't (like `RegisterInterchainAccount`).

Cases like the second might need some refactoring out since they have a very dubious existence.

e2e/tests/transfer/base_test.go
colin-axner: That's correct. I can't remember exactly where the SDK is constructing the returned error. The test is passing a sdk.Msg with an extra field the chain doesn't understand, so when it tries to unmarshal the sdk.Msg, the unknown field error is returned. I think it's fine to leave as is. It would be nice if we knew the root error though. I wonder if it is this [error](https://github.com/cosmos/cosmos-sdk/blob/7b10ada768f4524edbbbcb78026903c324bdae5e/types/errors/errors.go#L12). Ah I think I found [it](https://github.com/cosmos/cosmos-sdk/blob/7b10ada768f4524edbbbcb78026903c324bdae5e/x/auth/tx/decoder.go#L31)!

Is the updated change the right ordering? Looks like the `Contains` [godoc](https://pkg.go.dev/github.com/stretchr/testify/require#Contains) has the elem in question as the second argument?

colin-axner (commented):
Nice work!! Left minor comments, but otherwise everything looks great to me :)

e2e/tests/transfer/authz_test.go
```diff
@@ -307,10 +300,9 @@ func (suite *AuthzTransferTestSuite) TestAuthz_InvalidTransferAuthorizations() {
 				Msgs:    []*codectypes.Any{protoAny},
 			}
 
-			resp, err := suite.BroadcastMessages(context.TODO(), chainA, granteeWallet, msgExec)
+			resp := suite.BroadcastMessages(context.TODO(), chainA, granteeWallet, msgExec)
 			suite.Require().NotEqual(0, resp.Code)
 			suite.Require().Contains(resp.RawLog, sdkerrors.ErrInvalidAddress.Error())
```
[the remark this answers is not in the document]
colin-axner: Kind of. In the last SDK release, they moved their errors to be internal and recommended that each module define it's own errors rather than using SDK base errors, so we copied over only the SDK errors that were applicable to us. Since they are technically "new" errors, we gave them different error codes to have an ordering that was logical (rather than having an error list with gaps in between the codes)

I'm not sure how I feel about having error codes that cannot be referenced externally. I'm not sure I see the benefit of our errors being internal? But, I think we should leave as is and worry about it another day if another issue arises?

colin-axner (approved):
ACK, thanks @DimitrisJim!

e2e/tests/transfer/authz_test.go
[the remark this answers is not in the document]
DimitrisJim: > In the last SDK release, they moved their errors to be internal and recommended that each module define it's own errors rather than using SDK base errors

can't seem to be able to find this in the sdk :disappointed: . I see most of the module errors being defined, typically, in their `types` directory and sdk errors in `types/errors`. If you have time to give more context I'd love that.

> I'm not sure how I feel about having error codes that cannot be referenced externally. I'm not sure I see the benefit of our errors being internal?  But, I think we should leave as is and worry about it another day if another issue arises?

I'd assume these should be part of the API since they are returned by public facing functions, happy to leave as-is if you think that's fine atm.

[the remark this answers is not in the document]
colin-axner: See #2896:
> We did not move the default error types over cause the idea is that users/app devs will create their own with their custom codespace.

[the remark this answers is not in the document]
colin-axner: > I'd assume these should be part of the API since they are returned by public facing functions, happy to leave as-is if you think that's fine atm.

I think I agree, but maybe we could address in a separate issue. Would love to get @damiannolan's and @chatton's opinions here. We could open an issue and discuss at a future point? Just to have an explicit decision?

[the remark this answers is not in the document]
DimitrisJim: Yup, will open an issue on this.

DimitrisJim:
note: would be nice if #3552 is handled first in order to get the tests that require specific errors from there to use them here, see https://github.com/cosmos/ibc-go/pull/3475#discussion_r1182639669

damiannolan (approved):
Thanks @DimitrisJim 🙏

colin-axner (approved):

Issue #3448, opened 2023-04-13
Modify UpgradeTry to use new structure


chatton:
## Summary 

Part of #3438 

Reference [this commit](https://github.com/cosmos/ibc-go/pull/3438/commits/d38cc189e4597ff2954f589e3de4acb4e9f0b3d0)

Note: **these changes should go into [this branch](https://github.com/cosmos/ibc-go/tree/cian/issue%231619-implement-channelupgradetry-functionality-for-04-channel)**.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Closing as the msg field types have been modified, see #1619 to track the implementation of the upgrade try handler

PR #3478, opened 2023-04-18
Refactor ChanUpgradeTry to use new upgrade type 

charleenfei:
## Description

closes: #1619 
closes: #3448 

---

write a little note why.


e2e/tests/core/04-channel/channel_test.go
```diff
@@ -43,8 +54,49 @@ func (s *ChannelUpgradeTestSuite) TestChannelUpgrade() {
...
+		//s.Require().NoError(err)
+		//s.AssertValidTxResponse(txResp)
+		//
+		//s.Require().NoError(testsuite.UnmarshalMsgResponses(txResp, &msgChanUpgradeTryRes))
+		//s.Require().True(msgChanUpgradeTryRes.Success)
+		//
+		//channel, err := s.QueryChannel(ctx, chainB, chainBChannel.PortID, chainBChannel.ChannelID)
+		//s.Require().NoError(err)
+		//s.Require().Equal(channeltypes.TRYUPGRADE, channel.State)
+	})
```
chatton, line 101: we can either just remove this, or leave it commented out, I don't have a strong preference, this is a WIP e2e anyway

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,81 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		msg.ProposedUpgradeFields,
+		msg.ProposedUpgradeTimeout,
+		msg.CounterpartyProposedUpgrade,
+		msg.CounterpartyUpgradeSequence,
+		msg.ProofChannel,
+		msg.ProofUpgrade,
+		msg.ProofHeight,
+	)
+
+	if errorsmod.IsOf(err, channeltypes.ErrInvalidUpgrade) {
```
chatton, line 779: [q] do we want to do the changes that were done in #3418 in this PR or will we do it in a follow up?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +72,157 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+		proposedUpgrade, err := k.constructProposedUpgrade(
+			ctx,
+			portID,
+			channelID,
+			proposedUpgradeFields,
+			proposedUpgradeTimeout,
+		)
+		if err != nil {
+			return 0, errorsmod.Wrap(types.ErrInvalidUpgrade, "failed to construct proposed upgrade")
+		}
```
damiannolan, line 168: we can wrap the returned error here rather than throw away

```suggestion
		if err != nil {
			return 0, errorsmod.Wrap(err, "failed to construct proposed upgrade")
		}
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +72,157 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+		errorReceipt := types.NewErrorReceipt(channel.UpgradeSequence, err)
+		// TODO: emit error receipt events
+		k.SetUpgradeErrorReceipt(ctx, portID, channelID, errorReceipt)
+		return 0, errorsmod.Wrapf(types.ErrInvalidUpgrade, "upgrade timeout has passed, error receipt written for upgrade sequence: %d", channel.UpgradeSequence)
+	}
+
+	// happy path case
+	// increment upgrade sequence appropriately
+	if channel.State == types.OPEN {
+		if counterpartyUpgradeSequence > channel.UpgradeSequence {
```
damiannolan, line 153: I think this comment is a little misleading, a valid crossing hello is still a happy path

```suggestion
	if channel.State == types.OPEN {
	        // increment upgrade sequence appropriately
		if counterpartyUpgradeSequence > channel.UpgradeSequence {
```

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -280,22 +280,25 @@ message MsgChannelUpgradeTry {
...
+  string                    signer                        = 10;
 }
 
 // MsgChannelUpgradeTryResponse defines the MsgChannelUpgradeTry response type
-message MsgChannelUpgradeTryResponse {}
+message MsgChannelUpgradeTryResponse {
+  string channel_id       = 1;
+  string version          = 2;
+  uint64 upgrade_sequence = 3;
+  bool   success          = 4;
```
damiannolan, line 300: Could we reuse [`ResponseResultType`](https://github.com/cosmos/ibc-go/blob/main/proto/ibc/core/channel/v1/tx.proto#L46-L55) here and possibly add a new enum value `FAILED` or something similar?

what do ya'll think?

damiannolan (commented):
Looks great. Left a couple of intial comments, will continue reviewing tomorrow!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +72,157 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+			channelID,
+			proposedUpgradeFields,
+			proposedUpgradeTimeout,
+		)
+		if err != nil {
+			return 0, errorsmod.Wrap(types.ErrInvalidUpgrade, "failed to construct proposed upgrade")
+		}
+
+		k.SetChannel(ctx, portID, channelID, channel)
+		k.SetUpgrade(ctx, portID, channelID, proposedUpgrade)
```
crodriguezvega, line 171: Why do we need to `SetUpgrade` here if we do it in `msg_server` in `WriteUpgradeTryChannel` after executing the application callback?

modules/core/04-channel/types/msgs.go
```diff
@@ -569,29 +567,25 @@ func (msg MsgChannelUpgradeTry) ValidateBasic() error {
...
 		return errorsmod.Wrap(commitmenttypes.ErrInvalidProof, "cannot submit an empty channel proof")
 	}
-	if len(msg.ProofUpgradeTimeout) == 0 {
-		return errorsmod.Wrap(commitmenttypes.ErrInvalidProof, "cannot submit an empty upgrade timeout proof")
+	if len(msg.ProofUpgrade) == 0 {
+		return errorsmod.Wrap(commitmenttypes.ErrInvalidProof, "cannot submit an empty upgrade proof")
 	}
-	if len(msg.ProofUpgradeSequence) == 0 {
-		return errorsmod.Wrap(commitmenttypes.ErrInvalidProof, "cannot submit an empty upgrade sequence proof")
+	if msg.ProofHeight.IsZero() {
```
crodriguezvega, line 587: Should we allow here as well zero proof heights? As this was done for other messages in https://github.com/cosmos/ibc-go/pull/2746.

crodriguezvega (commented):
Looking good, @charleenfei. Thank you!

e2e/tests/core/04-channel/channel_test.go
chatton: let's maybe leave it actually so it is there as a reference (squashing the commit will remove this completely from git history if it is removed here)

modules/core/keeper/msg_server.go
chatton: actually let's wait and do this in a follow up

proto/ibc/core/channel/v1/tx.proto
charleenfei: good idea, will change :)

modules/core/04-channel/keeper/upgrade.go
charleenfei: ah, nice catch. i think the proposed upgrade currently being passed to msg_server is incorrect, it should not be the counterparty upgrade. will update!

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -71,6 +71,8 @@ enum ResponseResultType {
   RESPONSE_RESULT_TYPE_NOOP = 1 [(gogoproto.enumvalue_customname) = "NOOP"];
   // The message was executed successfully
   RESPONSE_RESULT_TYPE_SUCCESS = 2 [(gogoproto.enumvalue_customname) = "SUCCESS"];
+  // The message was executed unsuccessfully
+  RESPONSE_RESULT_TYPE_FAILED = 3 [(gogoproto.enumvalue_customname) = "FAILED"];
```
damiannolan, line 75: I wonder if this should be `FAILURE` instead of `FAILED`... I'm just thinking if it were `FAILED` then I would expect `SUCCESS` to rather be `SUCCEEDED`

what do ya'll think?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +72,158 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+	portID,
+	channelID string,
+	proposedUpgradeFields types.UpgradeFields,
+	proposedUpgradeTimeout types.UpgradeTimeout,
+	counterpartyProposedUpgrade types.Upgrade,
+	counterpartyUpgradeSequence uint64,
+	proofCounterpartyChannel,
+	proofUpgrade []byte,
+	proofHeight clienttypes.Height,
+) (proposedUpgrade types.Upgrade, err error) {
```
damiannolan, line 91: I would be in favour of removing the named return args if possible

damiannolan (commented):
Left some additional comments/suggestions! I think this is almost good to merge, will also review the test code a bit more in depth

Awesome work ❤️

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,84 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+	module, _, err := k.ChannelKeeper.LookupModuleByChannel(ctx, msg.PortId, msg.ChannelId)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrap(err, "could not retrieve module from port-id"))
+		return nil, errorsmod.Wrap(err, "could not retrieve module from port-id")
+	}
+	cbs, ok := k.Router.GetRoute(module)
+	if !ok {
+		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
+		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+	}
```
colin-axner, line 764:
```suggestion
	ctx := sdk.UnwrapSDKContext(goCtx)
	
	module, _, err := k.ChannelKeeper.LookupModuleByChannel(ctx, msg.PortId, msg.ChannelId)
	if err != nil {
		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrap(err, "could not retrieve module from port-id"))
		return nil, errorsmod.Wrap(err, "could not retrieve module from port-id")
	}
	
	cbs, ok := k.Router.GetRoute(module)
	if !ok {
		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
	}
```

modules/core/04-channel/types/msgs_test.go
```diff
@@ -593,15 +572,29 @@ func (suite *TypesTestSuite) TestMsgChannelUpgradeTryValidateBasic() {
 	for _, tc := range testCases {
 		tc := tc
 		suite.Run(tc.name, func() {
+			proposedUpgrade := types.NewUpgrade(
+				types.NewUpgradeFields(types.UNORDERED, []string{ibctesting.FirstChannelID}, mock.Version),
+				types.NewUpgradeTimeout(clienttypes.NewHeight(0, 10000), timeoutTimestamp),
+				1,
+			)
```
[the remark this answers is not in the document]
chatton: yeah I think this would be better, I think this is a left over from when we passed in the upgrade.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +72,158 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+	// their sequence as well.
+	// We must then increment our sequence so both sides start the next upgrade with a fresh sequence.
+	if counterpartyUpgradeSequence < channel.UpgradeSequence {
+		errorReceipt := types.NewErrorReceipt(channel.UpgradeSequence, errorsmod.Wrapf(types.ErrInvalidUpgrade, "counterparty chain upgrade sequence <= upgrade sequence (%d <= %d)", counterpartyUpgradeSequence, channel.UpgradeSequence))
+		channel.UpgradeSequence++
+		// TODO: emit error receipt events
+		k.SetUpgradeErrorReceipt(ctx, portID, channelID, errorReceipt)
+		k.SetChannel(ctx, portID, channelID, channel)
+
+		return types.Upgrade{}, errorsmod.Wrapf(types.ErrInvalidUpgrade, "counterparty chain upgrade sequence <= upgrade sequence (%d <= %d)", counterpartyUpgradeSequence, channel.UpgradeSequence)
```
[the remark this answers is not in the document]
chatton: Yeah in a follow up I'd like us to add the ErrorReceipt type as an `error` implementation do do this abort in one place.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +72,158 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+	}
+
+	counterpartyConnectionHops := []string{connectionEnd.GetCounterparty().GetConnectionID()}
+
+	expectedCounterpartyChannel := types.Channel{
+		State:           types.INITUPGRADE,
+		Counterparty:    types.NewCounterparty(channel.Counterparty.PortId, channel.Counterparty.ChannelId),
+		Ordering:        channel.Ordering,
+		ConnectionHops:  counterpartyConnectionHops,
+		Version:         channel.Version,
```
chatton, line 119: Do we need to provide the counter party version here too? In the case asymmetric versions this value will not be correct.

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,84 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+
+	// set version to return value of callback
+	proposedUpgrade.Fields.Version = proposedUpgradeVersion
+
+	k.ChannelKeeper.WriteUpgradeTryChannel(
+		ctx,
+		msg.PortId,
+		msg.ChannelId,
+		channel,
```
colin-axner, line 822: this is unsafe atm since the channel is obtained before the call to ChanUpgradeTry which will modify the channel upgrade sequence. I think for safety we should have each Write function get a fresh channel even if it is extra reads to disk

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,84 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		msg.ProofChannel,
+		msg.ProofUpgrade,
+		msg.ProofHeight,
+	)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade try failed", "error", errorsmod.Wrap(err, "channel handshake upgrade try failed"))
+		return &channeltypes.MsgChannelUpgradeTryResponse{
+			ChannelId:          msg.ChannelId,
+			Version:            proposedUpgrade.Fields.Version,
+			UpgradeSequence:    channel.UpgradeSequence,
```
colin-axner, line 788: this value could be incorrect if the channel state was updated in `ChanUpgradeTry`. It looks like the response might be discarded in case of msg execution [failure](https://github.com/cosmos/cosmos-sdk/blob/main/baseapp/baseapp.go#L792)? Can we confirm if this response will be used?

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,84 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+			ResponseResultType: channeltypes.FAILED,
+		}, err
+	}
+
+	proposedUpgradeVersion, err := cbs.OnChanUpgradeTry(
+		ctx,
+		msg.CounterpartyProposedUpgrade.Fields.Ordering,
+		msg.CounterpartyProposedUpgrade.Fields.ConnectionHops,
+		msg.PortId,
+		msg.ChannelId,
```
colin-axner, line 798: In a followup pr should we move these arguments to be first after ctx?

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,84 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		channel.UpgradeSequence,
+		channel.Counterparty,
```
colin-axner, line 800: I think this field can be removed in a followup

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,84 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		channel.Version,
+		msg.CounterpartyProposedUpgrade.Fields.Version,
+	)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade try callback failed", "port-id", msg.PortId, "channel-id", msg.ChannelId, "error", err.Error())
+		// TODO: commit error receipt to state and abort channel upgrade
+		return &channeltypes.MsgChannelUpgradeTryResponse{
+			ChannelId:          msg.ChannelId,
+			Version:            proposedUpgradeVersion,
+			UpgradeSequence:    channel.UpgradeSequence,
```
colin-axner, line 810: ditto

modules/core/keeper/msg_server.go
colin-axner: ahh, if we are writing an error receipt we should not be returning an error (state will be reverted)

modules/core/04-channel/keeper/upgrade.go
chatton: this looks to not be the case, but this value should be coming from the application callback from the INIT step.

modules/core/04-channel/keeper/upgrade.go
charleenfei: i guess now it's a bit more explicit than when it was just the seq number

modules/core/04-channel/keeper/upgrade.go
charleenfei: i guess this value should actually be from the counterparty proposed upgrade, since it's the expected counterparty channel?

chatton: I think the method signature needs to be added to include this

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,101 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+	}
+
+	// set version to return value of callback
+	proposedUpgrade.Fields.Version = proposedVersion
+
+	// grab channel her to get latest version after UpgradeTry
+	channel, found := k.ChannelKeeper.GetChannel(ctx, msg.PortId, msg.ChannelId)
+	if !found {
+		return nil, errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "port ID (%s) channel ID (%s)", msg.PortId, msg.ChannelId)
+	}
```
chatton, line 834: Should we maybe just do this get inside `WriteUpgradeTryChannel`? This way we can never accidentally pass in a stale channel to the function.

chatton (commented):
Took a quick pass, will look more in depth later today!

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,101 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
 
 // ChannelUpgradeTry defines a rpc handler method for MsgChannelUpgradeTry.
 func (k Keeper) ChannelUpgradeTry(goCtx context.Context, msg *channeltypes.MsgChannelUpgradeTry) (*channeltypes.MsgChannelUpgradeTryResponse, error) {
```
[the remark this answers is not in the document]
charleenfei: it contains already the entire proposed counterparty upgrade tho -- or did you mean our own proposed version (it contains as well)

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +73,187 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+
+		return types.Upgrade{}, types.Channel{}, errorsmod.Wrapf(types.ErrInvalidUpgrade, "counterparty chain upgrade sequence <= upgrade sequence (%d <= %d)", counterpartyUpgradeSequence, channel.UpgradeSequence)
+	}
+
+	var proposedUpgrade types.Upgrade
+
+	// non-crossing hellos case
+	if channel.State == types.OPEN {
+		if counterpartyUpgradeSequence > channel.UpgradeSequence {
+			channel.UpgradeSequence = counterpartyUpgradeSequence - 1
```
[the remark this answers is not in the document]
charleenfei: it's bc the upgrade sequence of the counterparty should be ours + 1 as they are already in init. Init would increment the upgrade sequence so that we are on the same one (afaiu)

modules/core/04-channel/keeper/upgrade.go
charleenfei: i think it's already in the proposed upgrade as per the new refactor

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +73,187 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+			k.SetChannel(ctx, portID, channelID, channel)
+		}
+		// ChanUpgradeInit will increment the upgrade sequence
```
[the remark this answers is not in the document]
charleenfei: cc @colin-axner

modules/core/keeper/msg_server.go
charleenfei: could make sense...but then WriteUpgradeTry is doing much more than just writing the upgrade try, and we should change the method name

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +73,187 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+		proposedUpgrade, err = k.ChanUpgradeInit(
```
AdityaSripal, line 189: I think we should have helper functions: `initUpgradeHandshake`, `blockUpgradeHandshake`, and `openHandshake` and compose them together in these handshake functions as shown in the spec PR: https://github.com/cosmos/ibc/pull/967/files

modules/core/04-channel/keeper/upgrade.go
AdityaSripal: No it shouldn't be. Per the new refactor, the channel stays with the original parameters until we finally go to OPEN and then we change the parameters.

So in this step, the version in the channel path should be the last successfully negotiated version

modules/core/04-channel/keeper/upgrade.go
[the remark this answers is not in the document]
AdityaSripal: It should also perform the same checks, set the upgrade path and set the final correct sequence.

But you could also just do the necessary checks (some are redundant i believe) and set the sequence correctly in the first place in the lines above (remove the `- 1`, and just set the upgradePath here directly

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +73,187 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+	// We must then increment our sequence so both sides start the next upgrade with a fresh sequence.
+	if counterpartyUpgradeSequence < channel.UpgradeSequence {
+		errorReceipt := types.NewErrorReceipt(channel.UpgradeSequence, errorsmod.Wrapf(types.ErrInvalidUpgrade, "counterparty chain upgrade sequence <= upgrade sequence (%d <= %d)", counterpartyUpgradeSequence, channel.UpgradeSequence))
+		channel.UpgradeSequence++
+		// TODO: emit error receipt events
+		k.SetUpgradeErrorReceipt(ctx, portID, channelID, errorReceipt)
+
+		// if this code path is reached in a crossing hellos scenario, we must reset the channel state to OPEN
+		if channel.State == types.INITUPGRADE {
+			channel.State = types.OPEN
```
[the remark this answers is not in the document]
charleenfei: because it's an error path, see @damiannolan's comment here: https://github.com/cosmos/ibc-go/pull/3478#discussion_r1172625309

modules/core/04-channel/keeper/upgrade.go
charleenfei: maybe we can do that refactor after this PR is in?

modules/core/keeper/msg_server.go
colin-axner: Lets move into `WriteUpgradeTry`. This mimics how we do it in the opening handshake. I'm perfectly happy for `WriteUpgradeTry` to get the channel in order to update it's state

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,101 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		msg.CounterpartyProposedUpgrade.Fields.Ordering,
+		msg.CounterpartyProposedUpgrade.Fields.ConnectionHops,
+		msg.CounterpartyProposedUpgrade.Fields.Version,
+	)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade try callback failed", "port-id", msg.PortId, "channel-id", msg.ChannelId, "error", err.Error())
+		// TODO: commit error receipt to state and abort channel upgrade
+		return &channeltypes.MsgChannelUpgradeTryResponse{
+			ChannelId: msg.ChannelId,
+			Version:   proposedVersion,
```
colin-axner, line 822: I would leave as empty on failed upgrades

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,101 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		msg.ProofUpgrade,
+		msg.ProofHeight,
+	)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade try failed", "error", errorsmod.Wrap(err, "channel handshake upgrade try failed"))
+		return &channeltypes.MsgChannelUpgradeTryResponse{
+			ChannelId: msg.ChannelId,
+			Version:   proposedUpgrade.Fields.Version,
+			Result:    channeltypes.FAILURE,
+		}, err
```
colin-axner, line 786: return nil error

colin-axner (commented):
Brief review, will review the spec first on wednesday then continue here

modules/core/04-channel/keeper/upgrade.go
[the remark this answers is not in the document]
chatton: to me it feels more natural to re-assign `channel.UpgradeSequence = counterpartyUpgradeSequence` rather than doing a -1 and incrementing later

modules/core/04-channel/keeper/upgrade.go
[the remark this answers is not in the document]
chatton: this and the following SetChannel call should end up happening at the message server layer after an error receipt is returned. Maybe we can add a TODO to move them back out once we add the `ErrorReceipt` as an `error` followup?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +73,189 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+
+		k.SetChannel(ctx, portID, channelID, channel)
+
+		return types.Upgrade{}, types.Channel{}, errorsmod.Wrapf(types.ErrInvalidUpgrade, "counterparty chain upgrade sequence <= upgrade sequence (%d <= %d)", counterpartyUpgradeSequence, channel.UpgradeSequence)
+	}
+
+	var proposedUpgrade types.Upgrade
+	proposedUpgradeFields := types.NewUpgradeFields(
+		counterpartyProposedUpgrade.Fields.Ordering,
+		proposedConnectionHops,
```
chatton, line 180: why do we need this to be a separate field?

modules/core/04-channel/keeper/upgrade.go
chatton: I'm a fan of splitting it up into helper functions, but am fine with a follow up

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -68,3 +73,189 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+		if !reflect.DeepEqual(currentUpgrade.Fields, proposedUpgradeFields) {
+			return types.Upgrade{}, types.Channel{}, errorsmod.Wrap(types.ErrInvalidUpgrade, "proposed upgrade fields have changed since UpgradeInit")
+		}
+
+		// if the counterparty sequence is greater than the current sequence, we fast forward to the counterparty sequence
+		// so that both channel ends are using the same sequence for the current upgrade
+		if counterpartyUpgradeSequence > channel.UpgradeSequence {
+			channel.UpgradeSequence = counterpartyUpgradeSequence
+			k.SetChannel(ctx, portID, channelID, channel)
+		}
```
chatton, line 219: maybe a candidate for a small helper function

charleenfei:
closing this for now due to spec refactors/changes, will open again with new version

PR #3515, opened 2023-04-24
docs: Add docs to Acknowledgement.Success

### LINKED ISSUE #3434: Define what Success() in the Acknowledgement interface means
In ibc-go we have the interface

```go
type Acknowledgement interface {
	Success() bool
	Acknowledgement() []byte
}
```

In wasmd [we try to use that](https://github.com/CosmWasm/wasmd/pull/1299) (not sure if that's the right way though).

Now it is not clear to me from the interface definition what `Success()` means. Does it mean a non-error acknowledgement? Does it mean the ack is committed to state and can be either success or error? We can only find [this implementation details](https://github.com/cosmos/ibc-go/blob/v7.0.0/modules/core/keeper/msg_server.go#L453). But we don't want to derive the intended meaning from the implementation.

Ideally the `Acknowledgement` interface does not make a different between error and success and just stores an app specific binary acknowledgement. It is easy to envision non-binary acknowledgements in applications. In our case in wasmd we don't know the meaning if an acknowledgement that we receive from a contract.

webmaster128:
Closes #3434


modules/core/exported/channel.go
```diff
@@ -34,6 +34,10 @@ type PacketI interface {
 // Acknowledgement defines the interface used to return
 // acknowledgements in the OnRecvPacket callback.
 type Acknowledgement interface {
+	// Success tells ibc-go if the given Acknowledgement instance should be written to state or not.
```
alpe, line 37: Technically the ack is always written to state when not nil. The `Success()` is about the callback data. You can see this best in the [code](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/msg_server.go#L455-L472)

webmaster128: What would be a better rephasing of this sentence?

alpe:
```suggestion
	// Success tells ibc-go if the state should be persisted. During the "OnRecvPacket" callback execution, all state is hold in a cache store and committed later. Note: events are always persisted.
```

modules/core/exported/channel.go
```diff
@@ -34,6 +34,10 @@ type PacketI interface {
 // Acknowledgement defines the interface used to return
 // acknowledgements in the OnRecvPacket callback.
 type Acknowledgement interface {
+	// Success tells ibc-go if the state should be persisted. During the "OnRecvPacket" callback execution, all state is hold in a cache store and committed later. Note: events are always persisted.
+	// This is independent of application level success/error which is encoded in the acknowledgement
+	// bytes in a protocol specific way.
+	// See https://github.com/cosmos/ibc-go/blob/main/docs/ibc/apps.md for further explanations.
```
damiannolan, line 40:
```suggestion
	// Success determines if the IBC application state should be persisted when handling `RecvPacket`. During `OnRecvPacket` IBC application callback execution, all state changes are held in a cache store and committed later. Note: IBC application callback events are always persisted.
	// This is independent of application level success/error which is encoded in the acknowledgement
	// bytes in a protocol specific way.
	// See https://github.com/cosmos/ibc-go/blob/main/docs/ibc/apps.md for further explanations.
```

damiannolan (approved):
Thanks @webmaster128 🙏 

Small suggestion to be slightly more explicit :) 

Perhaps we could be even more explicit:
> Note: IBC application callback events are always persisted so long as `RecvPacket` succeeds without error.

modules/core/exported/channel.go
```diff
@@ -34,6 +34,14 @@ type PacketI interface {
...
 // acknowledgements in the OnRecvPacket callback.
 type Acknowledgement interface {
+	// Success determines if the IBC application state should be persisted when handling `RecvPacket`.
+	// During `OnRecvPacket` IBC application callback execution, all state changes are held in a cache store and committed later.
+	//
+	// Note 1: IBC application callback events are always persisted so long as `RecvPacket` succeeds without error.
+	//
+	// Note 2: This method is independent of application level success/error which is encoded in the acknowledgement bytes in a protocol specific way.
+	//
+	// See https://github.com/cosmos/ibc-go/blob/main/docs/ibc/apps.md for further explanations.
```
damiannolan, line 44: side note, we've been trying to avoid links to `main` in code comments but we are in the process of migrating documentation to docusaurus https://github.com/cosmos/ibc-go/pull/3511 so I think we can handle updating this out of band to a versioned doc link :)

damiannolan (commented):
Looks great! Thanks again @webmaster128 :) 

LGTM!

modules/core/exported/channel.go
webmaster128: Makes sense. But what exactly do you prefer to do here and now? Link to a tag or commit? Leave it as is?

damiannolan: Happy to leave as is, or you can drop a tag in if you want. Maybe https://github.com/cosmos/ibc-go/blob/v7.0.0/docs/ibc/apps.md

modules/core/exported/channel.go
```diff
@@ -34,6 +34,14 @@ type PacketI interface {
 // Acknowledgement defines the interface used to return
 // acknowledgements in the OnRecvPacket callback.
 type Acknowledgement interface {
+	// Success determines if the IBC application state should be persisted when handling `RecvPacket`.
+	// During `OnRecvPacket` IBC application callback execution, all state changes are held in a cache store and committed later.
+	//
```
colin-axner, line 39:
```suggestion
	// During `OnRecvPacket` IBC application callback execution, all state changes are held in a cache store and committed if:
	// - the acknowledgement.Success() returns true
	// - a nil acknowledgement is returned (asynchronous acknowledgements)
	// 
```

modules/core/exported/channel.go
```diff
@@ -34,6 +34,14 @@ type PacketI interface {
...
+	// Note 1: IBC application callback events are always persisted so long as `RecvPacket` succeeds without error.
+	//
+	// Note 2: This method is independent of application level success/error which is encoded in the acknowledgement bytes in a protocol specific way.
```
colin-axner, line 42: I'm not sure what this note means? Could you elaborate?

modules/core/exported/channel.go
webmaster128: Done. Version pinning is always a good idea.

colin-axner:
Thank you @webmaster128! This is a great addition, we definitely should have added this earlier!

modules/core/exported/channel.go
webmaster128: You have a public type called "Acknowledgement" which is not an IBC acknowledgement. And `acknowledgement.Success` has nothing to do with success acknowledgements. This is a source of confusion for users that I tried to resolve.

colin-axner: Ah, by "This method is independent of application level success/error" this is referencing [result](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/acknowledgement.go#L22) and error acknowledgements, not the success/error of a transaction at the application layer?

> And acknowledgement.Success has nothing to do with success acknowledgements

Hmm, I'm not sure I agree with this point. If by "success acknowledgement" you mean "result acknowledgement" then the only reason it is referred to "success" is because it returns true when fulfilling the `acknowledgement.Success()` interface function. The response/error acknowledgements are not required, they are sane default acknowledgements IBC apps may use if they wish not to define their own custom types

After reading this again, are you referring to [this](https://github.com/cosmos/ibc-go/blob/main/modules/apps/29-fee/types/ack.go#L12)?

webmaster128: > by "This method is independent of application level success/error" this is referencing [result](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/acknowledgement.go#L22) and error acknowledgements, not the success/error of a transaction at the application layer?

Right. I used "success" as the opposite to an error acknowledgement.

I think my primary source of confusion is that the "Acknowledgement" interface **is not** the acknowledgement but **contains** the acknowledgement.

modules/core/exported/channel.go
```diff
@@ -34,6 +34,16 @@ type PacketI interface {
 // Acknowledgement defines the interface used to return
 // acknowledgements in the OnRecvPacket callback.
 type Acknowledgement interface {
+	// Success determines if the IBC application state should be persisted when handling `RecvPacket`.
+	// During `OnRecvPacket` IBC application callback execution, all state changes are held in a cache store and committed if:
+	// - the acknowledgement.Success() returns true
+	// - a nil acknowledgement is returned (asynchronous acknowledgements)
+	//
```
DimitrisJim, line 41: wildest suggestion ever but this is what's causing the linter to fail (extra space)

```suggestion
	//
```

modules/core/exported/channel.go
```diff
@@ -34,6 +34,16 @@ type PacketI interface {
 // Acknowledgement defines the interface used to return
 // acknowledgements in the OnRecvPacket callback.
```
colin-axner, line 35:
```suggestion
// Acknowledgement defines the interface used to return acknowledgements in the OnRecvPacket callback.
// The Acknowledgement interface is used by core IBC ensure partial state changes are not committed 
// when packet receives have not properly succeeded (typically resulting in an error acknoweldgement being returned).
// The interface also allows core IBC to obtain the acknowledgement bytes whose encoding is determined by each IBC application or middleware. 
// Each custom acknowledgement type must implement this interface.
```

modules/core/exported/channel.go
colin-axner: Okay I see now. Thanks for the followup comments. I see how this can be confusing, especially given that an acknowledgement can wrap other acknowledgements. The idea of the interface is to convey information about the acknowledgements so that core IBC can do some useful logic without having reference to the actual type. That's why we need the `Success()` function and the `Acknowledgement()` function. So that core IBC can do some useful handling of state during receive packets and also set the marshalled bytes

Maybe we could make some more adjustments to clear up the confusion (I also see we need to add docs about wrapping acknowledgements). I will add some suggestions

modules/core/exported/channel.go
```diff
@@ -34,6 +34,16 @@ type PacketI interface {
...
 type Acknowledgement interface {
+	// Success determines if the IBC application state should be persisted when handling `RecvPacket`.
+	// During `OnRecvPacket` IBC application callback execution, all state changes are held in a cache store and committed if:
+	// - the acknowledgement.Success() returns true
+	// - a nil acknowledgement is returned (asynchronous acknowledgements)
+	//
+	// Note 1: IBC application callback events are always persisted so long as `RecvPacket` succeeds without error.
+	//
+	// Note 2: This method is independent of application level success/error which is encoded in the acknowledgement bytes in a protocol specific way.
```
colin-axner, line 44:
```suggestion
	// Note 2: The return value should account for the success of underlying IBC application/middleware. Thus the Success is representative of the entire IBC stack's success when receiving a packet. The individual success of each acknowledgement associated with an IBC application or middleware must be determined by obtaining the actual acknowledgement type after decoding the acknowledgement bytes
```

colin-axner (commented):
Thanks again for you patience @webmaster128. I added some additional suggestions that should hopefully clear up any confusion for readers. Let me know what you think!

colin-axner (approved):
Thanks @webmaster128! Sorry for the back and forth

webmaster128:
Merci

Issue #3504, opened 2023-04-22
Migrate `27-interchain-accounts` host params to be self managed


crodriguezvega:
Part of https://github.com/cosmos/ibc-go/issues/2010. Please read that issue for more information.

aleem1314:
@crodriguezvega can I take this issue?

crodriguezvega:
> @crodriguezvega can I take this issue?

Yes, @aleem1314. Thank you very much! Please check #2010 and [this comment](https://github.com/cosmos/ibc-go/issues/2010#issuecomment-1399618502) for guidance on how to go about it. :)

Issue #3222, opened 2023-03-01
Consider simplifying migration code by removing migrator types


chatton:
## Summary

Refactor migrators to use simple functions instead.

## Problem Definition

We currently use the `Migrator` construct (not enforced in anyway, just a pattern) in order to register migration functions. I think these are unnecessary and the migration registration code can be made more concise without them.

## Proposal

The current registration process involves a few steps.

1. Create a new package / location for migrations for a specific version.
2. Create a [migrator type](https://github.com/cosmos/ibc-go/blob/e093d85b533ab3572b32a7de60b88a0816bed4af/modules/apps/transfer/keeper/migrations.go#L12-L19).
3. Create an instance with a keeper and [register the function](https://github.com/cosmos/ibc-go/blob/682ba5501d2230c8e4cc369612f0ccc926440d26/modules/apps/transfer/module.go#L106-L109)

I propose that we simply keep migrations as standalone functions.

For example, we currently have 

```go
// RegisterServices registers module services.
func (am AppModule) RegisterServices(cfg module.Configurator) {
	// register types

	m := keeper.NewMigrator(am.keeper)
	if err := cfg.RegisterMigration(types.ModuleName, 1, m.MigrateTraces); err != nil {
		panic(fmt.Sprintf("failed to migrate transfer app from version 1 to 2: %v", err))
	}
}
```

which could be replaced with

```go


func MigrateTraces(keeper keeper.Keeper) func(sdk.Context) error {
	return func(ctx sdk.Context) error {
            // migration code
	}
}

// RegisterServices registers module services.
func (am AppModule) RegisterServices(cfg module.Configurator) {
        // register types
	if err := cfg.RegisterMigration(types.ModuleName, 1, MigrateTraces(am.keeper)); err != nil {
		panic(fmt.Sprintf("failed to migrate transfer app from version 1 to 2: %v", err))
	}
}
```

This reduces the overhead of creating new types for every type of migration we want to do.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
If all our instances of migrators only take in the keeper, I wonder if we can simply just use the keeper:
```go

// RegisterServices registers module services.
func (am AppModule) RegisterServices(cfg module.Configurator) {
        // register types
	if err := cfg.RegisterMigration(types.ModuleName, 1, am.keeper.MigrateTraces); err != nil {
[...]

PR #3520, opened 2023-04-25
feat(ica/host)!: migrate ICA host params to be self managed

aleem1314:
## Description

With this PR:

- [x] The `ica/host` submodule now self-manages its parameters.
- [x] Unit and integration tests have been added for the `ica/host` submodule's parameters.
- [x] Migration code has been implemented for moving parameters from `x/params`' subspace to the `ica/host` submodule's keeper.
- [x] Migration unit and integration tests have been added.

closes: #3504 

addresses: #2010 

```
feat(ica/host)!: migrate ica/host to params to be self managed
```

---

write a little note why.


modules/apps/27-interchain-accounts/host/migrations/v8/migrations.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+const (
+	SubModuleName = "icahost"
+)
+
+var ParamsKey = []byte{0x00}
+
+// Migrate migrates the 27-interchain-accounts host submodule state from the consensus version 1 to
+// version 2. Specifically, it takes the parameters that are currently stored
```
aleem1314, line 20:
```suggestion
// Migrate migrates the 27-interchain-accounts host submodule state from the consensus version 2 to
// version 3. Specifically, it takes the parameters that are currently stored
```

modules/apps/27-interchain-accounts/host/keeper/params.go
srdtrk, line 1: Imo, it makes more sense for these functions to be in `keeper.go` instead.

Remove the `IsHostEnabled` function all together, imo, it is wrong to return false if there is a problem unmarshalling. This case should be handled in `GetParams`, it seems like you're re-implementing `GetParams`, unnecessary code duplication

proto/ibc/applications/interchain_accounts/host/v1/tx.proto
srdtrk, line 1: You should register the new message in `types/codec.go` (which you need to create)

testing/simapp/app.go
srdtrk, line 1: You should also initialise the key table for migration in line 915. See example [here](https://github.com/cosmos/ibc-go/pull/3553/files#diff-d18972debee5e64f16e40807b2ae112ddbe609504a93ea5e1c80a5d489c3a08aR913)

srdtrk (changes requested):

modules/apps/27-interchain-accounts/host/types/keys.go
```diff
@@ -4,6 +4,9 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+// ParamsKey is the key to use for the storing params.
+var ParamsKey = []byte{0x00}
```
colin-axner, line 8: I wonder if we should just use a string value here? Not a big deal either way:

pros:
- byte savings (5-10 bytes), but key is only written once

cons:
- harder to maintain, not possible to assert another key doesn't use the same value
- not a const

Our codebase is a bit split, but the majority of the time we using strings

modules/apps/27-interchain-accounts/module.go
```diff
@@ -97,25 +98,31 @@ type AppModule struct {
...
 	controllerKeeper *controllerkeeper.Keeper
 	hostKeeper       *hostkeeper.Keeper
+
+	// legacyHostSubspace is used solely for migration of x/params managed parameters
+	legacyHostSubspace exported.Subspace
 }
 
 // NewAppModule creates a new IBC interchain accounts module
-func NewAppModule(controllerKeeper *controllerkeeper.Keeper, hostKeeper *hostkeeper.Keeper) AppModule {
+func NewAppModule(controllerKeeper *controllerkeeper.Keeper, hostKeeper *hostkeeper.Keeper, hss exported.Subspace) AppModule {
```
colin-axner, line 107: This seems like an unnecessary field to add to the API since it will be removed. On one hand, it's nice to ensure the app.go file is providing the subspace, but it'd be nice if there's a way we could get the migrations to work by default without the app.go file changing?

colin-axner: Maybe we should leave the subspace in the host keeper, add a `GetSubspace` function (since it is private) and remove the subspace from the keeper api after we do the migration. I'd prefer to limit app.go changes (otherwise upgrading ibc-go becomes a little more annoying) especially if those changes would need to be done again in the next release. Seems better to delay and have the removal of the subspace happen in the release after?

colin-axner (commented):
Thank you @aleem1314! Really cool that you picked up this work for us :pray: I left various comments and only had time to review half the pr, I will hopefully pick up my review soon

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### State Machine Breaking
 
+* (modules/apps/27-interchain-accounts/host) [\#3520](https://github.com/cosmos/ibc-go/pull/3520) Migrate `modules/apps/27-interchain-accounts/host` to self-managed parameters and deprecate it's usage of `x/params`.
```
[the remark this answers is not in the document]
srdtrk: implemented

modules/apps/27-interchain-accounts/host/keeper/keeper.go
```diff
@@ -199,3 +196,33 @@ func (k Keeper) SetInterchainAccountAddress(ctx sdk.Context, connectionID, portI
...
+// GetParams returns the total set of the host submodule parameters.
+func (k Keeper) GetParams(ctx sdk.Context) types.Params {
+	var params types.Params
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.ParamsKey)
+	if bz == nil {
+		return types.Params{}
+	}
+
+	k.cdc.MustUnmarshal(bz, &params)
```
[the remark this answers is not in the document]
srdtrk: implemented

srdtrk:
@colin-axner, this PR is not ready for merging yet (see my past comments). I can notify you once it is ready over slack, so you can review it then?

srdtrk (approved):
I'm approving this PR after the changes I've added :) Thanks for contributing @aleem1314. This PR is now ready for review @colin-axner.

modules/apps/27-interchain-accounts/host/types/params_legacy.go
```diff
@@ -0,0 +1,58 @@
+/*
+NOTE: Usage of x/params to manage parameters is deprecated in favor of x/gov
+controlled execution of MsgUpdateParams messages. These types remains solely
+for migration purposes and will be removed in a future release.
```
colin-axner, line 4: Maybe we should open a tracking issue to decide in what version this code can be removed

modules/apps/27-interchain-accounts/host/types/params.go
```diff
@@ -46,29 +32,7 @@ func (p Params) Validate() error {
...
-	return nil
-}
-
-func validateAllowlist(i interface{}) error {
-	allowMsgs, ok := i.([]string)
-	if !ok {
-		return fmt.Errorf("invalid parameter type: %T", i)
-	}
-
+func validateAllowlist(allowMsgs []string) error {
```
colin-axner, line 35: the code:
```go
    if err := validateEnabledType(p.HostEnabled); err != nil {
	return err
    }
```

can be removed in my opinion. I believe this code only executed as extra defense to how the x/params module worked. Given self managed params, there should be no instance when the params type has a non boolean value for the enabled type (I'm not sure how this could occur)

This would also allow `params_legacy.go` to be deleted without additional changes as the code is written currently

modules/apps/27-interchain-accounts/host/types/params_legacy.go
```diff
@@ -0,0 +1,58 @@
...
+func validateEnabledType(i interface{}) error {
+	_, ok := i.(bool)
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+
+	return nil
+}
+
+func validateAllowlistLegacy(i interface{}) error {
```
colin-axner, line 45: Are `validateEnabledType` and `validateAllowlistLegacy` still necessary? My understanding is that only the interface functions need to be kept around for migrations?

modules/apps/27-interchain-accounts/host/types/codec.go
```diff
@@ -0,0 +1,14 @@
+package types
+
+import (
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// RegisterInterfaces registers the interchain accounts controller message types using the provided InterfaceRegistry
```
colin-axner, line 8:
```suggestion
// RegisterInterfaces registers the interchain accounts host message types using the provided InterfaceRegistry
```

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
```diff
@@ -0,0 +1,36 @@
+package keeper
+
+import (
+	"context"
+
+	"cosmossdk.io/errors"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/ibc-go/v7/modules/apps/27-interchain-accounts/host/types"
```
colin-axner, line 8:
```suggestion
	sdk "github.com/cosmos/cosmos-sdk/types"
	
	"github.com/cosmos/ibc-go/v7/modules/apps/27-interchain-accounts/host/types"
```

modules/apps/27-interchain-accounts/host/keeper/migrator.go
colin-axner, line 1: ibc-go convention has this as `migrations.go`. What do you think about renaming it? We haven't seen much benefit in the migrator type so far, see #3222

colin-axner (commented):
Thanks @srdtrk for the updates! Continuing my review, reviewed most files except keeper.go, migrations.go, and module.go. Will review once there is a resolution on limiting API breakage

modules/apps/27-interchain-accounts/host/types/params_legacy.go
srdtrk: We could add it as a list item issue to the epic and link it here. Let me know if you'd like that?

modules/apps/27-interchain-accounts/host/types/params.go
srdtrk: Thanks, I was confused why this was there, I'll push this to all other param migration PRs as well

modules/apps/27-interchain-accounts/host/types/params_legacy.go
srdtrk: You are right, they are not needed.

srdtrk: they are actually needed. See lines 31 and 32. `x/params` module expects a validator function for each registered param.

srdtrk: However, I don't know if we **actually** need them since params "should" already be registered if we're doing a migration. But maybe we need to do this each time the chain is restarted. Keeping them certainly makes testing much easier. I think we should keep them since this file is ought to be removed anyway.

modules/apps/27-interchain-accounts/host/keeper/migrator.go
srdtrk: I believe this is the convention in the sdk too, I've renamed it.

modules/apps/27-interchain-accounts/host/types/codec.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
srdtrk: done

modules/apps/27-interchain-accounts/module.go
srdtrk: I've done this change

modules/apps/27-interchain-accounts/host/keeper/keeper.go
```diff
@@ -199,3 +205,33 @@ func (k Keeper) SetInterchainAccountAddress(ctx sdk.Context, connectionID, portI
...
+	var params types.Params
+	k.cdc.MustUnmarshal(bz, &params)
+	return params
+}
+
+// SetParams sets the total set of the host submodule parameters.
+func (k Keeper) SetParams(ctx sdk.Context, params types.Params) error {
+	if err := params.Validate(); err != nil {
+		return err
+	}
```
colin-axner, line 231: @damiannolan do you have any thoughts on this? This pattern goes against our normal behaviour (normally all validation is expected to happen before calling `Set`). I'm fine with it but it will likely lead to redundant checks

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
```diff
@@ -0,0 +1,37 @@
...
+
+type msgServer struct {
+	*Keeper
+}
+
+// NewMsgServerImpl returns an implementation of the ICS27 host MsgServer interface
+// for the provided Keeper.
+func NewMsgServerImpl(keeper *Keeper) types.MsgServer {
+	return &msgServer{Keeper: keeper}
+}
```
colin-axner, line 23: I believe this construction was only used for the controller submodule to avoid name collision of the `SendTx` and `RegisterInterchainAccount` functions (one function signature is legacy/deprecated now)

Maybe we should just use the keeper directly?

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
```diff
@@ -0,0 +1,37 @@
+package keeper
+
+import (
+	"context"
+
+	"cosmossdk.io/errors"
```
colin-axner, line 6: errorsmod is the import alias we use

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
```diff
@@ -0,0 +1,37 @@
...
+// NewMsgServerImpl returns an implementation of the ICS27 host MsgServer interface
+// for the provided Keeper.
+func NewMsgServerImpl(keeper *Keeper) types.MsgServer {
+	return &msgServer{Keeper: keeper}
+}
+
+// UpdateParams updates the host submodule's params.
+func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
+	if m.authority != msg.Authority {
+		return nil, errors.Wrapf(ibcerrors.ErrInvalidAddress, "invalid authority; expected %s, got %s", m.authority, msg.Authority)
```
colin-axner, line 28: I wonder if we should have a `ibcerrors.ErrInvalidAuthority`?

modules/apps/27-interchain-accounts/host/types/params.go
```diff
@@ -1,29 +1,15 @@
 package types
 
 import (
-	"fmt"
+	fmt "fmt"
```
colin-axner, line 4:
```suggestion
	"fmt"
```

modules/apps/27-interchain-accounts/module.go
colin-axner, line 1: missing msg server registration:

```diff
diff --git a/modules/apps/27-interchain-accounts/module.go b/modules/apps/27-interchain-accounts/module.go
index 4a421acd..35682483 100644
--- a/modules/apps/27-interchain-accounts/module.go
+++ b/modules/apps/27-interchain-accounts/module.go
@@ -139,6 +139,7 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
        }
 
        if am.hostKeeper != nil {
+               hosttypes.RegisterMsgServer(cfg.MsgServer(), hostkeeper.NewMsgServerImpl(am.hostKeeper))
                hosttypes.RegisterQueryServer(cfg.QueryServer(), am.hostKeeper)
        }
 
```

we should add an e2e for param change msgs so that this sort of integration bug is caught

modules/apps/27-interchain-accounts/host/keeper/keeper.go
```diff
@@ -199,3 +205,33 @@ func (k Keeper) SetInterchainAccountAddress(ctx sdk.Context, connectionID, portI
...
+func (k Keeper) GetAuthority() string {
+	return k.authority
+}
+
+// GetParams returns the total set of the host submodule parameters.
+func (k Keeper) GetParams(ctx sdk.Context) types.Params {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get([]byte(types.ParamsKey))
+	if bz == nil {
+		return types.Params{}
```
colin-axner, line 219: We should panic here? An unset parameter seems unsafe to me cc @damiannolan

testing/simapp/app.go
colin-axner, line 1: nice :) love how little diffs there are on the app.go now

colin-axner (commented):
Wonderful! Code looks good! I've reviewed all the files, so this should be the last of my requested changes. Mostly left suggestions for some final touches/code coverage

modules/apps/27-interchain-accounts/host/types/params_legacy.go
colin-axner: Yes, that sounds perfect to me :+1:

modules/apps/27-interchain-accounts/host/types/params_legacy.go
colin-axner: oh yea, lets leave it then :+1:

modules/apps/27-interchain-accounts/host/keeper/genesis.go
```diff
@@ -27,7 +27,9 @@ func InitGenesis(ctx sdk.Context, keeper Keeper, state genesistypes.HostGenesisS
 		keeper.SetInterchainAccountAddress(ctx, acc.ConnectionId, acc.PortId, acc.AccountAddress)
 	}
 
-	keeper.SetParams(ctx, state.Params)
+	if err := keeper.SetParams(ctx, state.Params); err != nil {
+		panic(fmt.Sprintf("could not set ica host params at genesis: %v", err))
+	}
```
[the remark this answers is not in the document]
srdtrk: I'm not sure how to test panics

modules/apps/27-interchain-accounts/host/keeper/keeper.go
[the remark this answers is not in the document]
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/migrations_test.go
```diff
@@ -0,0 +1,41 @@
...
+		msg            string
+		malleate       func()
+		expectedParams icahosttypes.Params
+	}{
+		{
+			"success: default params",
+			func() {
+				params := icahosttypes.DefaultParams()
+				subspace := suite.chainA.GetSimApp().GetSubspace(icahosttypes.SubModuleName) // get subspace
+				subspace.SetParamSet(suite.chainA.GetContext(), &params)                     // set params
```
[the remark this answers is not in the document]
srdtrk: It won't pass, it is not setting up the paramspace, it is setting the parameters in the param space. We removed the code that sets the parameters in paramspace from genesis

modules/apps/27-interchain-accounts/host/types/params.go
srdtrk: done

modules/apps/27-interchain-accounts/module.go
srdtrk: added registration

modules/apps/27-interchain-accounts/host/keeper/keeper.go
srdtrk: I am in favour of panic too. Added

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/keeper.go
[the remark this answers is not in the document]
srdtrk: added test case

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
```diff
@@ -0,0 +1,37 @@
...
+// NewMsgServerImpl returns an implementation of the ICS27 host MsgServer interface
+// for the provided Keeper.
+func NewMsgServerImpl(keeper *Keeper) types.MsgServer {
+	return &msgServer{Keeper: keeper}
+}
+
+// UpdateParams updates the host submodule's params.
+func (m msgServer) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
+	if m.authority != msg.Authority {
+		return nil, errors.Wrapf(ibcerrors.ErrInvalidAuthority, "invalid authority; expected %s, got %s", m.authority, msg.Authority)
```
colin-axner, line 28:
```suggestion
		return nil, errors.Wrapf(ibcerrors.ErrInvalidAuthority, "expected %s, got %s", m.authority, msg.Authority)
```

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
colin-axner, line 6:
```suggestion
	errorsmod "cosmossdk.io/errors"
```

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
```diff
@@ -256,3 +262,20 @@ func (suite *KeeperTestSuite) TestParams() {
...
+	// this state should not be possible outside of tests
+	chainID := "testchain"
+	header := tmproto.Header{
+		ChainID: chainID,
+		Height:  1,
+		Time:    suite.coordinator.CurrentTime.UTC(),
+	}
+
+	app := simapp.NewSimApp(log.NewNopLogger(), dbm.NewMemDB(), nil, true, map[int64]bool{}, simapp.DefaultNodeHome, 5, simapp.MakeTestEncodingConfig(), simtestutil.EmptyAppOptions{}, baseapp.SetChainID(chainID))
+	ctx := app.GetBaseApp().NewContext(true, header)
```
colin-axner, line 276: another approach could be to delete the key:
```go
store := suite.chainA.GetContext().KVStore(suite.chainA.GetSimApp().GetKey(hosttypes.SubModuleName))
store.Delete(icatypes.ParamsKey)

```

modules/apps/27-interchain-accounts/host/keeper/keeper.go
colin-axner: After thinking about this some more, I have a slight preference for maintaining our existing design pattern of doing validation before calling the set function. I'm not sure there's a benefit to doing an extra validation before setting (unless you assume there's a missing validation check in our code)

colin-axner (approved):
Thanks @srdtrk for all the followup changes! Dropping my approval as the changes look good to me. I only have a few minor outstanding comments left. Let me know if you have any questions about them

We should have a second approval from someone who didn't work on the code

modules/apps/27-interchain-accounts/host/keeper/keeper.go
damiannolan: imho I prefer to have setters with a void return arg.

modules/apps/27-interchain-accounts/host/keeper/keeper.go
```diff
@@ -199,3 +205,33 @@ func (k Keeper) SetInterchainAccountAddress(ctx sdk.Context, connectionID, portI
...
+}
+
+// GetParams returns the total set of the host submodule parameters.
+func (k Keeper) GetParams(ctx sdk.Context) types.Params {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get([]byte(types.ParamsKey))
+	if bz == nil {
+		panic("ica/host params have not been set")
+	}
+
```
damiannolan, line 221: Afaik we normally check `len(bz) == 0`

If `len(bz) == 0` then should we be panicking or return `types.DefaultParams()`?

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
damiannolan: agree, we should avoid shadowing pkgs from std lib

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
damiannolan: I'm kind of indifferent on this - using the unexported `msgServer` type does reduce the surface area for issues of other random modules doing something they shouldn't be 🤷

modules/core/errors/errors.go
```diff
@@ -60,4 +60,7 @@ var (
 
 	// ErrNotFound defines an error when requested entity doesn't exist in the state.
 	ErrNotFound = errorsmod.Register(codespace, 16, "not found")
+
+	// ErrInvalidAuthority defines an error when the authority address is invalid.
+	ErrInvalidAuthority = errorsmod.Register(codespace, 17, "invalid authority")
```
damiannolan, line 65: mega nit tbh, but I don't see why we couldn't use `ErrUnauthorized` 

happy to keep it for parity with sdk

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
damiannolan: is there a need to create a new app in this test case when everything should be bootstrapped by `suite.SetupTest()`?

damiannolan (commented):
Excellent work on this, looking super clean. I left a couple of minor additional comments but happy to drop my approval after the `SetParams` API has been updated! 🚀

modules/apps/27-interchain-accounts/host/keeper/msg_server.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
srdtrk: done

modules/apps/27-interchain-accounts/module.go
```diff
@@ -136,13 +140,19 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
...
 
 	m := controllerkeeper.NewMigrator(am.controllerKeeper)
 	if err := cfg.RegisterMigration(types.ModuleName, 1, m.AssertChannelCapabilityMigrations); err != nil {
 		panic(fmt.Sprintf("failed to migrate interchainaccounts app from version 1 to 2: %v", err))
 	}
+
+	hostm := hostkeeper.NewMigrator(am.hostKeeper)
+	if err := cfg.RegisterMigration(types.ModuleName, 2, hostm.MigrateParams); err != nil {
+		panic(fmt.Sprintf("failed to migrate interchainaccounts host params from version 2 to 3: %v", err))
+	}
```
crodriguezvega, line 155: We should bump the number returned by function `ConsensusVersion` in line 186 of this file.

modules/apps/27-interchain-accounts/module.go
```diff
@@ -136,13 +140,19 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
...
 	}
 
 	m := controllerkeeper.NewMigrator(am.controllerKeeper)
 	if err := cfg.RegisterMigration(types.ModuleName, 1, m.AssertChannelCapabilityMigrations); err != nil {
 		panic(fmt.Sprintf("failed to migrate interchainaccounts app from version 1 to 2: %v", err))
 	}
+
+	hostm := hostkeeper.NewMigrator(am.hostKeeper)
+	if err := cfg.RegisterMigration(types.ModuleName, 2, hostm.MigrateParams); err != nil {
+		panic(fmt.Sprintf("failed to migrate interchainaccounts host params from version 2 to 3: %v", err))
```
crodriguezvega, line 154: Since we will probably migrate together both controller and host params, maybe we can do something here to bump the version only once. But maybe we can deal with that in a later PR after merging also the migration of the controller params.

```suggestion
		panic(fmt.Sprintf("failed to migrate interchainaccounts app from version 2 to 3: %v", err))
```

modules/apps/27-interchain-accounts/host/keeper/msg_server_test.go
```diff
@@ -0,0 +1,56 @@
...
+		{
+			"invalid authority address",
+			func(authority string) {
+				msg.Authority = "authority"
+				msg.Params = types.DefaultParams()
+			},
+			false,
+		},
+		{
+			"success",
```
crodriguezvega, line 25: Nit: by convention we usually place the success cases first.

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
```diff
@@ -218,3 +219,52 @@ func (suite *KeeperTestSuite) TestSetInterchainAccountAddress() {
...
+	suite.Require().Equal(expParams, params)
+
+	testCases := []struct {
+		name    string
+		input   types.Params
+		expPass bool
+	}{
+		{"success: set default params", types.DefaultParams(), true},
+		{"success: non-default params", types.NewParams(!types.DefaultHostEnabled, []string{"/cosmos.staking.v1beta1.MsgDelegate"}), true},
+		{"success: set empty allowMsg", types.NewParams(true, nil), true},
```
crodriguezvega, line 236:
```suggestion
		{"success: set empty byte for allow messages", types.NewParams(true, nil), true},
```

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
```diff
@@ -218,3 +219,52 @@ func (suite *KeeperTestSuite) TestSetInterchainAccountAddress() {
...
+		{"failure: set empty string", types.NewParams(true, []string{""}), false},
```
crodriguezvega, line 237:
```suggestion
		{"failure: set empty string for allow messages", types.NewParams(true, []string{""}), false},
```

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
```diff
@@ -218,3 +219,52 @@ func (suite *KeeperTestSuite) TestSetInterchainAccountAddress() {
...
+		{"failure: set space string", types.NewParams(true, []string{" "}), false},
```
crodriguezvega, line 238:
```suggestion
		{"failure: set space string for allow messages", types.NewParams(true, []string{" "}), false},
```

modules/apps/27-interchain-accounts/host/keeper/genesis_test.go
```diff
@@ -41,11 +41,71 @@ func (suite *KeeperTestSuite) TestInitGenesis() {
...
 
+func (suite *KeeperTestSuite) TestGenesisParams() {
+	testCases := []struct {
+		name    string
+		input   types.Params
+		expPass bool
+	}{
+		{name: "success: set default params", input: types.DefaultParams(), expPass: true},
+		{name: "success: non-default params", input: types.NewParams(!types.DefaultHostEnabled, []string{"/cosmos.staking.v1beta1.MsgDelegate"}), expPass: true},
+		{name: "success: set empty allowMsg", input: types.NewParams(true, nil), expPass: true},
```
crodriguezvega, line 57:
```suggestion
		{name: "success: set empty byte for allow messages", input: types.NewParams(true, nil), expPass: true},
```

modules/apps/27-interchain-accounts/host/keeper/genesis_test.go
```diff
@@ -41,11 +41,71 @@ func (suite *KeeperTestSuite) TestInitGenesis() {
...
+		{name: "failure: set empty string", input: types.NewParams(true, []string{""}), expPass: false},
```
crodriguezvega, line 58:
```suggestion
		{name: "failure: set empty string for allow messages", input: types.NewParams(true, []string{""}), expPass: false},
```

modules/apps/27-interchain-accounts/host/keeper/genesis_test.go
```diff
@@ -41,11 +41,71 @@ func (suite *KeeperTestSuite) TestInitGenesis() {
...
+		{name: "failure: set space string", input: types.NewParams(true, []string{" "}), expPass: false},
```
crodriguezvega, line 59:
```suggestion
		{name: "failure: set space string for allow messages", input: types.NewParams(true, []string{" "}), expPass: false},
```

crodriguezvega (commented):
Great work, @aleem1314! I just left a few nits.

modules/apps/27-interchain-accounts/host/keeper/genesis_test.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/genesis_test.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/genesis_test.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
srdtrk: done

modules/apps/27-interchain-accounts/host/keeper/msg_server_test.go
srdtrk: done

modules/apps/27-interchain-accounts/module.go
srdtrk: done, decided to increase the version now, I assumed the two would be done together but this is safer

modules/apps/27-interchain-accounts/module.go
srdtrk: done

modules/apps/27-interchain-accounts/module.go
damiannolan: Yep, the migration handler can be configured like this with an anonymous function

```go
if err := cfg.RegisterMigration(types.ModuleName, 2, func(ctx sdk.Context) error {
    if err := hostMigrator.MigrateParams(ctx); err != nil {
        return err
    }

    if err := controllerMigrator.MigrateParams(ctx); err != nil {
        return err
    }
}); err != nil {
	panic(fmt.Sprintf("failed to migrate interchainaccounts app from version 2 to 3: %v", err))
}
```

damiannolan (approved):
Thanks @aleem1314 for your work on this, and @srdtrk for seeing it to completion

modules/apps/27-interchain-accounts/module.go
damiannolan: this can be done later, lgtm

colin-axner:
I almost forgot, we need a note in the migration doc for the breaking change in app.go. It should be pretty short with a small diff noting that the authority needs to be added

modules/apps/27-interchain-accounts/host/types/msgs.go
```diff
@@ -0,0 +1,25 @@
...
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+var _ sdk.Msg = (*MsgUpdateParams)(nil)
+
+// ValidateBasic implements sdk.Msg
+func (msg MsgUpdateParams) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Authority)
+	if err != nil {
+		return err
```
DimitrisJim, line 11: seeing something similar in the transfer PR (different error message used there though)

```suggestion
		return errorsmod.Wrapf(ibcerrors.ErrInvalidAddress, "string could not be parsed as address: %v", err)
```

proto/ibc/applications/interchain_accounts/host/v1/tx.proto
```diff
@@ -0,0 +1,28 @@
...
+import "ibc/applications/interchain_accounts/host/v1/host.proto";
+
+// Msg defines the 27-interchain-accounts/host Msg service.
+service Msg {
+  // UpdateParams defines a rpc handler for MsgUpdateParams.
+  rpc UpdateParams(MsgUpdateParams) returns (MsgUpdateParamsResponse);
+}
+
+// MsgUpdateParams defines the payload for Msg/UpdateParams
+message MsgUpdateParams {
```
DimitrisJim, line 17: ah, also adding the message signer option or else we'll totally miss it after https://github.com/cosmos/ibc-go/pull/3627

proto/ibc/applications/interchain_accounts/host/v1/tx.proto
```diff
@@ -0,0 +1,28 @@
...
+
+package ibc.applications.interchain_accounts.host.v1;
+
+option go_package = "github.com/cosmos/ibc-go/v7/modules/apps/27-interchain-accounts/host/types";
+
+import "gogoproto/gogo.proto";
+import "ibc/applications/interchain_accounts/host/v1/host.proto";
+
+// Msg defines the 27-interchain-accounts/host Msg service.
+service Msg {
```
DimitrisJim, line 11: and the equivalent one for `services` https://github.com/cosmos/ibc-go/pull/3630

proto/ibc/applications/interchain_accounts/host/v1/tx.proto
srdtrk: Shouldn't their PRs take care of these when their PR is merging. If I implement it now, it can be confusing. Even if they have to manually resolve the merge conflict. I think it should be fine?

modules/apps/27-interchain-accounts/host/types/msgs.go
srdtrk: done

Issue #3483, opened 2023-04-18
Add Timeout type


colin-axner:
## Summary

Apply suggested [spec change](https://github.com/cosmos/ibc/issues/957). There is no rush, so maybe it makes sense to add a `Timeout` type directly to main and then adjust 04-channel-upgrades to use it?

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Expanding on this. While reviewing #3333, I proposed adding a helper function to bundle the timeout logic into a single function, see https://github.com/cosmos/ibc-go/pull/3333#discussion_r1158416582. Since we are adding a timeout type which bundles the height and timestamp of a timeout, it seemed useful to me to add a function which handles timeout checks for both types. 

04-channel handlers are primarily interested in if the timeout has been reached or not, aside from pretty error messages, it does not care which if the timeout height or timestamp was reached. 

The original proposed API included a boolean and a string to provide information about the timeout relative to the provided timestamp/height:
```go
func (t Timeout) Status(height clienttypes.Height, timestamp uint64) (bool, string)
```

The intention was that the bool returns if the timeout has passed and the string provides relevant information that can be used in the returned error.

This was then modified to be:
```go
func (t Timeout) HasPassed(ctx sdk.Context) (bool, error) {
```

I raised the concern (see https://github.com/cosmos/ibc-go/pull/3535#discussion_r1183460674) about using the error as a secondary return value since an error occurs based on the calling context, not whether the timeout has been reached or not. 

Further concerns were raised (see https://github.com/cosmos/ibc-go/pull/3535#issuecomment-1535505979) about returning relevant information in the secondary return. A proposal was made to use `Before()/After()` functions to align with the golang time package:

```go
func (t Timeout) AfterHeight(height clienttypes.Height) bool 
func (t Timeout) AfterTimestamp(timestamp uint64) bool 
```

This was then implemented in #3535

Upon reviewing the changes, I had differing opinions on proposed direction. Given all the back and forth design discussion on the pr's and offline, we decided to move the discussion to an issue so we could come to consensus on the design introduced. As stated in the first comment of this issue, the timeout type API is a convenience to improve code readability and safety, so it should not need to block implementation of channel upgradability. It's design can be agreed upon and merged to main and then updated on channel upgradability to be used. Channel upgradability only needs the type for verifiability which does not require additional API's being discussed here.

---

[...]

PR #3535, opened 2023-04-27
feat: util function for checking if an UpdateTimeout has passed, more testing for Upgrade.ValidateBasic()

charleenfei:
NOTE that a test for `HasPassed()` requires the `ctx` to be passed in from a chain, so the test for this function will be part of the upgrade tests for the ChanUpgradeTry handler, in a future PR.

pulls out changes from https://github.com/cosmos/ibc-go/pull/3478

ref: https://github.com/cosmos/ibc-go/issues/1619
ref: https://github.com/cosmos/ibc-go/issues/3448

---

write a little note why.


modules/core/04-channel/types/upgrade.go
```diff
@@ -69,3 +71,22 @@ func (uf UpgradeFields) ValidateBasic() error {
 func (ut UpgradeTimeout) IsValid() bool {
 	return !ut.Height.IsZero() || ut.Timestamp != 0
 }
+
+// HasPassed returns true if the upgrade has passed the timeout height or timestamp
+func (ut UpgradeTimeout) HasPassed(ctx sdk.Context) (bool, error) {
```
damiannolan, line 76: Could we add some unit tests for this? 

It should be possible to pass a context by either by adding a coordinator and test chain to the `TypesTestSuite` or maybe using the new [testutil pkg func](https://github.com/cosmos/cosmos-sdk/blob/main/testutil/context.go#L20) from the sdk. 

At a glance it looks like it would be easier to use the coordinator to malleate time

charleenfei: i was thinking about this method but wasnt sure if it was worth reworking the TypesTestSuite just for this one test. but i can update!

modules/core/04-channel/types/upgrade.go
```diff
@@ -69,3 +71,22 @@ func (uf UpgradeFields) ValidateBasic() error {
...
+	if !ut.IsValid() {
+		return true, errorsmod.Wrap(ErrInvalidUpgrade, "upgrade timeout cannot be empty")
```
DimitrisJim, line 78: dropping the comment from #3536 here

shouldn't this be `ErrInvalidUpgradeTimeout`? To mirror its usage in ValidateBasic of MsgChannelUpgradeTry:

```suggestion
		return true, errorsmod.Wrap(ErrInvalidUpgradeTimeout, "timeout height or timeout timestamp must be non-zero")
```

modules/core/04-channel/types/msgs_test.go
```diff
@@ -103,6 +109,16 @@ func (suite *TypesTestSuite) SetupTest() {
...
 }
 
+// SetupTestCoordinator creates a coordinator with 2 test chains.
+func (suite *TypesTestSuite) SetupTestCoordinator() {
+	suite.coordinator = ibctesting.NewCoordinator(suite.T(), 2)
+	suite.chainA = suite.coordinator.GetChain(ibctesting.GetChainID(1))
+	suite.chainB = suite.coordinator.GetChain(ibctesting.GetChainID(2))
+	// commit some blocks so that QueryProof returns valid proof (cannot return valid query if height <= 1)
+	suite.coordinator.CommitNBlocks(suite.chainA, 2)
+	suite.coordinator.CommitNBlocks(suite.chainB, 2)
```
colin-axner, line 119: I think this setup code is outdated. Could we try removing it and see if tests pass?

modules/core/04-channel/types/upgrade.go
```diff
@@ -69,3 +71,22 @@ func (uf UpgradeFields) ValidateBasic() error {
...
 }
+
+// HasPassed returns true if the upgrade has passed the timeout height or timestamp
+func (ut UpgradeTimeout) HasPassed(ctx sdk.Context) (bool, error) {
+	if !ut.IsValid() {
+		return true, errorsmod.Wrap(ErrInvalidUpgradeTimeout, "upgrade timeout cannot be empty")
+	}
+
+	selfHeight, timeoutHeight := clienttypes.GetSelfHeight(ctx), ut.Height
+	if selfHeight.GTE(timeoutHeight) && timeoutHeight.GT(clienttypes.ZeroHeight()) {
```
colin-axner, line 82: I would switch this ordering:
```suggestion
	if !timeoutHeight.IsZero() && selfHeight.GTE(timeoutHeight)  {
```

modules/core/04-channel/types/upgrade.go
colin-axner, line 76: I'm not sure how I feel about this API, it feels arbitrary to me to return an error when checking if the timeout has passed. I would prefer to ensure we never call `HasPassed` on an invalid timeout and then just return supplementary information when checking if the timeout has passed. It isn't an error if the timeout has passed or not. Was there a reason we didn't use:
```go
func (ut UpgradeTimeout) HasPassed(ctx sdk.Context) (bool, string) 
```
where string is information about the timeout relative to the current time. Then:
```go
// recv packet
if hasPassed, info := timeout.HasPassed(ctx); hasPassed {
    return errorsmod.Wrapf(ErrTimeoutPassed, info)
}
``` 
or 
```go
// timeout packet
if hasPassed, info := timeout.HasPassed(ctx); !hasPassed {
    return errorsmod.Wrapf(ErrInvalidTimeout, "timeout has not passed: %s", info)
}

**Review by colin-axner — CHANGES_REQUESTED** [review]
Thanks for splitting out this work and adding tests!

Using request changes as I don't agree with the current API and would like discussion on it

File: modules/core/04-channel/types/upgrade.go (hunk 1 above)
Comment (damiannolan) [reply modules/core/04-channel/types/upgrade.go:76]: I was also going to raise a comment about this API. To be honest, I initially expected this function to just return a `bool`, and was unsure of the value of the "info" string. 

But I see now it can add valuable error info indicating whether its the block height or timestamp which has elapsed. 

I'd be happy with `bool` or `(bool, string)` but I do agree that `(bool, error)` seems a little off

Comment (charleenfei) [reply modules/core/04-channel/types/upgrade.go:76]: I am happy with `bool, info` as I do feel like it is important to indicate which timeout has elapsed.

File: modules/core/04-channel/types/upgrade_test.go [hunk 5]
```diff
@@ -0,0 +1,161 @@
...
+
+			tc.malleate()
+
+			passed, info := upgrade.Timeout.HasPassed(suite.chainA.GetContext())
+
+			if tc.expPass {
+				suite.Require().False(passed)
+				suite.Require().Equal("", info)
+			} else {
+				suite.Require().True(passed)
```
DimitrisJim, line 157: do we also want to check the specific info matches in the err case?

modules/core/04-channel/types/upgrade.go
```diff
@@ -69,3 +72,22 @@ func (uf UpgradeFields) ValidateBasic() error {
 func (ut UpgradeTimeout) IsValid() bool {
 	return !ut.Height.IsZero() || ut.Timestamp != 0
 }
+
+// HasPassed returns true if the upgrade has passed the timeout height or timestamp
+func (ut UpgradeTimeout) HasPassed(ctx sdk.Context) (bool, string) {
+	if !ut.IsValid() {
+		return true, "upgrade timeout cannot be empty"
+	}
```
colin-axner, line 80: I think we can remove this. We should validate a timeout before using `HasPassed`

modules/core/04-channel/types/upgrade.go
```diff
@@ -69,3 +72,22 @@ func (uf UpgradeFields) ValidateBasic() error {
 func (ut UpgradeTimeout) IsValid() bool {
 	return !ut.Height.IsZero() || ut.Timestamp != 0
 }
+
+// HasPassed returns true if the upgrade has passed the timeout height or timestamp
+func (ut UpgradeTimeout) HasPassed(ctx sdk.Context) (bool, string) {
```
colin-axner, line 77:
```suggestion
// HasPassed returns true if the upgrade has passed the timeout height or timestamp.
// A string is returned containing information about the timeout relative to the context's time and height. 
func (ut UpgradeTimeout) HasPassed(ctx sdk.Context) (bool, string) {
```

modules/core/04-channel/types/upgrade.go
```diff
@@ -69,3 +72,22 @@ func (uf UpgradeFields) ValidateBasic() error {
...
+		return true, "upgrade timeout cannot be empty"
+	}
+
+	selfHeight, timeoutHeight := clienttypes.GetSelfHeight(ctx), ut.Height
+	if selfHeight.GTE(timeoutHeight) && timeoutHeight.GT(clienttypes.ZeroHeight()) {
+		return true, fmt.Sprintf("block height >= upgrade timeout height (%s >= %s)", selfHeight, timeoutHeight)
+	}
+
+	selfTime, timeoutTimestamp := uint64(ctx.BlockTime().UnixNano()), ut.Timestamp
+	if selfTime >= timeoutTimestamp && timeoutTimestamp > 0 {
```
colin-axner, line 88:
```suggestion
	if timeoutTimestamp != 0  && selfTime >= timeoutTimestamp &&  {
```

modules/core/04-channel/types/upgrade.go
```diff
@@ -69,3 +72,22 @@ func (uf UpgradeFields) ValidateBasic() error {
...
+		return true, fmt.Sprintf("block timestamp >= upgrade timeout timestamp (%s >= %s)", ctx.BlockTime(), time.Unix(0, int64(timeoutTimestamp)))
+	}
+
+	return false, ""
```
colin-axner, line 92: In the false case, we should still return a string which contains information about the timeout not being past. Maybe we should be collecting up the hasPassed, info in each if statement:

```go
var (
    hasPassed bool
    info string
)

if !timeoutHeight.IsZero() {
    hasPassed = selfHeight.GTE(timeoutHeight)
     info = fmt.Sprintf("block height %s, timeout height %s", selfHeight, timeoutHeight)
}

if timestamp != 0 {
    hasPassed = hasPassed || selfTime >= timeoutTimestamp
    info = fmt.Sprintf("%s: block timestamp %s, timeout timestamp %s", info, selfTime, timeoutTimestamp)
}

return hasPassed, info
```

Or something similar to this. I'm not sure exactly how we should format the information returned

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+			proposedConnectionHops = []string{path.EndpointB.ConnectionID}
+
+			upgrade = types.NewUpgrade(
+				types.NewUpgradeFields(
+					types.UNORDERED, proposedConnectionHops, fmt.Sprintf("%s-v2", mock.Version),
+				),
+				types.NewUpgradeTimeout(path.EndpointA.Chain.GetTimeoutHeight(), 0),
+				0,
+			)
```
colin-axner, line 147: I don't think we need to build an upgrade type. Just using the `Timeout` should be sufficient in my opinion (I think UpgradeTimeout should be renamed to Timeout and used for packet timeouts as well)

colin-axner (commented):
Thanks for switching the API, I left some code structure suggestions. Let me know what you think!

modules/core/04-channel/types/upgrade_test.go
charleenfei: i can open another issue for the UpgradeTimeout -> Timeout name change, since wouldn't this change the SendPacket API?

modules/core/04-channel/types/upgrade.go
charleenfei: is there any harm in keeping this check though? otherwise we end up returning a false negative in the event the timeout validation doesn't occur.

modules/core/04-channel/types/upgrade.go
```diff
@@ -69,3 +71,31 @@ func (uf UpgradeFields) ValidateBasic() error {
 func (ut UpgradeTimeout) IsValid() bool {
 	return !ut.Height.IsZero() || ut.Timestamp != 0
 }
+
+// HasPassed returns true if the upgrade has passed the timeout height or timestamp.
+// A string is returned containing information about the timeout relative to the context's time and height.
+func (ut UpgradeTimeout) HasPassed(ctx sdk.Context) (bool, string) {
```
AdityaSripal, line 77: Is this the cleanest return signature? feels like an odd pattern. Why don't we just return an sdk error with an error string

charleenfei: because there was [a preference for an info string](https://github.com/cosmos/ibc-go/pull/3535#discussion_r1183460674) rather than returning an error if the timeout had passed

damiannolan: do you mean an `error` only? no bool? @AdityaSripal

damiannolan:
There seems to be some unclarity around the API in this PR and I've been brainstorming through some different thoughts and ideas. I'll try to highlight and reiterate some of the themes I'm seeing coming up:

`HasPassed(ctx sdk.Context) (bool, string)`

This method returns a boolean as well as a string, intended to indicate some information regarding whether the timeout has elapsed or not. The format of the this string is unclear, it also seems that when this method returns `false`, that there is little value for it. We have also looked at `(bool, error)` for example, but this seems a little ambiguous as @colin-axner points out https://github.com/cosmos/ibc-go/pull/3535#discussion_r1183460674. 

Some other less desirable alternatives are returning just `bool`, but then we lose valuable information to return to the caller such as which value triggered the timeout - height or timestamp. We could simply return `error` but that feels off given the conventions of `HasXyz` functions.

I'd like to propose something like this [draft gist](https://gist.github.com/damiannolan/8b8b008264c2231978e15a553301547a), where we follow similar naming conventions to stdlib, e.g. `time.After()`, `time.Before()` and continue to do both height and timestamp checks explicitly in the handler while still introducing a `Timeout` type addressing this PR as well as #3483. 
I've left a small example usage in the gist. To me it feels more natural to check in the handler and return errors with the appropriate info (height or timestamp). What do ya'll think?

DimitrisJim:
I think I'd prefer https://github.com/cosmos/ibc-go/pull/3535#issuecomment-1535505979 but, just to offer another alternative for completeness'/consideration sake, the return value could also be an `enum` representing the different info states.

charleenfei:
**NOTE** please first review https://github.com/cosmos/ibc-go/pull/3572

I've updated the API now following the suggestions from @damiannolan. As discussed above, returning just an error would not make sense in instances where the usage of the handler calling this method would actually be to time out a packet. Returning just a bool will lose valuable UX information for the caller about which timeout has passed (and by how many blocks/how much time). Returning both seems ambiguous, as pointed out by @colin-axner

modules/core/04-channel/types/packet.go
```diff
@@ -133,3 +133,36 @@ func (p PacketId) Validate() error {
...
+	return t.Timestamp > timestamp
+}
+
+// IsValid validates the Timeout. It ensures that either height or timestamp is set.
+func (t Timeout) IsValid() bool {
+	return !t.ZeroHeight() || !t.ZeroTimestamp()
+}
+
+// ZeroHeight returns true if Timeout height is zero, otherwise false.
+func (t Timeout) ZeroHeight() bool {
```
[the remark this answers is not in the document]
DimitrisJim: I think `IsZeroHeight` makes sense here (and similarly for `Timestamp`). `Zero` can be confused for a constructing function that returns a zero height plus there's some backing for the `Is` variant by looking at [equivalent functions](https://pkg.go.dev/time#Time.IsZero) in stdlib's `time`.

modules/core/04-channel/types/timeout.go
```diff
@@ -0,0 +1,38 @@
...
+func (t Timeout) IsValid() bool {
+	return !t.IsZeroHeight() || !t.IsZeroTimestamp()
+}
+
+// IsZeroHeight returns true if Timeout height is zero, otherwise false.
+func (t Timeout) IsZeroHeight() bool {
+	return t.Height.IsZero()
+}
+
+// ZeroTimestamp returns true if Timeout timestamp is zero, otherwise false.
```
DimitrisJim, line 35:
```suggestion
// IsZeroTimestamp returns true if Timeout timestamp is zero, otherwise false.
```

testing/endpoint.go
```diff
@@ -569,7 +569,7 @@ func (endpoint *Endpoint) ChanUpgradeInit(timeout channeltypes.Timeout) error {
 		endpoint.ChannelConfig.PortID,
 		endpoint.ChannelID,
 		channeltypes.NewUpgradeFields(endpoint.ChannelConfig.Order, []string{endpoint.ConnectionID}, endpoint.ChannelConfig.Version),
-		channeltypes.NewUpgradeTimeout(timeout.Height, timeout.Timestamp),
+		channeltypes.NewTimeout(timeout.Height, timeout.Timestamp),
```
damiannolan, line 572: I think we're passing in timeout as a function arg here and recreating a new timeout. We can use the arg directly

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -0,0 +1,81 @@
+package types_test
+
+import (
+	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
+	"github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+	"github.com/cosmos/ibc-go/v7/testing/mock"
+)
+
+func (suite *TypesTestSuite) TestUpgradeTimeout() {
```
damiannolan, line 9: This looks like a test for `ValidateBasic`?

modules/core/04-channel/types/timeout_test.go
```diff
@@ -0,0 +1,109 @@
...
+			if tc.expPass {
+				suite.Require().True(passed)
+			} else {
+				suite.Require().False(passed)
+			}
+		})
+	}
+}
+
+func (suite *TypesTestSuite) TestTimeout() {
```
damiannolan, line 66: Maybe something like:

```suggestion
func (suite *TypesTestSuite) TestTimeoutIsValid() {
```

side note: but this looks like its being covered as part of testing `Upgrade.ValidateBasic()` but I'm happy to keep this also.

modules/core/04-channel/types/timeout_test.go
```diff
@@ -0,0 +1,109 @@
...
+	var timeout types.Timeout
+	var valid bool
```
damiannolan, line 68: `valid` doesn't need to be exposed at the top level scope of this function, we can use `valid := timeout.IsValid()` on line 100

modules/core/04-channel/types/timeout_test.go
```diff
@@ -0,0 +1,109 @@
+package types_test
+
+import (
+	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
+	"github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+)
+
+func (suite *TypesTestSuite) TestTimeoutPassed() {
+	var timeout types.Timeout
+	var passed bool
```
damiannolan, line 10: I'm surprised the linter isn't asking for:

```go
var (
    timeout types.Timeout
    passed  bool
)
```

modules/core/04-channel/types/timeout_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	testCases := []struct {
+		name     string
+		malleate func()
+		expPass  bool
+	}{
+		{
+			"client height is not after timeout height",
+			func() {
+				passed = timeout.AfterHeight(clienttypes.NewHeight(0, 75))
```
damiannolan, line 20: I think it could be worth adding some more testcases like where heights are equal, where heights have different revision numbers but equal heights etc. 

Similarly for timestamp, where timestamps are equal it should return false

damiannolan (commented):
Just left some nits on testing, will ACK once changes are applied. 

Thank you @charleenfei! 🙏

modules/core/04-channel/types/timeout_test.go
charleenfei: yea, i thought about this but also thought it would make more sense to have the unit testing more explicitly associated with what is being tested, rather than as part of the tests for another method.

modules/core/04-channel/types/timeout_test.go
```diff
@@ -0,0 +1,131 @@
...
+			true,
+		},
+		{
+			"invalid timeout",
+			func() {
+				timeout.Height = clienttypes.ZeroHeight()
+				timeout.Timestamp = 0
+			},
+			false,
+		},
```
damiannolan, line 109: we should add some more cases - e.g. height is valid but timestamp is not, as well as the inverse

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		suite.Run(tc.name, func() {
+			upgrade = types.NewUpgrade(
+				types.NewUpgradeFields(
+					types.ORDERED,
+					[]string{"connection-0"},
```
damiannolan, line 60: mega nit: there is a const in the testing pkg `ibctesting.FirstConnectionID`

damiannolan (approved):
Awesome lgtm! Thanks for pushing this through after all of the chopping and changing! 🚀 

Could you applied the test name change suggestion in https://github.com/cosmos/ibc-go/pull/3535/files#r1196075862

colin-axner:
I've recapped the design discussion so far and added my thoughts as well on the `AfterTimestamp`, `AfterHeight` approach, https://github.com/cosmos/ibc-go/issues/3483#issuecomment-1553007654. Maybe we could come to consensus on an approach there? Since this is a util function, it should not block channel upgradability (just use the existing approach used by the other 04-channel handlers)

PR #3541, opened 2023-05-02
chore: script to automate updates of compatibility test configs for new releases

charleenfei:
## Description

Currently, we have had to manually replace release versions with new ones if there is EOL or a security issue with one of the versions, and new releases necessitated the manual insertion of the release version into each compatibility text matrix config file.

This script automates each one of these paths depending on user input. It also creates a backup directory before any of the automations are run in case of something going wrong.

```bash
chore: script to automate updates of compatibility test configs for new releases
```

---

write a little note why.


.github/compatibility-test-matrices/release-script.sh
```diff
@@ -0,0 +1,50 @@
...
+# Create a backup directory with the current date and time
+backup_dir="${directory_path}_backup_$(date +%Y-%m-%d-%H-%M-%S)"
+mkdir "$backup_dir"
+
+# Copy the contents of the directory to the backup directory
+cp -R "$directory_path"/* "$backup_dir"
+
+# Step 1: Replace the release version in JSON files
+echo "Enter the release version to replace (ie: v4.4.0, leave empty to skip):"
+read old_release_version
```
damiannolan, line 15: Looks like this script is doing two things, right? Do you think it warrants being split up?

If keeping as a single script then a general UX remark could be to present a menu and select (1) or (2) to choose which action to perform rather than a CR input to skip here.

charleenfei: yea, it was the two possible options that i saw carlos doing with release versions (either replacing due to security) or adding a new one. I can split it up, good idea

.github/compatibility-test-matrices/release-script.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Get the directory path where JSON files are located
+directory_path=$(dirname "$0")
+
+# Create a backup directory with the current date and time
+backup_dir="${directory_path}_backup_$(date +%Y-%m-%d-%H-%M-%S)"
+mkdir "$backup_dir"
+
+# Copy the contents of the directory to the backup directory
+cp -R "$directory_path"/* "$backup_dir"
```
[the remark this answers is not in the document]
charleenfei: yes, it is

.github/compatibility-test-matrices/release-script.sh
```diff
@@ -0,0 +1,50 @@
...
+          # Use jq to replace the old release version with the new release version
+          updated_object=$(echo "$object" | jq ".[] |= map(if . == \"$old_release_version\" then \"$new_release_version\" else . end)")
+          # Use jq to update the object in the JSON file
+          jq --argjson updated_object "$updated_object" '. |= $updated_object' "$file" > tmp.json && mv tmp.json "$file"
+          echo "Replaced ${old_release_version} with ${new_release_version} in ${file}"
+      done
+  done
+fi
+
+# Step 2: Add new release version to testing configuration files
```
colin-axner, line 32: Should we have a remove script as well?

.github/compatibility-test-matrices/release-script.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# loop through all json files in directory and its subdirectories
+for file in $(find "$directory_path" -name "*.json"); do
+
+  # parse the json file and search for the recent_version string
+  json=$(cat $file)
+  if [[ $json == *"$recent_version"* ]]; then
+
+    # add new_version string to the json array containing recent_version string
+    updated_json=$(echo $json | jq ".[] |= if(index(\"$recent_version\")) then . + [\"$new_version\"] else . end")
```
DimitrisJim, line 44: also, re running this. It currently appends to the list of versions after finding the most recent version. In practice, I guess this just means it will run the compat test with the most recent version _after_ running it with older versions, i.e after adding `v6.2.0` as the new release version and selecting `v6.1.0` as most recent we go from:

```json
 "chain-b": [ "release-v4.3.x", "v6.1.0", "v5.2.0", "v4.3.0", "v4.2.0", "v4.1.1"]
```

to:

```json
 "chain-b": [ "release-v4.3.x", "v6.1.0", "v5.2.0", "v4.3.0", "v4.2.0", "v4.1.1", "v6.2.0" ]
```
probably not of much consequence but we might want to keep these in sorted order?

DimitrisJim: also, `jq` pretty prints by default, looking at [this SO answer](https://stackoverflow.com/a/33181419) adding `-c` to `jq` could have it keep the previous formatting (haven't checked this tho)

.github/compatibility-test-matrices/release-script.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Create a backup directory with the current date and time
+backup_dir="${directory_path}_backup_$(date +%Y-%m-%d-%H-%M-%S)"
+mkdir "$backup_dir"
+
+# Copy the contents of the directory to the backup directory
+cp -R "$directory_path"/* "$backup_dir"
+
+# Step 1: Replace the release version in JSON files
+echo "Enter the release version to replace (ie: v4.4.0, leave empty to skip):"
```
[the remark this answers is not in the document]
charleenfei: yes, if we have security issue. but i guess actually for this situation, the delete script is probably better than a replace script

.github/compatibility-test-matrices/release-script.sh
```diff
@@ -0,0 +1,50 @@
...
+      done
+  done
+fi
+
+# Step 2: Add new release version to testing configuration files
+read -p "Enter the new release version that you would like to insert (ie: v4.5.0): " new_version
+read -p "Enter the most recent release version (ie: v4.4.0): " recent_version
+
+# loop through all json files in directory and its subdirectories
+for file in $(find "$directory_path" -name "*.json"); do
```
srdtrk, line 37:
```suggestion
# shellcheck disable=SC2044
for file in $(find "$directory_path" -name "*.json"); do
```
shellcheck complains here with [SC2044](https://www.shellcheck.net/wiki/SC2044). This means that certain filenames can produce unintended behaviour. You can follow the suggestions in the link, or if you think this is not a big deal since we plan to keep the naming consistent, then you can apply my suggestion.

srdtrk: Also the script operates on the backup files too, is this really intended?

DimitrisJim: if you run twice w/o deleting the back-up, right? I'd say that wouldn't be intended.

.github/compatibility-test-matrices/release-script.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
DimitrisJim: also also, other scripts set a couple of options wrt error handling: [set -e, -u, -o, -x pipefail](https://gist.github.com/vncsna/64825d5609c146e80de8b1fd623011ca). I don't think we do `-x` in the others so far, mostly `set -euo pipefail`.

.github/compatibility-test-matrices/release-script.sh
damiannolan: I thought `-c` was for "compact" i.e. on a single line.

I would also like to keep the ordering if possible, I think `jq` have some kind of `|=sort_by()` functionality which could do lexicographical ordering but I'd need to play around with it to figure it out

DimitrisJim: this should do the trick with keeping it sorted (going with reverse lexicographic order since that is what currently exists):

```suggestion
    updated_json=$(echo $json | jq ".[] |= if(index(\"$recent_version\")) then (. + [\"$new_version\"] | sort | reverse) else . end")
```

.github/compatibility-test-matrices/release-script.sh
```diff
@@ -0,0 +1,50 @@
...
+
+  # parse the json file and search for the recent_version string
+  json=$(cat $file)
+  if [[ $json == *"$recent_version"* ]]; then
+
+    # add new_version string to the json array containing recent_version string
+    updated_json=$(echo $json | jq ".[] |= if(index(\"$recent_version\")) then . + [\"$new_version\"] else . end")
+
+    # write the updated json to file
+    echo $updated_json > $file
```
DimitrisJim, line 47: Adding this suggestion from the other PR so it doesn't get lost:

```suggestion
    echo $updated_json | jq '.' > $file
```

this is to retain the pretty-printing `jq` does.

.github/compatibility-test-matrices/compat.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/python3
```
DimitrisJim, line 1: Added an alternative in Python. Though bash does the trick, the script seemed like it was doing enough to warrant a language that doesn't make your eyes hurt.

I'm generally open to sticking with bash but find this solution both more readable, more maintainable and less error prone.

Still need to get some feedback from Carlos to make sure the script does what is typically required. Other than that, willing to drop this and go with bash if people prefer that.

.github/compatibility-test-matrices/compat.py
```diff
@@ -0,0 +1,104 @@
...
+        dirs[:] = (d for d in dirs if not d.startswith(ignores))
+        for file_ in files:
+            if file_.endswith(".json"):
+                yield os.path.join(root, file_)
+
+def has_release_version(json_file, keys, version):
+    """ Check if the json file has the version in question. """
+    return any(version in row for row in (json_file[key] for key in keys))
+
+def update_version(json_file, keys, args):
```
chatton, line 25: technically handled in the parse args, code, but maybe we can raise a `ValueError` if the operation is unknown in this function too just to be safe and explode when bad inputs are used.

.github/compatibility-test-matrices/compat.py
```diff
@@ -0,0 +1,104 @@
...
+        action="store_true",
+        help="Allow for verbose output",
+        default=False,
+    )
+
+    args = parser.parse_args()
+    require_version(args)
+    return args
+
+def main(args):
```
chatton, line 90: one thing I noticed was that when we run the script giving an version that doesn't exists, it executes successfully but is just silent. I think it would be no harm either logging that the version wasn't found, or raise a ValueError with the input not having been found.

chatton (commented):
Reviewing the python version, I really like the look of this script being in python. I think it would be a lot more maintainable (and even testable if we wanted to go to those lengths)

If we do go forward with the python version, I'd like to see type hints introduced since we'd be starting with a very small amount of code it would be easy to do.

.github/compatibility-test-matrices/compat.py
```diff
@@ -0,0 +1,104 @@
...
+KEYS = ("chain-a", "chain-b")
+# Toggle if required. Indent = 4 matches jq.
+DUMP_ARGS = {"indent": 4, "sort_keys": False, "ensure_ascii": False}
+# Suggestions for recent and new versions.
+SUGGESTION = "for example, v4.3.0 or 4.3.0"
+
+def find_json_files(directory, ignores=(".",)):
+    """ Find JSON files in a directory. By default, ignore hidden directories."""
+    for root, dirs, files in os.walk(directory):
+        dirs[:] = (d for d in dirs if not d.startswith(ignores))
```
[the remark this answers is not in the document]
DimitrisJim: totally ugly, agree there but this in-place assignment of the returned directories allows for skipping any filtered entries when `topdown=True`. Alternative would be an iteration over the dirs and manual `del` of the correct indices.

In general, `ignores` was added based on the initial bash script considering a `.backup` dir being present. Since I haven't added that in (thinking a `git restore` could do the trick) this could also be removed).

.github/compatibility-test-matrices/compat.py
```diff
@@ -0,0 +1,104 @@
...
+        for file_ in files:
+            if file_.endswith(".json"):
+                yield os.path.join(root, file_)
+
+def has_release_version(json_file, keys, version):
+    """ Check if the json file has the version in question. """
+    return any(version in row for row in (json_file[key] for key in keys))
```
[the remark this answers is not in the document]
DimitrisJim: but don't you know that one-liners always win? Will split up, serves no real point.

.github/compatibility-test-matrices/compat.py
DimitrisJim: yeah, can see the value in that

.github/compatibility-test-matrices/compat.py
```diff
@@ -2,78 +2,93 @@
...
 import json
+import enum
+from typing import Tuple, Generator, Optional, Dict, Any
 
 # JSON keys to search in.
-KEYS = ("chain-a", "chain-b")
+KEYS: Tuple[str, str] = ("chain-a", "chain-b")
 # Toggle if required. Indent = 4 matches jq.
-DUMP_ARGS = {"indent": 4, "sort_keys": False, "ensure_ascii": False}
+DUMP_ARGS: Dict[Any, Any] = {
```
DimitrisJim, line 11: unfortunate but `my;py` throws a big hissy-fit here. `**kwargs` in general are a pain in typing iirc.

.github/compatibility-test-matrices/compat.py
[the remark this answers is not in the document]
chatton: one liners win until someone else needs to modify it :D

.github/compatibility-test-matrices/compat.py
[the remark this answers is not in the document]
chatton: oh I see so we're actually modifying the dirs that we're already iterating over

scripts/compat.py
```diff
@@ -0,0 +1,191 @@
...
+# Toggle if required. Indent = 2 matches our current formatting.
+DUMP_ARGS: Dict[Any, Any] = {
+    "indent": 2,
+    "sort_keys": False,
+    "ensure_ascii": False,
+}
+# Suggestions for recent and new versions.
+SUGGESTION: str = "for example, v4.3.0 or 4.3.0"
+# Supported Operations.
+Operation = enum.Enum("Operation", ["ADD", "REMOVE", "REPLACE"])
```
chatton, line 52: ❤️

chatton (approved):
LGTM thank you so much @charleenfei  & @DimitrisJim , this will make the release process a lot simpler 🥇 

One final nit, `compat.py` is a bit ambiguous, maybe we could name the file `update_compatibility_tests.py` to be more explicit, not a blocker at all though.

scripts/update_compatibility_tests.py
```diff
@@ -0,0 +1,191 @@
...
+
+def has_release_version(json_file: Any, keys: Tuple[str, str], version: str) -> bool:
+    """Check if the json file has the version in question."""
+    rows = (json_file[key] for key in keys)
+    return any(version in row for row in rows)
+
+
+def sorter(key: str) -> str:
+    """Since 'main' < 'vX.X.X' and we want to have 'main' as the first entry
+    in the list, we return a version that is considerably large. If ibc-go
```
charleenfei, line 74: 😄

charleenfei (commented):
tysm for making a much more readable/maintainable version of my hack <3

scripts/update_compatibility_tests.py
```diff
@@ -0,0 +1,191 @@
...
+    reaches this version I'll wear my dunce hat and go sit in the corner.
```
damiannolan, line 75: I will also eat my shorts on live stream

damiannolan (approved):
Awesome stuff! The python code looks pretty clean, its been a while since I looked at any! I think its more maintainable than bash.

LGTM 🚀

scripts/update_compatibility_tests.py
crodriguezvega: At our current pace of 3 major releases a year we can get there in 33330 years. Think about the poor person on whose face this time bomb will explode, man.

DimitrisJim: a quantum commit in the year 35353 "F*** you, DimitrisJim."

Issue #3575, opened 2023-05-12
Create an API on the transfer module that allows to construct an IBC denom given the channel, port and original denom


srdtrk:
## Summary

Design an API to Return Full IBC Denomination Based on Channel, Port, and Original Denomination.

## Problem Definition

External modules or applications that need to generate the IBC denomination often find themselves replicating code from the transfer module. An example of such replication can be found [here](https://github.com/cosmos/interchain-security/blob/ab5201f67d4c138a1dd14383ebda8c35986f2210/x/ccv/consumer/keeper/distribution.go#L174-L188)

## Proposal

To address this issue, it is proposed to create an API that returns the full IBC denomination given the channel, port, and original denomination. The solution would look like (and reside in `types/coin.go`):

```go
// GetIBCDenom generates the full IBC denomination string based on
// the provided portID, channelID, and baseDenom.
func GetIBCDenom(portID, channelID, baseDenom string) string {
	denomTrace := ParseDenomTrace(GetPrefixedDenom(portID, channelID, baseDenom))
	return denomTrace.IBCDenom()
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
See [comment](https://github.com/cosmos/ibc-go/pull/3579#issuecomment-1637662636).

PR #3579, opened 2023-05-15
refactor(transfer): created 'GetIBCDenom' api function

srdtrk:
## Description

This PR:

- [x] Creates the API function `GetIBCDenom` and refactors `GetTransferCoin` function to use it.

closes: #3575 

Notes:

- No additional tests needed.
- Other functions need not be factored to use `GetIBCDenom`.
- No docs additions are needed.

```
refactor(transfer): created 'GetIBCDenom' api function
```

---

write a little note why.

- [x] ~Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).~
- [x] ~Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).~


modules/apps/transfer/types/coin.go
```diff
@@ -40,9 +40,15 @@ func GetPrefixedDenom(portID, channelID, baseDenom string) string {
 	return fmt.Sprintf("%s/%s/%s", portID, channelID, baseDenom)
 }
 
+// GetIBCDenom generates the full IBC denomination string based on
+// the provided portID, channelID, and baseDenom.
+func GetIBCDenom(portID, channelID, baseDenom string) string {
```
DimitrisJim, line 45: true to the purpose of this PR, I'm seeing a couple places in our code where this might be used. I grepped for usages of `ParseDenomTrace`, there's some noise there but not too much (I think its also used in tests which also means indirect testing!).

crodriguezvega:
@srdtrk Looking at the [code in interchain-security](https://github.com/cosmos/interchain-security/blob/ab5201f67d4c138a1dd14383ebda8c35986f2210/x/ccv/consumer/keeper/distribution.go#L174-L188) that is linked in the issue. I am wondering if it could be simplified to something like this:

```go
for _, denom := range k.GetProviderRewardDenoms(ctx) {
  prefixedDenom := transfertypes.GetPrefixedDenom(transfertypes.PortID, k.GetDistributionTransmissionChannel(ctx), denom)

  // construct the denomination trace from the full raw denomination
  denomTrace := transfertypes.ParseDenomTrace(prefixedDenom)

  // append the IBC denom to the list of allowed reward denoms
  rewardDenoms = append(rewardDenoms, denomTrace.IBCDenom())
}
```

With this they also don't need to duplicate code from ibc-go to generate the IBC denomination and they can just use existing APIs that are available. What do you think?

srdtrk:
@crodriguezvega I think denomTrace should be for internal usage and external developers need not understand it. They wouldn't have asked if it was easy for them to figure this out on their own. And why should we have them use an intermediary object that they don't need. But I also see your point, so I am open to not merging this too.

charleenfei:
to be honest, i do see the UX value of having an explicit API for getting IBC denoms, rather than relying on devs understanding how to use `ParseDenomTrace`. I would vote for merging this in.

modules/apps/transfer/types/coin.go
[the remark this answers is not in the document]
srdtrk: yes I agree. I've made this change.

srdtrk: I don't want this function to be used a lot in the internal workings of the modules. This is meant to be useful to external developers who want to import our types.

I think internally, denom traces make sense, and `ParseDenomTrace` may have other reasons to exist (although you did mention there isn't that much noise). I think it being used in `GetTransferCoin` will indirectly test it.

I can refactor further like you suggested if others agree.

crodriguezvega:
Thanks for the feedback, @srdtrk and @charleenfei.

I would still see maybe the value in external developers have access to the `DenomTrace` type. Getting the IBC denom is one thing they can do, but `DenomTrace` also provides other methods that might be useful (like `GetFullDenomPath`, or getting the full path in case base denom is a denomination that also contains path information), and they would also have access to any new functions that we may add to the type in the future.

More than just adding a helper function I think we could add a constructor function that hides the complexity of calling `GetPrefixedDenom` and `ParseDenomTrace`. For example:

```go
func NewDenomTrace(portID, channelID, baseDenom string) DenomTrace {
  return ParseDenomTrace(GetPrefixedDenom(portID, channelID, baseDenom))
}
```

Then in ICS they can just do:

```
for _, denom := range k.GetProviderRewardDenoms(ctx) {
  // construct the denomination trace from the full raw denomination
  denomTrace := transfertypes.NewDenomTrace(transfertypes.PortID, k.GetDistributionTransmissionChannel(ctx), denom)

  // append the IBC denom to the list of allowed reward denoms
  rewardDenoms = append(rewardDenoms, denomTrace.IBCDenom())
}
```

With this constructor function `NewDenomTrace` we could potentially make `ParseDenomTrace` unexported.

srdtrk:
I like this idea too @crodriguezvega, were you thinking this as an addition to this PR or to replace the current function I implemented?

crodriguezvega:
In my honest opinion, I think we could replace the `GetIBCDenom` function with what I propose above. What do you think?

srdtrk:
Larger discussion needed around transfer api

Issue #3568, opened 2023-05-08
Failed to send localhost ibc transfer


jtieri:
## Summary of Bug

We got localhost channel handshakes working with the go-relayer and I started to test ics-20 transfers but the cmd to initiate the transfer fails every time.

The cmd being executed:
```
simd tx ibc-transfer transfer transfer channel-0 cosmos1cdu2wky305myqfgpr5v47nuwkhrldhhycvfzal 1000stake --from cosmos1aw37tvhsvugtxz674smtxzkez9nlyyp305k3dg --gas-prices 0.0stake --gas-adjustment 1.2 --keyring-backend test --output json -y --home /var/cosmos-chain/simd --node tcp://chain-a-val-0-TestLocalhostIBC:26657 --chain-id chain-a
```

Error logs:
```
Error:  Received unexpected error:
        send ibc transfer: exit code 1:  Error: rpc error: code = NotFound desc = rpc error: code = NotFound desc = client-id: 09-localhost: consensus state not found: key not found
```

I'm using the ibc-go-simd docker image from GHCR built from the main branch.

Looks like the error is returned from here:
https://github.com/cosmos/ibc-go/blob/8571b04d40e58b70caa940c466055138ab18d926/modules/core/04-channel/keeper/packet.go#L68-L71

## Version

89ecf25f3619

## Steps to Reproduce

I have prepared a test case in the relayer repo here:
https://github.com/cosmos/relayer/blob/fb9b953cbc81e6c01774c719adbe35f220f809f8/interchaintest/localhost_client_test.go#L21

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Thank you for reporting this error, @jtieri. We will look into it, but I just wanted to share that we have an [e2e test](https://github.com/cosmos/ibc-go/blob/main/e2e/tests/transfer/localhost_test.go#L32) that makes an ICS20 transfer over the localhost client and it has been passing for us. The test doesn't use the golang relayer though; we broadcast all the handshake messages manually. But maybe we could do a bit of quick refactoring and plugin the relayer if you have a tag that we can use?

damiannolan:
[...]

PR #3587, opened 2023-05-16
fix: add relative timeout support for localhost clients using the transfer CLI client

damiannolan:
## Description

This PR adds a little bit more code to the CLI handler in order to accommodate localhost clients. 
I've implemented it as so to avoid breaking APIs which will allow it be backported for `v7.1.0`

- Adds relative timeout support for localhost clients in transfer CLI
- ~Updates incorrect error return in 04-channel `SendPacket`~ https://github.com/cosmos/ibc-go/pull/3593

closes: #3568 

```bash
fix: add relative timeout support for localhost clients using the transfer CLI client
```

---

write a little note why.


modules/core/04-channel/keeper/packet.go
```diff
@@ -67,7 +67,7 @@ func (k Keeper) SendPacket(
 
 	clientState, found := k.clientKeeper.GetClientState(ctx, connectionEnd.GetClientID())
 	if !found {
-		return 0, clienttypes.ErrConsensusStateNotFound
+		return 0, clienttypes.ErrClientNotFound
```
colin-axner, line 70: This is state machine breaking :eyes: since the code is changing. Can we make the change in a separate pr in case this gets backported?

modules/apps/transfer/client/cli/tx.go
```diff
@@ -85,24 +86,51 @@ corresponding to the counterparty channel. Any timeout set to 0 is disabled.`),
...
+				if !ok {
+					return fmt.Errorf("invalid height type. expected type: %T, got: %T", clienttypes.Height{}, clientState.GetLatestHeight())
+				}
+
+				var consensusState exported.ConsensusState
+				if clientState.ClientType() != exported.Localhost {
+					consensusStateRes, err := channelutils.QueryChannelConsensusState(clientCtx, srcPort, srcChannel, clientHeight, false)
+					if err != nil {
+						return err
+					}
```
colin-axner, line 109: it might be faster to just query for the consensus state using the clientRes.ClientID? Otherwise I think another channel query, connection query is performed

modules/apps/transfer/client/cli/tx.go
```diff
@@ -85,24 +86,51 @@ corresponding to the counterparty channel. Any timeout set to 0 is disabled.`),
...
 				if !timeoutHeight.IsZero() {
-					absoluteHeight := height
+					absoluteHeight := clientHeight
 					absoluteHeight.RevisionNumber += timeoutHeight.RevisionNumber
 					absoluteHeight.RevisionHeight += timeoutHeight.RevisionHeight
 					timeoutHeight = absoluteHeight
 				}
 
+				// use local clock time as reference time if it is later than the
+				// consensus state timestamp of the counter party chain, otherwise
```
colin-axner, line 124:
```suggestion
				// consensus state timestamp of the counterparty chain, otherwise
```

modules/apps/transfer/client/cli/tx.go
```diff
@@ -85,24 +86,51 @@ corresponding to the counterparty channel. Any timeout set to 0 is disabled.`),
...
+					return err
+				}
+
+				clientHeight, ok := clientState.GetLatestHeight().(clienttypes.Height)
+				if !ok {
+					return fmt.Errorf("invalid height type. expected type: %T, got: %T", clienttypes.Height{}, clientState.GetLatestHeight())
+				}
+
+				var consensusState exported.ConsensusState
+				if clientState.ClientType() != exported.Localhost {
```
colin-axner, line 105: I wonder if this command works for solo machines?

modules/core/04-channel/keeper/packet.go
damiannolan: Yup, the fix for the CLI is intended for `v7.1.0` which will require a co-ordinated upgrade anyway. But I agree with pulling out this error change to a separate PR 👍

modules/apps/transfer/client/cli/tx.go
damiannolan: > I wonder if this command works for solo machines?

I think it should. 

> Could we avoid checking explicitly for Localhost here? 

We could and just use local clock time as the standard. But I'm happy to defer that til later, I'll open an issue for it. 
But I think we should just proceed with this approach and backport for 7.1. I don't see many future client implementations not having the concept of consensus states, I think localhost is an outlier

modules/apps/transfer/client/cli/tx.go
damiannolan: nice catch, I've updated to make use of `clientutils.QueryConsensusState` using `clientRes.ClientID` as you've suggested

modules/core/04-channel/keeper/packet.go
damiannolan: https://github.com/cosmos/ibc-go/pull/3593

charleenfei (approved):
nice work and catch on this issue!

crodriguezvega (approved):

modules/apps/transfer/client/cli/tx.go
colin-axner: Solo machines don't set consensus states so it shouldn't work? This appears to be true already so happy to see it fixed in #3594

PR #3620, opened 2023-05-19
nit: remove connection hops check in`ChannelOpenTry`

crodriguezvega:
## Description

Isn't this check redundant since the same check [is already done in `ValidateBasic` of `Channel`](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/channel.go#L62) when [validating `MsgChannelOpenTry`](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/msgs.go#L122)?

closes: #XXXX

```bash
N/A
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/handshake.go
+++ b/modules/core/04-channel/keeper/handshake.go
@@ -113,11 +113,6 @@ func (k Keeper) ChanOpenTry(
 	proofInit []byte,
 	proofHeight exported.Height,
 ) (string, *capabilitytypes.Capability, error) {
-	// connection hops only supports a single connection
-	if len(connectionHops) != 1 {
-		return "", nil, errorsmod.Wrapf(types.ErrTooManyConnectionHops, "expected 1, got %d", len(connectionHops))
-	}
-
 	// generate a new channel
 	channelID := k.GenerateChannelIdentifier(ctx)
 


damiannolan (commented):
I agree! This check is redundant, but I think it may be worth keeping until we make the handshake handlers private/unexported. I suppose the only concern is some module importing ibc and calling this method directly (bypassing `ValidateBaisc`) with empty or nil connection hops and causing a panic - as we assume its safe to access `connectionHops[0]` most of the time.

The chances of the above are slim in reality to be honest, and somebody importing the channel keeper and calling this directly doesn't seem like a valid use-case off the top of my head.

crodriguezvega:
> I suppose the only concern is some module importing ibc and calling this method directly (bypassing ValidateBaisc) with empty or nil connection hops and causing a panic

True, I also thought about that possibility, but thought that if that was to happen then I guess we would need to validate also other things besides `connectionHops`...

colin-axner:
It'd opt to leave this for now. I think we can spend some time removing redundant checks when we restructure our validation. I think the benefit of leaving this check is that it is easier to compare the handler against the spec. If there's strong desire to remove this now, a comment should be added indicating where the check is performed

Issue #3500, opened 2023-04-22
Migrate `02-client` params to be self managed


crodriguezvega:
Part of https://github.com/cosmos/ibc-go/issues/2010. Please read that issue for more information.

PR #3640, opened 2023-05-24
feat(client): migrate client params

DimitrisJim:
## Description

closes: #3500 

```bash
feat(client): migrate client params
```

---

write a little note why.


modules/core/keeper/keeper.go
```diff
@@ -61,7 +63,9 @@ func NewKeeper(
 		panic(fmt.Errorf("cannot initialize IBC keeper: empty scoped keeper"))
 	}
 
-	clientKeeper := clientkeeper.NewKeeper(cdc, key, paramSpace, stakingKeeper, upgradeKeeper)
+	authority := authtypes.NewModuleAddress(govtypes.ModuleName).String()
```
DimitrisJim, line 66: q: any reason why this shouldn't be done here? This results in no changes to `app.go` but I'm not completely sure if there's some gotcha.

modules/core/02-client/keeper/migrations.go
```diff
@@ -31,3 +32,17 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
 func (m Migrator) Migrate3to4(ctx sdk.Context) error {
 	return v7.MigrateLocalhostClient(ctx, m.keeper)
 }
+
+// MigrateParams migrates from consensus version 4 to 5.
+// This migration takes the parameters that are currently stored and managed by x/params
+// and stores them directly in the ibc module's state.
+func (m Migrator) MigrateParams(ctx sdk.Context) error {
```
DimitrisJim, line 39: Talked a bit with Carlos and he mentioned structuring this migration as we do for `v7`. Though it is doable, it requires a separate `expected_keepers.go` and some restructuring in order to pass the `legacySubspace` to the migration.

Thoughts?

modules/core/keeper/keeper.go
srdtrk: I think the gotcha is that this won't register if the authority address changes by governance vote. Also, this is generating a new address and not using already existing address.

modules/core/02-client/keeper/migrations.go
srdtrk: I don't see why you need a separate `expected_keepers.go`

modules/core/02-client/exported/exported.go
srdtrk, line 1: We don't want (since we don't need) an `exported.go`, this was discussed in the other migration PRs. Shouldn't be hard to remove as we kept the subspace in the keeper anyway

modules/core/02-client/types/params_legacy.go
```diff
@@ -0,0 +1,31 @@
+package types
```
srdtrk, line 1:
```suggestion
/*
NOTE: Usage of x/params to manage parameters is deprecated in favor of x/gov
controlled execution of MsgUpdateParams messages. These types remains solely
for migration purposes and will be removed in a future release.
[#3621](https://github.com/cosmos/ibc-go/issues/3621)
*/

package types
```

Add tracking issue for removal of these types.

modules/core/02-client/genesis.go
srdtrk, line 1: maybe we can add a genesis test? for codecov

modules/core/02-client/types/keys.go
```diff
@@ -25,6 +25,9 @@ const (
 	// KeyNextClientSequence is the key used to store the next client sequence in
 	// the keeper.
 	KeyNextClientSequence = "nextClientSequence"
+
+	// ParamsKey is the store key for the IBC client parameters
+	ParamsKey = "clientParams"
```
srdtrk, line 30: nit: in other PRs, this is called `"params"`. So maybe we should keep doing this?

modules/core/keeper/keeper.go
damiannolan: I guess the authority field should be passed to the ibc core keeper in app.go and propagated to submodules

modules/core/02-client/keeper/params.go
```diff
@@ -6,19 +6,28 @@
...
+	return p.AllowedClients
 }
 
 // GetParams returns the total set of ibc-client parameters.
 func (k Keeper) GetParams(ctx sdk.Context) types.Params {
-	return types.NewParams(k.GetAllowedClients(ctx)...)
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get([]byte(types.ParamsKey))
+	if bz == nil {
+		panic("ibc client params are not set in store")
```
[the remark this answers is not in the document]
damiannolan: This is called in BeginBlocker when updating the localhost client. Personally I'd rather avoid panics entirely in these situations.

damiannolan (commented):
Nice work! Left some comments/suggestions. Looks pretty clean so far!

modules/core/02-client/types/keys.go
damiannolan: IBC submodules share the store space (all use the core IBC store), so this would cause collisions with connection params etc

modules/core/keeper/msg_server.go
```diff
@@ -697,3 +698,16 @@ func (k Keeper) Acknowledgement(goCtx context.Context, msg *channeltypes.MsgAckn
...
+
+// UpdateClientParams defines a rpc handler method for MsgUpdateClientParams.
+func (k Keeper) UpdateClientParams(goCtx context.Context, msg *clienttypes.MsgUpdateClientParams) (*clienttypes.MsgUpdateClientParamsResponse, error) {
+	ck := k.ClientKeeper
+	if ck.GetAuthority() != msg.Authority {
+		return nil, errorsmod.Wrapf(ibcerrors.ErrUnauthorized, "expected %s, got %s", ck.GetAuthority(), msg.Authority)
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	ck.SetParams(ctx, msg.Params)
```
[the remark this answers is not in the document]
srdtrk: `Validate()` is called in `ValidateBasic` instead. (Another reason it should go XD)

modules/core/keeper/msg_server.go
```diff
@@ -697,3 +698,16 @@ func (k Keeper) Acknowledgement(goCtx context.Context, msg *channeltypes.MsgAckn
 
 	return &channeltypes.MsgAcknowledgementResponse{Result: channeltypes.SUCCESS}, nil
 }
+
+// UpdateClientParams defines a rpc handler method for MsgUpdateClientParams.
+func (k Keeper) UpdateClientParams(goCtx context.Context, msg *clienttypes.MsgUpdateClientParams) (*clienttypes.MsgUpdateClientParamsResponse, error) {
+	ck := k.ClientKeeper
```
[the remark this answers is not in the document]
DimitrisJim: definitely can. Don't have strong opinions either way.

modules/core/module.go
```diff
@@ -137,6 +137,10 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	}); err != nil {
 		panic(err)
 	}
+
+	if err := cfg.RegisterMigration(exported.ModuleName, 4, clientMigrator.MigrateParams); err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
DimitrisJim: yup, that's the plan. connection params will be rebased on main when client is added.

modules/core/02-client/keeper/params.go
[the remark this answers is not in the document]
DimitrisJim: Gotcha, will switch to err. Out of curiosity, what's the difference in semantics between handling an error/panic in begin blocker?

DimitrisJim:
re: And a docs entry is needed in docs/migrations/ just a reference to this issue mentioning how app.go needs to be modified.

I saw that mentioned by Colin and was wondering if just adding a single note for all modules that params have been migrated and an example of what needs to change would suffice. Seems like having a diff for all 5(?) modules would be a lot of unecessary duplication. Thoughts?

modules/core/02-client/genesis.go
DimitrisJim: iirc its already tested in `genesis_test.go`.

modules/core/02-client/types/params_test.go
```diff
@@ -8,6 +8,24 @@ import (
...
+	testCases := []struct {
+		name       string
+		clientType string
+		params     Params
+		expPass    bool
+	}{
+		{"valid client", exported.Tendermint, DefaultParams(), true},
+		{"valid client with custom params", exported.Tendermint, NewParams(exported.Tendermint), true},
+		{"invalid blank client", " ", DefaultParams(), false},
+		{"invalid client with custom params", exported.Localhost, NewParams(exported.Tendermint), false},
```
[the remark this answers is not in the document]
DimitrisJim: ah, it got away from me. nice catch.

modules/core/keeper/msg_server.go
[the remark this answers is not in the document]
DimitrisJim: oh wait, no. We still need to access `GetAuthority` in two places so this shorthand makes sense.

proto/ibc/core/client/v1/tx.proto
```diff
@@ -4,6 +4,9 @@ package ibc.core.client.v1;
 
 option go_package = "github.com/cosmos/ibc-go/v7/modules/core/02-client/types";
 
+import "cosmos_proto/cosmos.proto";
+import "cosmos/msg/v1/msg.proto";
+import "ibc/core/client/v1/client.proto";
 import "gogoproto/gogo.proto";
 import "google/protobuf/any.proto";
```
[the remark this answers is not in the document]
damiannolan: No worries, we can always look into this for a future PR. I agree, there isn't much consistency to import ordering in our protobuf files so all g! 👍

modules/core/02-client/keeper/params.go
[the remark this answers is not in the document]
damiannolan: Is empty params what the sdk does? I know @colin-axner mentioned briefly that he had a minor preference for panicking, but it would be nice if we could avoid it imo. 

For context, panicking in the BeginBlocker abci call is not caught by sdk's baseapp. Panicking in msg server handlers is "fine" because its caught by the grpc recovery interceptor, but doing so in BeginBlocker would crash nodes and ultimately halt a chain

proto/ibc/core/client/v1/tx.proto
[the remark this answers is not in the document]
DimitrisJim: I'll open an issue w a diff to show what buf basically thinks is cool when formatting proto files. Just to keep track of this.

modules/core/02-client/keeper/params.go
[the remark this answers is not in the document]
crodriguezvega: Ah, good point, @damiannolan! I believe the [SDK does not panic](https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/mint/keeper/keeper.go#L100).

modules/core/keeper/keeper.go
damiannolan: Yep, address generation is deterministic

modules/core/keeper/msg_server.go
```diff
@@ -697,3 +698,16 @@ func (k Keeper) Acknowledgement(goCtx context.Context, msg *channeltypes.MsgAckn
...
+	if ck.GetAuthority() != msg.Authority {
```
damiannolan, line 705: since the entrypoint(s) to the msg server for ibc core submodules is in the core keeper, it might make sense for `GetAuthority` to just exist at the level of the core ibc keeper rather than passing the authority along to each nested keeper (client, connection). 

I don't see a world right now in which client would have authority x and connection would have authority y.

Thoughts? cc. @colin-axner

modules/core/02-client/keeper/params.go
```diff
@@ -6,19 +6,28 @@ import (
...
-	return types.NewParams(k.GetAllowedClients(ctx)...)
+	var p types.Params
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get([]byte(types.ParamsKey))
+	if bz == nil {
+		return p
+	}
+
+	k.cdc.MustUnmarshal(bz, &p)
+	return p
```
damiannolan, line 25: nit: I have a mild preference for consistency with other getter funcs

```suggestion
	store := ctx.KVStore(k.storeKey)
	bz := store.Get([]byte(types.ParamsKey))
	if bz == nil {
		return p
	}

        var params types.Params
	k.cdc.MustUnmarshal(bz, &params)
	return params
```

modules/core/02-client/keeper/params.go
damiannolan, line 1: seems like the trend so far was to remove the params.go file and move to a single `GetParams` method within keeper.go. 

And changing references of `GetAllowedClients` to `GetParams(ctx).AllowedClients`

modules/core/02-client/keeper/params.go
damiannolan: I also don't really like named return args! 

It is, I thought I edited it before anyone saw 🙈

modules/core/02-client/keeper/params.go
damiannolan: Personally I would've probably kept params.go and helpers such as `GetAllowedClients` / `GetSendEnabled` for fields which exist on params but I don't feel super strongly about it tbh.

modules/core/keeper/msg_server.go
DimitrisJim: that's seems like a good point to me. I'll wait for Colin's thoughts on this before making changes.

DimitrisJim: I've made this change. Code definitely looks better. Will revert if required.

modules/core/keeper/keeper.go
```diff
@@ -33,13 +33,17 @@ type Keeper struct {
 	ChannelKeeper    channelkeeper.Keeper
 	PortKeeper       portkeeper.Keeper
 	Router           *porttypes.Router
+
+	// the address capable of executing a MsgUpdateClientParams message. Typically, this
+	// should be the x/gov module account.
+	authority string
```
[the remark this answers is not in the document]
DimitrisJim: completely fine w removal, iirc that's the way I initially had it (albeit, not for the same reason)

modules/core/02-client/keeper/params.go
damiannolan: could we apply this nit? i.e. move the var declaration below and rename `p` to `params`

DimitrisJim: done, flew under the radar for some reason

proto/ibc/core/client/v1/tx.proto
```diff
@@ -98,3 +104,19 @@ message MsgSubmitMisbehaviour {
 // MsgSubmitMisbehaviourResponse defines the Msg/SubmitMisbehaviour response
 // type.
 message MsgSubmitMisbehaviourResponse {}
+
+// MsgUpdateClientParams defines the sdk.Msg type to update the client parameters.
+message MsgUpdateClientParams {
+  option (cosmos.msg.v1.signer) = "authority";
```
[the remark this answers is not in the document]
DimitrisJim: brought this back as per Carlos's comment https://github.com/cosmos/ibc-go/pull/3627#issuecomment-1567952346

proto/ibc/core/client/v1/tx.proto
```diff
@@ -98,3 +104,19 @@ message MsgSubmitMisbehaviour {
...
+
+  // authority is the address of the governance account.
+  string authority = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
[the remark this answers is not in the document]
DimitrisJim: Removed this for now and opened https://github.com/cosmos/ibc-go/issues/3694 to keep track of it.

modules/core/02-client/keeper/keeper.go
```diff
@@ -413,3 +413,23 @@ func (k Keeper) GetClientStatus(ctx sdk.Context, clientState exported.ClientStat
 	}
 	return clientState.Status(ctx, k.ClientStore(ctx, clientID), k.cdc)
 }
+
+// GetParams returns the total set of ibc-client parameters.
+func (k Keeper) GetParams(ctx sdk.Context) types.Params {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get([]byte(types.ParamsKey))
+	if bz == nil {
```
damiannolan, line 421: mega nit:

```suggestion
	if len(bz) == 0 {
```

modules/core/keeper/keeper.go
```diff
@@ -92,6 +95,11 @@ func (k *Keeper) SetRouter(rtr *porttypes.Router) {
 	k.Router.Seal()
 }
 
+// GetAuthority returns the client module's authority.
```
damiannolan, line 98:
```suggestion
// GetAuthority returns the client submodule's authority.
```

damiannolan (approved):
Awesome work, thanks @DimitrisJim 🚀 

I think we should go ahead and add the migration doc entry in this PR.

modules/core/02-client/keeper/keeper.go
srdtrk: In the other PRs, we decided to use `bz == nil` since we also decided to panic. But I suppose this is ok here

modules/core/keeper/msg_server_test.go
```diff
@@ -774,3 +774,52 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest()
+			_, err := keeper.Keeper.UpdateClientParams(*suite.chainA.App.GetIBCKeeper(), suite.chainA.GetContext(), tc.msg)
+			if tc.expPass {
+				suite.Require().NoError(err)
```
srdtrk, line 819: This is something I omitted in my early migration PR too, but let's add a check here that the params have actually changed

srdtrk (approved):
Great, lgtm

chatton (approved):
LGTM nice job!

modules/core/02-client/keeper/keeper.go
```diff
@@ -413,3 +413,23 @@ func (k Keeper) GetClientStatus(ctx sdk.Context, clientState exported.ClientStat
...
 	return clientState.Status(ctx, k.ClientStore(ctx, clientID), k.cdc)
 }
+
+// GetParams returns the total set of ibc-client parameters.
+func (k Keeper) GetParams(ctx sdk.Context) types.Params {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get([]byte(types.ParamsKey))
+	if len(bz) == 0 {
+		return types.Params{}
+	}
```
colin-axner, line 423: we should panic on unset params? Similar to how the other param prs do

DimitrisJim: there's been quite some back and forth here w/o much consensus. See some discussions on connections params https://github.com/cosmos/ibc-go/pull/3650#discussion_r1210398066 and here in this pr https://github.com/cosmos/ibc-go/pull/3640#discussion_r1204210620.

modules/core/02-client/keeper/params.go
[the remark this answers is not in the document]
colin-axner: My view is that unset params are a sign of a misconfigured chain which should be halted. It seems unsafe to process any transactions if certain parameters on chain haven't been decided on. My recommendation would be defense in depth, as we may choose to add params in the future which might not have a similar sane default value as allowed clients

modules/core/02-client/keeper/keeper.go
DimitrisJim: done https://github.com/cosmos/ibc-go/pull/3721

Issue #3648, opened 2023-05-24
Implement startFlushUpgradeHandshake


colin-axner:
## Summary

Implement the spec subprotocol `startFlushUpgradeHandshake` as outlined in [#967](https://github.com/cosmos/ibc/pull/967)

Do not include checks for pending in flight packets. Note: the current structure in the spec is subject to change

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3671, opened 2023-05-26
startFlushUpgradeHandshake

charleenfei:
## Description

closes: #3648

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -1,11 +1,15 @@
 package keeper
 
 import (
+	"reflect"
+
 	errorsmod "cosmossdk.io/errors"
 	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
+	"github.com/cosmos/ibc-go/v7/internal/collections"
 	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
+	connectiontypes "github.com/cosmos/ibc-go/v7/modules/core/03-connection/types"
 	"github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
 )
 
@@ -96,4 +100,114 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
 		Timeout:            upgradeTimeout,
 		LatestSequenceSend: seq - 1,
 	}, nil
+}
+
+// startFlushUpgradeSequence will verify that the channel is in a valid precondition for calling the startFlushUpgradeHandshake
+// and that the desiredChannelState is valid
+// it will verify the proofs of the counterparty channel and upgrade
+// it will verify that the upgrades on both ends are mutually compatible
+// it will set the channel to desiredChannel state and move to flushing mode
+// if flush is already complete, it will automatically set flushStatus to FLUSHCOMPLETE
+//
+//lint:ignore U1000 Ignore unused function temporarily for debugging
+func (k Keeper) startFlushUpgradeHandshake(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+	proposedUpgradeFields types.UpgradeFields,
+	counterpartyChannel types.Channel,
+	counterpartyUpgrade types.Upgrade,
+	desiredChannelState types.State,
+	// TODO: add flush state here when enum is present
+	proofCounterpartyChannel,
+	proofUpgrade []byte,
+	proofHeight clienttypes.Height,
+) error {
+
+	if !collections.Contains(desiredChannelState, []types.State{types.TRYUPGRADE, types.ACKUPGRADE}) {
+		return errorsmod.Wrapf(types.ErrInvalidChannelState, "expected one of [%s, %s], got %s", types.TRYUPGRADE, types.ACKUPGRADE, desiredChannelState)
+	}
+
+	channel, found := k.GetChannel(ctx, portID, channelID)
+	if !found {
+		return errorsmod.Wrapf(types.ErrChannelNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	connection, err := k.GetConnection(ctx, channel.ConnectionHops[0])
+	if err != nil {
+		return errorsmod.Wrap(err, "failed to retrieve connection using the channel connection hops")
+	}
+
+	// make sure connection is OPEN
+	if connection.GetState() != int32(connectiontypes.OPEN) {
+		return errorsmod.Wrapf(
+			connectiontypes.ErrInvalidConnectionState,
+			"connection state is not OPEN (got %s)", connectiontypes.State(connection.GetState()).String(),
+		)
+	}
+
+	// verify the counterparty channel state containing the upgrade sequence
+	if err := k.connectionKeeper.VerifyChannelState(
+		ctx,
+		connection,
+		proofHeight,
+		proofCounterpartyChannel,
+		channel.Counterparty.PortId,
+		channel.Counterparty.ChannelId,
+		counterpartyChannel,
+	); err != nil {
+		return err
+	}
+
+	// verifies the proof that a particular proposed upgrade has been stored in the upgrade path of the counterparty
+	if err := k.connectionKeeper.VerifyChannelUpgrade(ctx, connection, proofHeight, proofUpgrade, channel.Counterparty.PortId,
+		channel.Counterparty.ChannelId, counterpartyUpgrade); err != nil {
+		return err
+	}
+
+	if !reflect.DeepEqual(proposedUpgradeFields, counterpartyUpgrade.Fields) {
+		k.restoreChannel(portID, channelID)
+	}
+
+	// connectionHops can change in a channelUpgrade, however both sides must still be each other's counterparty.
+	proposedConnection, found := k.connectionKeeper.GetConnection(ctx, proposedUpgradeFields.ConnectionHops[0])
+	if !found {
+		k.restoreChannel(portID, channelID)
+	}
+
+	if proposedConnection.GetState() != int32(connectiontypes.OPEN) {
+		k.restoreChannel(portID, channelID)
+	}
+
+	if counterpartyUpgrade.Fields.ConnectionHops[0] != proposedConnection.Counterparty.ConnectionId {
+		k.restoreChannel(portID, channelID)
+	}
+
+	// set the channel to the desired state
+	channel.State = desiredChannelState
+	// TODO: set flushstate
+
+	// if k.pendingInflightPackets(portID, channelID) == nil {
+	// if there are no packets in flight, then flush is complete
+	// TODO: set flushstate to FLUSHCOMPLETE
+	// }
+
+	k.SetChannel(ctx, portID, channelID, channel)
+	// k.setChannelCounterpartyLastPacketSequenceSend
+
+	return nil
+}
+
+// restoreChannel will write an error receipt, set the channel back to its original state and
+// delete upgrade information when the executing channel needs to abort the upgrade handshake and return to the original parameters.
+func (k Keeper) restoreChannel(portID, channelID string) {
+	// TODO
+}
+
+// pendingInflightPackets returns the packet sequences sent on this end that have not had their lifecycle completed
+//
+//lint:ignore U1000 Ignore unused function temporarily for debugging
+func (k Keeper) pendingInflightPackets(portID, channelID string) []uint64 {
+	// TODO
+	return []uint64{}
 }
--- a/modules/core/04-channel/types/expected_keepers.go
+++ b/modules/core/04-channel/types/expected_keepers.go
@@ -71,6 +71,15 @@ type ConnectionKeeper interface {
 		channelID string,
 		nextSequenceRecv uint64,
 	) error
+	VerifyChannelUpgrade(
+		ctx sdk.Context,
+		connection exported.ConnectionI,
+		height exported.Height,
+		proof []byte,
+		portID,
+		channelID string,
+		upgrade Upgrade,
+	) error
 }
 
 // PortKeeper expected account IBC port keeper


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -104,3 +107,104 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+		// error on the higher sequence so that both chains synchronize on a fresh sequence
+		channel.UpgradeSequence = math.Max(counterpartyChannel.UpgradeSequence, channel.UpgradeSequence)
+		k.SetChannel(ctx, portID, channelID, channel)
+
+		return types.NewUpgradeError(channel.UpgradeSequence, errorsmod.Wrapf(
+			types.ErrIncompatibleCounterpartyUpgrade, "expected upgrade sequence (%d) to match counterparty upgrade sequence (%d)", channel.UpgradeSequence, counterpartyChannel.UpgradeSequence),
+		)
+	}
+
+	// assert that both sides propose the same channel ordering
```
colin-axner, line 178: if we think we might need to add more checks in the future, we could consider making a helper function for all the below checks to validate the counterparty upgrade against the self upgrade, maybe something like:
```go
    err := k.checkUpgradesForCompatibility(proposedUpgrade, counterpartyUpgrade)
    if err != nil {
        return types.NewUpgradeError(channel.UpgradeSequence, err)
    }
```

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -124,4 +128,180 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest()
+
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+			suite.coordinator.Setup(path)
+
+			path.EndpointA.ChannelConfig.ProposedUpgrade.Fields.Version = fmt.Sprintf("%s-v2", mock.Version)
+			err := path.EndpointA.ChanUpgradeInit()
+			suite.Require().NoError(err)
```
colin-axner, line 258: will open a separate pr with these changes
```diff
diff --git a/testing/config.go b/testing/config.go
index 9a6ace904..90353f3bc 100644
--- a/testing/config.go
+++ b/testing/config.go
@@ -47,9 +47,10 @@ func NewConnectionConfig() *ConnectionConfig {
 }
 
 type ChannelConfig struct {
-       PortID  string
-       Version string
-       Order   channeltypes.Order
+       PortID          string
+       Version         string
+       Order           channeltypes.Order
+       ProposedUpgrade channeltypes.Upgrade
 }
 
 func NewChannelConfig() *ChannelConfig {
diff --git a/testing/endpoint.go b/testing/endpoint.go
index fb278c241..2c82c1581 100644
--- a/testing/endpoint.go
+++ b/testing/endpoint.go
@@ -564,12 +564,30 @@ func (endpoint *Endpoint) TimeoutOnClose(packet channeltypes.Packet) error {
        return endpoint.Chain.sendMsgs(timeoutOnCloseMsg)
 }
 
-func (endpoint *Endpoint) ChanUpgradeInit(timeout channeltypes.Timeout) error {
+func (endpoint *Endpoint) ChanUpgradeInit() error {
+       timeout := endpoint.ChannelConfig.ProposedUpgrade.Timeout
+       if !timeout.IsValid() {
+               timeout = channeltypes.NewUpgradeTimeout(endpoint.Counterparty.Chain.GetTimeoutHeight(), 0)
+       }
+
+       upgradeFields := endpoint.ChannelConfig.ProposedUpgrade.Fields
+       if upgradeFields.Ordering == channeltypes.NONE {
+               upgradeFields.Ordering = endpoint.ChannelConfig.Order
+       }
+
+       if upgradeFields.Version == "" {
+               upgradeFields.Version = endpoint.ChannelConfig.Version
+       }
+
+       if len(upgradeFields.ConnectionHops) == 0 {
+               upgradeFields.ConnectionHops = []string{endpoint.ConnectionID}
+       }
+
        msg := channeltypes.NewMsgChannelUpgradeInit(
                endpoint.ChannelConfig.PortID,
                endpoint.ChannelID,
-               channeltypes.NewUpgradeFields(endpoint.ChannelConfig.Order, []string{endpoint.ConnectionID}, endpoint.ChannelConfig.Version),
-               channeltypes.NewUpgradeTimeout(timeout.Height, timeout.Timestamp),
+               upgradeFields,
+               timeout,
                endpoint.Chain.SenderAccount.GetAddress().String(),
        )
```

colin-axner: To clean this code up further, we can add a helper func on endpoint to get the proposed upgrade from the config

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -104,3 +107,104 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+		counterpartyUpgrade,
+	); err != nil {
+		return errorsmod.Wrap(err, "failed to verify counterparty upgrade")
+	}
+
+	// the current upgrade handshake must only continue if both channels are using the same upgrade sequence,
+	// otherwise an error receipt must be written so that the upgrade handshake may be attempted again with synchronized sequences
+	if counterpartyChannel.UpgradeSequence != channel.UpgradeSequence {
+		// error on the higher sequence so that both chains synchronize on a fresh sequence
+		channel.UpgradeSequence = math.Max(counterpartyChannel.UpgradeSequence, channel.UpgradeSequence)
```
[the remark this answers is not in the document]
damiannolan: ohhh! that makes sense, maybe we could use `sdkmath`, I think that's how it is in some other places iirc

modules/core/04-channel/keeper/upgrade_test.go
colin-axner: implemented in #3719

modules/core/04-channel/keeper/export_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
+	"github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+)
+
+// StartFlushUpgradeHandshake is an exported version of startFlushUpgradeHandshake to be used for unit testing
+func (k Keeper) StartFlushUpgradeHandshake(
```
colin-axner, line 11: ongoing discussion on this construction [here](https://github.com/cosmos/ibc-go/pull/3687#discussion_r1209408650), but needed to run unit tests and keep this function private

modules/core/04-channel/keeper/upgrade.go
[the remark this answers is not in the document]
colin-axner: opened #3720, a quick grep tells me we don't alias it at all rn

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -104,3 +106,107 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+//lint:ignore U1000 Ignore unused function temporarily for debugging
+func (k Keeper) startFlushUpgradeHandshake(
+	ctx sdk.Context,
+	portID,
+	channelID string,
+	proposedUpgradeFields types.UpgradeFields,
+	counterpartyChannel types.Channel,
+	counterpartyUpgrade types.Upgrade,
+	proofCounterpartyChannel,
+	proofUpgrade []byte,
```
colin-axner, line 124: we should align naming here. Do we want:
```go
proofChannel, 
proofUpgrade, 
```

or 
```
proofCounterpartyChannel, 
proofCounterpartyUpgrade, 
```
?

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -124,4 +128,180 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+			commitmenttypes.ErrInvalidProof,
+		},
+		{
+			"upgrade sequence mismatch, endpointB channel upgrade sequence is ahead",
+			func() {
+				channel := path.EndpointB.GetChannel()
+				channel.UpgradeSequence++
+				path.EndpointB.SetChannel(channel)
+			},
+			types.ErrIncompatibleCounterpartyUpgrade,
```
colin-axner, line 200: I'm not sure yet if there's a way we could return an `NewUpgradeError` here with the `ErrIncompatibleCounterpartyUpgrade` embedded and then do a check that the returned error is an upgrade error and contains the same parent err

modules/core/04-channel/keeper/upgrade.go
charleenfei: I prefer `proofCounterpartyChannel, proofCounterpartyUpgrade` as it is more explicit and because we do also have our own `proposedUpgrade` in the mix

colin-axner: sounds good to me. Will update

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -126,6 +129,185 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+
+			// ensure proof verification succeeds
+			err = path.EndpointB.UpdateClient()
+			suite.Require().NoError(err)
+
+			// same construction used by endpoint.ChanUpgradeTry
+			channelKey := host.ChannelKey(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+			proofChannel, proofHeight := path.EndpointA.Chain.QueryProof(channelKey)
+			upgradeKey := host.ChannelUpgradeKey(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+			proofUpgrade, _ := path.EndpointA.Chain.QueryProof(upgradeKey)
```
colin-axner, line 270: #3722

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -109,6 +110,110 @@ func (k Keeper) WriteUpgradeTryChannel(
...
+	if proposedUpgradeFields.Ordering != counterpartyUpgrade.Fields.Ordering {
+		return types.NewUpgradeError(channel.UpgradeSequence, errorsmod.Wrapf(
+			types.ErrIncompatibleCounterpartyUpgrade, "expected upgrade ordering (%s) to match counterparty upgrade ordering (%d)", proposedUpgradeFields.Ordering, counterpartyUpgrade.Fields.Ordering),
+		)
+	}
+
+	proposedConnection, err := k.GetConnection(ctx, proposedUpgradeFields.ConnectionHops[0])
+	if err != nil {
+		// NOTE: this error is expected to be unreachable as the proposed upgrade connectionID should have been
+		// validated in the upgrade INIT and TRY handlers
```
crodriguezvega, line 193: What's the reason to validate in both places? The spec seems to be doing the validation only in this function and not in the handlers, right? If we keep validation in the handlers then behaviour of the implementation deviates from the spec because the implementation will fail the transaction, but not write the error receipt, right?

charleenfei: we need to use the proposedConnection in the handlers to construct the `UpgradeFields` so they should be verified there at least.

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -124,4 +128,183 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+		upgrade             types.Upgrade
+		counterpartyChannel types.Channel
+		counterpartyUpgrade types.Upgrade
+	)
+
+	testCases := []struct {
+		name     string
+		malleate func()
+		// TODO: currently the code asserts based on an expected sdk error
+		// Allow support for asserting either the sdk error type or that the error is an upgrade error
```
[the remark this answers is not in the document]
damiannolan: Isn't `UpgradeError` a valid `error` implementation. Could we just add `types.NewUpgradeError` as `expError`?

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -126,6 +131,185 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+			"upgrade ordering is not the same on both sides",
+			func() {
+				upgrade.Fields.Ordering = types.ORDERED
+			},
+			types.ErrIncompatibleCounterpartyUpgrade,
+		},
+		{
+			"proposed connection is not found",
+			func() {
+				upgrade.Fields.ConnectionHops[0] = invalidConnection
```
damiannolan, line 215: nit: I think we have `ibctesting.InvalidID` or something, maybe we could use it here?

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -126,6 +131,185 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+
+	for _, tc := range testCases {
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest()
+
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+			suite.coordinator.Setup(path)
+
+			upgradeVersion := fmt.Sprintf("%s-v2", mock.Version)
```
damiannolan, line 259: could be worth adding another const to the mock pkg for this since its probably reused quite a bit. maybe `mock.UpgradeVersion` or something similar?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -109,6 +110,110 @@ func (k Keeper) WriteUpgradeTryChannel(
...
+
+		return types.NewUpgradeError(channel.UpgradeSequence, errorsmod.Wrapf(
+			types.ErrIncompatibleCounterpartyUpgrade, "expected upgrade sequence (%d) to match counterparty upgrade sequence (%d)", prevUpgradeSequence, counterpartyChannel.UpgradeSequence),
+		)
+	}
+
+	// assert that both sides propose the same channel ordering
+	if proposedUpgradeFields.Ordering != counterpartyUpgrade.Fields.Ordering {
+		return types.NewUpgradeError(channel.UpgradeSequence, errorsmod.Wrapf(
+			types.ErrIncompatibleCounterpartyUpgrade, "expected upgrade ordering (%s) to match counterparty upgrade ordering (%d)", proposedUpgradeFields.Ordering, counterpartyUpgrade.Fields.Ordering),
```
damiannolan, line 186: I think if %d is used here the enums int32 value will be formatted into the string, right?

```suggestion
			types.ErrIncompatibleCounterpartyUpgrade, "expected upgrade ordering (%s) to match counterparty upgrade ordering (%s)", proposedUpgradeFields.Ordering, counterpartyUpgrade.Fields.Ordering),
```

modules/core/04-channel/types/errors.go
```diff
@@ -37,13 +37,14 @@ var (
...
-	ErrInvalidUpgradeTimeout   = errorsmod.Register(SubModuleName, 27, "either timeout height or timeout timestamp must be non-zero")
-	ErrUpgradeSequenceNotFound = errorsmod.Register(SubModuleName, 28, "upgrade sequence not found")
-	ErrUpgradeErrorNotFound    = errorsmod.Register(SubModuleName, 29, "upgrade error receipt not found")
-	ErrInvalidUpgrade          = errorsmod.Register(SubModuleName, 30, "invalid upgrade")
-	ErrInvalidUpgradeSequence  = errorsmod.Register(SubModuleName, 31, "invalid upgrade sequence")
-	ErrUpgradeNotFound         = errorsmod.Register(SubModuleName, 32, "upgrade not found")
+	ErrInvalidChannelVersion           = errorsmod.Register(SubModuleName, 24, "invalid channel version")
+	ErrPacketNotSent                   = errorsmod.Register(SubModuleName, 25, "packet has not been sent")
+	ErrInvalidTimeout                  = errorsmod.Register(SubModuleName, 26, "invalid packet timeout")
+	ErrInvalidUpgradeTimeout           = errorsmod.Register(SubModuleName, 27, "either timeout height or timeout timestamp must be non-zero")
```
damiannolan, line 43: Unrelated to your changes, but we could maybe remove this error now? Or it could be useful to keep and just change the string to `"invalid upgrade timeout"` as `"invalid packet timeout"` is used above for `ErrInvalidTimeout`

modules/core/04-channel/types/errors.go
```diff
@@ -37,13 +37,14 @@ var (
...
+	ErrUpgradeSequenceNotFound         = errorsmod.Register(SubModuleName, 28, "upgrade sequence not found")
```
damiannolan, line 44: maybe this one could be removed as the upgrade sequence now exists on the channel? I don't think it can be not found.

we can address all of these errors in a follow up housekeeping PR

modules/core/04-channel/keeper/upgrade_test.go
damiannolan: Nice, agree this would improve readability and clutter of the test body :)

damiannolan (approved):
Awesome work! Left some minor comments, nothing major. Happy to move forward with this and iterate on it as things develop! I assume the in-flight packets checks and its associated funcs will be added separately later on. 

Thanks @charleenfei and @colin-axner ❤️

modules/core/04-channel/keeper/upgrade_test.go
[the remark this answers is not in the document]
colin-axner: Yes, but the way `upgradeError.Is()` is implemented, it disregards the actual `UpgradeError` type, so I believe passing in `channeltypes.ErrChannelNotFound` would have the same result as passing in `channeltypes.NewUpgradeError(0, channeltypes.ErrChannelNotFound`. Ideally, the former would not succeed if an upgrade error is returned, I guess we could do an assertion:
```go
if ok := tc.expError.(channeltypes.UpgradeError); ok {
    suite.Require().True(err.(channeltypes.UpgradeError))
}
```

modules/core/04-channel/keeper/upgrade.go
colin-axner: The spec does this check [twice](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#utility-functions):
```go

// initUpgradeHandshake 

...

// proposedConnection must exist and be in OPEN state for 
// channel upgrade to be accepted
proposedConnection = provableStore.Get(connectionPath(proposedUpgradeFields.connectionHops[0])
abortTransactionUnless(proposedConnection != null && proposedConnection.state == OPEN)
``

and 

```
```go
// startFlushUpgradeHandshake 

...

// connectionHops can change in a channelUpgrade, however both sides must still be each other's counterparty.
proposedConnection = provableStore.get(connectionPath(proposedUpgradeFields.connectionHops[0])
if (proposedConnection == null || proposedConnection.state != OPEN) {
    restoreChannel(portIdentifier, channelIdentifier)
}
```

modules/core/04-channel/keeper/upgrade_test.go
colin-axner: opened #3734

modules/core/04-channel/types/errors.go
colin-axner: opened #3734

modules/core/04-channel/types/errors.go
colin-axner: opened #3734

colin-axner:
>  I assume the in-flight packets checks and its associated funcs will be added separately later on.

Yup! I will open an issue for it. In-flight packets will still need another spec pr as flushing to flushing complete check needs to happen whenever we delete packet commitments

modules/core/04-channel/keeper/upgrade_test.go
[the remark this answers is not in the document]
damiannolan: Okay, yeah I get you! The assertion in your snippet looks good to me

colin-axner:
requires #3736

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -109,6 +110,110 @@ func (k Keeper) WriteUpgradeTryChannel(
 	emitChannelUpgradeTryEvent(ctx, portID, channelID, channel, proposedUpgrade)
 }
 
+// startFlushUpgradeHandshake will verify the counterparty proposed upgrade and the current channel state.
+// Once the counterparty information has been verified, it will be validated against the self proposed upgrade.
+// If any of the proposed upgrade fields are incompatible, an upgrade error will be returned resulting in an
+// aborted upgrade.
+//
+//lint:ignore U1000 Ignore unused function temporarily for debugging
+func (k Keeper) startFlushUpgradeHandshake(
```
AdityaSripal, line 119: My understanding is that the writes that happen in the spec will happen in a separate function in the implementation?

AdityaSripal (approved):

modules/core/04-channel/keeper/upgrade.go
charleenfei: they will happen in the `WriteUpgradeInit/Try/Ack` etc functions

chatton (approved):
Looks great! Left a few small comments and suggestions but nothing major, feel free to disregard or open up follow up issues to not block this PR.

modules/core/04-channel/keeper/export_test.go
```diff
@@ -7,9 +7,25 @@ package keeper
...
+	channelID string,
+	proposedUpgradeFields types.UpgradeFields,
+	counterpartyChannel types.Channel,
+	counterpartyUpgrade types.Upgrade,
+	proofCounterpartyChannel,
+	proofCounterpartyUpgrade []byte,
+	proofHeight clienttypes.Height,
+) error {
+	return k.startFlushUpgradeHandshake(ctx, portID, channelID, proposedUpgradeFields, counterpartyChannel, counterpartyUpgrade, proofCounterpartyChannel, proofCounterpartyUpgrade, proofHeight)
+}
```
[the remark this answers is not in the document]
damiannolan: I did a quick check on this since it wasn't clear if having the function in a `_test.go` file would prevent it being used. 

I added the method sig to the expected keeper interface of `ChannelKeeper` in transfer and there was no compiler issues and you could call the method with no issues. So to me, it seems like just making it exported from the get go is the same thing

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -126,6 +129,184 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+			if tc.expError != nil {
+				suite.Require().Error(err)
+
+				if expUpgradeError, ok := tc.expError.(types.UpgradeError); ok {
+					upgradeError, ok := err.(types.UpgradeError)
+					suite.Require().True(ok)
+					suite.Require().Equal(expUpgradeError.GetErrorReceipt(), upgradeError.GetErrorReceipt())
+				}
+
+				suite.Require().True(errorsmod.IsOf(err, tc.expError), err)
```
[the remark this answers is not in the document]
charleenfei: this makes sense to me!

modules/core/04-channel/keeper/export_test.go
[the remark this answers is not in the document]
DimitrisJim: wait, the function should not be added to the binary in non-test mode, this is in `go help build`:

> When compiling packages, build ignores files that end in '_test.go'.

the compiler as in the LSP that does checking or does a `go build` succeed? (the second case would sound like a serious bug :smile:)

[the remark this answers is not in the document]
damiannolan: You sir, are correct. It was the LSP that has me misled. A `make install` fails rightly as you say!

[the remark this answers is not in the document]
colin-axner: `startFlushUpgradeHandshake` should only be used during the upgrade handshake and thus does not need to be referenced outside of the `04-channel/keeper` package. Given that our versioning is dependent upon API breaking changes, it gives us more flexibility to handle fixes without incrementing our major version if this function is unexported. In my opinion, only the `Msg`'s and grpc handlers should be exported. It is unsafe to reference this function outside of the context it is called in (the upgrade message handlers). I think in general we should aim to limit our public API as it gives us maximum development flexibility while also establishing a clear public API external developers can rely on. Unless there's a reason to make the function public, I think it is healthy to default to leaving the function private

Issue #3686, opened 2023-05-29
Move ValidateUpgradeFields to upgrade.go


colin-axner:
## Summary

Upgrade checks/handlers should stay within the upgrade file. In addition, we should make the function private

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3628, opened 2023-05-22
feat: handler for `ChanUpgradeTry`

charleenfei:
This PR introduces the [ChanUpgradeTry handler.](https://github.com/cosmos/ibc/pull/967/files#diff-69d7c4ac1d4cdeeabcc757d3a34e85945ebc5613a034a52ebd0769cd8a0fc560R442)

closes: #3739 

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#pull-request-targeting)).
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/11-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [commit message](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) to be used for the changelog entry in the PR description for review.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.
- [ ] Review `Codecov Report` in the comment section below once CI passes.

PR #3687, opened 2023-05-29
refactor: make ValidateUpgradeFields unexported

colin-axner:
## Description

closes: #3686

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/export_test.go
+++ b/modules/core/04-channel/keeper/export_test.go
@@ -1,0 +1,15 @@
+package keeper
+
+/*
+	This file is to allow for unexported functions to be accessible to the testing package.
+*/
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+)
+
+func (k Keeper) ValidateUpgradeFields(ctx sdk.Context, proposedUpgrade types.UpgradeFields, currentChannel types.Channel) error {
+	return k.validateUpgradeFields(ctx, proposedUpgrade, currentChannel)
+}
--- a/modules/core/04-channel/keeper/keeper.go
+++ b/modules/core/04-channel/keeper/keeper.go
@@ -2,7 +2,6 @@ package keeper
 
 import (
 	"encoding/binary"
-	"reflect"
 	"strconv"
 	"strings"
 
@@ -538,36 +537,6 @@ func (k Keeper) SetUpgrade(ctx sdk.Context, portID, channelID string, upgrade ty
 	store := ctx.KVStore(k.storeKey)
 	bz := k.cdc.MustMarshal(&upgrade)
 	store.Set(host.ChannelUpgradeKey(portID, channelID), bz)
-}
-
-// ValidateUpgradeFields validates the proposed upgrade fields against the existing channel.
-// It returns an error if the following constraints are not met:
-// - there exists at least one valid proposed change to the existing channel fields
-// - the proposed order is a subset of the existing order
-// - the proposed connection hops do not exist
-// - the proposed version is non-empty (checked in UpgradeFields.ValidateBasic())
-// - the proposed connection hops are not open
-func (k Keeper) ValidateUpgradeFields(ctx sdk.Context, proposedUpgrade types.UpgradeFields, currentChannel types.Channel) error {
-	currentFields := extractUpgradeFields(currentChannel)
-
-	if reflect.DeepEqual(proposedUpgrade, currentFields) {
-		return errorsmod.Wrap(types.ErrChannelExists, "existing channel end is identical to proposed upgrade channel end")
-	}
-
-	connectionID := proposedUpgrade.ConnectionHops[0]
-	connection, err := k.GetConnection(ctx, connectionID)
-	if err != nil {
-		return errorsmod.Wrapf(connectiontypes.ErrConnectionNotFound, "failed to retrieve connection: %s", connectionID)
-	}
-
-	if connection.GetState() != int32(connectiontypes.OPEN) {
-		return errorsmod.Wrapf(
-			connectiontypes.ErrInvalidConnectionState,
-			"connection state is not OPEN (got %s)", connectiontypes.State(connection.GetState()).String(),
-		)
-	}
-
-	return nil
 }
 
 // extractUpgradeFields returns the upgrade fields from the provided channel.
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -1,11 +1,14 @@
 package keeper
 
 import (
+	"reflect"
+
 	errorsmod "cosmossdk.io/errors"
 	"github.com/cosmos/cosmos-sdk/telemetry"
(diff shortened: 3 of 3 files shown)


modules/core/04-channel/keeper/export_test.go
```diff
@@ -0,0 +1,15 @@
+package keeper
+
+/*
+	This file is to allow for unexported functions to be accessible to the testing package.
```
colin-axner, line 4: `ValidateUpgradeFields` is used by our tests

DimitrisJim: I do wonder if there's a better way to do this in go. Is there no way to have intra-module public symbols? If not, this sort of wrapping with a good docstring is a +1 from me.

crodriguezvega (approved):
LGTM! Alternatively we could test `ValidateUpgradeFields` implicitly while testing `ChanUpgradeInit` and check the returned error type.

modules/core/04-channel/keeper/export_test.go
damiannolan: can we move [this test code](https://github.com/cosmos/ibc-go/blob/04-channel-upgrades/modules/core/04-channel/keeper/keeper_test.go#L489) to upgrade.go and rename to `TestValidateUpgradeFields` ?

modules/core/04-channel/keeper/keeper.go
```diff
@@ -540,36 +539,6 @@ func (k Keeper) SetUpgrade(ctx sdk.Context, portID, channelID string, upgrade ty
...
-			connectiontypes.ErrInvalidConnectionState,
-			"connection state is not OPEN (got %s)", connectiontypes.State(connection.GetState()).String(),
-		)
-	}
-
-	return nil
-}
-
 // extractUpgradeFields returns the upgrade fields from the provided channel.
 func extractUpgradeFields(channel types.Channel) types.UpgradeFields {
```
damiannolan, line 543: should this also be moved alongside?

modules/core/04-channel/keeper/export_test.go
chatton: I'm actually not a huge fan of this approach, in general I don't really like the pattern of defining methods on types outside the file the type is defined in, I find it makes it harder to reason about and I often find myself needing to rely on tools to determine what methods a type actually has. I think this is particularly tough as lots of types have the same name. E.g. `Keeper` so there is already some cognitive overhead to figure out what is going on. 

I know this isn't a new pattern being introduced in this PR, but I thought I would share my opinion on it here!

I do agree that it would be nicer if there was a standard way of having an unexported method be called from a test (without resorting to the `reflect` package )

I don't think it's the worst thing to have this method exported as is, but I do not feel super strongly about this, and am happy to go with this approach if the majority of the team likes it!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -85,6 +88,36 @@ func (k Keeper) WriteUpgradeTryChannel(
 	// grab channel inside this function to get most current channel status
 }
 
+// validateUpgradeFields validates the proposed upgrade fields against the existing channel.
+// It returns an error if the following constraints are not met:
+// - there exists at least one valid proposed change to the existing channel fields
+// - the proposed order is a subset of the existing order
+// - the proposed connection hops do not exist
+// - the proposed version is non-empty (checked in UpgradeFields.ValidateBasic())
```
crodriguezvega, line 96: Is this comment still true for this function?

modules/core/04-channel/keeper/export_test.go
colin-axner: > I do wonder if there's a better way to do this in go.

If you figure something out, I'd love to give it a go! From my knowledge there isn't a workaround, until I recently saw this construction I didn't realize there was a way you could unit test private functions

> but also happy to cover the code from the larger handler tests if needs be.

In this situation, it'd be totally fine to test in a larger handler, but in general I think we should be capable of unit testing private functions. If we manage to limit our exported API to msg server handler + grpc, we will end up with very large tests trying to test multiple layers. It's a lot more sustainable in my opinion to have unit tests which directly test a functions functionality. 

> I'm actually not a huge fan of this approach, in general I don't really like the pattern of defining methods on types outside the file the type is defined in, I find it makes it harder to reason about and I often find myself needing to rely on tools to determine what methods a type actually has. I think this is particularly tough as lots of types have the same name. E.g. Keeper so there is already some cognitive overhead to figure out what is going on.

I agree, but I do see a lot of value in being able to test unexported functions, in addition these public functions are defined in the testing file, so I don't think they should show up on go docs, as they shouldn't be exported outside of this directory. Happy to go with another approach, but I think it'd be good to move forward with the ability to test unexported functions for the time being. 

@chatton what if I rename the public function to `k.TestValidateUpgradeFields` or some prefix indicating that the function is for testing purposes only?

modules/core/04-channel/keeper/keeper.go
colin-axner: yes, will move

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -85,6 +88,36 @@ func (k Keeper) WriteUpgradeTryChannel(
...
+// - the proposed connection hops are not open
+func (k Keeper) validateUpgradeFields(ctx sdk.Context, proposedUpgrade types.UpgradeFields, currentChannel types.Channel) error {
```
charleenfei, line 98: shoud we add the [`proposedConnectionHops` checks](https://github.com/cosmos/ibc-go/pull/3628#discussion_r1210307663) to this validation function?

modules/core/04-channel/keeper/export_test.go
DimitrisJim: Ah right. I remember I was reading about this yesterday and wanted to comment.

This is the typical approach for doing this, e.g [go's `fmt` does the same thing](https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/fmt/export_test.go) with a couple of function. 

I guess you could also simplify by re-binding the names instead of re-declaring a public symbol but for methods this results in a call where the receiver is passed in explicitly as the first argument, i.e instead of:

```go
ValidateUpgradeFields(args)
```

that we have now it would be:

```go
// k is keeper
ValidateUpgradeFields(k, args)
```

not sure if this is something people really like (I personally don't) but its an alternative.

modules/core/04-channel/keeper/upgrade.go
colin-axner: Yes it is checked in `MsgChannelUpgradeInit.ValidateBasic()` which calls `msg.Fields.ValidateBasic()`, would you like me to update the comment to be more specific?

modules/core/04-channel/keeper/upgrade.go
colin-axner: There are three types of validation we can do on an upgrade:
- stateless validation (upgrade.ValidateBasic())
- stateful validation of self upgrade
- stateful validation of self upgrade relative to counterparty upgrade

The proposed connection hop check falls into the third category, which is not the aim of this function. This function's purpose is to do stateful validation of the self upgrade

Issue #3689, opened 2023-05-29
Use ErrorReceipt as Error type


colin-axner:
## Summary

As demonstated in #3418, use the ErrorReceipt as an Error type, so all our API's which return an `error` can return an `ErrorReceipt`. Once the msg server receives the `ErrorReceipt` it can call `abortUpgradeHandshake` which does app callbacks and writes the error receipt

I like this flow because it makes it very clear to the msg server when the handshake should be aborted and when it succeeded, otherwise a nil error could be returned by `channelKeeper.ChanUpgradeTry` even though the handshake should not continue

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Here's how I imagine things:
1. chanUpgradeTry enters `startFlushUpgradeHandshake` subprotcol and hits error receipts situation
2. The code returns an error receipt:
```go
if err := func(); err != nil {
    return types.NewErrorReceipt(err)
}
```
3. The msg server checks for an error receipt:
```go
if ok := err.(types.ErrorReceipt); ok {
    abortUpgradeHandsahke(err)
}
```
4. restoreChannel emits/logs the error
```go
    ctx.Log(errorReceipt.FullError())
    emitErrorReceiptEvent()
    
    // restore channel ensure non-determinstic info is discarded from Message field
    errorReceipt = types.NewErrorReceipt(channel.Sequence, err)
    k.SetErrorReceipt(errorReceipt)
```

One part I'm unsure about, in the initial return, I think it is nice for it to look like a normal error type (just with a ErrorReceipt prefix), but I would like to do logs/emit events later in the restore channel function. So creating an API which creates an error receipt type that has full non-determinstic error info that can then be converted into the verifiable error receipt type

This is just my ideal handling, open to other ideas

chatton:
Should the actual emission of events / setting error receipt happen in the restore function? I was thinking something like

```golang
// msgServer
err := ChanUpgradeTry(...)
if isErrReceipt(err) {
   abortUpgradeHandshake(err)
   return ... 
}

func abortUpgradeHandshake(err) {
   log(err)
   restoreChannel()
   emitErrorReceiptEvent()
   emitEvents(err)
}
```



[...]

PR #3700, opened 2023-05-30
Add Error implementation for ErrorReceipt

chatton:
## Description

closes: #3689 

N/A feature branch

---

write a little note why.


modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +77,48 @@ func (ut Timeout) IsValid() bool {
 	return !ut.Height.IsZero() || ut.Timestamp != 0
 }
 
+// UpgradeError defines an error that occurs during an upgrade.
+type UpgradeError struct {
+	// underlyingError is the underlying error that caused the upgrade to fail.
+	// this error should not be written to state.
+	underlyingError error
```
damiannolan, line 84: what do you think about `err`, `cause` or `reason` as possible alternatives for this field?

modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +77,48 @@ func (ut Timeout) IsValid() bool {
...
+	// upgradeSequence is the sequence number of the upgrade that failed.
+	upgradeSequence uint64
```
damiannolan, line 86: maybe we could just go with `sequence` since its obvious that its upgrade sequence?

modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +77,48 @@ func (ut Timeout) IsValid() bool {
...
+func NewUpgradeError(upgradeSequence uint64, err error) UpgradeError {
+	return UpgradeError{
+		underlyingError: err,
+		upgradeSequence: upgradeSequence,
+	}
+}
+
 // NewErrorReceipt returns an error receipt with the code from the provided error type stripped
 // out to ensure changes of the error message don't cause state machine breaking changes.
 func NewErrorReceipt(upgradeSequence uint64, err error) ErrorReceipt {
```
damiannolan, line 112: I think we already discussed removing this constructor, but just noting while this is still a WIP

damiannolan (commented):
dropping some small suggestions that we can pick up on tomorrow :)

modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +77,48 @@ func (ut Timeout) IsValid() bool {
...
-		Error:    fmt.Sprintf("ABCI code: %d: %s", code, restoreErrorString),
+		Message:  fmt.Sprintf("ABCI code: %d: %s", code, restoreErrorString),
 	}
 }
+
+var _ error = &ErrorReceipt{}
+
+func (e *ErrorReceipt) Error() string {
+	return e.Message
+}
```
colin-axner, line 124: To support usage of `errormod.IsOf` could we add the following functions?
```diff
+func (u *UpgradeError) Is(err error) bool {
+       if x, ok := u.underlyingError.(interface{ Is(error) bool }); ok && x.Is(err) {
+               return true
+       }
+
+       return false
+}
+
+// Unwrap implements the built-in errors.Unwrap
+func (u *UpgradeError) Unwrap() error {
+       return u.underlyingError
+}
+
```

feel free to modify the `Is` code, I copied from stderrors library

modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +77,48 @@ func (ut Timeout) IsValid() bool {
...
+	// this error should not be written to state.
+	underlyingError error
+	// upgradeSequence is the sequence number of the upgrade that failed.
+	upgradeSequence uint64
+}
+
+func (u *UpgradeError) Error() string {
+	return u.underlyingError.Error()
+}
+
```
colin-axner, line 92: To support usage of `errormod.IsOf` could we add the following functions?
```diff
+func (u *UpgradeError) Is(err error) bool {
+       if x, ok := u.underlyingError.(interface{ Is(error) bool }); ok && x.Is(err) {
+               return true
+       }
+
+       return false
+}
+
+// Unwrap implements the built-in errors.Unwrap
+func (u *UpgradeError) Unwrap() error {
+       return u.underlyingError
+}
+
```

feel free to modify the `Is` code, I copied from stderrors library

modules/core/04-channel/types/upgrade.go
chatton: yep absolutely, I think this name was mostly to show intent in the github issue! happy to just have this as even just `err`

modules/core/04-channel/types/upgrade.go
chatton: yeah that works for me

modules/core/04-channel/types/upgrade.go
chatton: yep we can remove this or make it unexported maybe

modules/core/04-channel/types/upgrade.go
chatton: I think this could be nice!

modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +77,48 @@ func (ut Timeout) IsValid() bool {
...
+// UpgradeError defines an error that occurs during an upgrade.
+type UpgradeError struct {
+	// underlyingError is the underlying error that caused the upgrade to fail.
+	// this error should not be written to state.
+	underlyingError error
+	// upgradeSequence is the sequence number of the upgrade that failed.
+	upgradeSequence uint64
+}
+
+func (u *UpgradeError) Error() string {
```
colin-axner, line 89: I think this needs to match the return value of `NewUpgradeError`? When testing I got this error:
> keeper/upgrade.go:172:10: cannot use types.NewUpgradeError(channel.UpgradeSequence, errorsmod.Wrapf(types.ErrIncompatibleCounterpartyUpgrade, "expected upgrade sequence (%d) to match counterparty upgrade sequence (%d)", channel.UpgradeSequence, counterpartyChannel.UpgradeSequence)) (value of type "[github.com/cosmos/ibc-go/v7/modules/core/04-channel/types](http://github.com/cosmos/ibc-go/v7/modules/core/04-channel/types)".UpgradeError) as error value in return statement: "[github.com/cosmos/ibc-go/v7/modules/core/04-channel/types](http://github.com/cosmos/ibc-go/v7/modules/core/04-channel/types)".UpgradeError does not implement error (method Error has pointer receiver)

Until I removed the pointer receiver in this function and the error receipt one

modules/core/04-channel/types/upgrade.go
chatton: it looks like this test passes with no code changes

```golang

func (suite *KeeperTestSuite) TestUpgradeErrorIsOf() {
	ue := channeltypes.UpgradeError{}
	suite.Require().True(errorsmod.IsOf(ue, channeltypes.UpgradeError{}))
	suite.Require().False(errorsmod.IsOf(ue, channeltypes.ErrInvalidChannel))
}
```

modules/core/04-channel/types/upgrade.go
chatton: ended up updating to remove method receivers, tests seem to be passing now, let me know if there are any other issues

modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +71,36 @@ func (ut Timeout) IsValid() bool {
...
+
+func (u UpgradeError) Error() string {
+	return u.err.Error()
+}
+
+// GetErrorReceipt returns an error receipt with the code from the underlying error type stripped.
+func (u UpgradeError) GetErrorReceipt() ErrorReceipt {
+	// restoreErrorString defines a string constant included in error receipts.
+	// NOTE: Changing this const is state machine breaking as it is written into state.
+	const restoreErrorString = "restored channel to pre-upgrade state"
```
chatton, line 91: moved this into the function to make it more obvious that this should not be changed if this function is modified in the future.

damiannolan: agree, makes more sense here imo!

modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +71,37 @@ func (ut Timeout) IsValid() bool {
...
+	}
+}
+
+// Error implements the error interface, returning the underlying error which caused the upgrade to fail.
+func (u UpgradeError) Error() string {
+	return u.err.Error()
+}
+
+// GetErrorReceipt returns an error receipt with the code from the underlying error type stripped.
+func (u UpgradeError) GetErrorReceipt() ErrorReceipt {
```
damiannolan, line 97: we could add a small unit test for this I guess

damiannolan (approved):
Looks good to me! Really like this idea 🚀

modules/core/04-channel/types/upgrade.go
colin-axner: maybe assert that 2 of the same base errors wrapped differently still return the same error receipt?

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -0,0 +1,16 @@
...
+	errorsmod "cosmossdk.io/errors"
+
+	channeltypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+)
+
+func (suite *TypesTestSuite) TestUpgradeErrorIsOf() {
+	ue := channeltypes.UpgradeError{}
+	suite.Require().True(errorsmod.IsOf(ue, channeltypes.UpgradeError{}))
+	suite.Require().False(errorsmod.IsOf(ue, channeltypes.ErrInvalidChannel))
+
```
colin-axner, line 13:
```suggestion
	ue := channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)
	suite.Require().True(errorsmod.IsOf(ue, channeltypes.UpgradeError{}))
	suite.Require().False(errorsmod.IsOf(ue, channeltypes.ErrInvalidChannel))
```
what happens in this case? (this is closer to what we would use in practice

colin-axner (approved):
Looks great! Thank you :)

modules/core/04-channel/types/upgrade.go
damiannolan: hmm, interesting. If the same base error is wrapped differently will we get the code from the base error or the top level

modules/core/04-channel/types/upgrade_test.go
colin-axner: Ah, this works:
```diff
-       ue := channeltypes.UpgradeError{}
-       suite.Require().True(errorsmod.IsOf(ue, channeltypes.UpgradeError{}))
+       ue := channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)
+       suite.Require().True(errorsmod.IsOf(ue, channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)))
        suite.Require().False(errorsmod.IsOf(ue, channeltypes.ErrInvalidChannel))
 
        wrappedErr := errorsmod.Wrap(ue, "wrapped upgrade error")
-       suite.Require().True(errorsmod.IsOf(wrappedErr, channeltypes.UpgradeError{}))
+       suite.Require().True(errorsmod.IsOf(wrappedErr, channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)))
 }
```

Looks like without a custom `Is` function, it just does a reflect, minus the wrapped string. If the sequences mismatch it'll return false

colin-axner: Leaving off the custom `Is` function should work fine

modules/core/04-channel/types/upgrade.go
colin-axner: A wrapped error should only have 1 code (the base error), wrapping just adds a string to the `wrappedError.description`

damiannolan: Yep, we had to modify unwrapping such that the base error code is always returned. When we started writing a test we could see a different code returned in `GetErrorReceipt` as it was taking the top level code as opposed to the base error code

modules/core/04-channel/types/upgrade.go
```diff
@@ -77,12 +72,55 @@ func (ut Timeout) IsValid() bool {
...
+// Unwrap returns the base error that caused the upgrade to fail.
+func (u UpgradeError) Unwrap() error {
+	for {
+		if err := errors.Unwrap(u.err); err != nil {
+			u.err = err
+		} else {
+			return u.err
+		}
+	}
+}
```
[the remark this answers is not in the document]
colin-axner: I guess this would help in `Is` since it would unwrap an upgrade error if it is provided as the target?

modules/core/04-channel/types/upgrade.go
```diff
@@ -93,13 +94,31 @@ func (u UpgradeError) Error() string {
 	return u.err.Error()
 }
 
+
+// Is returns true if the underlying error is of the given err type.
+func (u UpgradeError) Is(err error) bool {
+	return errors.Is(u.err, err)
```
colin-axner, line 99: Combining this with `Unwrap` allows us to disregard the `UpgradeError` type wrapper?

colin-axner: Would it make sense to actually enforce that the target error is also an `UpgradeError`?

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	errorsmod "cosmossdk.io/errors"
+
+	channeltypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+)
+
+func (suite *TypesTestSuite) TestUpgradeErrorIsOf() {
+	ue := channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)
+	suite.Require().True(errorsmod.IsOf(ue, channeltypes.ErrInvalidChannel))
+	suite.Require().False(errorsmod.IsOf(ue, channeltypes.UpgradeError{}))
```
colin-axner, line 12: We could add test cases also for:
- two empty upgrade errors with IsOf
- the err is empty upgrade error and target is non-empty

But I'm happy as is as well

modules/core/04-channel/types/upgrade.go
damiannolan: Yeah, the issue was when wrapping the `UpgradeError` itself, as `Unwrap` wasn't implemented like it is for errors registered with errorsmod.

damiannolan: For example this is fine:

```go
upgradeErr := channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)
wrappedUpgradeErr := channeltypes.NewUpgradeError(1, errorsmod.Wrap(channeltypes.ErrInvalidChannel, "testing"))

suite.Require().Equal(upgradeErr.GetErrorReceipt().Message, wrappedUpgradeErr.GetErrorReceipt().Message)
```

But this isn't (without `Unwrap` implemented on `UpgradeError`):

```go
ue := channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)
wrappedErr := errorsmod.Wrap(ue, "wrapped upgrade error")
suite.Require().True(errorsmod.IsOf(wrappedErr, channeltypes.ErrInvalidChannel))

ue2 := channeltypes.NewUpgradeError(1, wrappedErr)
suite.Require().Equal(ue2.GetErrorReceipt().Message, ue.GetErrorReceipt().Message)
```

modules/core/04-channel/types/upgrade.go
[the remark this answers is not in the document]
chatton: so it looks like an issue we were running into is that if you are just wrapping an sdk error, eg. `ErrInvalidChannel` you don't need to worry about `Unwrap`. The following test code works without implementing `Unwrap` in the `UpgradeError` type

```golang
func (suite *TypesTestSuite) TestGetErrorReceipt() {
	upgradeErr := channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)
	wrappedUpgradeErr := channeltypes.NewUpgradeError(1, errorsmod.Wrap(channeltypes.ErrInvalidChannel, "testing"))

	suite.Require().Equal(upgradeErr.GetErrorReceipt().Message, wrappedUpgradeErr.GetErrorReceipt().Message)
}
```

However, if you want to be able to wrap the `UpgradeError` itself (which I think there is no harm providing the option) the `Unwrap` method needs to be implemented.

For example, if the `Unwrap` method is removed from `UpgradeError`, the following test fails.

```golang
func (suite *TypesTestSuite) TestGetErrorReceipt() {
	ue := channeltypes.NewUpgradeError(1, channeltypes.ErrInvalidChannel)

	wrappedErr := errorsmod.Wrap(ue, "wrapped upgrade error")
	suite.Require().True(errorsmod.IsOf(wrappedErr, channeltypes.ErrInvalidChannel))

	ue2 := channeltypes.NewUpgradeError(1, wrappedErr)

	suite.Require().Equal(ue2.GetErrorReceipt().Message, ue.GetErrorReceipt().Message)
}
```

with the error

```
-ABCI code: 1: restored channel to pre-upgrade state
+ABCI code: 4: restored channel to pre-upgrade state
```

I we should be able to wrap the `UpgradeError` type, so we don't need to worry about special casing it in any way.

WDYT?

modules/core/04-channel/types/upgrade.go
damiannolan: lol, see https://github.com/cosmos/ibc-go/pull/3700#discussion_r1213197560, we can close this thread

modules/core/04-channel/types/upgrade.go
chatton: I think I would prefer if the `UpgradeError` was effectively identical to other sdk errors as far as the `errorsmod` package is concerned.

I think this might make usage clunky, possibly needing to construct an `UpgradeError` from a specific sdk error in order to do an error comparison. 

```golang
if errorsmod.Is(err, types.UpgradeError{types.ErrInvalidChannel}) {}
```

modules/core/04-channel/types/upgrade.go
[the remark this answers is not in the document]
colin-axner: Oh yea! It's because `UpgradeError` doesn't implement the `Cause` interface function. See https://github.com/cosmos/ibc-go/pull/3700#discussion_r1213170980

modules/core/04-channel/types/upgrade_test.go
chatton: added! And refactored to use table tests

colin-axner (approved):
ACK, thank you :)

Issue #3631, opened 2023-05-23
add FlushStatus enum


charleenfei:
## Summary

Update the `ChannelState` to include a `FlushStatus` enum, as per [the spec for channel upgradeability.](https://github.com/cosmos/ibc/pull/967/files/a8a419c3fdde8df5e1ab6cb3c27bbd1c8aa09493#diff-69d7c4ac1d4cdeeabcc757d3a34e85945ebc5613a034a52ebd0769cd8a0fc560R67) Please note that the naming for the different flushing states is still under discussion.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
completed in #3702

PR #3702, opened 2023-05-30
imp: add flush status enum

charleenfei:
## Description

closes: #3631 

---

write a little note why.


proto/ibc/core/channel/v1/channel.proto
```diff
@@ -79,6 +81,18 @@ enum State {
   STATE_ACKUPGRADE = 7 [(gogoproto.enumvalue_customname) = "ACKUPGRADE"];
 }
 
+// FlushStatus defines the status of a channel end pertaining to in-flight packets.
+enum FlushStatus {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // Default status, the channel is not in an upgrade handshake
+  FLUSH_STATUS_NOTINFLUSH_UNSPECIFIED = 0 [(gogoproto.enumvalue_customname) = "NOTINFLUSH"];
```
damiannolan, line 89: should `UNSPECIFIED` / `UNINITIALIZED` be a separate field to `NOTINFLUSH`? I think its probably fine to treat them the same

proto/ibc/core/channel/v1/channel.proto
```diff
@@ -79,6 +81,18 @@ enum State {
   STATE_ACKUPGRADE = 7 [(gogoproto.enumvalue_customname) = "ACKUPGRADE"];
 }
 
+// FlushStatus defines the status of a channel end pertaining to in-flight packets.
```
damiannolan, line 84: nit: should we explicitly mention upgrade handshake? you can technically "flush" packets when you're in an active/open state

```suggestion
// FlushStatus defines the status of a channel end pertaining to in-flight packets during an upgrade handshake.
```

damiannolan (approved):
LGTM 🙏

proto/ibc/core/channel/v1/channel.proto
colin-axner: I think it is fine to treat them the same since the default (0 value) needs to be set after the opening handshake to ensure backwards compatibility (otherwise proofs on older versions would fail). It would be odd to me to place the channel in `UNINITIALIZED` until the first upgrade occurs and then in `NOTINFLUSH` after that. What do you think?

damiannolan: Yep, I agree!

Issue #3649, opened 2023-05-24
Implement restoreChannel


colin-axner:
## Summary

Implement spec subprotocol `restoreChannel` (name subject to change on our end). Note, spec code is subject to change. One difference between our code and the spec is that calling the restore channel will likely happen at the top level and the error receipt will already be constructed, so the API should likely take in the port/channel ID and an error receipt

We will likely want two functions
1. higher level func, which calls the (2) function and the `onChanUpgradeRestore` callback (handling logic)
2. sets the channel state/flush status back to normal, writes the error receipt, and deletes unnecessary info (disk read/write)

Maybe we can start by implementing the second function which is similar to the `restoreChannel` func in the spec

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3728, opened 2023-06-01
Implement Restore Channel

chatton:
## Description

ref: #3649

This PR adds the abort handshake method. The usage of this function within the message server will be added in a future PR.

N/A feature branch

---

write a little note why.


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -265,3 +264,61 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+
+	// TODO: could we want to abort but have upgrade error be non-nil?
+	if upgradeError != nil {
+		if err := k.writeErrorReceipt(ctx, portID, channelID, upgrade, upgradeError); err != nil {
+			return err
+		}
+	}
+
+	// TODO: callback execution
+	// cbs.OnChanUpgradeRestore()
```
chatton, line 289: this likely won't be here, this will be at the message server layer, I can remove this TODO if it's confusing

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,98 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+			err := channelKeeper.AbortHandshake(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, upgradeError)
+
+			if tc.expPass {
+				suite.Require().NoError(err)
+
+				channel, found := channelKeeper.GetChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().True(found, "channel should be found")
+
+				suite.Require().Equal(types.OPEN, channel.State, "channel state should be OPEN")
+				suite.Require().Equal(types.NOTINFLUSH, channel.FlushStatus, "channel flush status should be NOTINFLUSH")
```
crodriguezvega, line 455: nit: instead of hardcoding `OPEN` and `NOTINFLUSH` in the message, we could use the string conversion methods of those enums.

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,98 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+
+				_, found = channelKeeper.GetUpgrade(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().False(found, "upgrade info should be deleted")
+
+				errorReceipt, found := channelKeeper.GetUpgradeErrorReceipt(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				if upgradeError != nil {
+					suite.Require().True(found, "error receipt should be found")
+					suite.Require().Equal(upgradeError.GetErrorReceipt(), errorReceipt, "error receipt should be stored")
```
crodriguezvega, line 463: Since presence or not of the error receipt is checked above.

```suggestion
					suite.Require().Equal(upgradeError.GetErrorReceipt(), errorReceipt, "error receipt does not match expected error receipt")
```

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,98 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+				}
+
+				_, found = channelKeeper.GetCounterpartyLastPacketSequence(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().False(found, "counterparty last packet sequence should be found")
```
crodriguezvega, line 467:
```suggestion
				suite.Require().False(found, "counterparty last packet sequence should not be found")
```

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,98 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+			} else {
+				suite.Require().Error(err)
+				channel, found := channelKeeper.GetChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				if tc.channelPresent {
+					suite.Require().True(found, "channel should be found")
+					suite.Require().Equal(types.INITUPGRADE, channel.State, "channel state should not be restored to OPEN")
+				}
+
+				_, found = channelKeeper.GetUpgradeErrorReceipt(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().False(found, "error receipt should not be written")
```
crodriguezvega, line 478:
```suggestion
				suite.Require().False(found, "error receipt should not be found")
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -265,3 +264,61 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+	upgrade, found := k.GetUpgrade(ctx, portID, channelID)
+	if !found {
+		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	if err := k.restoreChannel(ctx, portID, channelID); err != nil {
+		return err
+	}
+
+	// TODO: could we want to abort but have upgrade error be non-nil?
```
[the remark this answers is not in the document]
crodriguezvega: I guess aborting is fine, since if `upgradeError` is `nil` that would mean a bug in our code?

crodriguezvega (commented):
Just left a few nits. Thanks, @chatton!

modules/core/04-channel/keeper/upgrade.go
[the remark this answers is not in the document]
chatton: maybe we can just error out in teh case of `upgradeError` being nil then

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -265,3 +264,61 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+	}
+
+	// TODO: callback execution
+	// cbs.OnChanUpgradeRestore()
+
+	return nil
+}
+
+// restoreChannel will restore the channel state and flush status to their pre-upgrade state so that upgrade is aborted
+// it write an error receipt to state so counterparty can restore as well.
```
colin-axner, line 295:
```suggestion
// restoreChannel will restore the channel state and flush status to their pre-upgrade state so that upgrade is aborted
// It will write an error receipt to state so that the counterparty can restore as well.
```

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,106 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+	for _, tc := range tests {
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest()
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+			suite.coordinator.Setup(path)
+
+			channelKeeper := suite.chainA.GetSimApp().IBCKeeper.ChannelKeeper
+
+			path.EndpointA.ChannelConfig.Version = mock.Version + "-v2"
```
colin-axner, line 449: We can use `mock.UpgradeVersion

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,106 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+				_, found = channelKeeper.GetCounterpartyLastPacketSequence(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().False(found, "counterparty last packet sequence should not be found")
+
+			} else {
+				suite.Require().Error(err)
+				channel, found := channelKeeper.GetChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				if tc.channelPresent {
+					suite.Require().True(found, "channel should be found")
+					suite.Require().Equal(types.INITUPGRADE, channel.State, "channel state should not be restored to %s", types.INITUPGRADE.String())
+				}
```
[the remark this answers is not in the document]
colin-axner: The state machine code should never ever delete a channel so `tc.channelPresent` is accounting for behaviour of the setup code which can be determined simply via the returned bool. The only time `tc.channelPresent` is true and `found` is false is if we write the test case incorrectly

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,106 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+
+				suite.Require().Equal(types.OPEN, channel.State, "channel state should be %s", types.OPEN.String())
+				suite.Require().Equal(types.NOTINFLUSH, channel.FlushStatus, "channel flush status should be %s", types.NOTINFLUSH.String())
+				_, found = channelKeeper.GetUpgrade(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().False(found, "upgrade info should be deleted")
+
+				errorReceipt, found := channelKeeper.GetUpgradeErrorReceipt(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				if upgradeError != nil {
+					suite.Require().True(found, "error receipt should be found")
+					suite.Require().Equal(upgradeError.GetErrorReceipt(), errorReceipt, "error receipt does not match expected error receipt")
```
colin-axner, line 473: If the test passes, the upgradeError should not be nil?

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,106 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+				channel, found := channelKeeper.GetChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				if tc.channelPresent {
+					suite.Require().True(found, "channel should be found")
+					suite.Require().Equal(types.INITUPGRADE, channel.State, "channel state should not be restored to %s", types.INITUPGRADE.String())
+				}
+
+				_, found = channelKeeper.GetUpgradeErrorReceipt(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().False(found, "error receipt should not be found")
+
+				// TODO: assertion that GetCounterpartyLastPacketSequence is present and correct
```
colin-axner, line 490: Should we check that the upgrade is present as well?

colin-axner (approved):
Great work! Looks good to me, mostly left a few suggestions on the testing code

modules/core/04-channel/keeper/keeper.go
```diff
@@ -532,13 +532,25 @@ func (k Keeper) SetCounterpartyLastPacketSequence(ctx sdk.Context, portID, chann
 	store.Set(host.ChannelCounterpartyLastPacketSequenceKey(portID, channelID), bz)
 }
 
+// deleteCounterpartyLastPacketSequence deletes the counterparty last packet sequence from the store.
+func (k Keeper) deleteCounterpartyLastPacketSequence(ctx sdk.Context, portID, channelID string) {
```
DimitrisJim, line 536: ping: added these separately in https://github.com/cosmos/ibc-go/pull/3780 so as to re-use them in any pr's required.

modules/core/04-channel/keeper/upgrade_test.go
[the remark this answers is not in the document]
chatton: love it, great suggestion

modules/core/04-channel/keeper/upgrade_test.go
chatton: good catch, this was a left over from when this was a valid situation

modules/core/04-channel/keeper/upgrade_test.go
chatton: yes!

AdityaSripal (approved):
Nice work!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -326,3 +325,61 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
 		LatestSequenceSend: nextSequenceSend - 1,
 	}, nil
 }
+
+// abortHandshake will restore the channel state and flush status to their pre-upgrade state so that upgrade is aborted.
+// any unnecessary state is deleted. An error receipt is written, and the OnChanUpgradeRestore callback is called.
+func (k Keeper) abortHandshake(ctx sdk.Context, portID, channelID string, upgradeError *types.UpgradeError) error {
```
damiannolan, line 331: discussed with @chatton about this API yesterday - I think maybe it needs to accept a built-in `error` instead of `types.UpgradeError` as we can't make any guarantees that applications will always return `types.UpgradeError`

also nit: rename to `AbortUpgrade` rather than handshake as that has spill over with the opening channel handshake naming

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -607,3 +607,118 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
 	suite.Require().True(errorsmod.IsOf(actualError, expError), actualError)
 }
+
+// TestAbortHandshake tests that when the channel handshake is aborted, the channel state
+// is restored the previous state and that an error receipt is written, and upgrade state which
+// is no longer required is deleted.
+func (suite *KeeperTestSuite) TestAbortHandshake() {
+	var (
+		path         *ibctesting.Path
+		upgradeError error
```
chatton, line 617: note: this changed to `error` instead of `types.UpgradeError` because of [how nil checks on errors work](https://go.dev/doc/faq#nil_error) in go.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -326,3 +325,73 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+	// in this case we need to construct one in order to write the error receipt.
+	upgradeError, ok := err.(*types.UpgradeError)
+	if !ok {
+		channel, found := k.GetChannel(ctx, portID, channelID)
+		if !found {
+			return errorsmod.Wrapf(types.ErrChannelNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+		}
+
+		upgradeError = types.NewUpgradeError(channel.UpgradeSequence, err)
+	}
```
chatton, line 355: this feels a bit ugly, open to suggestions on improving this.

chatton (commented):
Made the suggested changes let me know what you think 

cc @damiannolan / @colin-axner

modules/core/04-channel/keeper/upgrade.go
damiannolan: the channel is already looked up in `restoreChannel`, right? we could move that logic to be scoped within this function such that you don't have to do the lookup twice (would also save gas)

is `restoreChannel()` intended to only be called via `abortUpgrade()`?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -377,3 +376,73 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
 		LatestSequenceSend: nextSequenceSend - 1,
 	}, nil
 }
+
+// abortUpgrade will restore the channel state and flush status to their pre-upgrade state so that upgrade is aborted.
+// any unnecessary state is deleted. An error receipt is written, and the OnChanUpgradeRestore callback is called.
+func (k Keeper) abortUpgrade(ctx sdk.Context, portID, channelID string, err error) error {
```
damiannolan, line 382: I think we'll need to make this public/exported

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -377,3 +376,73 @@ func (k Keeper) constructProposedUpgrade(ctx sdk.Context, portID, channelID stri
...
+	if err == nil {
+		return errorsmod.Wrap(types.ErrInvalidUpgradeError, "cannot abort upgrade handshake with nil error")
+	}
+
+	upgrade, found := k.GetUpgrade(ctx, portID, channelID)
+	if !found {
+		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	if err := k.restoreChannel(ctx, portID, channelID); err != nil {
```
charleenfei, line 392: should we wrap this err with some sort of error indicating it came from restoring the channel? since the error returned from `RestoreChannel` is just if the channel was not found.

modules/core/04-channel/keeper/upgrade.go
chatton: yep good point, will do a small refactor

modules/core/04-channel/keeper/upgrade.go
chatton: yep good call

chatton:
thanks for getting this over the line @charleenfei 🥳

PR #3729, opened 2023-06-01
chore: make UpgradeError a reference type instead of value type

chatton:
## Description

There is a small issue with leaving the `UpgradeError` as a value type.

Since we want the method signature of the  keeper functions to be `error`, we would need to return a struct value instead of a method receiver. This would in turn break any `err != nil` checks. My moving the functions to methods, nil checks should work as expected.

I made a small modification to the `Unwrap` function so as to not accidentally mutate state (`u.err`).

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/types/upgrade.go
+++ b/modules/core/04-channel/types/upgrade.go
@@ -81,31 +81,34 @@ type UpgradeError struct {
 	sequence uint64
 }
 
+var _ error = &UpgradeError{}
+
 // NewUpgradeError returns a new UpgradeError instance.
-func NewUpgradeError(upgradeSequence uint64, err error) UpgradeError {
-	return UpgradeError{
+func NewUpgradeError(upgradeSequence uint64, err error) *UpgradeError {
+	return &UpgradeError{
 		err:      err,
 		sequence: upgradeSequence,
 	}
 }
 
 // Error implements the error interface, returning the underlying error which caused the upgrade to fail.
-func (u UpgradeError) Error() string {
+func (u *UpgradeError) Error() string {
 	return u.err.Error()
 }
 
 // Is returns true if the underlying error is of the given err type.
-func (u UpgradeError) Is(err error) bool {
+func (u *UpgradeError) Is(err error) bool {
 	return errors.Is(u.err, err)
 }
 
 // Unwrap returns the base error that caused the upgrade to fail.
-func (u UpgradeError) Unwrap() error {
+func (u *UpgradeError) Unwrap() error {
+	var baseError = u.err
 	for {
-		if err := errors.Unwrap(u.err); err != nil {
-			u.err = err
+		if err := errors.Unwrap(baseError); err != nil {
+			baseError = err
 		} else {
-			return u.err
+			return baseError
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


modules/core/04-channel/types/upgrade_test.go
```diff
@@ -94,30 +94,22 @@ func (suite *TypesTestSuite) TestUpgradeErrorIsOf() {
...
 				wrappedErr := errorsmod.Wrap(upgradeError, "wrapped upgrade error")
-				upgradeError = types.NewUpgradeError(1, wrappedErr)
-			},
-			expPass: true,
-		},
-		{
-			msg: "nil underlying error and target",
-			malleate: func() {
-				upgradeError = types.UpgradeError{}
-				intputErr = types.UpgradeError{}
```
chatton, line 113: this test case was removed, because we don't actually need to compare a nil error to see if it's a nil error.

The `errors.Is` check in stdlib does not actually return true if two nil errors are both nil but are of different implementations. E.g. the following test function does not pass

```golang
func TestFoo(t *testing.T) {
   var err1 *channeltypes.UpgradeError = nil
   var err2 error = nil
   require.Nil(t, err1)
   require.Nil(t, err2)
   res := errorsmod.IsOf(err1, err2)
   require.True(t, res)
}
```

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -4,13 +4,13 @@ import (
 	errorsmod "cosmossdk.io/errors"
 
 	clienttypes "github.com/cosmos/ibc-go/v7/modules/core/02-client/types"
-	"github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
+	channeltypes "github.com/cosmos/ibc-go/v7/modules/core/04-channel/types"
```
colin-axner, line 7: typically we allow the `types` directory to be used without a prefix when we are in that package or sub package. So `types` in the 04-channel subdirectory would be just `types`, but I'm happy to be more explicit as well!

modules/core/04-channel/types/upgrade.go
```diff
@@ -67,31 +67,34 @@ type UpgradeError struct {
...
+func (u *UpgradeError) Unwrap() error {
+	baseError := u.err
 	for {
-		if err := errors.Unwrap(u.err); err != nil {
-			u.err = err
+		if err := errors.Unwrap(baseError); err != nil {
+			baseError = err
 		} else {
-			return u.err
+			return baseError
```
colin-axner, line 97: nice anticipation! If the mutation bug was accidentally introduced, do we have a test that would fail? Maybe we should have an `Unwrap()` test which asserts that the base type is not mutated after being called?

colin-axner (approved):
Nice catch! Thanks for pushing a fix so fast

charleenfei (approved):
nice work, lgtm

damiannolan (approved):
Nice catch and quick fix!

modules/core/04-channel/types/upgrade_test.go
chatton: happy to change back, think I just picked one that was in the merge conflicts but I can just use `types`

modules/core/04-channel/types/upgrade.go
chatton: sure I'll add a quick Unwrap test! Nice suggestion

damiannolan (approved):
ACK lgtm

DimitrisJim (approved):
lgtm!

Issue #3644, opened 2023-05-24
Implement openUpgradeHandshake subprotocol


colin-axner:
## Summary

Implement `openUpgradeHandshake` subprotocol in the spec. The function name may be changed to align better with our codebase (cc @damiannolan if you have any ideas)

The check for deleting the counterparty last packet sequence send may be skipped for now as the mapping may not exist when this issue is implemented

Spec [pr](https://github.com/cosmos/ibc/pull/967)

Code from spec:
```
// caller must do all relevant checks before calling this function
function openUpgradeHandshake(
    portIdentifier: Identifier,
    channelIdentifier: Identifier,
) {
    currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
    upgrade = provableStore.get(channelUpgradePath(portIdentifier, channelIdentifier))

    // switch channel fields to upgrade fields
    // and set channel state to OPEN
    currentChannel.ordering = upgrade.fields.ordering
    currentChannel.version = upgrade.fields.version
    currentChannel.connectionHops = upgrade.fields.connectionHops
    currentchannel.state = OPEN
    currentChannel.flushStatus = NOTINFLUSH
    provableStore.set(channelPath(portIdentifier, channelIdentifier), currentChannel)

    // delete auxilliary state
    provableStore.delete(channelUpgradePath(portIdentifier, channelIdentifier))
    privateStore.delete(channelCounterpartyLastPacketSequencePath(portIdentifier, channelIdentifier))
}
```

Code is subject to change based on pr reviews at the time of writing

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
`OpenUpgradeHandshake` to me sounds a little bit synonymous to `InitUpgradeHandshake` in some ways. Maybe we could use something like "complete" or "finalise" instead. If I think of something else I'll drop it here

damiannolan:
I guess we can roll with that for now. I still have some reservations with "open" terminology - there is also `ChanOpenInit` as part of the channel opening handshake. 

I guess we can nail down namings later, but as a general remark part of me slightly feels that if we are changing one handler name (confirm->open) then it may be less confusing if we changed another making it two instead of one outlier. 

Some food for thought could be:

```go
func ChanUpgradePropose()

func ChanUpgradeTry()

func ChanUpgradeAck()

[...]

PR #3759, opened 2023-06-05
Implement openUpgradeHandshake subprotocol

DimitrisJim:
## Description

Add the `openUpgradeHandshake` sub protocol (renamed to `WriteUpgradeOpenChannel` as per the comments in the issue.

closes: #3644 

```bash
Implement openUpgradeHandshake subprotocol
```

---

write a little note why.


modules/core/04-channel/keeper/keeper.go
```diff
@@ -539,6 +539,18 @@ func (k Keeper) SetUpgrade(ctx sdk.Context, portID, channelID string, upgrade ty
 	store.Set(host.ChannelUpgradeKey(portID, channelID), bz)
 }
 
+// deleteCounterpartyLastPacketSequence deletes the counterparty last packet sequence from the store.
+func (k Keeper) deleteCounterpartyLastPacketSequence(ctx sdk.Context, portID, channelID string) {
```
DimitrisJim, line 543: taken from #3728 temporarily.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -109,6 +109,37 @@ func (k Keeper) WriteUpgradeTryChannel(
 	emitChannelUpgradeTryEvent(ctx, portID, channelID, channel, proposedUpgrade)
 }
 
+// WriteUpgradeOpenChannel writes the agreed upon upgrade fields to the channel fields. This can be called in one of two cases:
+// - In the UpgradeAck step of the handshake both sides have already flushed any in-flight packets.
+// - In the UpgradeOpen step of the handshake.
+func (k Keeper) WriteUpgradeOpenChannel(ctx sdk.Context, portID, channelID string) {
```
DimitrisJim, line 115: Question. It would be nice if we could get information about who called us (`Ack` or `Open`) since that can be used in the error messages and when incrementing the telemetry counter. Would a bool (or enum :^)) be the preferred approach?

I don't think we can use the previousState here to make assumptions about who called us but I haven't thought too hard about it yet.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -109,6 +109,37 @@ func (k Keeper) WriteUpgradeTryChannel(
 	emitChannelUpgradeTryEvent(ctx, portID, channelID, channel, proposedUpgrade)
 }
 
+// WriteUpgradeOpenChannel writes the agreed upon upgrade fields to the channel fields. This can be called in one of two cases:
+// - In the UpgradeAck step of the handshake both sides have already flushed any in-flight packets.
```
charleenfei, line 113:
```suggestion
// - In the UpgradeAck step of the handshake if both sides have already flushed all in-flight packets.
```

modules/core/04-channel/keeper/upgrade.go
charleenfei: why can't we use the `previousState`?

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: yikes :smile:

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: If my understanding of the state transitions are correct (ref https://github.com/cosmos/ibc/pull/978/files). We can end up in the subprotocol from both open and from ack with channel state being `ACKUPGRADE` for channel `A`.

charleenfei: are you talking about the case where this subprotocol is being called on chain B, where chain A has automatically moved from ACK -> OPEN bc the pending in flight packets check results in an automatic chan OPEN?

modules/core/04-channel/keeper/keeper.go
colin-axner: Maybe we should open a quick pr to merge these changes?

colin-axner (approved):
Nice! LGTM

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -109,6 +109,38 @@ func (k Keeper) WriteUpgradeTryChannel(
 	emitChannelUpgradeTryEvent(ctx, portID, channelID, channel, proposedUpgrade)
 }
 
+// WriteUpgradeOpenChannel writes the agreed upon upgrade fields to the channel, sets the channel flush status to `NOTINFLUSH` and sets the channel state back to `OPEN`. This can be called in one of two cases:
+// - In the UpgradeAck step of the handshake if both sides have already flushed all in-flight packets.
+// - In the UpgradeOpen step of the handshake.
+func (k Keeper) writeUpgradeOpenChannel(ctx sdk.Context, portID, channelID string) {
```
damiannolan, line 115: will this method not be called from the core msg server and need to be unexported?

damiannolan: can be updated when its used

modules/core/04-channel/keeper/keeper.go
DimitrisJim: good idea https://github.com/cosmos/ibc-go/pull/3780

charleenfei (approved):

AdityaSripal (approved):
One note is that I didn't use checks in spec since they would be duplicated. So in spec it is caller responsibility. Up to implementation team whether it makes sense to have checks in the function or not

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: wouldn't be be in `TRYUPGRADE`. To be a bit more verbose about my confusion, during `ChanUpgradeAck` and after `startFlushUpgradeHandshake` the states of `A` and `B` will be `(ACKUPGRADE, TRYUPGRADE)`. If at this point both A and B have finished flushing, `openUpgradeHandShake` will be called with those two states for A and B.

OTOH, if they have not finished flushing, after they do, the relayer will submit the `ChanUpgradeClose` messages to both `A` and `B` and the states, having not changed, will still be ``(ACKUPGRADE, TRYUPGRADE)`.

I feel I'm missing something here but can't see it yet. Either way, this doesn't affect the core logic, its mostly for the telemetry counter and the err messages.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -308,6 +308,38 @@ func (k Keeper) startFlushUpgradeHandshake(
 	return nil
 }
 
+// WriteUpgradeOpenChannel writes the agreed upon upgrade fields to the channel, sets the channel flush status to `NOTINFLUSH` and sets the channel state back to `OPEN`. This can be called in one of two cases:
```
crodriguezvega, line 311: If I'm not mistaken, good doesn't support markdown syntax, so the backticks will not have the effect of formatting as code, unfortunately...

```suggestion
// WriteUpgradeOpenChannel writes the agreed upon upgrade fields to the channel, sets the channel flush status to NOTINFLUSH and sets the channel state back to OPEN. This can be called in one of two cases:
```

crodriguezvega (commented):
Nice, @DimitrisJim!

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: It actually does but a super simplified version for some silly reason, can't find a specification on what it supports but backticks dont seem to be in the subset. 

p.s Doc links (i.e godoc creates an automatic link if you enclose things in brackets i.e `[Status.OPEN]`) could work here and would be nice to have but that's another subject.

Issue #3752, opened 2023-06-05
Implement `WriteUpgradeCancelChannel`


colin-axner:
## Summary

`WriteUpgradeCancelChannel` should apply the state changes of a successful upgrade cancellation. It should emit events/logs. This function will be similar to `WriteUpgradeTimeoutChannel`, lets implement them separately for now and refactor later to reduce duplicate code. The function name may be changed, but using existing convention for now

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3773, opened 2023-06-07
Add implementation for WriteUpgradeCancelChannel

chatton:
## Description

~This PR uses some methods introduced in #3728 so should be reviewed after.~

closes: #3752

N/A Feature Branch

---

write a little note why.


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -110,6 +110,29 @@ func (k Keeper) WriteUpgradeTryChannel(
...
+	}
+
+	upgrade, found := k.GetUpgrade(ctx, portID, channelID)
+	if !found {
+		panic(fmt.Sprintf("could not find existing upgrade when cancelling channel upgrade, channelID: %s, portID: %s", channelID, portID))
+	}
+
+	if err := k.restoreChannel(ctx, portID, channelID); err != nil {
+		return errorsmod.Wrap(types.ErrUpgradeRestoreFailed, fmt.Sprintf("failed to restore channel, channelID: %s, portID: %s", channelID, portID))
+	}
```
chatton, line 130: do we want to just panic here too? I think mixing errors/panics is worse than using one or the other. Is a regular state reversion an okay outcome if a restore fails?

The current implementation of restore is using an error return, but I guess this could be a panic as well

charleenfei: i like the idea of returning an error, because as we discussed before then it is clear from the method signature that `WriteUpgrade...` could fail.

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -389,3 +389,106 @@ func (suite *KeeperTestSuite) assertUpgradeError(actualError, expError error) {
...
+				_, found = channelKeeper.GetCounterpartyLastPacketSequence(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().False(found, "counterparty last packet sequence should not be found")
+
+			} else {
+				suite.Require().Error(err)
+				channel, found := channelKeeper.GetChannel(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				if tc.channelPresent {
+					suite.Require().True(found, "channel should be found")
+					suite.Require().Equal(types.INITUPGRADE, channel.State, "channel state should not be restored to %s", types.INITUPGRADE.String())
+				}
```
crodriguezvega, line 485: Should there be here an `else` block to check `suite.Require().False(found, "channel should not be found")` in case the test case has `channelPresent` set to false?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -110,6 +110,29 @@ func (k Keeper) WriteUpgradeTryChannel(
...
 }
 
+// WriteUpgradeCancelChannel writes a channel which has canceled the upgrade process. Auxiliary upgrade state is
+// also deleted.
+func (k Keeper) WriteUpgradeCancelChannel(ctx sdk.Context, portID, channelID string) error {
+	defer telemetry.IncrCounter(1, "ibc", "channel", "upgrade-cancel")
+
+	channel, found := k.GetChannel(ctx, portID, channelID)
+	if !found {
+		panic(fmt.Sprintf("could not find existing channel when cancelling channel upgrade, channelID: %s, portID: %s", channelID, portID))
```
crodriguezvega, line 120: Is it really necessary to add `when cancelling channel upgrade` to the error message? Wouldn't there be any other information in the log (like a stack trace) to figure that out?

crodriguezvega (commented):
Thank you, @chatton.

modules/core/04-channel/keeper/upgrade.go
chatton: yeah you're right it's not really required

AdityaSripal (approved):

DimitrisJim (approved):
lgtm

Issue #3642, opened 2023-05-24
Implement WriteUpgradeAckChannel


colin-axner:
## Summary

Implement the function `WriteUpgradeAckChannel`. Requires #3632. See `WriteUpgradeTryChannel` for reference, currently in #3628 

```go
func (k Keeper) WriteUpgradeAckChannel(ctx sdk.Context, portID, channelID string) 
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

charleenfei:
looking at this issue now for implementation, it seems like the sole purpose of this function would be to emit the events associate with the handshake?  the latest version of the spec has the [channel `State` changing in the `startFlushUpgrade` function.](https://github.com/cosmos/ibc/pull/967/files#diff-69d7c4ac1d4cdeeabcc757d3a34e85945ebc5613a034a52ebd0769cd8a0fc560R344)

is this correct? if so, maybe we should rename this and related functions, or consider not even having this function at all and using the event emitter directly in the handshake handler. see [my related comment](https://github.com/cosmos/ibc-go/pull/3628/files#r1206826570) on the `chanUpgradeTry` pr.

colin-axner:
I think my preference is to keep with our existing design pattern. Our `startFlushUpgrade` doesn't need to write the state change (we are allowed to differ from the spec in this regard), if we decide to combine these functions later, I think that's fine

charleenfei:
wouldn't it be safer to actually write the state change in startFlush though, since that is the method that is actually verify the proofs which the state change would be based on?

PR #3787, opened 2023-06-07
WriteUpgradeAck method

charleenfei:
## Description

closes: #3642 

---

write a little note why.


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -257,6 +257,33 @@ func (k Keeper) startFlushUpgradeHandshake(
...
 }
 
+// WriteUpgradeAckChannel writes a channel which has successfully passed the UpgradeAck handshake step as well as
+// setting the upgrade for that channel.
+// An event is emitted for the handshake step.
+func (k Keeper) WriteUpgradeAckChannel(
+	ctx sdk.Context,
+	portID, channelID string,
+	proposedUpgrade types.Upgrade,
+	flushStatus types.FlushStatus,
```
AdityaSripal, line 267: I think its better to determine flushStatus inside this function rather than passing it in as argument

AdityaSripal: You can do this by creating an iterator over packet commitments and seeing if it is empty

charleenfei: okay, that makes sense!

damiannolan: We can add the iterator logic to determine that in a follow up! :)

charleenfei: ah, i just used the `GetAllPacketCommitmentsAtChannel` method from the 04-channel keeper, is there a reason why this should not suffice?

crodriguezvega (commented):
Looks good, @charleenfei! Should we also have some tests for this function?

charleenfei:
> Looks good, @charleenfei! Should we also have some tests for this function?

if you look at the other `Write` functions, we usually test them as part of testing the handler (in this case, [`UpgradeAck`](https://github.com/cosmos/ibc-go/tree/damian/1620-chan-upgrade-ack) which has not been submitted as a PR yet. if we want to change this testing pattern going forward, im happy to accomodate (we can try to test it as we test validate basic), but i think it's a bit easier to test this when the handler function is in as we rely on the packets to be flushed, etc. cc @damiannolan @colin-axner

modules/core/04-channel/keeper/upgrade.go
AdityaSripal: that should work since its doing the work for you. Please add tests to double check. Nice work keeping it DRY

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -204,6 +204,37 @@ func (k Keeper) WriteUpgradeTryChannel(
...
+
+	channel, found := k.GetChannel(ctx, portID, channelID)
+	if !found {
+		panic(fmt.Sprintf("could not find existing channel when updating channel state in successful ChanUpgradeAck step, channelID: %s, portID: %s", channelID, portID))
+	}
+
+	previousState := channel.State
+	channel.State = types.ACKUPGRADE
+
+	if k.GetAllPacketCommitmentsAtChannel(ctx, portID, channelID) != nil {
```
AdityaSripal, line 225: My only nit is that this is inefficient. Since it will do the full iteration and give you all packet commitments when really you just need to check if there exists at least one.

You can maybe make a helper function to do this

AdityaSripal: Can be done in a followup

colin-axner, line 225: I would prefer to add this in a followup

chatton: I'd be happy to resolve this in a follow up 👍

chatton (approved):
LGTM!

PR #3791, opened 2023-06-07
[wip] scaffold msg server for chanUpgradeTry

damiannolan:
## Description

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -203,6 +203,10 @@ func (k Keeper) WriteUpgradeTryChannel(
 	emitChannelUpgradeTryEvent(ctx, portID, channelID, channel, proposedUpgrade)
 }
 
+func (k Keeper) AbortUpgrade(ctx sdk.Context, portID, channelID string, err error) error {
+	return nil
+}
+
 // startFlushUpgradeHandshake will verify the counterparty proposed upgrade and the current channel state.
 // Once the counterparty information has been verified, it will be validated against the self proposed upgrade.
 // If any of the proposed upgrade fields are incompatible, an upgrade error will be returned resulting in an
--- a/modules/core/keeper/msg_server.go
+++ b/modules/core/keeper/msg_server.go
@@ -751,7 +751,62 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
 
 // ChannelUpgradeTry defines a rpc handler method for MsgChannelUpgradeTry.
 func (k Keeper) ChannelUpgradeTry(goCtx context.Context, msg *channeltypes.MsgChannelUpgradeTry) (*channeltypes.MsgChannelUpgradeTryResponse, error) {
-	return nil, nil
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	module, _, err := k.ChannelKeeper.LookupModuleByChannel(ctx, msg.PortId, msg.ChannelId)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade init failed", "port-id", msg.PortId, "error", errorsmod.Wrap(err, "could not retrieve module from port-id"))
+		return nil, errorsmod.Wrap(err, "could not retrieve module from port-id")
+	}
+
+	cbs, ok := k.Router.GetRoute(module)
+	if !ok {
+		ctx.Logger().Error("channel upgrade init failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
+		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+	}
+
+	upgrade, err := k.ChannelKeeper.ChanUpgradeTry(ctx, msg.PortId, msg.ChannelId, msg.ProposedUpgradeConnectionHops, msg.UpgradeTimeout, msg.CounterpartyProposedUpgrade, msg.CounterpartyUpgradeSequence, msg.ProofChannel, msg.ProofUpgrade, msg.ProofHeight)
+	if err != nil {
+		if upgradeErr, ok := err.(*channeltypes.UpgradeError); ok {
+			if err := k.ChannelKeeper.AbortUpgrade(ctx, msg.PortId, msg.ChannelId, upgradeErr); err != nil {
+				return nil, err
+			}
+
+			// NOTE: a failure Result is indicated to the client and an error receipt is written to state
+			return &channeltypes.MsgChannelUpgradeTryResponse{
+				Result: channeltypes.FAILURE,
+			}, nil
+		}
+
+		// TODO: add meaningful comment about type of error and rejecting full tx + revert state
+		return nil, err
+	}
+
+	cacheCtx, writeFn := ctx.CacheContext()
+	version, err := cbs.OnChanUpgradeTry(cacheCtx, msg.PortId, msg.ChannelId, upgrade.Fields.Ordering, upgrade.Fields.ConnectionHops, upgrade.Fields.Version)
+	if err != nil {
+		if err := k.ChannelKeeper.AbortUpgrade(ctx, msg.PortId, msg.ChannelId, err); err != nil {
+			return nil, err
+		}
+
+		return &channeltypes.MsgChannelUpgradeTryResponse{
+			Result: channeltypes.FAILURE,
+		}, nil
+	}
+
+	writeFn()
+
+	upgrade.Fields.Version = version
+
+	// TODO: determine where flushStatus is computed, pendingInflightPackets could be checked within WriteUpgradeTryChannel(?)
+	// the spec currently defines this in startFlushUpgradeHandshake(), if done here it would need to be returned and propagated to this call
+	k.ChannelKeeper.WriteUpgradeTryChannel(ctx, msg.PortId, msg.ChannelId, upgrade, channeltypes.FLUSHING)
+
+	return &channeltypes.MsgChannelUpgradeTryResponse{
+		ChannelId: msg.ChannelId,
+		Version:   "", // todo: return upgrade and upgrade sequence similarly to INIT
+		Result:    channeltypes.SUCCESS,
+	}, nil
 }
 
 // ChannelUpgradeAck defines a rpc handler method for MsgChannelUpgradeAck.
--- a/modules/core/keeper/msg_server.go
+++ b/modules/core/keeper/msg_server.go
@@ -755,13 +755,13 @@ func (k Keeper) ChannelUpgradeTry(goCtx context.Context, msg *channeltypes.MsgCh
 
 	module, _, err := k.ChannelKeeper.LookupModuleByChannel(ctx, msg.PortId, msg.ChannelId)
 	if err != nil {
-		ctx.Logger().Error("channel upgrade init failed", "port-id", msg.PortId, "error", errorsmod.Wrap(err, "could not retrieve module from port-id"))
+		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrap(err, "could not retrieve module from port-id"))
 		return nil, errorsmod.Wrap(err, "could not retrieve module from port-id")
 	}
 
 	cbs, ok := k.Router.GetRoute(module)
 	if !ok {
-		ctx.Logger().Error("channel upgrade init failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
+		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
 		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
 	}
 
@@ -804,7 +804,7 @@ func (k Keeper) ChannelUpgradeTry(goCtx context.Context, msg *channeltypes.MsgCh
 
 	return &channeltypes.MsgChannelUpgradeTryResponse{
 		ChannelId: msg.ChannelId,
-		Version:   "", // todo: return upgrade and upgrade sequence similarly to INIT
+		Version:   "", // todo: return upgrade and upgrade sequence aligning with return args of INIT
 		Result:    channeltypes.SUCCESS,
 	}, nil
 }


modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,62 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+
+	upgrade.Fields.Version = version
+
+	// TODO: determine where flushStatus is computed, pendingInflightPackets could be checked within WriteUpgradeTryChannel(?)
+	// the spec currently defines this in startFlushUpgradeHandshake(), if done here it would need to be returned and propagated to this call
+	k.ChannelKeeper.WriteUpgradeTryChannel(ctx, msg.PortId, msg.ChannelId, upgrade, channeltypes.FLUSHING)
+
+	return &channeltypes.MsgChannelUpgradeTryResponse{
+		ChannelId: msg.ChannelId,
+		Version:   "", // todo: return upgrade and upgrade sequence aligning with return args of INIT
```
damiannolan, line 807: I'll open another PR for this

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,62 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		}, nil
+	}
+
+	writeFn()
+
+	upgrade.Fields.Version = version
+
+	// TODO: determine where flushStatus is computed, pendingInflightPackets could be checked within WriteUpgradeTryChannel(?)
+	// the spec currently defines this in startFlushUpgradeHandshake(), if done here it would need to be returned and propagated to this call
+	k.ChannelKeeper.WriteUpgradeTryChannel(ctx, msg.PortId, msg.ChannelId, upgrade, channeltypes.FLUSHING)
```
damiannolan, line 803: I think it could make more sense to compute the `flushStatus` inside this method

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -195,13 +190,20 @@ func (k Keeper) WriteUpgradeTryChannel(
 
 	previousState := channel.State
 	channel.State = types.TRYUPGRADE
-	channel.FlushStatus = flushStatus
+	// TODO: determine flush status
+	// channel.FlushStatus = flushStatus
```
damiannolan, line 194: should be addressed in a future PR.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -180,12 +180,7 @@ func (k Keeper) ChanUpgradeTry(
 
 // WriteUpgradeTryChannel writes a channel which has successfully passed the UpgradeTry handshake step.
 // An event is emitted for the handshake step.
-func (k Keeper) WriteUpgradeTryChannel(
-	ctx sdk.Context,
-	portID, channelID string,
-	proposedUpgrade types.Upgrade,
-	flushStatus types.FlushStatus,
-) {
+func (k Keeper) WriteUpgradeTryChannel(ctx sdk.Context, portID, channelID string, upgrade types.Upgrade, upgradeVersion string) {
```
damiannolan, line 183: I can open another PR to replicate this on INIT - accepting the `upgradeVersion` as an arg and updating the upgrade inside this method as opposed to in the msg server

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,54 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+
+			// NOTE: a FAILURE result is returned to the client and an error receipt is written to state.
+			return &channeltypes.MsgChannelUpgradeTryResponse{Result: channeltypes.FAILURE}, nil
+		}
+
+		// NOTE: an error is returned to baseapp and transaction state is not committed.
+		return nil, err
+	}
+
+	cacheCtx, writeFn := ctx.CacheContext()
```
damiannolan, line 783: state changes made the by app are not committed on error of the app callback, and instead, an error receipt is written for the current upgrade sequence.

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,54 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+
+	writeFn()
+
+	k.ChannelKeeper.WriteUpgradeTryChannel(ctx, msg.PortId, msg.ChannelId, upgrade, upgradeVersion)
+
+	return &channeltypes.MsgChannelUpgradeTryResponse{
+		Result:    channeltypes.SUCCESS,
+		ChannelId: msg.ChannelId,
+		Upgrade:   upgrade,
+	}, nil
```
damiannolan, line 801: Upgrade sequence is defined in the response but requires a channel lookup in the msg server as its no longer returned from `ChanUpgradeTry`. I'm not a massive fan of doing that lookup at this level. Upgrade sequence is still emitted in events.

anyone have any thoughts on this?

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,54 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+
+		// NOTE: an error is returned to baseapp and transaction state is not committed.
+		return nil, err
+	}
+
+	cacheCtx, writeFn := ctx.CacheContext()
+	upgradeVersion, err := cbs.OnChanUpgradeTry(cacheCtx, msg.PortId, msg.ChannelId, upgrade.Fields.Ordering, upgrade.Fields.ConnectionHops, upgrade.Fields.Version)
+	if err != nil {
+		if err := k.ChannelKeeper.AbortUpgrade(ctx, msg.PortId, msg.ChannelId, err); err != nil {
+			return nil, err
```
DimitrisJim, line 787: guess we should also be logging this error case? Similarly log a successful upgradetry before we return in 796.

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,54 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+	}
+
+	writeFn()
+
+	k.ChannelKeeper.WriteUpgradeTryChannel(ctx, msg.PortId, msg.ChannelId, upgrade, upgradeVersion)
+
+	return &channeltypes.MsgChannelUpgradeTryResponse{
+		Result:    channeltypes.SUCCESS,
+		ChannelId: msg.ChannelId,
+		Upgrade:   upgrade,
```
AdityaSripal, line 800: It looks like this will return an upgrade with the passed-in version. Doesn't it make sense to return with the version set by callback?

modules/core/keeper/msg_server.go
AdityaSripal: Im ok with removing that from response

modules/core/keeper/msg_server.go
damiannolan: aha, yes indeed! nice catch

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,54 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+	}
+
+	upgrade, err := k.ChannelKeeper.ChanUpgradeTry(ctx, msg.PortId, msg.ChannelId, msg.ProposedUpgradeConnectionHops, msg.UpgradeTimeout, msg.CounterpartyProposedUpgrade, msg.CounterpartyUpgradeSequence, msg.ProofChannel, msg.ProofUpgrade, msg.ProofHeight)
+	if err != nil {
+		if upgradeErr, ok := err.(*channeltypes.UpgradeError); ok {
+			if err := k.ChannelKeeper.AbortUpgrade(ctx, msg.PortId, msg.ChannelId, upgradeErr); err != nil {
+				return nil, err
+			}
+
+			// NOTE: a FAILURE result is returned to the client and an error receipt is written to state.
```
colin-axner, line 775: nit: I think this inline comment could give more detail on why this is happening

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,54 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+
+	cbs, ok := k.Router.GetRoute(module)
+	if !ok {
+		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
+		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+	}
+
+	upgrade, err := k.ChannelKeeper.ChanUpgradeTry(ctx, msg.PortId, msg.ChannelId, msg.ProposedUpgradeConnectionHops, msg.UpgradeTimeout, msg.CounterpartyProposedUpgrade, msg.CounterpartyUpgradeSequence, msg.ProofChannel, msg.ProofUpgrade, msg.ProofHeight)
+	if err != nil {
+		if upgradeErr, ok := err.(*channeltypes.UpgradeError); ok {
```
colin-axner, line 770: What do we think about `channeltypes.IsUpgradeError(err)`? Just throwing the idea out there

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,54 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
+		return nil, errorsmod.Wrap(err, "could not retrieve module from port-id")
+	}
+
+	cbs, ok := k.Router.GetRoute(module)
+	if !ok {
+		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
+		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+	}
+
+	upgrade, err := k.ChannelKeeper.ChanUpgradeTry(ctx, msg.PortId, msg.ChannelId, msg.ProposedUpgradeConnectionHops, msg.UpgradeTimeout, msg.CounterpartyProposedUpgrade, msg.CounterpartyUpgradeSequence, msg.ProofChannel, msg.ProofUpgrade, msg.ProofHeight)
```
colin-axner, line 768: Note: if this contained state changes, we would need to be doing a cache context around it. I wonder if we should do this defensively?

colin-axner: Food for thought

It looks like we have a pattern:
```go
writeFn() := cacheCtx()
err := funcCall()
if isUpgradError() {
    return Abort()
}

writeFn()
```

I wonder if we can abstract this into a helper function? Like `handleUpgrade()`? 

I guess the OnChanUpgradeTry func would need to provide:
```go
func() {
    err := OnChanUpgradeTry()
    if err != nil {
        return types.NewUpgradeError(err)
    }
    
    return nil
}
```

modules/core/keeper/msg_server.go
colin-axner: Agreed. Upgrade set here should have updated version

modules/core/keeper/msg_server.go
colin-axner: I agree with not doing the lookup here, but maybe we could abstract into a function:
```go
return formatUpgradeTrySuccessResponse()
```
or something. No strong preference, happy to remove now and add in the future upon request

modules/core/keeper/msg_server_test.go
```diff
@@ -774,3 +777,143 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
...
+			"application callback error writes upgrade error receipt",
+			func() {
+				suite.chainB.GetSimApp().IBCMockModule.IBCApp.OnChanUpgradeTry = func(
+					ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, counterpartyVersion string,
+				) (string, error) {
+					return "", fmt.Errorf("mock app callback failed")
+				}
+			},
+			func(res *channeltypes.MsgChannelUpgradeTryResponse, err error) {
+				suite.Require().NoError(err)
```
colin-axner, line 863: should this assert no state change occurred?

modules/core/keeper/msg_server_test.go
```diff
@@ -774,3 +777,143 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
...
+			counterpartyUpgrade, found := suite.chainA.GetSimApp().GetIBCKeeper().ChannelKeeper.GetUpgrade(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+			suite.Require().True(found)
+
+			proofChannel, proofUpgrade, proofHeight := path.EndpointB.QueryChannelUpgradeProof()
+
+			msg = &channeltypes.MsgChannelUpgradeTry{
+				PortId:                        path.EndpointB.ChannelConfig.PortID,
+				ChannelId:                     path.EndpointB.ChannelID,
+				ProposedUpgradeConnectionHops: []string{ibctesting.FirstConnectionID},
+				UpgradeTimeout:                channeltypes.NewTimeout(clienttypes.NewHeight(1, 100), 0),
```
colin-axner, line 903: `path.EndpointA.Chain.GetTimeoutHeight()`?

colin-axner (approved):
Ay! Looks super clean! I left several suggestions, but happy to push any you agree with into issues and handle them in followups

modules/core/keeper/msg_server.go
```diff
@@ -751,7 +751,54 @@ func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgC
...
 // ChannelUpgradeTry defines a rpc handler method for MsgChannelUpgradeTry.
 func (k Keeper) ChannelUpgradeTry(goCtx context.Context, msg *channeltypes.MsgChannelUpgradeTry) (*channeltypes.MsgChannelUpgradeTryResponse, error) {
-	return nil, nil
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	module, _, err := k.ChannelKeeper.LookupModuleByChannel(ctx, msg.PortId, msg.ChannelId)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade try failed", "port-id", msg.PortId, "error", errorsmod.Wrap(err, "could not retrieve module from port-id"))
+		return nil, errorsmod.Wrap(err, "could not retrieve module from port-id")
+	}
```
chatton, line 760: [nit] should we add the port ID as a format argument here?

modules/core/keeper/msg_server.go
chatton: I was thinking something similar actually,I think it would be nice if we could do something like

```golang
defer handleAbort(...)
err := OnChanUpgradeTry(...)
...
```

I think we can leave this for now though and maybe consider it later.

modules/core/keeper/msg_server.go
chatton: The check happens enough times (in tests and in callbacks) that this might be no harm. and help a bit in readability. It could also handle the nil check maybe to reduce levels of indentation here.

```golang
func IsUpgradeError(err error) bool {
   if err == nil {
     return false
   }
  _, ok := err.(*channeltypes.UpgradeError)
  return ok
}

```

modules/core/keeper/msg_server_test.go
```diff
@@ -774,3 +777,143 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
...
+func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
+	var (
+		path *ibctesting.Path
+		msg  *channeltypes.MsgChannelUpgradeTry
+	)
+
+	cases := []struct {
+		name      string
+		malleate  func()
+		expResult func(res *channeltypes.MsgChannelUpgradeTryResponse, err error)
```
chatton, line 790: [nit] usually I would imagine `expResult` is the expected result, but this is an assertion function. Maybe `assertionFn`?

chatton (approved):
Nice job this looks great! There were just a few small nits but I don't think anything should block this going forward.  Main thing to change that's been pointed out is making sure the correct version is returned in the response 👍

modules/core/keeper/msg_server_test.go
damiannolan: Yes! I had forgotten about doing this. I added an arbitrary `store.Set()` in the app callback and asserted false for `store.Has()` in the expected result func

modules/core/keeper/msg_server.go
damiannolan: I'm fine with this as is, but if we want to make additional changes we can open a follow up

modules/core/keeper/msg_server.go
damiannolan: added as additional context

> // This signals to the relayer to begin the cancel upgrade handshake subprotocol.

modules/core/keeper/msg_server.go
damiannolan: this is consistent with the rest of the handlers, the wrapped error bubbled up from the scopedKeepers `LookupModules` method (https://github.com/cosmos/ibc-go/blob/main/modules/capability/keeper/keeper.go#L439).

I think the `name` should be in the wrapped error which should give the user port info

modules/core/keeper/msg_server_test.go
damiannolan: that's a fair point, I think we have this test setup in some extra places in the codebase using `expResult`, so maybe we can do one swoop to change all if people feel its worth it. 

(I know we have inconsistencies with `expPass`, `expError` and `expResult` in different places - maybe we can keep some housekeeping work for a rainy day (?))

modules/core/keeper/msg_server.go
damiannolan: Two potential ways of handling this:

1. pass `Upgrade` to the `WriteUpgradeTryChannel` method as a reference, pointer type.
2. Return the `Upgrade` or even possibly the entire response from the `WriteUpgradeTryChannel` method.

do you guys favour either?

damiannolan: We could also go for this returning the upgrade and the sequence for the response

```go
func (k Keeper) WriteUpgradeTryChannel(...) (Upgrade, uint64)
```

modules/core/keeper/msg_server.go
colin-axner: opened #3821 to move discussion there

modules/core/keeper/msg_server_test.go
colin-axner: `assertionFn` sounds nice to me. Agree it should be done in a single swoop

modules/core/keeper/msg_server.go
damiannolan: Agree using `cacheCtx` here provides more defence in depth but restoring the channel in `AbortUpgrade` is supposed to delete the upgrade and set the channel back to its open and operational state.

Let's open an issue to discuss this later. @colin-axner made great points about INIT events being emitted as false positives in an error situation from [here](https://github.com/cosmos/ibc-go/blob/04-channel-upgrades/modules/core/04-channel/keeper/upgrade.go#L134)

modules/core/keeper/msg_server.go
AdityaSripal: You could return the entire response I think

modules/core/keeper/msg_server.go
colin-axner: #3823

colin-axner: #3824

colin-axner (approved):
Followup ACK

chatton (approved):
Followup ACK 👍

PR #3808, opened 2023-06-09
sdk 50

faddat:
This is a very rough draft pr that upgrades ibc-go to sdk v0.50.x -- the idea is to have 
this software available to speed up deployment on chains and in common ecosystem 
libraries like wasmd.

- sdk 50 first go
- add go workspace


crodriguezvega:
Thank you @faddat for kicking off this upgrade! I just created a feature branch for the upgrade to v0.50 ([feat/upgrade-sdk-v0.50](https://github.com/cosmos/ibc-go/tree/feat/upgrade-sdk-v0.50)) and I have re-targeted the PR to that branch. Let's target PRs for the upgrade to v0.50 to that branch.

faddat:
Awesome, thank you @crodriguezvega :)

modules/core/02-client/abci_test.go
```diff
@@ -67,12 +66,9 @@ func (suite *ClientTestSuite) TestBeginBlockerConsensusState() {
 	err := suite.chainA.GetSimApp().UpgradeKeeper.SetUpgradedClient(newCtx, plan.Height, []byte("client state"))
 	suite.Require().NoError(err)
 
-	req := abci.RequestBeginBlock{Header: newCtx.BlockHeader()}
```
faddat, line 70: this is what I have been removing

modules/apps/transfer/keeper/migrations_test.go
```diff
@@ -194,8 +194,8 @@ func (suite *KeeperTestSuite) TestMigrateTotalEscrowForDenom() {
 				coin2 := sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(100))
 
 				// funds the escrow accounts to have balance
-				suite.Require().NoError(banktestutil.FundAccount(suite.chainA.GetSimApp().BankKeeper, suite.chainA.GetContext(), escrowAddress1, sdk.NewCoins(coin1)))
-				suite.Require().NoError(banktestutil.FundAccount(suite.chainA.GetSimApp().BankKeeper, suite.chainA.GetContext(), escrowAddress2, sdk.NewCoins(coin2)))
+				suite.Require().NoError(banktestutil.FundAccount(suite.chainA.GetContext(), suite.chainA.GetSimApp().BankKeeper, escrowAddress1, sdk.NewCoins(coin1)))
```
faddat, line 197: Here's another spot where I think I got things right but I'm not sure @tac0turtle

tac0turtle:
> In this PR I have removed a number of BeginBlock and EndBlock calls that are found in tests. I think that this is the correct thing to do because they are now called automatically, but I am not 100% sure.

you are correct here

modules/apps/transfer/keeper/migrations_test.go
tac0turtle: what is the question here?

faddat: I'm sorry, on second glance I think that this was not the correct thing.  I will mark it again when I reencounter and thanks a ton for your reply!

crodriguezvega:
@faddat, can you please rebase your branch with the latest from the [feature branch](https://github.com/cosmos/ibc-go/tree/feat/upgrade-sdk-v0.50)?

tac0turtle (commented):
some of these changes like adoption of cosmossdk.io/math could be done on main to reduce the amount of changes in the feature branch. It could help reviewers with reduced noise in their reviews

alpe:
Thanks @faddat for starting the upgrade and moving x/capability over. This is helpful for the wasmd upgrade, too. 
Dependencies are still not in sync. Can you ensure the correct project path is set for `capability` on the modules?

`"github.com/cosmos/ibc-go/modules/capability` should be `github.com/cosmos/ibc-go/v7/modules/capability`

I guess  v7 is likely v8 later but to be consistent for now

faddat:
@tac0turtle after looking through the cosmos-sdk, it seems like we should get the validator set updates from the endblocker.  

Is that correct?

Also I'm not really sure how to do that -- all issues are now confined to two files:

chain.go and app.go in the testing folder (not simapp)

faddat:
@alpe the tidy problem that you mentioned seems to have something to do with the migration code and when I might try doing is just checking out the main branch I'm not really sure what's wrong with it

faddat:
Hey, the biggest remaining issue is this:

```
    /Users/faddat/ibc-go/modules/apps/transfer/keeper/suite.go:87: test panicked: type_url /ibc.applications.fee.v1.MsgRegisterPayee has not been registered yet. Before calling RegisterService, you must register all interfaces by calling the `RegisterInterfaces` method on module.BasicManager. Each module should call `msgservice.RegisterMsgServiceDesc` inside its `RegisterInterfaces` method with the `_Msg_serviceDesc` generated by proto-gen
```

And interestingly I have hit this several times recently, always when dealing with situations where I've rebuilt / redid protos.  Anyone have any ideas?

faddat:
The next most frequent error is:

```
                Error:          Received unexpected error:
                                invalid height: 1; expected: 2
                Test:           TestTypesTestSuite/TestMarshalClientUpdateProposalProposal
    --- FAIL: TestTypesTestSuite/TestMarshalConsensusStateWithHeight (0.06s)
        chain.go:292: 
                Error Trace:    /Users/faddat/ibc-go/testing/chain.go:292
                                                        /Users/faddat/ibc-go/testing/coordinator.go:185
                                                        /Users/faddat/ibc-go/testing/chain.go:158
                                                        /Users/faddat/ibc-go/testing/chain.go:187
                                                        /Users/faddat/ibc-go/testing/coordinator.go:40
                                                        /Users/faddat/ibc-go/modules/core/02-client/types/msgs_test.go:28
                                                        /Users/faddat/go/pkg/mod/github.com/stretchr/testify@v1.8.4/suite/suite.go:187
```

Anyone know where we set expected for this?

tac0turtle:
> after looking through the cosmos-sdk, it seems like we should get the validator set updates from the endblocker.
> 
> Is that correct?

updating the docs here https://github.com/cosmos/cosmos-sdk/pull/16501. just about done but it has the begin and endblocker changes

modules/capability/go.mod
```diff
@@ -1,18 +1,18 @@
-module github.com/cosmos/ibc-go/modules/capability
+module github.com/cosmos/ibc-go/modules/capability/v7
```
alpe, line 1: The path is not correct. It should resolve to `github.com/cosmos/ibc-go/v7/modules/capability` instead. I am wondering why you need a `go.mod` here at all? The other packets in "module" don't have a `go.mod` and inherit the module path from the root go.mod.

faddat: I dont think that you're right here, because it is a multi module repository.... I think

And the reason there's a module there is the work on modularization going on in the sdk and also in ibc

alpe: This project is has `go.mod` files in root and e2e dirs. Everything within `modules/` is within the same `github.com/cosmos/ibc-go/v7` Go module path.
It can make sense to split every subpackage in their own Go module but this should not be part of this PR.
Let's focus on a working example. We should not assume the release cycle of `capability` is different than the rest of the project. Using different Go modules adds quite some complexity to the project.
The SDK uses the `cosmossdk.io` domain and different tags for example to not conflict on the root path.

faddat: the modularization isn't a part of this PR, it's present in the main branch and the branch we're targeting, and isn't something new to this PR

alpe: Apologies, you are right with the main branch. Seems that I was looking at an older version. I am a bit lost now. 🤷 
There is a versioned path for the main repo but no versioned path for the capability. I would appreciate some guidance how to integrate the capability now. I  👀  3 different Go paths
* github.com/cosmos/ibc-go/v7
* github.com/cosmos/ibc-go/modules/capability (main)
* github.com/cosmos/ibc-go/modules/capability/v7 (notional)

faddat: might as well make ours match main, my changes were mainly to see if it solved the issue that you were facing.

alpe: quick heads up: it resolves with 2 replace statements now. Thanks @faddat !
```
github.com/cosmos/ibc-go/modules/capability => github.com/notional-labs/ibc-go/modules/capability v0.0.0-20230616092119-6e16d6223af4
github.com/cosmos/ibc-go/v7 => github.com/notional-labs/ibc-go/v7 v7.0.0-20230616092119-6e16d6223af4
```

faddat:
My most recent commit breaks a lot of things.  I think there may be a design question for @crodriguezvega & the IBC team -- The thing that this seems to break most is the testing style that is used here in ibc-go. 

Maybe we line up a call?  

Note: I made this commit in part by using the goland side by side view and looking at the diff between the SDK's simapp.  Where possible, I made ibc's simapp look like sdk's simapp.

faddat:
nb:

https://github.com/cosmos/cosmos-sdk/blob/13e61bd0c5c7950d137bbdfb171f95d87179ed29/CHANGELOG.md?plain=1#L43

This is at a minimum going to affect how we work with packet.go

```go
// GetBytes returns the JSON marshalled interchain account packet data.
func (iapd InterchainAccountPacketData) GetBytes() []byte {
	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&iapd))
}

// GetBytes returns the JSON marshalled interchain account CosmosTx.
func (ct CosmosTx) GetBytes() []byte {
	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&ct))
}
```

I assume that is a determinsim hack to ensure all nodes see the same thing?

faddat:
Yesterday I did a pretty interesting exercise where I tried to make the simapp folder match the SDK's as closely as I could.

I am going to do the same today, but with the main branch.  

I think this will reduce the changeset.

faddat:
Also might start this branch over, depends on findings from that.

faddat:
@crodriguezvega hey, I think that what I will do here, is cherry pick commits beginning iwth the upgrade-sdk-v0.50 branch, just to keep this cleaner.  I have merged main here, for example and I don't think that is desired.

faddat:
closing in favor of a new branch

PR #3809, opened 2023-06-10
ci: use thelper

faddat:
This PR removes the depguard linter and enables thelper. It will make upgrading 
to v50 easier. 

- tidy and enable thelper
- use thelper and disable depguard
- use latest golangci-lint in ci


crodriguezvega:
Hey @faddat, can you please give us a bit more of context?

- What's the reason to remove depguard?
- What benefits does thelper bring?
- Why would these changes make upgrading to SDK v0.50 easier?

Thanks!

faddat:
Sure I do very happy to tell you a bit more about this one, so the thelper linter really just reminds devs to use the call t.Helper()

And when you do that, your IDE can give you some additional information for debugging. I would like to use it during the upgrade to SDK 50 because that useful information just makes the job easier.

depguard .... I got to tell you I think it never worked. But you see them the thing is it started working and when it began to work it started doing some kind of bad stuff, like blocking every single cosmos import, so I pulled it out.

Anyhow this should make every upgrade easier, not just SDK 50

modules/apps/27-interchain-accounts/host/client/cli/tx_test.go
```diff
@@ -137,6 +141,7 @@ func TestGeneratePacketData(t *testing.T) {
 }
 
 func assertMsgBankSend(t *testing.T, msg sdk.Msg) {
+	t.Helper()
```
chatton, line 144: I'm unsure if this should actually be flagged a helper function, this is an assertion function, with the same purpose as something like `require.True(t, ok)`. What is the benefit of marking this as a helper? (same applies to the functions that begin with `assert`)

DimitrisJim: Am I understanding this correctly that `t.Helper` will point up the call stack instead of the specific line the error occurs at? The idea being multiple functions call the helper?

DimitrisJim:
> What's the reason to remove depguard?

Looking into this can't see why we already have `depguard`, there's no configuration for it here or in the sdk. Seems like it can safely be removed.

modules/apps/27-interchain-accounts/host/client/cli/tx_test.go
chatton: I think we would want the test to fail at a line in the assertionFn though right, .e.g. I would want to know that this line failed

`require.Equal(t, "cosmosvaloper1qnk2n4nlkpw9xfqntladh74w6ujtulwnmxnh3k", msgDelegate.ValidatorAddress)`

rather than

`tc.assertMsgDelegate(t, msg)`

unless I am misunderstanding something

DimitrisJim: Am understanding the same thing so I'm now less inclined to believe I'm not understanding something.

+1 on wanting the error trace to point at most relevant spot.

faddat: how abvout the assertMsgDelegate below?

Should that have it?

chatton: So my initial thought was that it the same logic should apply, however I just ran some tests that I modified to fail, and the stack trace points exactly where I would expect with and without the `t.Helper()` call. (at the failing assertion within the `assertMsgDelegate` )

I'm not actually noticing a difference in the output.

DimitrisJim (approved):
lgtm!

PR #3810, opened 2023-06-10
barberry

faddat:
This PR patches ibc-go for barberry, which also means making a few other version 
bumps, too.


go.mod
```diff
@@ -1,4 +1,4 @@
-go 1.19
+go 1.20
```
crodriguezvega, line 1: Could we bump go version in a separate PR? I believe this is not strictly needed for SDK 0.47.3, right?

crodriguezvega (changes requested):
Thanks, @faddat! I only left one comment: I think we should bump go version in a separate PR.

faddat:
Sure thing, sorry about the duplication here. I'll take care of that right now also

faddat:
@crodriguezvega hey another thing I should make you aware of with this one -- it moves go.work.example to go.work, I am going to make that a separate PR though. 

I think we should do that because it is the way that Osmosis is handling this, and it is really nice.   Basically, we can:

keep go.work.sum in .gitignotre
commit go.work

this way when a dev sits down to work, they can 

```bash
go work sync
```

and everything goes great.  I usually label the resulting commit "sync"

colin-axner (approved):
Thanks @faddat

PR #3813, opened 2023-06-12
ci: use go.work while ignoring go.work.sum

faddat:
This PR is aimed at making go workspaces easier for developers to use. 

* editors will recognize the workspace and use it
* develoers can go work sync instead of updating many dependencies at once


DimitrisJim:
I'd personally be slightly opposed to this. We currently don't have any sync-up issues and the go.work.example suffices for anyone needing it. I personally think this is a case of if it ain't broke, don't fix it.

IIRC VSCode really has an issue with this not being present but other editors do not seem to mind that much (at the moment, at least). I'd at least want other folks who don't use VSCode to chime in and give the ok that pulling this change doesn't result in them needing to dig around in some configuration hole for their editor/language server.

Issue #3643, opened 2023-05-24
Make e2e tests a required check


DimitrisJim:
## Summary

Currently, e2e tests aren't a required check before merging/automerging. This could cause some headaches in cases where we merge from main and then automerge, an error might be introduced which isn't noticed until after the automerge has been performed.

This wasn't done previously due to some non-deterministic failures which, as far as I'm aware, have been resolved.

## Proposal

Make e2e tests a required check before merging.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Looks like we need to explicitly skip fork and non-fork e2e's in certain situations. Right now, enabling e2e's for all will cause pr's to block as fork e2e's don't run on a non-fork. See [github docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks)

DimitrisJim:
hopefully this should be straight-forward enough that I can add these workflows until Monday. It seems we just need to `NOT` the conditions defined in each but I'll open a couple of test PRs at some point to check.

DimitrisJim:
Some additional comments here.

It seems that currently, due to the way e2e's are set up, we aren't able to mark them as required. The problem is due to the fact that e2e's for normal PRs and e2e's for forks are launched differently:

 - e2es for normal PRs basically [launch sub-jobs](https://github.com/cosmos/ibc-go/blob/c1f737db34add198c9410476272133d52e20ac5e/.github/workflows/e2e.yaml#L67) that get executed. This means that from Github's POV, the job that launched them is independent from the launched jobs. Skipping the main job does not skip the sub jobs but instead marks them as pending.
 - e2es for fork-prs are run as part of the `e2e` job using a matrix, [reference](https://github.com/cosmos/ibc-go/blob/c1f737db34add198c9410476272133d52e20ac5e/.github/workflows/e2e-fork.yml#L40). The end result being that when the `e2e` job gets skipped, they all get skipped and are marked as successful.

As a result, we can't set the `e2e` job as a requirement for branch protection, using it results in false-positives since, for example, on a normal PR we'd have:

 1. e2e job marked as required.
 2, e2e job for the fork-pr workflow is skipped and considered passing.
 3. e2e job marked as required is then considered successful even if a particular e2e test fails.

[...]

PR #3818, opened 2023-06-12
Mark e2e tests as successful when non-code PRs are pushed.

DimitrisJim:
## Description

A suggestion. Due to the way github actions works with path ignores, PRs that don't target them will be unable to fulfill the required checks qualification. Adds workflows to mark the required e2e tests as passing if a PR that doesn't target their required paths is opened.

requires: #3884 

closes: #3643

```bash
chore: mark e2e tests as successful when non-code PRs are pushed
```

---

write a little note why.

### DIFF
--- a/.github/workflows/ignore-e2e-fork.yml
+++ b/.github/workflows/ignore-e2e-fork.yml
@@ -1,0 +1,21 @@
+# This workflow is added in order for the required e2e/build job to run successfully even
+# when the PR targets non-code files (e.g. docs, LICENSE, etc). This is done because of the
+# way Github required checks work, reference:
+#
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: Tests / E2E Fork
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "docs/**"
+      - "**.md"
+      - "LICENSE"
+
+jobs:
+    e2e:
+        runs-on: ubuntu-latest
+        steps:
+            - run: echo "E2E-fork tests skipped for docs."
--- a/.github/workflows/ignore-e2e-fork.yml
+++ b/.github/workflows/ignore-e2e-fork.yml
@@ -1,4 +1,4 @@
-# This workflow is added in order for the required e2e/build job to run successfully even
+# This workflow is added in order for the required e2e-fork job to run successfully even
 # when the PR targets non-code files (e.g. docs, LICENSE, etc). This is done because of the
 # way Github required checks work, reference:
 #
@@ -15,7 +15,7 @@ on:
       - "LICENSE"
 
 jobs:
-    e2e:
+    e2e-fork:
         runs-on: ubuntu-latest
         steps:
             - run: echo "E2E-fork tests skipped for docs."
--- a/.github/workflows/ignore-e2e-fork.yml
+++ b/.github/workflows/ignore-e2e-fork.yml
@@ -1,4 +1,4 @@
-# This workflow is added in order for the required e2e-fork job to run successfully even
+# This workflow is added in order for the required e2e-fork-status job to complete successfully (skipped) even
 # when the PR targets non-code files (e.g. docs, LICENSE, etc). This is done because of the
 # way Github required checks work, reference:
 #
@@ -15,7 +15,7 @@ on:
       - "LICENSE"
 
 jobs:
-    e2e-fork:
+    e2e-fork-status:
         runs-on: ubuntu-latest
         steps:
             - run: echo "E2E-fork tests skipped for docs."
--- a/.github/workflows/ignore-e2e.yml
+++ b/.github/workflows/ignore-e2e.yml
@@ -23,4 +23,4 @@ jobs:
     e2e:
         runs-on: ubuntu-latest
         steps:
-            - run: echo "E2E-fork tests skipped for docs."
+            - run: echo "E2E tests skipped for docs."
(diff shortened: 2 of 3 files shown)


.github/workflows/ignore-e2e.yml
```diff
@@ -0,0 +1,26 @@
...
+    paths:
+      - "docs/**"
+      - "**.md"
+      - "LICENSE"
+
+jobs:
+    e2e:
+        runs-on: ubuntu-latest
+        steps:
+            - run: echo "E2E-fork tests skipped for docs."
```
colin-axner, line 26:
```suggestion
            - run: echo "E2E tests skipped for docs."
```

.github/workflows/ignore-e2e-fork.yml
```diff
@@ -0,0 +1,21 @@
+# This workflow is added in order for the required e2e/build job to run successfully even
+# when the PR targets non-code files (e.g. docs, LICENSE, etc). This is done because of the
+# way Github required checks work, reference:
+#
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: Tests / E2E Fork
+on:
```
colin-axner, line 7: should we run this on non-fork branches as well?

.github/workflows/ignore-e2e.yml
```diff
@@ -0,0 +1,26 @@
+# This workflow is added in order for the required e2e/build job to run successfully even
+# when the PR targets non-code files (e.g. docs, LICENSE, etc). This is done because of the
+# way Github required checks work, reference:
+#
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: Tests / E2E
+on:
```
colin-axner, line 7: should we run this on forks?

colin-axner (commented):
LGTM! See suggestions. Would like to hear @chatton's opinion

.github/workflows/ignore-e2e-fork.yml
DimitrisJim: I think this passes fine for tests whose workflow jobs are skipped with an `if` (i.e, the result of the job is Skipped and not Pending)

DimitrisJim: Seems Github makes a distinction on skipping the whole workflow and skipping all the jobs of a workflow.

Skipping the workflow as a whole is done with path filtering, branch filtering as mentioned in the [docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks). Doing this skip results in its status being reported as 'Pending' thereby blocking the required checks from being fulfilled.

Skipping jobs inside a workflow file can be done using expressions, we currently do this to not run our e2es on drafts e.g [here](https://github.com/cosmos/ibc-go/blob/b5dadfdd99addb381fbf0b8ca0e7070d5aff7891/.github/workflows/e2e.yaml#L21). Doing these skips marks the job as being 'Success', allowing the required checks to complete.

>  Just want to double check that if we mark e2e fork tests as required, if I open a pr against main which makes code changes, that test will be skipped for me

If my understanding of their docs is correct, if we mark e2e-fork as required, this shouldn't hold up a pull request.

colin-axner: That seems to make sense to me, but I'm just confused why marking both fork and non-fork e2e tests previously resulted in a "pending" state for pr's? I tried this with fork pr's by marking an e2e test as required and then updating a dependabot pr which should the test in "pending"

e2e/tests/transfer/base_test.go
```diff
@@ -43,6 +43,9 @@ func (s *TransferTestSuite) TestMsgTransfer_Succeeds_Nonincentivized() {
 	t := s.T()
 	ctx := context.TODO()
 
+	// TEST
+	s.Require().True(false)
```
colin-axner, line 47: Testing the discussion [here](https://github.com/cosmos/ibc-go/pull/3818#discussion_r1226639511)

colin-axner: requires renaming the fork e2e before trying again, looks like there is a naming collision, so even though `e2e` fails on native repo, it skips on fork which happens first

.github/workflows/ignore-e2e.yml
colin-axner: Sorry, my comment was asking if this should be run on forks for code changes, as it seems fork pr's currently hang on required E2E's

chatton: that's slightly tricker, if we want to make this run on code changes for forks it would run on code changes on regular PRs as well

chatton: Maybe we can make the existing E2E task `exit 0` if it's a fork and pass?

DimitrisJim: there's some missing context since Colins last comment, seems the hanging was caused, partly, by `e2e-fork.yml` and `e2e,yml` having the same job name for running the tests, `e2e`. A first step would be going with the renaming #3827 and adding the new job names in the branch protection rules.

I think the workflows to be added in this PR are a different issue that's unrelated. These just make sure that the job will be skipped correctly on non-code prs due to the way github reports a job status if it is skipped with a `path` filter on `on`.

DimitrisJim: hopefully, it definitely appears to be the first right step. After it gets merged we can trigger some runs/add it to branch protection and see how it plays.

DimitrisJim:
closing this for now, need to revisit given some time to consider wasm tests

Issue #3741, opened 2023-06-05
Implement `ChanUpgradeAck` handler


colin-axner:
## Summary

Implement `ChanUpgradeAck` handler as outlined by the spec

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3828, opened 2023-06-13
feat: adding `chanUpgradeAck` handler to 04-channel

damiannolan:
## Description

Adding 04-channel handler for `ChanUpgradeAck`, msg server implementation PR will follow this.

closes: #3741

---

write a little note why.


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -191,7 +191,7 @@ func (k Keeper) WriteUpgradeTryChannel(ctx sdk.Context, portID, channelID string
 	previousState := channel.State
 	channel.State = types.TRYUPGRADE
 	// TODO: determine flush status
-	// channel.FlushStatus = flushStatus
+	channel.FlushStatus = types.FLUSHING
```
damiannolan, line 194: currently setting to flushing as default for now, as flush status must be in flushing or flush complete to proceed within the handler tests.

testing/endpoint.go
```diff
@@ -790,6 +790,20 @@ func (endpoint *Endpoint) SetChannel(channel channeltypes.Channel) {
 	endpoint.Chain.App.GetIBCKeeper().ChannelKeeper.SetChannel(endpoint.Chain.GetContext(), endpoint.ChannelConfig.PortID, endpoint.ChannelID, channel)
 }
 
+// GetChannelUpgrade retrieves an IBC Channel Upgrade for the endpoint. The upgrade
+// is expected to exist otherwise testing will fail.
+func (endpoint *Endpoint) GetChannelUpgrade() channeltypes.Upgrade {
```
damiannolan, line 795: added these test helpers, can follow up with replacing usage in another PR:

```diff
- upgrade, found := suite.chainA.GetSimApp().GetIBCKeeper().ChannelKeeper.GetUpgrade(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
- suite.Require().True(found)
+ upgrade := path.EndpointA.GetChannelUpgrade()
```

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -325,6 +325,174 @@ func (suite *KeeperTestSuite) TestChanUpgradeTry() {
...
+		{
+			"upgrade not found",
+			func() {
+				store := suite.chainA.GetContext().KVStore(suite.chainA.GetSimApp().GetKey(exported.ModuleName))
+				store.Delete(host.ChannelUpgradeKey(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID))
+			},
+			false,
+		},
+		{
+			"channel end version mismatch on crossing hellos",
```
damiannolan, line 421: spoke with @AdityaSripal about this yesterday, testing this case as a full integration with the testing lib is tricky as its a timing sensitive state we end up, which would require a historical proof being submitted for the second chain to transition to TRY. 

I chose to override the channel state directly here in order to cover this case.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -208,12 +208,70 @@ func (k Keeper) AbortUpgrade(ctx sdk.Context, portID, channelID string, err erro
...
+			return types.NewUpgradeError(channel.UpgradeSequence, errorsmod.Wrap(types.ErrIncompatibleCounterpartyUpgrade, "both channel ends must agree on the same version"))
+		}
+	}
+
+	if err := k.startFlushUpgradeHandshake(ctx, portID, channelID, upgrade.Fields, counterpartyChannel, counterpartyUpgrade,
+		proofChannel, proofUpgrade, proofHeight); err != nil {
+		return err
+	}
+
+	return nil
```
damiannolan, line 268: Linter triggered me on this piece of code so I tweaked the settings. If people prefer returning directly with the func call then we can do that. Otherwise I'll open a PR to main with the linter config change

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -325,6 +325,174 @@ func (suite *KeeperTestSuite) TestChanUpgradeTry() {
...
+	var (
+		path                    *ibctesting.Path
+		counterpartyFlushStatus types.FlushStatus
+		counterpartyUpgrade     types.Upgrade
+	)
+
+	testCases := []struct {
+		name     string
+		malleate func()
+		expPass  bool
```
damiannolan, line 338: I can refactor this to `expError` if people think its worth doing, can also do it in a follow up PR to this one to keep the ball moving

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -208,12 +208,70 @@ func (k Keeper) AbortUpgrade(ctx sdk.Context, portID, channelID string, err erro
 	return nil
 }
 
+// ChanUpgradeAck is called by a module to accept the ACKUPGRADE handshake step of the channel upgrade protocol.
```
colin-axner, line 211: godoc is from the spec? I think it could be improved. Maybe we can indicate that this function should only be called by the core IBC msg server?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -208,12 +208,70 @@ func (k Keeper) AbortUpgrade(ctx sdk.Context, portID, channelID string, err erro
...
+		return errorsmod.Wrapf(types.ErrInvalidChannelState, "expected one of [%s, %s], got %s", types.INITUPGRADE, types.TRYUPGRADE, channel.State)
+	}
+
+	if !collections.Contains(counterpartyFlushStatus, []types.FlushStatus{types.FLUSHING, types.FLUSHCOMPLETE}) {
+		return errorsmod.Wrapf(types.ErrInvalidFlushStatus, "expected one of [%s, %s], got %s", types.FLUSHING, types.FLUSHCOMPLETE, counterpartyFlushStatus)
+	}
+
+	connection, err := k.GetConnection(ctx, channel.ConnectionHops[0])
+	if err != nil {
+		return errorsmod.Wrapf(err, "failed to retrieve connection: %s", channel.ConnectionHops[0])
```
colin-axner, line 237: I believe `GetConnection` returns the connection ID that could not be found. We should indicate that this connection is associated with the channel ID trying to perform the ACK step

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -208,12 +208,70 @@ func (k Keeper) AbortUpgrade(ctx sdk.Context, portID, channelID string, err erro
...
+
+	counterpartyHops := []string{connection.GetCounterparty().GetConnectionID()}
+	counterpartyChannel := types.Channel{
+		State:           types.TRYUPGRADE,
+		Ordering:        channel.Ordering,
+		ConnectionHops:  counterpartyHops,
+		Counterparty:    types.NewCounterparty(portID, channelID),
+		Version:         channel.Version,
+		UpgradeSequence: channel.UpgradeSequence,
+		FlushStatus:     counterpartyFlushStatus,
```
colin-axner, line 248: what do you think about adding a comment indicating this value is provided by the relayer?

.golangci.yml
```diff
@@ -60,3 +60,7 @@ linters-settings:
     allow-leading-space: true
     require-explanation: false
     require-specific: false
+  revive:
+    rules:
+      - name: if-return
+        disabled: true
```
colin-axner, line 66: ++

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -208,12 +208,70 @@ func (k Keeper) AbortUpgrade(ctx sdk.Context, portID, channelID string, err erro
...
+	if !found {
+		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "failed to retrieve channel upgrade: port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	// in the crossing hellos case, the versions returned by both on TRY must be the same
+	if channel.State == types.TRYUPGRADE {
+		if upgrade.Fields.Version != counterpartyUpgrade.Fields.Version {
+			return types.NewUpgradeError(channel.UpgradeSequence, errorsmod.Wrap(types.ErrIncompatibleCounterpartyUpgrade, "both channel ends must agree on the same version"))
+		}
+	}
```
colin-axner, line 261: we haven't verified the counterparty upgrade yet. This allows a relayer to cause a failed upgrade error. It should probably be moved into `startFlushUpgradeHandshake` and we should create a call to `checkUpgradesForCompatibility`?

For now it can be done after `startFlushUpgradeHandshake`

modules/core/04-channel/keeper/upgrade_test.go
colin-axner: I think it's good practice for us to have unit tests which make stronger assumptions

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -325,6 +325,174 @@ func (suite *KeeperTestSuite) TestChanUpgradeTry() {
...
+	}{
+		{
+			"success",
+			func() {},
+			true,
+		},
+		{
+			"success with later upgrade sequence",
```
colin-axner, line 346: What is this testing? It seems it is asserting that a upgrade sequence that isn't 1 can be used?

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -325,6 +325,174 @@ func (suite *KeeperTestSuite) TestChanUpgradeTry() {
...
+				path.EndpointA.SetChannel(channel)
+			},
+			false,
+		},
+		{
+			"invalid connection state",
+			func() {
+				connectionEnd := path.EndpointA.GetConnection()
+				connectionEnd.State = connectiontypes.UNINITIALIZED
+				suite.chainA.GetSimApp().GetIBCKeeper().ConnectionKeeper.SetConnection(suite.chainA.GetContext(), path.EndpointA.ConnectionID, connectionEnd)
```
colin-axner, line 399: `path.EndpointA.SetConnection(connectionEnd)`? This function doesn't exist?

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -325,6 +325,174 @@ func (suite *KeeperTestSuite) TestChanUpgradeTry() {
...
+		},
+		{
+			"connection not found",
+			func() {
+				channel := path.EndpointA.GetChannel()
+				channel.ConnectionHops = []string{"connection-100"}
+				path.EndpointA.SetChannel(channel)
+			},
+			false,
+		},
```
colin-axner, line 411: dup test case

modules/core/04-channel/keeper/upgrade.go
damiannolan: Me too! I'll make a PR to main with the linter config then :)

colin-axner (commented):
Excellent! Only requested change is moving the version check below `startFlushUpgradeHandshake`. Left many suggestions

modules/core/04-channel/keeper/upgrade_test.go
damiannolan: Essentially yes, if you think its redundant I can remove it

modules/core/04-channel/keeper/upgrade_test.go
damiannolan: indeed it does!

modules/core/04-channel/keeper/upgrade.go
damiannolan: Added similarly to `upgradeSequence` in TRY, good call 👍

modules/core/04-channel/keeper/upgrade.go
damiannolan: Added that it should only be called by the core IBC msg server. If you have any suggestions for better godoc feel free to throw 'em down :)

modules/core/04-channel/keeper/upgrade_test.go
damiannolan: Yeah, agree for more sensitive code like this we should aim for that! You happy if I make this refactor in a follow up PR?

modules/core/04-channel/keeper/upgrade_test.go
damiannolan: https://github.com/cosmos/ibc-go/issues/3831

modules/core/04-channel/keeper/upgrade_test.go
damiannolan: removed

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -230,12 +230,71 @@ func (k Keeper) WriteUpgradeAckChannel(
 	emitChannelUpgradeAckEvent(ctx, portID, channelID, channel, proposedUpgrade)
 }
 
+// ChanUpgradeAck is called by a module to accept the ACKUPGRADE handshake step of the channel upgrade protocol.
+// This method should only be called by the IBC core msg server.
+func (k Keeper) ChanUpgradeAck(
```
colin-axner, line 235:
```suggestion
// This method should only be called by the IBC core msg server.
// This method will verify that the counterparty has entered TRYUPGRADE
// and that its own upgrade is compatible with the selected counterparty version.
func (k Keeper) ChanUpgradeAck(
```

modules/core/04-channel/keeper/upgrade_test.go
colin-axner: yup definitely!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -230,12 +230,73 @@ func (k Keeper) WriteUpgradeAckChannel(
...
+
+	if !collections.Contains(channel.State, []types.State{types.INITUPGRADE, types.TRYUPGRADE}) {
+		return errorsmod.Wrapf(types.ErrInvalidChannelState, "expected one of [%s, %s], got %s", types.INITUPGRADE, types.TRYUPGRADE, channel.State)
+	}
+
+	if !collections.Contains(counterpartyFlushStatus, []types.FlushStatus{types.FLUSHING, types.FLUSHCOMPLETE}) {
+		return errorsmod.Wrapf(types.ErrInvalidFlushStatus, "expected one of [%s, %s], got %s", types.FLUSHING, types.FLUSHCOMPLETE, counterpartyFlushStatus)
+	}
+
+	connection, err := k.GetConnection(ctx, channel.ConnectionHops[0])
```
crodriguezvega, line 260: Should we do this instead? 

```suggestion
	connection, err := k.connectionKeeper.GetConnection(ctx, channel.ConnectionHops[0])
```

I believe this is the way we retrieve the connection from the channel keeper in the rest of the 04-channel codebase. If I'm not mistaken, the `GetConnection` function in channel keeper is only used in ICA, since it is not possible to access the connection keeper from there.

crodriguezvega (commented):
Thanks, @damiannolan!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -230,12 +230,73 @@ func (k Keeper) WriteUpgradeAckChannel(
...
+	}
+
+	upgrade, found := k.GetUpgrade(ctx, portID, channelID)
+	if !found {
+		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "failed to retrieve channel upgrade: port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	if err := k.startFlushUpgradeHandshake(ctx, portID, channelID, upgrade.Fields, counterpartyChannel, counterpartyUpgrade,
+		proofChannel, proofUpgrade, proofHeight); err != nil {
+		return err
```
chatton, line 283: should we wrap this with a specific error type? Maybe a new one for the flushing step

modules/core/04-channel/keeper/upgrade.go
chatton: wow very subtle nice catch, verification happened at different levels like this seems like it might be quite error prone, I think we can worry about this a bit later, but I feel like it might make the most sense to put verifications in `ChanUpgradeAck` instead of `startFlushUpgradeHandshake` - if this happened we could verify `counterpartyChannel` then just pass in the upgrade sequence as we don't need the entire counterparty channel in `startFlushUpgradeHandshake`

modules/core/04-channel/keeper/upgrade_test.go
chatton: I really like the error approach, it's happened a few times where a test was failing for a different reason that the test was actually expecting. Follow up works for me

testing/endpoint.go
```diff
@@ -790,6 +790,20 @@ func (endpoint *Endpoint) SetChannel(channel channeltypes.Channel) {
...
+	upgrade, found := endpoint.Chain.App.GetIBCKeeper().ChannelKeeper.GetUpgrade(endpoint.Chain.GetContext(), endpoint.ChannelConfig.PortID, endpoint.ChannelID)
+	require.True(endpoint.Chain.TB, found)
```
chatton, line 797: I don't know if we should do this assertion in a getter. I would like to be able to use this function to assert an upgrade was not there yet. Or has been removed e.g. in an AbortUpgrade Test.

Maybe either a `GetChannelUpgrade() (channeltypes.Upgrade, found)`

or `GetChannelUpgrade() *channeltypes.Upgrade` where `nil` means not found. WDYT?

chatton (approved):
LGTM! I left a few small suggestions, not of which are blockers and I'd be happy with followups if you'd like to get this merged 💪

modules/core/04-channel/keeper/upgrade.go
damiannolan: seems like all of the channel upgradability code is currently using `k.GetConnection()`, maybe we could address all together and choose one for consistency in a follow up PR.

modules/core/04-channel/keeper/upgrade.go
damiannolan: TRY is also returning the error directly so decided to keep that here. I think all the necessary information is put into an `UpgradeError` within this function call anyways

testing/endpoint.go
damiannolan: Decided to go with this to align with the `path.Endpoint.GetChannel()` and `path.Endpoint.GetConnection()` methods. 

Personally I'd like to keep this one for happy path testing where you don't have to worry about checking the `found` boolean. Maybe for absence we can just do direct keeper invocation or add an additional helper?

modules/core/04-channel/keeper/upgrade.go
damiannolan: opened https://github.com/cosmos/ibc-go/issues/3840 as I'd like to get this merged

testing/endpoint.go
chatton: SGTM 👍

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -230,12 +230,73 @@ func (k Keeper) WriteUpgradeAckChannel(
...
+	if err != nil {
+		return errorsmod.Wrap(err, "failed to retrieve connection using the channel connection hops")
+	}
+
+	counterpartyHops := []string{connection.GetCounterparty().GetConnectionID()}
+	counterpartyChannel := types.Channel{
+		State:           types.TRYUPGRADE,
+		Ordering:        channel.Ordering,
+		ConnectionHops:  counterpartyHops,
+		Counterparty:    types.NewCounterparty(portID, channelID),
```
charleenfei, line 270: should we use `channel.Counterparty.PortID`, `channel.Counterparty.ChannelID`?

damiannolan: The counterparty of the counterparty is the chain executing this handler right? so portID and channelID local to here is the counterparty of the counterparty channel! 

If we use `channel.Counterparty.PortID` and `channel.Counterparty.ChannelID` then things would be off

charleenfei: ahhhh. this makes sense 🙈  the extra counterparty threw me off

charleenfei (approved):

AdityaSripal (approved):

Issue #3747, opened 2023-06-05
Implement `ChanUpgradeTimeout` handler for 04-channel keeper


colin-axner:
## Summary

Should be the same as the spec's `timeoutChannelUpgrade` handler except for the module lookup (msg server) and store state changes (writeUpgradeTimeoutChannel)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3797, opened 2023-06-08
WriteUpgradeTimeout

charleenfei:
## Description

closes: #3750 

### Commit Message / Changelog Entry

```bash
type: commit message
```

see the [guidelines](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) for commit messages. (view raw markdown for examples)

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#pull-request-targeting)).
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/11-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [commit message](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) to be used for the changelog entry in the PR description for review.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.
- [ ] Review `Codecov Report` in the comment section below once CI passes.

PR #3829, opened 2023-06-13
`ChanUpgradeTimeout` handler for 04-channel

charleenfei:
## Description

closes: #3747 

---

write a little note why.


modules/core/04-channel/types/errors.go
```diff
@@ -45,4 +45,8 @@ var (
 	ErrInvalidUpgradeSequence          = errorsmod.Register(SubModuleName, 29, "invalid upgrade sequence")
 	ErrUpgradeNotFound                 = errorsmod.Register(SubModuleName, 30, "upgrade not found")
 	ErrIncompatibleCounterpartyUpgrade = errorsmod.Register(SubModuleName, 31, "incompatible counterparty upgrade")
+	ErrInvalidUpgradeError             = errorsmod.Register(SubModuleName, 32, "invalid upgrade error")
+	ErrUpgradeRestoreFailed            = errorsmod.Register(SubModuleName, 33, "restore failed")
+	ErrUpgradeTimeout                  = errorsmod.Register(SubModuleName, 34, "upgrade timed-out")
```
charleenfei, line 50: i find these two upgrade timeout errors a bit confusingly similar, but i wasnt really sure exactly how to modify it to make it better.

modules/core/04-channel/types/errors.go
```diff
@@ -45,4 +45,8 @@ var (
 	ErrInvalidUpgradeSequence          = errorsmod.Register(SubModuleName, 29, "invalid upgrade sequence")
 	ErrUpgradeNotFound                 = errorsmod.Register(SubModuleName, 30, "upgrade not found")
 	ErrIncompatibleCounterpartyUpgrade = errorsmod.Register(SubModuleName, 31, "incompatible counterparty upgrade")
+	ErrInvalidUpgradeError             = errorsmod.Register(SubModuleName, 32, "invalid upgrade error")
+	ErrUpgradeRestoreFailed            = errorsmod.Register(SubModuleName, 33, "restore failed")
```
charleenfei, line 49: this error is coming from the [WriteUpgradeTimeoutPR](https://github.com/cosmos/ibc-go/pull/3797/files#diff-0fca3dc84b5a25a4a314bceed2fdee673da45a5381aed16d164bb84cd891642e)

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -293,6 +268,130 @@ func (k Keeper) ChanUpgradeAck(
...
+		)
+		if err != nil {
+			return errorsmod.Wrap(err, "failed to verify absence of counterparty channel upgrade error receipt")
+		}
+	} else {
+		// timeout for this sequence can only succeed if the error receipt written into the error path on the counterparty
+		// was for a previous sequence by the timeout deadline.
+		upgradeSequence := channel.UpgradeSequence
+		if upgradeSequence < prevErrorReceipt.Sequence {
+			return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "previous counterparty error receipt sequence is greater than our current upgrade sequence: %d > %d", prevErrorReceipt.Sequence, upgradeSequence)
```
charleenfei, line 350: idea: maybe we could also just call ChanUpgradeCancel here in this case, instead of forcing the user to call.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -293,6 +268,130 @@ func (k Keeper) ChanUpgradeAck(
...
+		proofHeight, proofCounterpartyChannel,
+		channel.Counterparty.PortId,
+		channel.Counterparty.ChannelId,
+		counterpartyChannel,
+	); err != nil {
+		return errorsmod.Wrap(err, "failed to verify counterparty channel state")
+	}
+
+	// counterparty channel must be proved to still be in OPEN state or INITUPGRADE state (crossing hellos)
+	if !collections.Contains(counterpartyChannel.State, []types.State{types.OPEN, types.INITUPGRADE}) {
```
AdityaSripal, line 329: move this above the proof verification

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -484,6 +484,174 @@ func (suite *KeeperTestSuite) TestChanUpgradeAck() {
...
+			expPass := tc.expError == nil
+
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+			suite.coordinator.Setup(path)
+
+			path.EndpointA.ChannelConfig.ProposedUpgrade.Fields.Version = mock.UpgradeVersion
+			path.EndpointB.ChannelConfig.ProposedUpgrade.Fields.Version = mock.UpgradeVersion
+
+			// set timeout height to 1 to ensure timeout
+			path.EndpointA.ChannelConfig.ProposedUpgrade.Timeout.Height = clienttypes.NewHeight(1, 1)
```
AdityaSripal, line 619: Instead of always proving it against timeout height, could be nice to have a success test case with timeoutHeight and a success test case with timeout timestamp?

AdityaSripal (approved):
Approving for correctness, I have a minor request on test coverage

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -293,6 +268,130 @@ func (k Keeper) ChanUpgradeAck(
...
+	if channel.State != types.INITUPGRADE {
+		return errorsmod.Wrapf(types.ErrInvalidChannelState, "channel state is not INITUPGRADE (got %s)", channel.State)
+	}
+
+	upgrade, found := k.GetUpgrade(ctx, portID, channelID)
+	if !found {
+		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	connection, found := k.connectionKeeper.GetConnection(ctx, channel.ConnectionHops[0])
```
DimitrisJim, line 296: though we aren't that consistent with this all over the place, we typically also check that the connection is OPEN.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -293,6 +268,130 @@ func (k Keeper) ChanUpgradeAck(
...
+	// proof must be from a height after timeout has elapsed. Either timeoutHeight or timeoutTimestamp must be defined.
+	// if timeoutHeight is defined and proof is from before timeout height, abort transaction
+	proofTimestamp, err := k.connectionKeeper.GetTimestampAtHeight(ctx, connection, proofHeight)
+	if err != nil {
+		return err
+	}
+
+	if (upgrade.Timeout.Height.IsZero() || proofHeight.LT(upgrade.Timeout.Height)) &&
+		(upgrade.Timeout.Timestamp == 0 || proofTimestamp < upgrade.Timeout.Timestamp) {
+		return errorsmod.Wrap(types.ErrInvalidUpgradeTimeout, "timeout has not yet passed on counterparty chain")
```
DimitrisJim, line 313: a suggestion: `timeout := upgrade.Timeout` would reduce this condition's length considerably. stylistic so can be punted to later

modules/core/04-channel/keeper/upgrade.go
charleenfei: i guess it would take some more gas, but maybe worth checking...? wdyt

DimitrisJim: I'd vote for consistency. I remember from a talk with Colin that given that the channel is OPEN, the connection will be too, these were added in the initial channel opening handshake handlers when optimistic packet sends were a possibility (which isn't something we clearly support now afaik).

I guess we could decide at a later point if removing them from certain handlers makes sense.

damiannolan: shouldn't cost any extra gas? accessing the store will consume gas but once its retrieved there's not gonna be any gas metering on checking its state

charleenfei: ah yea, of course since we already have the connection

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -313,6 +314,112 @@ func (k Keeper) ChanUpgradeAck(
...
+	// proof must be from a height after timeout has elapsed. Either timeoutHeight or timeoutTimestamp must be defined.
+	// if timeoutHeight is defined and proof is from before timeout height, abort transaction
+	proofTimestamp, err := k.connectionKeeper.GetTimestampAtHeight(ctx, connection, proofHeight)
+	if err != nil {
+		return err
+	}
+
+	timeout := upgrade.Timeout
+	if (timeout.Height.IsZero() || proofHeight.LT(timeout.Height)) &&
+		(timeout.Timestamp == 0 || proofTimestamp < timeout.Timestamp) {
```
chatton, line 366: maybe we can split this up into some named booleans to aid readability.

```go

proofHeightIsInvalid := timeout.Height.IsZero() || proofHeight.LT(timeout.Height)
proofTimestampIsInvalid := timeout.Timestamp == 0 || proofTimestamp < timeout.Timestamp
if proofHeightIsInvalid && proofTimestampIsInvalid {
    return errorsmod.Wrap(types.ErrInvalidUpgradeTimeout, "timeout has not yet passed on counterparty chain")
} 
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -313,6 +314,112 @@ func (k Keeper) ChanUpgradeAck(
...
+		proofHeight, proofCounterpartyChannel,
+		channel.Counterparty.PortId,
+		channel.Counterparty.ChannelId,
+		counterpartyChannel,
+	); err != nil {
+		return errorsmod.Wrap(err, "failed to verify counterparty channel state")
+	}
+
+	// Error receipt passed in is either nil or it is a stale error receipt from a previous upgrade
+	if (prevErrorReceipt == types.ErrorReceipt{}) {
```
chatton, line 388: do we want to change the `prevErrorReceipt` to be of type `*types.ErrorReceipt`. This would be more in line with the spec. There is no functional difference anyway.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -313,6 +314,112 @@ func (k Keeper) ChanUpgradeAck(
...
+			ctx,
+			channel.Counterparty.PortId, channel.Counterparty.ChannelId,
+			connection,
+			proofErrorReceipt,
+			proofHeight,
+		)
+		if err != nil {
+			return errorsmod.Wrap(err, "failed to verify absence of counterparty channel upgrade error receipt")
+		}
+	} else {
```
chatton, line 399: [nit] to reduce indentation we can remove the else, and just add a `return nil` the line above this.

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -484,6 +484,197 @@ func (suite *KeeperTestSuite) TestChanUpgradeAck() {
...
+		},
+		{
+			"connection not found",
+			func() {
+				channel := path.EndpointA.GetChannel()
+				channel.ConnectionHops[0] = ibctesting.InvalidID
+				path.EndpointA.SetChannel(channel)
+			},
+			connectiontypes.ErrConnectionNotFound,
+		},
```
chatton, line 579: looks like a duplicate test case

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -484,6 +484,197 @@ func (suite *KeeperTestSuite) TestChanUpgradeAck() {
...
+		{
+			"unable to retrieve timestamp at proof height",
+			func() {
+				proofHeight = clienttypes.NewHeight(0, uint64(path.EndpointA.Chain.GetContext().BlockHeight()+100))
```
chatton, line 583:
```suggestion
				proofHeight = suite.chainA.GetTimeoutHeight()
```

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -484,6 +484,197 @@ func (suite *KeeperTestSuite) TestChanUpgradeAck() {
...
+			},
+			clienttypes.ErrConsensusStateNotFound,
+		},
+		{
+			"timeout has not passed",
+			func() {
+				upgrade := path.EndpointA.GetProposedUpgrade()
+				upgrade.Timeout.Height = clienttypes.NewHeight(1, uint64(path.EndpointA.Chain.GetContext().BlockHeight()+100))
```
chatton, line 591:
```suggestion
				upgrade.Timeout.Height = suite.chainA.GetTimeoutHeight()
```

chatton (approved):
LGTM! I think there are a few small things we can clean up to simplify the code in the handler + test cases, but overall this looks great!

And nice test coverage!

modules/core/04-channel/keeper/upgrade_test.go
charleenfei: whoops!

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: I also stared at this for a bit trying to grok it. Naming is definitely beneficial here.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -366,11 +366,103 @@ func (k Keeper) ChanUpgradeTimeout(
...
+		); err != nil {
+			return errorsmod.Wrap(err, "failed to verify absence of counterparty channel upgrade error receipt")
+		}
+
+		return nil
+	}
+	// timeout for this sequence can only succeed if the error receipt written into the error path on the counterparty
+	// was for a previous sequence by the timeout deadline.
+	upgradeSequence := channel.UpgradeSequence
+	if upgradeSequence < prevErrorReceipt.Sequence {
```
colin-axner, line 451: I think this needs to be `<=`. We should not timeout an upgrade where our sequence matches the counterparty upgrade sequence and there exists an error receipt for that upgrade attempt. That should indicate a cancellation

charleenfei: i can make a followup PR for this issue! but i think also, now looking at the two functions, im really wondering if there is a way to combine these two such that there is first a check on `proofHeight` to see if it has passed the `Timeout`. if so, then we follow the `Timeout` flow. Otherwise, we follow the `Cancel` flow. and then we just expose one API/msg_server function. wdyt?

PR #3835, opened 2023-06-13
begin math lib transition

faddat:
This PR will contain all of the changes made to the use of sdk.* to math.*


crodriguezvega:
We recently merged [a PR to use `sdkmath` as import alias for `cosmossdk.io/math`](https://github.com/cosmos/ibc-go/pull/3772) to avoid confusion with the built in math package. Can you please use the same convention here? 🙏

faddat:
Note on the conventions being established here in ibc-go -- I took errorsmod and ported it to the ecosystem :)

It's good :)

faddat:
Thanks to @tac0turtle too, was his idea :)

Issue #3754, opened 2023-06-05
Implement `MsgChannelUpgradeCancel` handler for core IBC msg server


colin-axner:
## Summary

Requires: #3753 

Implement the msg server handler for `MsgChannelUpgradeCancel`. It should call `channelKeeper.ChanUpgradeCancel`, `cbs.OnChanUpgradeRestore` and `channelKeeper.WriteUpgradeCancelChannel`. It should be accompanied by tests

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #3848, opened 2023-06-14
Implement MsgChannelUpgradeCancel message server handler

chatton:
## Description

depends on #3846

closes: #3754 

---

write a little note why.


modules/core/keeper/msg_server.go
```diff
@@ -820,5 +820,33 @@ func (k Keeper) ChannelUpgradeTimeout(goCtx context.Context, msg *channeltypes.M
...
+		ctx.Logger().Error("channel upgrade cancel failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
+		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+	}
+
+	if err := k.ChannelKeeper.ChanUpgradeCancel(ctx, msg.PortId, msg.ChannelId, msg.ErrorReceipt, msg.ProofErrorReceipt, msg.ProofHeight); err != nil {
+		ctx.Logger().Error("channel upgrade cancel failed", "port-id", msg.PortId, "error", err.Error())
+		return nil, err
+	}
+
+	if err := cbs.OnChanUpgradeRestore(ctx, msg.PortId, msg.ChannelId); err != nil {
```
crodriguezvega, line 842: Sorry if I missed the discussing about the naming, but what about `onChanUpgradeCancel` instead?

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +925,115 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+			expErr: channeltypes.ErrInvalidChannel,
+		},
+		{
+			name: "invalid port id",
+			malleate: func() {
+				msg.PortId = ibctesting.InvalidID
+			},
+			expErr: porttypes.ErrInvalidPort,
+		},
+	}
```
crodriguezvega, line 973: Maybe also a test where `OnChanUpgradeRestore` returns error?

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +925,115 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+			if expPass {
+				suite.Require().NoError(err)
+				channel := path.EndpointA.GetChannel()
+				suite.Require().Equal(preUpgradeVersion, channel.Version, "channel version should be reverted")
+				suite.Require().Equalf(channeltypes.OPEN, channel.State, "channel state should be %s", channeltypes.OPEN.String())
+				suite.Require().Equalf(channeltypes.NOTINFLUSH, channel.FlushStatus, "channel flush status should be %s", channeltypes.NOTINFLUSH.String())
+				suite.Require().Equal(errorReceipt.Sequence+1, channel.UpgradeSequence, "channel upgrade sequence should be incremented")
+			} else {
+				suite.Require().Error(err)
+				suite.Require().Nil(res)
```
crodriguezvega, line 1035: Should we check that the error matches `expErr`?

crodriguezvega (commented):
Thank you, @chatton!

modules/core/keeper/msg_server.go
charleenfei: i think it makes more sense for the applications to call a `restore` function, as they are just going to be handling any restoring to previous state, and not cancelling anything like a handshake on the IBC core level.

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +924,133 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+			suite.coordinator.Setup(path)
+
+			preUpgradeVersion := path.EndpointA.ChannelConfig.Version
+			// configure the channel upgrade version on testing endpoints
+			path.EndpointA.ChannelConfig.ProposedUpgrade.Fields.Version = ibcmock.UpgradeVersion
+			path.EndpointB.ChannelConfig.ProposedUpgrade.Fields.Version = ibcmock.UpgradeVersion
+
+			err := path.EndpointA.ChanUpgradeInit()
+			suite.Require().NoError(err)
```
charleenfei, line 1006: super duper nit and others below but feel free to ignore!
```suggestion
		suite.Require().NoError(path.EndpointA.ChanUpgradeInit())
```

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +924,133 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+			if expPass {
+				suite.Require().NoError(err)
+				channel := path.EndpointA.GetChannel()
+				suite.Require().Equal(preUpgradeVersion, channel.Version, "channel version should be reverted")
+				suite.Require().Equalf(channeltypes.OPEN, channel.State, "channel state should be %s", channeltypes.OPEN.String())
+				suite.Require().Equalf(channeltypes.NOTINFLUSH, channel.FlushStatus, "channel flush status should be %s", channeltypes.NOTINFLUSH.String())
+				suite.Require().Equal(errorReceipt.Sequence+1, channel.UpgradeSequence, "channel upgrade sequence should be incremented")
+			} else {
+				suite.Require().Nil(res)
+				suite.Require().ErrorIs(err, tc.expErr)
```
charleenfei, line 1052: i think maybe here we should check that the channel version, state, flushstatus has not been mutated/upgraded sequence has not been incremented

charleenfei (approved):
preliminary approval contingent on updating `!expPass` case for testing

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +924,140 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+
+				channel := path.EndpointA.GetChannel()
+
+				suite.Require().Equal(prevChannel.Version, channel.Version, "channel version should not be changed")
+				suite.Require().Equalf(prevChannel.State, channel.State, "channel state should be %s", prevChannel.State.String())
+				suite.Require().Equalf(prevChannel.FlushStatus, channel.FlushStatus, "channel flush status should be %s", prevChannel.FlushStatus.String())
+				// TODO
```
chatton, line 1058: this is a tricky one, there is a state update happening in ` k.ChannelKeeper.ChanUpgradeCancel` so when the application callback errors, the upgrade sequence is incremented but this should not actually be the case. This only happens here as we call `ChannelUpgradeCancel`. In practice this state would still be reverted when an error returns from the message server.

I think this is another reason to implement [this](https://github.com/cosmos/ibc-go/pull/3846#discussion_r1230766927)

What do people think about moving the upgrade sequence updating logic into the write function?

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +924,139 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+			} else {
+				suite.Require().Nil(res)
+				suite.Require().ErrorIs(err, tc.expErr)
+
+				channel := path.EndpointA.GetChannel()
+
+				suite.Require().Equal(prevChannel.Version, channel.Version, "channel version should not be changed")
+				suite.Require().Equalf(prevChannel.State, channel.State, "channel state should be %s", prevChannel.State.String())
+				suite.Require().Equalf(prevChannel.FlushStatus, channel.FlushStatus, "channel flush status should be %s", prevChannel.FlushStatus.String())
+				// TODO suite.Require().Equal(prevChannel.UpgradeSequence, channel.UpgradeSequence, "channel upgrade sequence should not incremented")
```
chatton, line 1058: this is a tricky one, there is a state update happening in ` k.ChannelKeeper.ChanUpgradeCancel` so when the application callback errors, the upgrade sequence is incremented but this should not actually be the case. This only happens here as we call `ChannelUpgradeCancel`. In practice this state would still be reverted when an error returns from the message server.

I think this is another reason to implement [this](https://github.com/cosmos/ibc-go/pull/3846#discussion_r1230766927)

What do people think about moving the upgrade sequence updating logic into the write function?

colin-axner (approved):
Wahoo! Great work on the whole cancellation flow/handlers @chatton!! Everything looks clean to me! :heart: 

Left some suggestions/notes

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -232,51 +232,48 @@ func (k Keeper) WriteUpgradeAckChannel(
...
 	if counterpartySequence < currentSequence {
-		return errorsmod.Wrap(types.ErrInvalidUpgradeSequence, "error sequence must be less than current sequence")
+		return 0, errorsmod.Wrap(types.ErrInvalidUpgradeSequence, "error sequence must be less than current sequence")
 	}
 
-	channel.UpgradeSequence = errorReceipt.Sequence + 1
-	k.SetChannel(ctx, portID, channelID, channel)
-
-	return nil
+	return errorReceipt.Sequence, nil
```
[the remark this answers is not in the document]
chatton: Yeah I think it's nice for this handler to perform all the logic and return the expected output, otherwise there are additional steps you need to do outside of the handler to actually perform the full handling logic.

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -1168,15 +1168,15 @@ func (suite *KeeperTestSuite) TestChanUpgradeCancel() {
...
 
 			expPass := tc.expError == nil
 			if expPass {
 				suite.Require().NoError(err)
-				channel := path.EndpointA.GetChannel()
-				suite.Require().Equal(errorReceipt.Sequence+1, channel.UpgradeSequence, "upgrade sequence should be incremented")
+				suite.Require().Equal(errorReceipt.Sequence, newUpgradeSequence, "upgrade sequence should be incremented")
 			} else {
 				suite.Require().ErrorIs(err, tc.expError)
+				suite.Require().Equal(uint64(0), newUpgradeSequence, "upgrade sequence should not be incremented in the case of an error")
```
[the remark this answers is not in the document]
chatton: yeah I think it's not providing much value, happy to remove

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +924,139 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+				msg.ProofHeight = proofHeight
+			},
+			expErr: channeltypes.ErrInvalidUpgradeSequence,
+		},
+		{
+			name: "application callback fails",
+			malleate: func() {
+				suite.chainA.GetSimApp().IBCMockModule.IBCApp.OnChanUpgradeRestore = func(ctx sdk.Context, portID, channelID string) error {
+					// return an error type that is not returned in the regular flow.
+					return upgradetypes.ErrIntOverflowUpgrade
```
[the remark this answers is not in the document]
chatton: maybe we could define an inline error just for this test

```golang
type testAppCallbackError struct{}

func (e testAppCallbackError) Error() string {
	return "test app callback error"
}


...

{
	name: "application callback fails",
	malleate: func() {
		suite.chainA.GetSimApp().IBCMockModule.IBCApp.OnChanUpgradeRestore = func(ctx sdk.Context, portID, channelID string) error {
			// return an error type that is not returned in the regular flow.
			return &testAppCallbackError{}
		}
	},
	// error should be what the application callback returned.
	expErr: &testAppCallbackError{},
},

```

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +924,139 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+			// configure the channel upgrade version on testing endpoints
+			path.EndpointA.ChannelConfig.ProposedUpgrade.Fields.Version = ibcmock.UpgradeVersion
+			path.EndpointB.ChannelConfig.ProposedUpgrade.Fields.Version = ibcmock.UpgradeVersion
+
+			suite.Require().NoError(path.EndpointA.ChanUpgradeInit())
+
+			// fetch the previous channel when it is in the INITUPGRADE state.
+			prevChannel := path.EndpointA.GetChannel()
+
+			suite.Require().NoError(path.EndpointB.UpdateClient())
```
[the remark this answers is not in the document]
chatton: nice catch!

modules/core/keeper/msg_server_test.go
```diff
@@ -924,3 +924,139 @@ func (suite *KeeperTestSuite) TestChannelUpgradeTry() {
...
+			} else {
+				suite.Require().Nil(res)
+				suite.Require().ErrorIs(err, tc.expErr)
+
+				channel := path.EndpointA.GetChannel()
+
+				suite.Require().Equal(prevChannel.Version, channel.Version, "channel version should not be changed")
+				suite.Require().Equalf(prevChannel.State, channel.State, "channel state should be %s", prevChannel.State.String())
+				suite.Require().Equalf(prevChannel.FlushStatus, channel.FlushStatus, "channel flush status should be %s", prevChannel.FlushStatus.String())
+				suite.Require().Equal(prevChannel.UpgradeSequence, channel.UpgradeSequence, "channel upgrade sequence should not incremented")
```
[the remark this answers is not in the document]
chatton: I think I prefer the explicit assertions on each field, but I don't feel too strongly about it. I think individual fields makes it easier understand the test failure.

modules/core/keeper/msg_server.go
```diff
@@ -820,5 +820,31 @@ func (k Keeper) ChannelUpgradeTimeout(goCtx context.Context, msg *channeltypes.M
...
+
+	newUpgradeSequence, err := k.ChannelKeeper.ChanUpgradeCancel(ctx, msg.PortId, msg.ChannelId, msg.ErrorReceipt, msg.ProofErrorReceipt, msg.ProofHeight)
+	if err != nil {
+		ctx.Logger().Error("channel upgrade cancel failed", "port-id", msg.PortId, "error", err.Error())
+		return nil, errorsmod.Wrap(err, "channel upgrade cancel failed")
+	}
+
+	cbs.OnChanUpgradeRestore(ctx, msg.PortId, msg.ChannelId)
+
+	k.ChannelKeeper.WriteUpgradeCancelChannel(ctx, msg.PortId, msg.ChannelId, newUpgradeSequence)
```
damiannolan, line 845: The naming feels a little janky with this `Write..` method. I would've expected something like `WriteUpgradeRestoreChannel` possibly (?).

If I look at the implementation of `WriteUpgradeCancelChannel` and `WriteUpgradeTimeoutChannel` they should be more or less the same right? they both restore the channel to its working state

modules/core/keeper/msg_server.go
```diff
@@ -820,5 +820,31 @@ func (k Keeper) ChannelUpgradeTimeout(goCtx context.Context, msg *channeltypes.M
 
 // ChannelUpgradeCancel defines a rpc handler method for MsgChannelUpgradeCancel.
 func (k Keeper) ChannelUpgradeCancel(goCtx context.Context, msg *channeltypes.MsgChannelUpgradeCancel) (*channeltypes.MsgChannelUpgradeCancelResponse, error) {
-	return nil, nil
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	module, _, err := k.ChannelKeeper.LookupModuleByChannel(ctx, msg.PortId, msg.ChannelId)
```
damiannolan, line 825: we should add a testcase for module not found here I think!

damiannolan (approved):
Excellent work! I left some comments/questions but nothing blocking, we can handle anthing in following PRs. LGTM

modules/core/keeper/msg_server.go
damiannolan: more context - https://github.com/cosmos/ibc-go/pull/3797/files#r1236572812

modules/core/keeper/msg_server.go
```diff
@@ -820,5 +820,31 @@ func (k Keeper) ChannelUpgradeTimeout(goCtx context.Context, msg *channeltypes.M
...
+		return nil, errorsmod.Wrap(err, "could not retrieve module from port-id")
+	}
+
+	cbs, ok := k.Router.GetRoute(module)
+	if !ok {
+		ctx.Logger().Error("channel upgrade cancel failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
+		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+	}
+
+	newUpgradeSequence, err := k.ChannelKeeper.ChanUpgradeCancel(ctx, msg.PortId, msg.ChannelId, msg.ErrorReceipt, msg.ProofErrorReceipt, msg.ProofHeight)
```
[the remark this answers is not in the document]
chatton: good catch! We used to do `errorReceiptSequence + 1` but @colin-axner pointed out that that isn't actually correct.  We should be able to use the sequence from the message directly.

testing/mock/ibc_module.go
```diff
@@ -14,6 +14,13 @@ import (
 	"github.com/cosmos/ibc-go/v7/modules/core/exported"
 )
 
+// applicationCallbackError is a custom error type that will be unique for testing purposes.
+type applicationCallbackError struct{}
```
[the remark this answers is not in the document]
chatton: This was added because there used to be a testcase that checked the exact error type returned by the `OnChanUpgradeRestore` application callback. However that function no longer returns an error.

The idea behind this was that we can do `suite.Require().ErrorIs(err)` to assert a custom error type with `fmt.Errorf`, it would just be a regular go error and we could run into false positives. 

While not used in this PR, I think we could replace the usage of our failed application callback test cases with this. WDYT?

modules/core/keeper/msg_server.go
chatton: Yeah I think maybe it does feel a little off, it matches the convention we've been using with the `WriteX` functions, but those have typically been named after the state the final channel will be in. 

I think `WriteUpgradeRestore(d)Channel` might make more sense. I can create a followup PR with the renames!

testing/mock/ibc_module.go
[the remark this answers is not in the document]
damiannolan: oh, nevermind, I see its exposed with assignment to `MockApplicationCallbackError` in mock.go 👍

Issue #3842, opened 2023-06-14
Add slack webhook notifications for e2e nightly cron jobs


damiannolan:
## Summary

We've added nightly cron jobs for running e2e tests on a schedule in https://github.com/cosmos/ibc-go/pull/3822

It may be nice to explore the idea of adding a dedicated slack channel or other means of notification method to nicely show us status info. Whether it be success or failure.

@chatton has linked - https://github.com/marketplace/actions/slack-notify

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

DimitrisJim:
beat ya to it. Keep this one, I'll remove #3841

DimitrisJim:
scope mostly updated since we don't have a hermes e2e cron job. We do have a cron job for upgrade tests and we can use it there (and anywhere else a cron job might be used in future)

PR #3873, opened 2023-06-16
Add reusable workflow for scheduled jobs to notify on slack.

DimitrisJim:
## Description

Pending discussion on which channel to post (configurable via the WebHook). Current message drops a lot of information for starters, some of which might not really be needed (see a preview of it on the action repo [here](https://github.com/8398a7/action-slack/tree/v3/)).

closes: #3842 

```bash
type: commit message
```

---

write a little note why.


.github/workflows/slack-notify.yaml
```diff
@@ -0,0 +1,24 @@
...
+        default: true
+
+jobs:
+  slack-notify:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: 8398a7/action-slack@v3
+        with:
+          status: ${{ job.status }}
+          fields: repo,message,commit,author,action,eventName,ref,workflow,job,took,pullRequest
```
DimitrisJim, line 21: very nice thing is it renders urls for many of these.

.github/workflows/slack-notify.yaml
```diff
@@ -0,0 +1,24 @@
...
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK }}
```
DimitrisJim, line 23: added this for a private channel Carlos created for now.

DimitrisJim:
note: better to add https://github.com/cosmos/ibc-go/pull/3884 first. This way, I could plug this into the job that aggregates each e2e test result.

chatton:
small update, we recently updated the `upgrade` workflows to run when the files/workflows related to upgrades have changed. It's probably still no harm to run them on a schedule if we want think we want to.

.github/workflows/slack-notify.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# To configure a Slack Webhook URL, follow the instructions here: https://api.slack.com/messaging/webhooks
+on:
+  workflow_call:
+    inputs:
+      # Configure whether to send a Slack notification or not
+      notify-slack:
+        required: true
+        type: boolean
+        default: true
```
chatton, line 12: if something is required, and has a default. Is it required 🤔

.github/workflows/slack-notify.yaml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  workflow_call:
+
+jobs:
+  slack-notify:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: 8398a7/action-slack@v3
+        with:
+          status: 'failure'
```
chatton, line 14: based on the docs maybe `status: ${{ job.status }}` is a better value here? Either that or maybe allow it as an input to be a bit more flexible. WDYT?

DimitrisJim: I guess you'd need a specific one of the jobs to reference. I'm mainly hardcoding it since we only wanna send a notification for failures.

chatton: happy enough to start out with that and we can worry about it if we need to in the future!

.github/workflows/e2e-upgrade.yaml
```diff
@@ -131,3 +131,10 @@ jobs:
       relayer-type: rly
       relayer-image: ghcr.io/cosmos/relayer
       relayer-tag: latest
+
+  slack-notify:
+    needs: [upgrade-v5-hermes, upgrade-v7-hermes, upgrade-v7_1-hermes, upgrade-v8-hermes, upgrade-v5-rly, upgrade-v7-rly, upgrade-v7_1-rly, upgrade-v8-rly]
+    uses: ./.github/workflows/slack-notify.yaml
+    secrets: inherit
+    # Only send a slack notification if any of the tests failed, run only on cron jobs
+    if: ${{ always() && contains(needs.*.result, 'failure') && github.event_name == 'schedule' }}
```
chatton, line 140: This works fine but there is a built in `{{ failure() }}` [example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-status-check-functions).

Should we use that instead?

Also, I don't think this lets us know which one failed (I don't think that's a huge problem since we will just click on the link and see for ourselves)

chatton (approved):
LGTM left one suggestion, but always happy to update as needed 👍

.github/workflows/e2e-upgrade.yaml
DimitrisJim: > This works fine but there is a built in {{ failure() }} [example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-status-check-functions).

does this also apply to jobs? I remember this is an annoying discrepancy and using the contains with `needs` was the only workaround to getting the behaviour "fail if any of these jobs failed".

> I don't think this lets us know which one failed

Yup, tried to do that first go around. Requires additional inputs to re-usable workflow and explicit passing of secrets with `inherit`, I didn't particularly like the way it was shaping out so went with the aggregate.

chatton: Ah I missed that it was for a step and not a job! I'm not sure if it can be used the same way. Happy to leave it for now, either way it achieves the same result.

.github/workflows/e2e-upgrade.yaml
```diff
@@ -131,3 +131,10 @@ jobs:
       relayer-type: rly
       relayer-image: ghcr.io/cosmos/relayer
       relayer-tag: latest
+
+  slack-notify:
+    needs: [upgrade-v5-hermes, upgrade-v7-hermes, upgrade-v7_1-hermes, upgrade-v8-hermes, upgrade-v5-rly, upgrade-v7-rly, upgrade-v7_1-rly, upgrade-v8-rly]
```
damiannolan, line 136: I guess this is slightly stale now!

damiannolan (commented):
Did you ever manage to get this working? what channel would it post to? 

Happy to approve if its g2g! :)

crodriguezvega:
@DimitrisJim Can we merge this one now? Or is there anything that we need to agree on or implement?

crodriguezvega (approved):
Let's merge it!

PR #3886, opened 2023-06-18
sort imports fully deterministically

faddat:
This PR sort imports in a fully 
deterministic fashion using the `gci` linter. 

I tried to follow convention here --

* first the SDK
* then, in this case, more sdk since we've got 
cosmossdk.io and cosmos/cosmos-sdk
* cometbft
* ibc-go

```yaml
  gci:
    sections:
      - standard # Standard section: captures 
all standard packages.
      - default # Default section: contains all 
imports that could not be matched to another 
section type.
      - blank # blank imports
      - dot # dot imports
      - prefix(cosmossdk.io)
      - prefix(github.com/cosmos/cosmos-sdk)
      - prefix(github.com/cometbft/cometbft)
      - prefix(github.com/cosmos/ibc-go)
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,54 +2,68 @@ run:
   tests: true
   #   # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 5m
+  allow-parallel-runners: true
 
 linters:
   disable-all: true
   enable:
+    - asciicheck
+    - bodyclose
     - dogsled
     - exportloopref
     - errcheck
+    - errchkjson
+    - gci
     - goconst
     - gocritic
     - gofumpt
     - gosec
     - gosimple
     - govet
+    - gosmopolitan
     - ineffassign
     - misspell
     - nakedret
     - staticcheck
+    - tenv
     - thelper
+    - typecheck
     - stylecheck
     - revive
     - typecheck
     - unconvert
     - unused
     - misspell
+    - zerologlint
 
 issues:
   exclude-rules:
-    - text: "unused-parameter"
-      linters:
-        - revive
-    - text: "SA1019:"
-      linters:
-        - staticcheck
-    - text: "Use of weak random number generator"
+    - text: 'Use of weak random number generator'
       linters:
         - gosec
-    - text: "ST1003:"
+    - text: 'ST1003:'
       linters:
         - stylecheck
     # FIXME: Disabled until golangci-lint updates stylecheck with this fix:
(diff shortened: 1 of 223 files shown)


.golangci.yml
```diff
@@ -2,54 +2,67 @@ run:
   tests: true
   #   # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 5m
+  allow-parallel-runners: true
 
 linters:
   disable-all: true
   enable:
+    - asciicheck
```
DimitrisJim, line 10: do we need the rest of these linters? Ideally, since `gci` does seem like a good addition, it would be good to separate from the rest, they could be decided in a separate PR.

.golangci.yml
```diff
@@ -2,54 +2,67 @@ run:
...
     - revive
     - typecheck
     - unconvert
     - unused
     - misspell
+    - zerologlint
 
 issues:
   exclude-rules:
-    - text: "unused-parameter"
```
DimitrisJim, line 32: similarly, I think reviewing removing these would be easier.

modules/light-clients/06-solomachine/misbehaviour_handle.go
```diff
@@ -19,7 +19,7 @@ func (cs ClientState) CheckForMisbehaviour(_ sdk.Context, _ codec.BinaryCodec, _
 	return false
 }
 
-func (cs ClientState) verifyMisbehaviour(ctx sdk.Context, cdc codec.BinaryCodec, clientStore sdk.KVStore, misbehaviour *Misbehaviour) error {
+func (cs ClientState) verifyMisbehaviour(cdc codec.BinaryCodec, clientStore sdk.KVStore, misbehaviour *Misbehaviour) error {
```
DimitrisJim, line 22: guess `unused-parameter` being removed complained about this? Looks like a right idea but again I'd think its better if reviewed separately.

DimitrisJim (commented):
I think `gci` is a nice addition to have. I'd like to see the other changes (adding linters, removing exclude rules) separated since the addition of gci touches pretty much every go file making the reviewing process already pretty involved.

.golangci.yml
faddat: great idea @DimitrisJim

modules/light-clients/06-solomachine/misbehaviour_handle.go
faddat: totally agree, I'll redo as described :)

PR #3888, opened 2023-06-18
fully enable revive

faddat:
This PR fully enables the revive linter.

So far it has been removing import shadwing 
caused by the use of the word suite as a 
method reciever instead of the letter s.

## Summary of changes

* suite -> s
* standardized naming of variables
* remove unused recievers
* keep unused parameters

```yaml
linters-settings:
  gci:
    sections:
      - standard # Standard section: captures all standard packages.
      - default # Default section: contains all imports that could not be matched to another section type.
      - blank # blank imports
      - dot # dot imports
      - prefix(cosmossdk.io)
      - prefix(github.com/cosmos/cosmos-sdk)
      - prefix(github.com/cometbft/cometbft)
      - prefix(github.com/cosmos/ibc-go)
    custom-order: true
  dogsled:
    max-blank-identifiers: 3
  maligned:
    # print struct with more effective memory layout or not, false by default
    suggest-new: true
  nolintlint:
    allow-unused: false
    allow-leading-space: true
    require-explanation: false
    require-specific: false
  revive:
    enable-all-rules: true
    rules:
      - name: if-return
        disabled: true
      - name: max-public-structs
        disabled: true
      - name: cognitive-complexity
        disabled: true
      - name: argument-limit
        disabled: true
      - name: cyclomatic
        disabled: true
      - name: file-header
        disabled: true
      - name: function-length
        disabled: true
      - name: function-result-limit
        disabled: true
      - name: line-length-limit
        disabled: true
      - name: flag-parameter
        disabled: true
      - name: add-constant
        disabled: true
      - name: empty-lines
        disabled: true
      - name: banned-characters
        disabled: true
      - name: deep-exit
        disabled: true
      - name: unused-parameter
        disabled: true
      - name: modifies-value-receiver
        disabled: true
      - name: early-return
        disabled: true
      - name: confusing-naming
        disabled: true
      - name: defer
        disabled: true
      - name: unhandled-error
        disabled: false
        arguments:
          - 'fmt.Printf'
          - 'fmt.Print'
          - 'fmt.Println'
          - 'myFunction'
```


faddat:
"Return in a defer function has no effect"

* https://github.com/faddat/ibc-go/blob/9620af965caa8e458f8ac7e72628d53173508c48/modules/core/02-client/keeper/keeper.go#L141-L142

modules/apps/29-fee/keeper/keeper.go
faddat, line 1: this is a good example of "return in defer has no effect"

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
```diff
@@ -75,15 +75,15 @@ func (im IBCMiddleware) OnChanOpenInit(
...
-	counterpartyVersion string,
+func (IBCMiddleware) OnChanOpenTry(
+	_ sdk.Context,
+	_ channeltypes.Order,
+	_ []string,
+	_,
+	_ string,
+	_ *capabilitytypes.Capability,
+	_ channeltypes.Counterparty,
+	_ string,
```
crodriguezvega, line 86: I think we should probably disable this rule that uses `_` for unused parameters, otherwise it's very difficult to know what each argument is.

faddat: ok, and thanks!

faddat:
@crodriguezvega -- if you are reviewing, can you pull this branch, and then run:

```bash
golangci-lint run ./... --fix
```

?

I think this may have revealed an issue (the thing about return not doing anything inside defer)

e2e/testsuite/grpc_query.go
```diff
@@ -32,7 +33,7 @@ import (
 
 // GRPCClients holds a reference to any GRPC clients that are needed by the tests.
 // These should typically be used for query clients only. If we need to make changes, we should
-// use E2ETestSuite.BroadcastMessages to broadcast transactions instead.
+// use E2ETests.BroadcastMessages to broadcast transactions instead.
```
chatton, line 36: looks like this was an accident :D `suite => s`

e2e/testsuite/testsuite.go
```diff
@@ -84,7 +84,7 @@ func (s *E2ETestSuite) GetRelayerUsers(ctx context.Context, chainOpts ...testcon
 
 // SetupChainsRelayerAndChannel create two chains, a relayer, establishes a connection and creates a channel
 // using the given channel options. The relayer returned by this function has not yet started. It should be started
-// with E2ETestSuite.StartRelayer if needed.
+// with E2ETests.StartRelayer if needed.
```
chatton, line 87: ditto

e2e/testsuite/testsuite.go
```diff
@@ -383,7 +383,7 @@ func GetIBCToken(fullTokenDenom string, portID, channelID string) transfertypes.
 // the test. If the test is running in CI, more nodes are used, when running locally a single node is used by default to
 // use less resources and allow the tests to run faster.
 // both the number of validators and full nodes can be overwritten in a config file.
-func getValidatorsAndFullNodes(chainIdx int) (int, int) {
+func getValidatorsAndFullNodes(chainIdx int) (vailidators int, fullNodes int) {
```
chatton, line 386: [q] is this a linting change? I generally dislike named return arguments since they initialize variables outisde of the function body. Does a linter require this?

modules/apps/27-interchain-accounts/controller/types/params_legacy.go
```diff
@@ -27,7 +27,7 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
 	}
 }
 
-func validateEnabledTypeLegacy(i interface{}) error {
+func validateEnabledTypeLegacy(i any) error {
```
chatton, line 30: which linter requires this? Is there any benefit here?

modules/apps/27-interchain-accounts/module.go
```diff
@@ -219,11 +219,11 @@ func (AppModule) GenerateGenesisState(simState *module.SimulationState) {
 }
 
 // WeightedOperations is unimplemented.
-func (am AppModule) WeightedOperations(_ module.SimulationState) []simtypes.WeightedOperation {
+func (AppModule) WeightedOperations(simState module.SimulationState) []simtypes.WeightedOperation {
```
chatton, line 222: should we require names for variables not being used? I would have thought the underscore is okay here?

modules/apps/transfer/ibc_module.go
```diff
@@ -10,7 +10,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	capabilitytypes "github.com/cosmos/ibc-go/modules/capability/types"
-	"github.com/cosmos/ibc-go/v7/modules/apps/transfer/keeper"
+	transferkeeper "github.com/cosmos/ibc-go/v7/modules/apps/transfer/keeper"
```
chatton, line 13: which linter is this? I actually like this change, I'm aware the convention is to use `keeper` everywhere, cc @damiannolan  any thoughts?

chatton (commented):
Hi @faddat sorry for the delay taking a look at this, I think everything looks good but there is just one change, I think we generally prefer more explicit method receiver names,  (specifically with `suite` in the tests).

Is there way to exclude this rule for the purpose of our tests? I think the single letter receiver is good in most other cases!

I've gone through about ~2/3 of it but need to head for the day! I will continue the review tomorrow!

testing/simapp/config.go
```diff
@@ -30,7 +30,7 @@ var (
 )
 
 // GetSimulatorFlags gets the values of all the available simulation flags
```
chatton, line 32: docstring should be updated too

testing/endpoint.go
```diff
@@ -311,7 +311,7 @@ func (endpoint *Endpoint) QueryConnectionHandshakeProof() (
 	connectionKey := host.ConnectionKey(endpoint.Counterparty.ConnectionID)
 	proofConnection, _ = endpoint.Counterparty.QueryProofAtHeight(connectionKey, proofHeight.GetRevisionHeight())
 
-	return
+	return clientState, proofClient, proofConsensus, consensusHeight, proofConnection, proofHeight
```
chatton, line 314: nice I like this one, much more explicit!

chatton (commented):
The rest of it looks good! Not much to add to what I said previously

modules/apps/transfer/ibc_module.go
damiannolan: I guess the more widely used convention with sdk modules is that local keeper is `keeper` and external keepers are `xyzkeeper`

modules/apps/27-interchain-accounts/controller/types/params_legacy.go
DimitrisJim: not much, `any` is a type alias of `interface{}`. It really makes no difference but we could choose either/or.

e2e/testsuite/testsuite.go
faddat: no, but I can swear there was some reason I made the change. After I get main merged, I'll try and figure out why.

modules/apps/transfer/ibc_module.go
faddat: all of this is revive, but I suspect that this had to do with import shadowing.

e2e/relayer/relayer.go
```diff
@@ -66,19 +69,19 @@ func newHermesRelayer(t *testing.T, tag string, logger *zap.Logger, dockerClient
 	)
 }
 
-// RelayerMap is a mapping from test names to a relayer set for that test.
-type RelayerMap map[string]map[ibc.Wallet]bool
+// Map is a mapping from test names to a relayer set for that test.
```
DimitrisJim, line 72: a `-0` on this but I see in go's stdlib that it isn't all that uncommon to do

e2e/testconfig/testconfig.go
```diff
@@ -453,13 +455,13 @@ func defaultGovv1ModifyGenesis() func(ibc.ChainConfig, []byte) ([]byte, error) {
 func defaultGovv1Beta1ModifyGenesis() func(ibc.ChainConfig, []byte) ([]byte, error) {
 	const appStateKey = "app_state"
 	return func(chainConfig ibc.ChainConfig, genbz []byte) ([]byte, error) {
-		genesisDocMap := map[string]interface{}{}
+		genesisDocMap := map[string]any{}
```
DimitrisJim, line 458: similar to Cians other comment, I'm generally in favor of `any` due to it being more succinct and commonplace in other langs but don't mind either way.

e2e/testsuite/testsuite.go
```diff
@@ -177,29 +177,29 @@ func (s *E2ETestSuite) GetPathName(idx int64) string {
 }
 
 // generatePath generates the path name using the test suites name
-func (s *E2ETestSuite) generatePath(ctx context.Context, relayer ibc.Relayer) string {
+func (s *E2ETestSuite) generatePath(ctx context.Context, ibcrelayer ibc.Relayer) string {
```
DimitrisJim, line 180: nit: relayer seems like a term highly specific to ibc, not sure about benefit of adding it to name here

e2e/testvalues/values.go
```diff
@@ -72,7 +74,7 @@ var SimdNewGenesisCommandsFeatureReleases = semverutil.FeatureReleases{
 	MajorVersion: "v8",
 }
 
-// TransferSelfParamsFeatureReleases represents the releases the transfer module started managing its own params.
+// SelfParamsFeatureReleases represents the releases the transfer module started managing its own params.
```
DimitrisJim, line 77: nice catch! Needs some additional tweaking:

```suggestion
// SelfParamsFeatureReleases represents the releases where the ibc modules started managing their own params.
```

modules/apps/27-interchain-accounts/module.go
DimitrisJim: and generally preferred, at least for me, w/o the underscore I spend some time looking around for usages.

modules/apps/29-fee/module.go
```diff
@@ -36,7 +36,7 @@ func (AppModuleBasic) Name() string {
 }
 
 // RegisterLegacyAminoCodec implements AppModuleBasic interface
-func (AppModuleBasic) RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {}
+func (AppModuleBasic) RegisterLegacyAminoCodec(*codec.LegacyAmino) {}
```
DimitrisJim, line 39: Is the linter special casing this? Would be nice to be consistent with the rest:

```suggestion
func (AppModuleBasic) RegisterLegacyAminoCodec(_ *codec.LegacyAmino) {}
```

modules/apps/29-fee/module.go
```diff
@@ -91,7 +91,7 @@ func NewAppModule(k keeper.Keeper) AppModule {
 }
 
 // RegisterInvariants implements the AppModule interface
-func (AppModule) RegisterInvariants(ir sdk.InvariantRegistry) {
+func (AppModule) RegisterInvariants(sdk.InvariantRegistry) {
```
DimitrisJim, line 94: ditto

```suggestion
func (AppModule) RegisterInvariants(_ sdk.InvariantRegistry) {
```

modules/apps/transfer/ibc_module.go
DimitrisJim: If we do it here, would be nice to do it for `icahost` too. Currently there's a mismatch between em.

modules/apps/transfer/keeper/keeper.go
```diff
@@ -7,7 +7,7 @@ import (
 	sdkmath "cosmossdk.io/math"
 
 	"github.com/cosmos/cosmos-sdk/codec"
-	"github.com/cosmos/cosmos-sdk/store/prefix"
+	storeprefix "github.com/cosmos/cosmos-sdk/store/prefix"
```
DimitrisJim, line 10: q: linter? We also have this same import path in other places and can't see it being changed (`grep -inr "github.com/cosmos/cosmos-sdk/store/prefix" . --include="*.go"`)

modules/apps/transfer/types/trace.go
```diff
@@ -84,10 +84,10 @@ func (dt DenomTrace) IsNativeDenom() bool {
 
 // extractPathAndBaseFromFullDenom returns the trace path and the base denom from
 // the elements that constitute the complete denom.
-func extractPathAndBaseFromFullDenom(fullDenomItems []string) (string, string) {
+func extractPathAndBaseFromFullDenom(fullDenomItems []string) (path, basedenom string) {
```
DimitrisJim, line 87: no need for named returns here? I do personally prefer the renaming of `path`, `baseDenom`.

modules/capability/module.go
```diff
@@ -71,7 +71,7 @@ func (AppModuleBasic) ValidateGenesis(cdc codec.JSONCodec, config client.TxEncod
 }
 
 // RegisterGRPCGatewayRoutes registers the gRPC Gateway routes for the capability module.
-func (a AppModuleBasic) RegisterGRPCGatewayRoutes(_ client.Context, _ *gwruntime.ServeMux) {
+func (a AppModuleBasic) RegisterGRPCGatewayRoutes(ctx client.Context, mux *gwruntime.ServeMux) {
```
DimitrisJim, line 74: nit: consistency. unsure if this is the linter just shouting at you randomly tho

modules/core/02-client/migrations/v7/solomachine.go
```diff
@@ -50,104 +50,104 @@ func (cs ConsensusState) UnpackInterfaces(unpacker codectypes.AnyUnpacker) error
...
 
 // ExportMetadata panics!
-func (cs ClientState) ExportMetadata(_ storetypes.KVStore) []exported.GenesisMetadata {
+func (ClientState) ExportMetadata(_ storetypes.KVStore) []exported.GenesisMetadata {
 	panic("legacy solo machine is deprecated!")
 }
 
 // CheckForMisbehaviour panics!
-func (cs ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCodec, clientStore storetypes.KVStore, msg exported.ClientMessage) bool {
+func (ClientState) CheckForMisbehaviour(ctx sdk.Context, cdc codec.BinaryCodec, clientStore storetypes.KVStore, msg exported.ClientMessage) bool {
```
DimitrisJim, line 88: nit: consistency

```suggestion
func (ClientState) CheckForMisbehaviour(_ sdk.Context, _ codec.BinaryCodec, _ storetypes.KVStore, _ exported.ClientMessage) bool {
```

tho I'm seeing most might need to be updated to follow a consistent

modules/core/24-host/keys.go
```diff
@@ -218,7 +218,7 @@ func PacketReceiptKey(portID, channelID string, sequence uint64) []byte {
 	return []byte(PacketReceiptPath(portID, channelID, sequence))
 }
 
-func channelPath(portID, channelID string) string {
+func parseChannelPath(portID, channelID string) string {
```
DimitrisJim, line 221: I'd revert this one, we're not parsing really, we're constructing the path.

modules/core/24-host/parse.go
```diff
@@ -10,7 +10,7 @@ import (
 // ParseIdentifier parses the sequence from the identifier using the provided prefix. This function
 // does not need to be used by counterparty chains. SDK generated connection and channel identifiers
 // are required to use this format.
-func ParseIdentifier(identifier, prefix string) (uint64, error) {
+func ParseIdentifier(identifier, prefix string) (sequence uint64, err error) {
```
DimitrisJim, line 13: ditto comments re named returns. I'd keep as is.

modules/light-clients/06-solomachine/client_state.go
```diff
@@ -43,7 +43,7 @@ func (cs ClientState) GetLatestHeight() exported.Height {
 
 // GetTimestampAtHeight returns the timestamp in nanoseconds of the consensus state at the given height.
 func (cs ClientState) GetTimestampAtHeight(
-	_ sdk.Context,
+	ctx sdk.Context,
```
DimitrisJim, line 46: nit: consistency.

DimitrisJim (commented):
I've left a couple of comments on this (tried to mainly go through non-test files). Main thing I'm seeing is some inconsistency with the naming, i.e sometimes `_` is used, others it is removed and a name is used. Unsure if these are linter quirks or not but would be nice to stay consistent.

Big thanks for doing this though :heart: If you need _any_ help getting this over the line, feel free to ping.

modules/apps/27-interchain-accounts/module.go
faddat: @DimitrisJim agree with you.

modules/light-clients/06-solomachine/client_state.go
faddat: love it.  finishing up with the merge, then will finish the rest here.

modules/apps/27-interchain-accounts/controller/types/params_legacy.go
faddat: use-any

it can be disabled, if we'd like.

e2e/testsuite/testsuite.go
faddat: @chatton after reviewing, I felt that this was clearer with the named returns.

faddat:
@DimitrisJim -- one comment here -- for whatever reason ci doesn't work on this branch.

faddat:
Going to mark this as draft to save review energy -- I will go through and catch things like @DimitrisJim 's "nit consistency" once more here.

DimitrisJim:
Sounds good to me @faddat! Also, thanks again for pushing strong with this :pray:

faddat:
I've always found linting long-haul useful 🖖

e2e/testconfig/testconfig.go
faddat: K, I'll keep use-any enabled.

chatton (approved):
All of these changes look good to me, I'm leaving my approval now because I will be gone for ~2 weeks after today. 


I'll leave final approval to @DimitrisJim  since he's also reviewed the changes.

The main change that would be great would be the revert of `s => suite`, otherwise I think everything else is good to go!

thanks @faddat 🚀

faddat:
@chatton quick clarificiation, do you want to use th eword suite or the letter s?

I changed to using the letter s because the word suite shadows the import.

DimitrisJim:
@faddat ya, Cian is off for a couple of days so let me fill you in. 

We discussed this internally and came to an agreement that `suite` helps significantly in this case since test functions tend to be on the larger size and can get pretty chaotic.

testing/endpoint.go
faddat: :), the linter responsible is called naked-return

modules/apps/29-fee/module.go
faddat: I'd actually like to keep the cdc, let me see if it can be happy with that.

modules/apps/29-fee/module.go
faddat: prefer using the im?

I'm thinking about your lookup issue.  Hope to open whole thing for review today.

modules/apps/transfer/types/trace.go
faddat: ah, this now sets off the confusing-results linter. 

confusing-results: unnamed results of the same type may be confusing, consider using named results (revive)

We can disable that or use the named returns.

modules/apps/transfer/ibc_module.go
faddat: I think I know the solution, thanks guys for the thorough review of this, since the goal is to set down some lasting standards :)

modules/apps/transfer/ibc_module.go
```diff
@@ -122,7 +122,7 @@ func (im IBCModule) OnChanOpenTry(
 }
 
 // OnChanOpenAck implements the IBCModule interface
-func (im IBCModule) OnChanOpenAck(
+func (IBCModule) OnChanOpenAck(
```
faddat, line 125: What should the _ string in this function be named?

modules/apps/transfer/types/trace.go
faddat: I think that for now, I will leave this as-is, just lmk how you'd like it.

modules/apps/transfer/module.go
```diff
@@ -160,6 +160,6 @@ func (am AppModule) RegisterStoreDecoder(sdr sdk.StoreDecoderRegistry) {
 }
 
 // WeightedOperations returns the all the transfer module operations with their respective weights.
-func (am AppModule) WeightedOperations(_ module.SimulationState) []simtypes.WeightedOperation {
+func (AppModule) WeightedOperations(_ module.SimulationState) []simtypes.WeightedOperation {
```
faddat, line 163: let me know if this is right.  I think that in the end the _'s are kinda an anti pattern, I've found myself searching for the right name.

modules/capability/module.go
faddat: it isn't.  At some point I figured it may make sense to bring back the _, and started to reinstate them.  Basically because it is instructive to folks what they could / should do.

modules/core/24-host/keys.go
faddat: done

modules/core/24-host/parse.go
```diff
@@ -104,7 +104,7 @@ func MustParseConnectionPath(path string) string {
 
 // MustParseChannelPath returns the port and channel ID from a full path. Panics
 // if the provided path is invalid.
-func MustParseChannelPath(path string) (string, string) {
+func MustParseChannelPath(path string) (portID string, channelID string) {
```
faddat, line 107: In the end I think that I agree with the linter, but happy to set things up for ibc-go team's preference.

modules/light-clients/06-solomachine/client_state.go
faddat: Jim, I've now disabled this linter and I think that things are better off named to reduce dev time to search for the things to be named.  In cases where we've got an interface where we never use a prameter, then I figure we just remove the parameter.

faddat:
### Comments upon deeply going through ibc-go

* I think that we should use named returns, not so much because of the confusing-results linter (it is disabled), but because of the changes in keys.go and parse.go that I needed to make in order to make the code more readable.
* Using the blanks is probably an anti-pattern, I've turned off that linter.  We want to give devs as much clarity as possible, which is also related to the point above about named returns.  

Many thanks to the ibc-go team for reviewing this.  It's been a great time.  I'm going to move on to the eden upgrade now.

faddat (commented):
trying to "finish" my review :).

faddat:
@chatton -- will revert suite -> s and remove the import shadowing linter tomorrow :)

DimitrisJim:
> will revert suite -> s and remove the import shadowing linter tomorrow :)

You could just change the `import` to bind the name to something else no? I.e:

```go
testsuite "github.com/stretchr/testify/suite"
```
this way the linter could still stick around.

faddat:
I found a linter that flags named returns and removed 10-20 named returns.

DimitrisJim (approved):
Made a couple of changes here, mostly to keep the diff localized and minimal.

1. Disable `use-any`. I think this is fine tbh but I don't think there's much benefit in enforcing a single usage. We can revisit at a later point if required.
2. Renamed `s` to `suite` in the test files.
3. Reverted the renaming of vars I.e `_ -> someName` and `someName -> _`. This usage was (and still is) inconsistent. A different issue should be opened to tackle this consistency nit.

Overall, LGTM, big thanks @faddat for the work on this and sorry for the slight delay in getting through with it (big diffs spook me).

chatton:
Thank you @faddat  and @DimitrisJim  ❤️

damiannolan (approved):

faddat:
oh amazing guys!  I've been busy with family and thrilled to see this in <3

Issue #3876, opened 2023-06-17
Add flush status check in `RecvPacket` 


crodriguezvega:
Pre-requisite:

- [x] https://github.com/cosmos/ibc/pull/982

We already [check if the channel state is `OPEN`](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/packet.go#L132-L137), so we need to add a check to also allow receiving if flush status is `FLUSHING` (independently of channel state).

crodriguezvega:
Closed by #3914.

PR #3914, opened 2023-06-21
Add FlushStatus Checks to RecvPacket

chatton:
## Description

closes: #3876

---

write a little note why.


modules/core/04-channel/keeper/packet.go
```diff
@@ -136,7 +136,8 @@ func (k Keeper) RecvPacket(
 		return errorsmod.Wrap(types.ErrChannelNotFound, packet.GetDestChannel())
 	}
 
-	if channel.State != types.OPEN {
+	counterpartyLastPacketSent, _ := k.GetCounterpartyLastPacketSequence(ctx, packet.GetDestPort(), packet.GetDestChannel())
+	if channel.State != types.OPEN && (channel.FlushStatus != types.FLUSHING || counterpartyLastPacketSent > packet.GetSequence()) {
```
crodriguezvega, line 140: I find it easier to reason about like this:

```suggestion
	if !(channel.State == types.OPEN || (channel.FlushStatus == types.FLUSHING &&  packet.GetSequence() <= counterpartyLastPacketSent)) {
```

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -318,6 +318,58 @@ func (suite *KeeperTestSuite) TestRecvPacket() {
 			// attempts to receive packet 2 without receiving packet 1
 			channelCap = suite.chainB.GetChannelCapability(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
 		}, true},
+		{
+			msg: "success with closed channel: FLUSHING status",
```
crodriguezvega, line 322: Hmm, maybe the condition in `RecvPacket` needs to be tweaked a bit more? Because I would find it strange that the channel is closed but you can still receive the packet... Is it possible to even be in such state (channel closed and in flushing status)? I would expect the channel state in the test to be more like INITUPGRADE or TRYUPGRADE, right?

crodriguezvega (commented):
Thanks, @chatton! I left a couple of comments.

modules/core/04-channel/keeper/packet.go
```diff
@@ -136,7 +136,8 @@ func (k Keeper) RecvPacket(
...
 		return errorsmod.Wrapf(
 			types.ErrInvalidChannelState,
 			"channel state is not OPEN (got %s)", channel.State.String(),
```
crodriguezvega, line 143: The error message should be updated as well, right?

modules/core/04-channel/keeper/packet.go
DimitrisJim: we typically go with the approach of negating the spec conditions right? I do agree with you though that this makes it easier to reason (and, as a result, implement)

modules/core/04-channel/keeper/packet_test.go
DimitrisJim: I think it currently is possible to be in such a state if a message to close the channel is submitted during the upgrade handshake? Even if I'm missing something and that isn't possible we should definitely amend the channel closing handlers to set the flushing status to NOTINFLUSH.

modules/core/04-channel/keeper/packet.go
DimitrisJim: let me know what you think of the updated message

DimitrisJim:
pushing here too with Cian's blessing

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -318,6 +318,75 @@ func (suite *KeeperTestSuite) TestRecvPacket() {
...
+				path.EndpointB.ChannelConfig.ProposedUpgrade.Fields.Version = ibcmock.UpgradeVersion
+
+				err = path.EndpointB.ChanUpgradeInit()
+				suite.Require().NoError(err)
+
+				err = path.EndpointA.ChanUpgradeTry()
+				suite.Require().NoError(err)
+
+				err = path.EndpointB.ChanUpgradeAck()
+				suite.Require().NoError(err)
```
DimitrisJim, line 341: by default ends up in `[ACKUPGRADE, FLUSHING]`. The fact that we have sent a packet before the upgrade should mean even after we make the changes in #3928, this will still be in `FLUSHING`. Correct me if I'm wrong (and we can set the flush status manually here).

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -318,6 +318,75 @@ func (suite *KeeperTestSuite) TestRecvPacket() {
...
+			},
+			true,
+		},
+		{
+			"failure with closed channel counterparty sequence > packet sequence",
```
DimitrisJim, line 346: only setting channel to closed, flush status stays `NOTINFLUSH` so we are only checking the counterparty sequence > packet sequence part of the if.

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -318,6 +318,75 @@ func (suite *KeeperTestSuite) TestRecvPacket() {
...
+				err = path.EndpointB.ChanUpgradeInit()
+				suite.Require().NoError(err)
+
+				err = path.EndpointA.ChanUpgradeTry()
+				suite.Require().NoError(err)
+
+				err = path.EndpointB.ChanUpgradeAck()
+				suite.Require().NoError(err)
+
+				// TODO: Manually set until #3928 is implemented.
```
DimitrisJim, line 381: linking #3928

DimitrisJim:
rebased since history got messy after the base branch was automatically changed and tweaked a final thing. believe this should be r4r

modules/core/04-channel/keeper/packet.go
```diff
@@ -133,10 +133,11 @@ func (k Keeper) RecvPacket(
 		return errorsmod.Wrap(types.ErrChannelNotFound, packet.GetDestChannel())
 	}
 
-	if channel.State != types.OPEN {
+	counterpartyLastPacketSent, _ := k.GetCounterpartyLastPacketSequence(ctx, packet.GetDestPort(), packet.GetDestChannel())
+	if channel.State != types.OPEN && (channel.FlushStatus != types.FLUSHING || counterpartyLastPacketSent > packet.GetSequence()) {
 		return errorsmod.Wrapf(
 			types.ErrInvalidChannelState,
-			"channel state is not OPEN (got %s)", channel.State.String(),
+			"packets cannot be received on channel with state (%s) and flush status (%s)", channel.State, channel.FlushStatus,
```
crodriguezvega, line 140: The error message will not provide enough information if the channel is not open and flushing, but the packet sequence is > `counterpartyLastPacketSent`. I guess it would be very informative if we have different error messages depending on the condition that was not satisfied?

modules/core/04-channel/keeper/packet.go
charleenfei: agree with @crodriguezvega it's easier to reason about this way.

modules/core/04-channel/keeper/packet_test.go
charleenfei: it's actually a bit weird if we can recv packets thru a CLOSED channel...?

modules/core/04-channel/keeper/packet.go
charleenfei: should we actually add a condition in the code that if the channel is CLOSED then we can't recv packets? kinda weird if we can receive packets then right?

DimitrisJim: we shouldn't be able to. If state is CLOSED, flush status should be NOTINFLUSH and the condition should resolve to true.

charleenfei: oh! didnt see that comment, thanks! yea. that makes sense.

modules/core/04-channel/keeper/packet_test.go
charleenfei: also, if the  counterparty sequence > packet sequence shouldn't the tx actually work? in this case it's only failing bc the flushstatus is NOTINFLUSH

chatton:
ACK the latest changes made by @charleenfei, the new condition looks good. I think it is still quite complex,  but the added context to the error message should provide enough clarity as to the purpose of the checks!

DimitrisJim:
Also lgtm but since I was also slightly involved with this I'd rather leave someone else to also give it a view.

modules/core/04-channel/keeper/packet.go
```diff
@@ -133,10 +140,16 @@ func (k Keeper) RecvPacket(
...
+
+	if channel.State != types.OPEN && !(channel.FlushStatus == types.FLUSHING && packet.GetSequence() <= counterpartyLastPacketSent) {
 		return errorsmod.Wrapf(
 			types.ErrInvalidChannelState,
-			"channel state is not OPEN (got %s)", channel.State.String(),
+			"packets cannot be received: either channel with non OPEN state (%s) & not in upgrade, or channel is in upgrade but flush status (%s) is not FLUSHING & packet sequence %d must be less than or equal to counterparty last packet sent sequence %d",
+			channel.State,
+			channel.FlushStatus,
+			packet.GetSequence(),
+			counterpartyLastPacketSent,
```
damiannolan, line 152: I definitely found this section pretty hard to parse and reason about. Looks like there was already some discussion in another thread about it. 

I find it a bit complex to mix a channel state assertion with the `&&` expression which opens the door for weird channel states to slip through. So here, we are only going to fail the recv packet tx if the channel state is not OPEN AND the channel state is not flushing and the sequencing is all good. 

So if we are in UPGRADETRY or UPGRADEACK, or any other state for that matter then the first part of the expression will evaluate as false.

What do you think about explicitly checking the expected channel states? I would be more confident in the code and see less room for unexpected states to slip through the cracks with something like:

```go
if !collections.Contains(channel.State, []types.State{types.OPEN, types.TRYUPGRADE, types.ACKUPGRADE}) {
	return errorsmod.Wrapf(types.ErrInvalidChannelState, "channel state is not OPEN (got %s)", channel.State.String())
}

if counterpartyLastSequenceSend, found := k.GetCounterpartyLastPacketSequence(ctx, packet.GetDestPort(), packet.GetDestChannel()); found {
	if channel.FlushStatus != types.FLUSHING && packet.GetSequence() <= counterpartyLastSequenceSend {
		return errorsmod.Wrap(types.ErrInvalidChannelState, "something meaningful")
	}
}
```

I think we can also make use of the `found` bool with last sequence send instead of discarding it

damiannolan: Slightly related https://github.com/cosmos/ibc-go/issues/4001 - I noticed this field was never being set in store within the upgrade handshake

modules/core/04-channel/keeper/packet.go
```diff
@@ -33,7 +33,14 @@ func (k Keeper) SendPacket(
 		return 0, errorsmod.Wrap(types.ErrChannelNotFound, sourceChannel)
 	}
 
-	if channel.State != types.OPEN {
+	if channel.FlushStatus != types.NOTINFLUSH {
```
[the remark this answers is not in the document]
chatton: Good catch, this check is duplicated, I think this is due to branching off the SendPacket branch before this was fixed. I think we can remove all changes to `SendPacket` in this PR.

modules/core/04-channel/keeper/packet.go
chatton: I'm a big fan of this change actually, it makes things far more readable. =

DimitrisJim: note the last conditional should be `||`, righto?

DimitrisJim: I think it might also make sense to add a comment that the second check for the last seq sent is only meaningful for `UPGRADETRY/ACK`.

chatton: Currently the tests pass with both `||` and `&&` so I think we don't have enough test coverage 😅

chatton: But yes I believe `||` is correct here! If either of those conditions are false we should error.

DimitrisJim: also, for anyone else reviewing, the original condition was completely flipped with Damians suggestion, hence why we can break the giant if in two smaller ones. `&&` cannot be broken directly like this but since it was flipped to an `||` it can, this kept me up a bit.

modules/core/04-channel/keeper/packet.go
```diff
@@ -133,11 +134,16 @@ func (k Keeper) RecvPacket(
...
-		)
+	if !collections.Contains(channel.State, []types.State{types.OPEN, types.TRYUPGRADE, types.ACKUPGRADE}) {
+		return errorsmod.Wrapf(types.ErrInvalidChannelState, "channel state was not one of [%s, %s, %s] (got %s)", types.OPEN.String(), types.TRYUPGRADE.String(), types.ACKUPGRADE.String(), channel.State.String())
+	}
+
+	// in the case of the channel being in TRYUPGRADE or ACKUPGRADE we need to ensure that the channel is not in flushing,
+	// and that the counterparty last sequence send is less than or equal to the packet sequence.
+	if counterpartyLastSequenceSend, found := k.GetCounterpartyLastPacketSequence(ctx, packet.GetDestPort(), packet.GetDestChannel()); found {
+		if channel.FlushStatus != types.FLUSHING || packet.GetSequence() > counterpartyLastSequenceSend {
+			return errorsmod.Wrapf(types.ErrInvalidChannelState, "expected channel flush status to not be (%s) and for the packet sequence (%d) to be less than or equal to the counterparty last sequence send (%d)", types.FLUSHING.String(), packet.GetSequence(), counterpartyLastSequenceSend)
```
damiannolan, line 145: I wonder could we improve this error and msg a bit. I'm not sure if this is much better but it would be good to indicate that we expect the channel to be flushing when the counterparty last seq send is set.

`"expected channel flush status to be FLUSHING when counterparty last sequence send (100) is set, failed to recv packet at sequence (99)"`

we could also change to `ErrInvalidFlushStatus` maybe? what do you think?

```suggestion
			return errorsmod.Wrapf(types.ErrInvalidFlushStatus, "expected channel flush status to be (%s) when counterparty last sequence send (%d) is set, failed to recv packet (%d)", types.FLUSHING, counterpartyLastSequenceSend, packet.GetSequence())
```

damiannolan (approved):
Nice! I just left some suggestions on error msgs :)

modules/core/04-channel/keeper/packet.go
charleenfei: won't this fail if the channel is just in a normal OPEN state and NOTINFLUSH? am i misreading something

charleenfei: like, we should only run this condition with the if condition of TRY and ACKUPGRADE

charleenfei: ah, it's only run if the counterpartyLastSeq send is found, which would only happen if we are in the upgrade process I guess.

charleenfei (approved):
nice one bringing it over the line <3

DimitrisJim (approved):

Issue #3879, opened 2023-06-17
Add packet flushing logic to timeout and timeout on close flows


crodriguezvega:
Pre-requisite:

- [x] https://github.com/cosmos/ibc/pull/982

After [deleting the packet commitment](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/timeout.go#L154), change flush status to `FLUSHCOMPLETE` if the flush status is `FLUSHING` and there are no pending in-flight packets. Opening the issue for both `Timeout` and `TimeoutOnClose` since they both use `TimeoutExecuted` function and that's where the packet commitment is deleted, in case we decide to implement this logic inside that function.

See also question [here](https://github.com/cosmos/ibc-go/issues/3878#issue-1761863772).

crodriguezvega:
Closed by #3923.

PR #3922, opened 2023-06-22
Amend AcknowledgePacket to handle flushing

DimitrisJim:
## Description

closes: #3878 and #3877

pending spec merge.

### Commit Message / Changelog Entry

```bash
chore: amend AcknowledgePacket to allow flushing of pre-upgrade packets
```

see the [guidelines](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) for commit messages. (view raw markdown for examples)

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#pull-request-targeting)).
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/11-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [commit message](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) to be used for the changelog entry in the PR description for review.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.
- [ ] Review `Codecov Report` in the comment section below once CI passes.

PR #3923, opened 2023-06-22
Amend timeout to handle in-flight packets.

DimitrisJim:
## Description

closes: #3879 

pending spec merge.

```bash
chore: amend TimeoutExecuted to handle in-flight packets
```

---

write a little note why.


modules/core/04-channel/keeper/timeout.go
```diff
@@ -158,6 +158,11 @@ func (k Keeper) TimeoutExecuted(
 		k.SetChannel(ctx, packet.GetSourcePort(), packet.GetSourceChannel(), channel)
 	}
 
+	if channel.FlushStatus == types.FLUSHING && !k.hasInflightPackets(ctx, packet.GetSourcePort(), packet.GetSourceChannel()) {
```
DimitrisJim, line 161: ditto https://github.com/cosmos/ibc-go/pull/3922#discussion_r1237915276

colin-axner, line 161: Just noting that I think the spec needs to be modified here https://github.com/cosmos/ibc/pull/982#discussion_r1238262526

DimitrisJim: keeping an eye on it, will update as required

AdityaSripal: Updated on spec

modules/core/04-channel/keeper/timeout.go
```diff
@@ -158,6 +158,11 @@ func (k Keeper) TimeoutExecuted(
...
+		channel.FlushStatus = types.FLUSHCOMPLETE
+		k.SetChannel(ctx, packet.GetSourcePort(), packet.GetSourceChannel(), channel)
+	}
+
```
charleenfei, line 165: I think line 179 below just needs to be updated in line with the spec https://github.com/cosmos/ibc/pull/982/files#diff-6d4b263fc3c751e8e3fbd765b2817241db22c8e3083c61685900fa4783669220L1019

DimitrisJim: just needed some slight re-ordering too

modules/core/04-channel/keeper/timeout.go
```diff
@@ -153,8 +153,14 @@ func (k Keeper) TimeoutExecuted(
 
 	k.deletePacketCommitment(ctx, packet.GetSourcePort(), packet.GetSourceChannel(), packet.GetSequence())
 
+	if channel.FlushStatus == types.FLUSHING && !k.hasInflightPackets(ctx, packet.GetSourcePort(), packet.GetSourceChannel()) {
```
DimitrisJim, line 156: P.s, these could also be:

```go
	switch channel.Ordering {
	case types.ORDERED:
		channel.State = types.CLOSED
		channel.FlushStatus = types.NOTINFLUSH
	case types.UNORDERED:
		if channel.FlushStatus == types.FLUSHING && !k.hasInflightPackets(ctx, packet.GetSourcePort(), packet.GetSourceChannel()) {
			channel.FlushStatus = types.FLUSHCOMPLETE
		}
	}
	k.SetChannel(ctx, packet.GetSourcePort(), packet.GetSourceChannel(), channel)
```

and, considering `ORDERED_ALLOW_TIMEOUT`, this could be amended in the future to possible be `case types.UNORDERED, types.ORDERED_ALLOW_TIMEOUT` since it looks like it would require similar handling.

Happy to leave as is but dropping this sionce I noticed it.

charleenfei: i think the switch statement is nice for readability! 

but i guess we probably need to check `if channel.FlushStatus == types.FLUSHING && !k.hasInflightPackets(ctx, packet.GetSourcePort(), packet.GetSourceChannel())` in both cases rather than just in the `UNORDERED` case, and then basically if it's ORDERED set to CLOSED/NOTINFLUSH, and if it's UNORDERED set to FLUSHCOMPLETE. but i'm not super opinionated on which style we go with.

DimitrisJim:
playing around with tests didn't lead me to nice places. `timeoutPacket` still requires the channel be in `OPEN` and `timeoutOnClose` creates a proof of the expected counterparty channel w/o considering the upgrade sequence (https://github.com/cosmos/ibc/issues/973 should be the issue tracking these, second point).

crodriguezvega (commented):
Looks good to me, but are we not adding tests or is this tested indirectly somewhere else?

DimitrisJim:
> but are we not adding tests or is this tested indirectly somewhere else?

TimeoutPacket needs to be amended, probably in a similar way to Recv, in order to allow timeouts on channels with state != OPEN. 

In a similar vein, TimeoutOnClose needs to be amended in order for its proof to work (flushing status being NOTINFLUSH and passing the upgradeSequence).

In theory I could add these. In practice, I'd prefer waiting on our spec guru to reason about the conditionals.

colin-axner (approved):
I like the current approach. The switch statement could be nice, but it doesn't handle additional ordering types (and ordered allow timeout will eventually be added). My vote is to leave as is

colin-axner:
Tests would be nice

modules/core/04-channel/keeper/timeout_test.go
```diff
@@ -247,46 +248,120 @@ func (suite *KeeperTestSuite) TestTimeoutPacket() {
...
 		path    *ibctesting.Path
 		packet  types.Packet
 		chanCap *capabilitytypes.Capability
 	)
 
-	testCases := []testCase{
-		{"success ORDERED", func() {
-			path.SetChannelOrdered()
-			suite.coordinator.Setup(path)
+	testCases := []struct {
```
DimitrisJim, line 260: completely revamped this test to allow checking for how the state of affairs is post timeoutexecuted (also, the error returned). I think this is a way better approach than what previously existed though its unfortunate it added an additional diff to check.

modules/core/04-channel/keeper/timeout_test.go
```diff
@@ -247,46 +248,120 @@ func (suite *KeeperTestSuite) TestTimeoutPacket() {
...
+		malleate  func()
+		expResult func(packetCommitment []byte, err error)
+	}{
+		{
+			"success ORDERED",
+			func() {
+				path.SetChannelOrdered()
+				suite.coordinator.Setup(path)
+
+				timeoutHeight := clienttypes.GetSelfHeight(suite.chainB.GetContext())
```
charleenfei, line 271: i think there's also a default timeout height we could use `suite.chainB.GetTimeoutHeight()`

charleenfei (approved):
just one nit on the tests

crodriguezvega (approved):
Thank you for adding the tests, @DimitrisJim!

DimitrisJim:
I was initially confused as to what to test, since I didn't notice the direct tests on TimeoutExecuted. This still doesn't go through the top handlers tho, comment https://github.com/cosmos/ibc-go/pull/3923#issuecomment-1614305961 still applies and I believe should be fixed.

modules/core/04-channel/keeper/timeout_test.go
DimitrisJim: ah yeah, works out with the suggestion. I'm seeing the rest of the tests here use the old version so might be better to open an issue and change this altogether in a later point? wdyt?

charleenfei: Definitely works for me!

PR #3982, opened 2023-06-29
Pass relayer-tag to e2e-test-workflow-call

DimitrisJim:
## Description

Should address the most recent cron-job failure. Testing this is annoyingly difficult.

closes: #XXXX

```bash
chore: pass relayer tag to workflow file.
```

---

write a little note why.


charleenfei (approved):
🙏

.github/workflows/e2e.yaml
```diff
@@ -54,8 +56,10 @@ jobs:
           if [ "${{ github.event_name }}" == "schedule" ]
           then
             echo "relayer=hermes" >> $GITHUB_OUTPUT
+            echo "relayer-tag=1.4.0" >> $GITHUB_OUTPUT
           else
             echo "relayer=rly" >> $GITHUB_OUTPUT
+            echo "relayer-tag=latest" >> $GITHUB_OUTPUT
```
damiannolan, line 62: if this is just setting a default as `latest` should we just define it [here](https://github.com/cosmos/ibc-go/blob/main/.github/workflows/e2e-test-workflow-call.yml#L58) in the e2e-test-workflow-call? 

Judging from the comment in the link, I think the tests are possibly expecting the empty string `''` and choosing a default whereas this would override to always be `latest` regardless of whether we're using the temp image (`damiannolan/rly:latest`) or a standard release version. Happy to add this now and tweak as necessary when we can pull in the next release version of cosmos/rly

damiannolan (approved):
🙏

.github/workflows/e2e.yaml
DimitrisJim: the empty space there also basically controls if the tag from the config file will be used or the tag from the env var ([relevant snippet](https://github.com/cosmos/ibc-go/blob/a88fc76f865a16d8bc1885e73e119961304ffb16/e2e/testconfig/testconfig.go#L199-L201) in testconfig.go). 

I guess this would cause an issue in the case where we move to explicit tags (i.e `2.3.0`) and we only update it in the config file. We'd be running things with latest (`2.3.1`) in our github e2es since the overwriting won't take place (which may or may not be an actual issue).

Happy to also wait for Cian to chime in (and ignore the next 4 notifications on the failed schedule run)

chatton: I think to preserve existing behaviour we should just set this to an empty string (let the configuration happen in the code you linked for now)

I think a nice solution based on the issue you linked, is to have a config with all the relayer configs, and just specify that here in the end.  

`latest` should break things on regular PRs right? If we were running `pr-1234` we'll now be running `latest` and that will be a different image.

DimitrisJim: > If we were running pr-1234 we'll now be running latest and that will be a different image.

wait, this is for relayer, not simd, so PR tag shouldn't be affected in any way?

chatton: ah sorry yeah getting mixed up now, should be fine with the relayer. I'm happy to pass empty string or `latest` in that case. 👍

chatton (approved):
LGTM

Issue #3987, opened 2023-06-30
amend the channel closing handlers to set the flushing status to NOTINFLUSH.


charleenfei:
I think it currently is possible to be in such a state if a message to close the channel is submitted during the upgrade handshake? Even if I'm missing something and that isn't possible we should definitely amend the channel closing handlers to set the flushing status to NOTINFLUSH.

_Originally posted by @DimitrisJim in https://github.com/cosmos/ibc-go/pull/3914#discussion_r1239540481_

PR #3995, opened 2023-07-03
Set channel state to NOTINFLUSH when Closing

chatton:
## Description

closes: #3987

Since we close the channel in several places, I created a helper function for this behaviour. There are a few other ways that we could do it that might be nicer.

```go
// don't mutate the channel, but instead return a new one
func (c Channel) Close() Channel {
        c.State = CLOSED
	c.FlushStatus = NOTINFLUSH
        return c
}

// no modification
func CloseChannel(channel Channel) Channel {
	channel.State = CLOSED
	channel.FlushStatus = NOTINFLUSH
	return channel
}
```

If people think this is overkill, we can just keep the re-assignments directly in-line. My concern is that it is now no longer sufficient to just modify the channel state to "close" the channel so it seems beneficial to create a function for this.

N/A Feature branch

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/types/channel.go
+++ b/modules/core/04-channel/types/channel.go
@@ -75,7 +75,6 @@ func (ch Channel) ValidateBasic() error {
 	return ch.Counterparty.ValidateBasic()
 }
 
-
 // CloseChannel modifies the Channel with the new State set to CLOSED and FlushStatus set to NOTINFLUSH.
 func CloseChannel(channel *Channel) {
 	channel.State = CLOSED
--- a/modules/core/04-channel/types/channel.go
+++ b/modules/core/04-channel/types/channel.go
@@ -75,6 +75,13 @@ func (ch Channel) ValidateBasic() error {
 	return ch.Counterparty.ValidateBasic()
 }
 
+
+// CloseChannel modifies the Channel with the new State set to CLOSED and FlushStatus set to NOTINFLUSH.
+func CloseChannel(channel *Channel) {
+	channel.State = CLOSED
+	channel.FlushStatus = NOTINFLUSH
+}
+
 // NewCounterparty returns a new Counterparty instance
 func NewCounterparty(portID, channelID string) Counterparty {
 	return Counterparty{
--- a/modules/core/04-channel/types/channel.go
+++ b/modules/core/04-channel/types/channel.go
@@ -75,10 +75,10 @@ func (ch Channel) ValidateBasic() error {
 	return ch.Counterparty.ValidateBasic()
 }
 
-// CloseChannel modifies the Channel with the new State set to CLOSED and FlushStatus set to NOTINFLUSH.
-func CloseChannel(channel *Channel) {
-	channel.State = CLOSED
-	channel.FlushStatus = NOTINFLUSH
+// Close modifies the Channel with the new State set to CLOSED and FlushStatus set to NOTINFLUSH.
+func (ch *Channel) Close() {
+	ch.State = CLOSED
+	ch.FlushStatus = NOTINFLUSH
 }
 
 // NewCounterparty returns a new Counterparty instance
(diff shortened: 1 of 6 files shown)


charleenfei (commented):
thank you! i guess because we are modifying the state of the channel on which the handler is operating, your suggestion to use a method seems to be nicer?

modules/core/04-channel/types/channel.go
```diff
@@ -75,6 +75,12 @@ func (ch Channel) ValidateBasic() error {
 	return ch.Counterparty.ValidateBasic()
 }
 
+// CloseChannel modifies the Channel with the new State set to CLOSED and FlushStatus set to NOTINFLUSH.
+func CloseChannel(channel *Channel) {
```
DimitrisJim, line 79: any reason you went with non-method option? I see we don't generally add mutating methods to these proto-generated types but this feels like it should definitely be one.

Also, and this is personal taste, `types.CloseChannel(channel)` reads oddly in relation to `channel.Close()`.

chatton: I don't have a strong preference, I guess the main reason would be that the following use looks fine

```golang

channel := NewChannel(...)

channel.Close() 

SetChannel(channel)

```

when it would in fact not do anything (linters would probably actually catch this though)

Correct usage would be

```golang
channel = channel.Close()
```

using a method receiver makes it obvious we're modifying values in-place. If there are preferences on the team I'd be happy to change to whatever people think is the easiest to work with while also being intuitive.

DimitrisJim: wait why not just take it by pointer? 

```go
func (ch *Channel) SetClosed() {
	ch.State = CLOSED
	ch.FlushStatus = NOTINFLUSH
}
```

?

chatton: the main reason is that we don't really deal with references to channels, e.g. NewChannel returns a struct type, and the SetChannel keeper fn accepts a struct. We would need to do awkward conversions back and forward to persist the changes.

chatton: I think I'd be happy with a `channel.Close()` in that case.

DimitrisJim (approved):
lgtm

charleenfei (approved):

modules/core/04-channel/types/channel.go
crodriguezvega: When I first had a quick look at the PR, I also thought that having a method on a pointer receiver would look better, but  then I read online that it seems like generally the consensus ([1](https://go.dev/tour/methods/8), [2](https://yourbasic.org/golang/pointer-vs-value-receiver/)) is to avoid mixing value and pointer receivers?

crodriguezvega:
If we go for the pointer receiver solution, it would be nice to update the description of the PR, @chatton :)

modules/core/04-channel/types/channel.go
```diff
@@ -75,6 +75,12 @@ func (ch Channel) ValidateBasic() error {
 	return ch.Counterparty.ValidateBasic()
 }
 
+// Close modifies the Channel with the new State set to CLOSED and FlushStatus set to NOTINFLUSH.
+func (ch *Channel) Close() {
+	ch.State = CLOSED
+	ch.FlushStatus = NOTINFLUSH
+}
```
damiannolan, line 82: if we refactor things to put the flush status on the `Upgrade` then I don't think we'll need this method

damiannolan (approved):

modules/core/04-channel/types/channel.go
chatton: yep good point, We can maybe just revert this PR if that happens

modules/core/04-channel/types/channel.go
DimitrisJim: Yeah, in general not a good idea to mix these when also dealing with interfaces since you might have cases where something looks like it implements an interface but in reality doesn't because it has a pointer receiver.

I honestly doubt we'll even have an interface for closing these, seems like it wouldn't be something we'd expose. Current interfaces in `exported` only deal with `get`ting things so we should be fine.

Mixing these can be done relatively safely if done with care, imho.

Issue #1900, opened 2022-08-06
Implement `29-fee` channel upgrade callbacks


damiannolan:
## Summary

With the addition #1831, the `04-channel` upgrade callbacks will be added to the `IBCModule` interface.
This enforces applications to implement the required callbacks.

The `29-fee` module must handle the upgrade callbacks appropriately and follow similar logic to the channel opening handshake:

- Attempt to unmarshal the provided `version` string to `types.Metadata`. 
- In the event of failure it should pass through to the underlying app. 
- The provided `FeeVersion` must be compatible with the app version: `FeeVersion == types.Version`
- The `OnChanUpgradeInit` and `OnChanUpgradeTry` callbacks should handle the underlying app version and set it in the JSON encoded `types.Metadata` string returned.
- Set the channel as fee enabled. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
Needs https://github.com/cosmos/ibc-go/issues/4125 for `OnChanUpgradeRestore`

colin-axner:
I think we got confused implementing this issue since it was written in 2022 and didn't account for the latest designs. The fee enablement change should only occur upon a successful complemention of the upgrade handshake. This would result in `OnChanUpgradeRestore` being a no-op. Given the latest spec changes, our handlers should perform the following functionality:
**onChanUpgradeInit** validates the upgrade version
**onChanUpgradeTry** validates the upgrade version 
**onChanUpgradeAck** validates the upgrade version
**onChanUpgradeOpen** upgrade to new fields (set fee enable)
**onChanUpgradeRestore** should no-op

damiannolan:
I agree, that given the new design structure (upgrade fields are not taking effect until `ChanUpgradeOpen`) that it probably makes more sense to not make application level changes until the `onOpen` callback. 
But we can't make any guarantees that other applications won't do what they want on each callback handler.

With the current API for callbacks I don't think we can actually make the above changes work:

```go
OnChanUpgradeOpen(ctx sdk.Context, portID, channelID string)
```

We get the `port/channel` from ibc core and that is it. I know we already had some discussions about changing these APIs, I'm not sure what would actually work the best for applications. But something like below might work better, it would also provide a more forwards compatible API if new fields are added for upgradability.

```go
[...]

PR #4019, opened 2023-07-05
feat: adding `OnChanUpgradeInit` handler implementation to `29-fee`

damiannolan:
## Description

- Adds the `OnChanUpgradeInit` handler to `29-fee`.
- Testing library changes are cherry-picked from #4018 

**NOTE**: this omits the allowance of the empty string which is used in the opening handshake to automatically select the version if none is provided by the client.

ref: #1900

---

write a little note why.

### DIFF
--- a/modules/apps/29-fee/ibc_middleware.go
+++ b/modules/apps/29-fee/ibc_middleware.go
@@ -324,9 +324,8 @@ func (im IBCMiddleware) OnTimeoutPacket(
 func (im IBCMiddleware) OnChanUpgradeInit(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, sequence uint64, version, previousVersion string) (string, error) {
 	var versionMetadata types.Metadata
 	if err := types.ModuleCdc.UnmarshalJSON([]byte(version), &versionMetadata); err != nil {
-		// Since it is valid for fee version to not be specified, the above middleware version may be for a middleware
-		// lower down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying
-		// application.
+		// Since it is valid for fee version to not be specified, the above upgrade version may be for a middleware
+		// or application further down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying application.
 		return im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, version, previousVersion)
 	}
 
--- a/modules/apps/29-fee/ibc_middleware.go
+++ b/modules/apps/29-fee/ibc_middleware.go
@@ -352,7 +352,7 @@ func (im IBCMiddleware) OnChanUpgradeInit(ctx sdk.Context, portID, channelID str
 
 // OnChanUpgradeTry implement s the IBCModule interface
 func (im IBCMiddleware) OnChanUpgradeTry(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, counterpartyVersion string) (string, error) {
-	return counterpartyVersion, nil
+	return im.app.OnChanUpgradeTry(ctx, portID, channelID, order, connectionHops, counterpartyVersion)
 }
 
 // OnChanUpgradeAck implements the IBCModule interface
--- a/modules/apps/29-fee/ibc_middleware.go
+++ b/modules/apps/29-fee/ibc_middleware.go
@@ -322,12 +322,63 @@ func (im IBCMiddleware) OnTimeoutPacket(
 
 // OnChanUpgradeInit implements the IBCModule interface
 func (im IBCMiddleware) OnChanUpgradeInit(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, sequence uint64, version, previousVersion string) (string, error) {
-	return im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, version, previousVersion)
+	var versionMetadata types.Metadata
+	if err := types.ModuleCdc.UnmarshalJSON([]byte(version), &versionMetadata); err != nil {
+		// Since it is valid for fee version to not be specified, the above middleware version may be for a middleware
+		// lower down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying
+		// application.
+		return im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, version, previousVersion)
+	}
+
+	if versionMetadata.FeeVersion != types.Version {
+		return "", errorsmod.Wrapf(types.ErrInvalidVersion, "expected %s, got %s", types.Version, versionMetadata.FeeVersion)
+	}
+
+	appVersion, err := im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, versionMetadata.AppVersion, previousVersion)
+	if err != nil {
+		return "", err
+	}
+
+	versionMetadata.AppVersion = appVersion
+	versionBz, err := types.ModuleCdc.MarshalJSON(&versionMetadata)
+	if err != nil {
+		return "", err
+	}
+
+	im.keeper.SetFeeEnabled(ctx, portID, channelID)
+
+	return string(versionBz), nil
 }
 
 // OnChanUpgradeTry implement s the IBCModule interface
 func (im IBCMiddleware) OnChanUpgradeTry(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, counterpartyVersion string) (string, error) {
-	return im.app.OnChanUpgradeTry(ctx, portID, channelID, order, connectionHops, counterpartyVersion)
+	var versionMetadata types.Metadata
+	if err := types.ModuleCdc.UnmarshalJSON([]byte(counterpartyVersion), &versionMetadata); err != nil {
+		// Since it is valid for fee version to not be specified, the above middleware version may be for a middleware
+		// lower down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying
+		// application.
+		return im.app.OnChanUpgradeTry(ctx, portID, channelID, order, connectionHops, counterpartyVersion)
+	}
+
+	if versionMetadata.FeeVersion != types.Version {
+		return "", errorsmod.Wrapf(types.ErrInvalidVersion, "expected %s, got %s", types.Version, versionMetadata.FeeVersion)
+	}
+
+	// call underlying app's OnChanOpenTry callback with the app versions
+	appVersion, err := im.app.OnChanUpgradeTry(ctx, portID, channelID, order, connectionHops, versionMetadata.AppVersion)
+	if err != nil {
+		return "", err
+	}
+
+	versionMetadata.AppVersion = appVersion
+	versionBz, err := types.ModuleCdc.MarshalJSON(&versionMetadata)
+	if err != nil {
+		return "", err
+	}
+
+	im.keeper.SetFeeEnabled(ctx, portID, channelID)
+
+	return string(versionBz), nil
 }
 
 // OnChanUpgradeAck implements the IBCModule interface
--- a/modules/apps/29-fee/ibc_middleware.go
+++ b/modules/apps/29-fee/ibc_middleware.go
@@ -352,33 +352,7 @@ func (im IBCMiddleware) OnChanUpgradeInit(ctx sdk.Context, portID, channelID str
 
 // OnChanUpgradeTry implement s the IBCModule interface
 func (im IBCMiddleware) OnChanUpgradeTry(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, counterpartyVersion string) (string, error) {
-	var versionMetadata types.Metadata
-	if err := types.ModuleCdc.UnmarshalJSON([]byte(counterpartyVersion), &versionMetadata); err != nil {
-		// Since it is valid for fee version to not be specified, the above middleware version may be for a middleware
-		// lower down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying
-		// application.
-		return im.app.OnChanUpgradeTry(ctx, portID, channelID, order, connectionHops, counterpartyVersion)
-	}
-
-	if versionMetadata.FeeVersion != types.Version {
-		return "", errorsmod.Wrapf(types.ErrInvalidVersion, "expected %s, got %s", types.Version, versionMetadata.FeeVersion)
-	}
-
-	// call underlying app's OnChanOpenTry callback with the app versions
-	appVersion, err := im.app.OnChanUpgradeTry(ctx, portID, channelID, order, connectionHops, versionMetadata.AppVersion)
-	if err != nil {
-		return "", err
-	}
-
-	versionMetadata.AppVersion = appVersion
-	versionBz, err := types.ModuleCdc.MarshalJSON(&versionMetadata)
-	if err != nil {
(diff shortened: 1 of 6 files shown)


modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -1004,6 +1004,90 @@ func (suite *FeeTestSuite) TestOnTimeoutPacket() {
...
+			nil,
+		},
+		{
+			"invalid upgrade version",
+			func() {
+				path.EndpointA.ChannelConfig.ProposedUpgrade.Fields.Version = "invalid-version"
+				path.EndpointB.ChannelConfig.ProposedUpgrade.Fields.Version = "invalid-version"
+
+				suite.chainA.GetSimApp().FeeMockModule.IBCApp.OnChanUpgradeInit = func(_ sdk.Context, _, _ string, _ channeltypes.Order, _ []string, _ uint64, _, _ string) (string, error) {
+					return "", ibcmock.MockApplicationCallbackError
```
damiannolan, line 1027: Intentionally force the error here so we can assert that a passthrough results in `isFeeEnabled == false`

modules/apps/29-fee/ibc_middleware.go
```diff
@@ -322,7 +322,32 @@ func (im IBCMiddleware) OnTimeoutPacket(
...
+		return "", err
+	}
+
+	versionMetadata.AppVersion = appVersion
+	versionBz, err := types.ModuleCdc.MarshalJSON(&versionMetadata)
+	if err != nil {
+		return "", err
+	}
+
+	im.keeper.SetFeeEnabled(ctx, portID, channelID)
```
charleenfei, line 348: do we already want to setFeeEnabled at this step?

damiannolan: we set is as enabled in `OnChanOpenInit` so I expected the upgrade handlers should follow suit, setting fee enabled on INIT and TRY, and then asserting that the handshake must finish as fee enabled in ACK and CONFIRM.

charleenfei: ah okay, this makes sense. thanks!

charleenfei: do you know the reasoning we set fee enabled on INIT and TRY already for `OnChanOpenInit`?

damiannolan: I'm not 100% sure, it's been that way since @AdityaSripal added the initial implementation in #307. 
I imagine it could be useful if the underlying app is doing something in the channel handshake handlers (INIT/TRY) based on whether fee is enabled or not.

modules/apps/29-fee/ibc_middleware.go
```diff
@@ -322,7 +322,32 @@ func (im IBCMiddleware) OnTimeoutPacket(
 
 // OnChanUpgradeInit implements the IBCModule interface
 func (im IBCMiddleware) OnChanUpgradeInit(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, sequence uint64, version, previousVersion string) (string, error) {
-	return im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, version, previousVersion)
+	var versionMetadata types.Metadata
+	if err := types.ModuleCdc.UnmarshalJSON([]byte(version), &versionMetadata); err != nil {
+		// Since it is valid for fee version to not be specified, the above middleware version may be for a middleware
+		// lower down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying
+		// application.
```
damiannolan, line 329:
```suggestion
		// Since it is valid for fee version to not be specified, the above upgrade version may be for a middleware
		// or application further down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying application.
```

modules/apps/29-fee/ibc_middleware.go
```diff
@@ -322,7 +322,31 @@ func (im IBCMiddleware) OnTimeoutPacket(
 
 // OnChanUpgradeInit implements the IBCModule interface
 func (im IBCMiddleware) OnChanUpgradeInit(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, sequence uint64, version, previousVersion string) (string, error) {
-	return im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, version, previousVersion)
+	var versionMetadata types.Metadata
+	if err := types.ModuleCdc.UnmarshalJSON([]byte(version), &versionMetadata); err != nil {
```
crodriguezvega, line 326: It's currently [not allowed to initiate an upgrade with an empty `version`](https://github.com/cosmos/ibc-go/blob/3203f07f10c76d17e094773754f5da10c49933c5/modules/core/04-channel/types/upgrade.go#L54-L56), but in [`OnChanOpenInit` of 29-fee if `version` is empty, then we use the default metadata](https://github.com/cosmos/ibc-go/blob/a9c97c4024e2bd45c370e6d0829279437a0e8eee/modules/apps/29-fee/ibc_middleware.go#L49-L54) (similar thing happens in `OnChanOpenInit` of transfer). To make the open an upgrade flows more similar, would it make sense to allow empty versions and then fill them in with a default in the application callbacks?

modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -1004,6 +1004,90 @@ func (suite *FeeTestSuite) TestOnTimeoutPacket() {
...
+	for _, tc := range testCases {
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest()
+
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+
+			// configure the initial path to create an unincentivized mock channel
+			path.EndpointA.ChannelConfig.PortID = ibctesting.MockFeePort
+			path.EndpointB.ChannelConfig.PortID = ibctesting.MockFeePort
```
crodriguezvega, line 1061: Do we need all these `path.EndpointB.ChannelConfig...`? Asking because we are only doing `ChanUpgradeInit` on endpoint A.

crodriguezvega (commented):
Thanks, @damiannolan. I left a couple of questions.

modules/apps/29-fee/ibc_middleware.go
damiannolan: We discussed potentially that the empty string in an `Upgrade` may be used to signal that the version should remain unchanged. Whether we implement that or not is tbd.

I decided to leave this out of the upgrade path for fee, as I think for upgrades you should have to be explicit about the version you are providing. I discussed this async with @AdityaSripal and he agreed

modules/apps/29-fee/ibc_middleware_test.go
damiannolan: Yes, because we use these fields in order to create the initial uncentivized channel without fee enabled.

crodriguezvega (approved):
Nice work, @damiannolan.

modules/apps/29-fee/ibc_middleware.go
```diff
@@ -322,7 +322,31 @@ func (im IBCMiddleware) OnTimeoutPacket(
 
 // OnChanUpgradeInit implements the IBCModule interface
 func (im IBCMiddleware) OnChanUpgradeInit(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, sequence uint64, version, previousVersion string) (string, error) {
```
colin-axner, line 324: super nit: should we rename `sequence` to `upgradeSequence`?

modules/apps/29-fee/ibc_middleware.go
```diff
@@ -322,7 +322,31 @@ func (im IBCMiddleware) OnTimeoutPacket(
...
-	return im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, version, previousVersion)
+	var versionMetadata types.Metadata
+	if err := types.ModuleCdc.UnmarshalJSON([]byte(version), &versionMetadata); err != nil {
+		// Since it is valid for fee version to not be specified, the above upgrade version may be for a middleware
+		// or application further down in the stack. Thus, if it is not a fee version we pass the entire version string onto the underlying application.
+		return im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, version, previousVersion)
+	}
```
colin-axner, line 330: We should ensure fee is disabled in this case, it might be nice to add some descriptive bool:

```go
err := types.ModuleCdc.UnmarshalJSON([]byte(version), &versionMetadata)
disableFees := err != nil

if disableFees {
    im.keeper.SetFeeDisabled(ctx, portID, channelID)
    return im.app.OnChanUpgradeInit(ctx, portID, channelID, order, connectionHops, sequence, version, previousVersion)
}

File: modules/apps/29-fee/ibc_middleware_test.go [hunk 8]
```diff
@@ -1004,6 +1004,90 @@ func (suite *FeeTestSuite) TestOnTimeoutPacket() {
...
+	testCases := []struct {
+		name     string
+		malleate func()
+		expError error
+	}{
+		{
+			"success",
+			func() {},
+			nil,
+		},
```
colin-axner, line 1019: add success case for going from fee enabled to disabled

modules/apps/29-fee/ibc_middleware_test.go
colin-axner: should we mention this in the test case?

colin-axner (commented):
Excellent work! Only requested change is that we should support downgrading channels

Thanks @damiannolan for taking this work on a splitting everything up into bite size pr's!

damiannolan:
Will try to pick up work on this again today

modules/apps/29-fee/ibc_middleware.go
damiannolan: changed the code structure a little by adding `types.MetadataFromVersion()` func, and now disabling fees on error

modules/apps/29-fee/ibc_middleware_test.go
damiannolan: yeah definitely, good call!

modules/apps/29-fee/types/metadata.go
```diff
@@ -0,0 +1,13 @@
...
+// an error is returned if it fails to do so.
+func MetadataFromVersion(version string) (Metadata, error) {
+	var metadata Metadata
+	err := ModuleCdc.UnmarshalJSON([]byte(version), &metadata)
+	if err != nil {
+		return Metadata{}, ErrInvalidVersion
+	}
+
+	return metadata, nil
+}
```
colin-axner, line 13: should we upstream this to main?

modules/apps/29-fee/types/metadata.go
```diff
@@ -0,0 +1,13 @@
+package types
+
+// MetadataFromVersion attempts to parse the given string into a fee version Metadata,
+// an error is returned if it fails to do so.
+func MetadataFromVersion(version string) (Metadata, error) {
```
colin-axner, line 5: Hmm, I feel the naming could be improved, but I don't have a good suggestion in mind. I think my hesitation lies around the fact that `Metadata` and `Version` aren't super well defined. Maybe something more explicit could help like:
* `ParseMetadataFromVersionString`
* `UnmarshalMetadataFromVersionString`

Another option is:
```go
var metadata Metadata
err := metadata.Unmarshal(version)
```

I guess the current naming is supposed to mimic `AccAddressFromBech32`

Given that I don't have a concrete suggestion, I'm fine to leave as is if you are happy with it

modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -1004,6 +1004,119 @@ func (suite *FeeTestSuite) TestOnTimeoutPacket() {
...
+		{
+			"invalid upgrade version",
+			func() {
+				expFeeEnabled = false
+				path.EndpointA.ChannelConfig.ProposedUpgrade.Fields.Version = "invalid-version"
+				path.EndpointB.ChannelConfig.ProposedUpgrade.Fields.Version = "invalid-version"
+
+				suite.chainA.GetSimApp().FeeMockModule.IBCApp.OnChanUpgradeInit = func(_ sdk.Context, _, _ string, _ channeltypes.Order, _ []string, _ uint64, _, _ string) (string, error) {
+					// intentionally force the error here so we can assert that a passthrough results in isFeeEnabled == false
```
colin-axner, line 1052:
```suggestion
					// intentionally force the error here so we can assert that a passthrough occurs when fees should not be enabled for this channel
```

colin-axner: this is the intention right? Otherwise I don't follow. The only thing is that a passthrough should eventually occur

modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -1004,6 +1004,119 @@ func (suite *FeeTestSuite) TestOnTimeoutPacket() {
...
+
+			tc.malleate()
+
+			err := path.EndpointA.ChanUpgradeInit()
+
+			isFeeEnabled := suite.chainA.GetSimApp().IBCFeeKeeper.IsFeeEnabled(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+
+			expPass := tc.expError == nil
+			if expPass {
+				suite.Require().Equal(expFeeEnabled, isFeeEnabled)
```
colin-axner, line 1109: this can be abstracted out, but fine to leave as is

modules/apps/29-fee/types/metadata.go
```diff
@@ -0,0 +1,13 @@
...
+	var metadata Metadata
+	err := ModuleCdc.UnmarshalJSON([]byte(version), &metadata)
+	if err != nil {
+		return Metadata{}, ErrInvalidVersion
```
colin-axner, line 9: Should we indicate that the version could not be unmarshaled into the metadata?

colin-axner (approved):
ACK, functionally looks good to me. Left minor nits on the tests and some comments on the helper function. Happy to see the helper function comments addressed in a followup pr to main (where it can be utilized in other parts of the code and rectified in the upgrade branch)

modules/apps/29-fee/types/metadata.go
damiannolan: Originally was thinking that this would be wrapped by the caller like for example: `errorsmod.Wrapf(err, "failed to unmarshal ICS29 counterparty version metadata: %s)`

But I like this suggestion! Changed it to return this here: `errorsmod.Wrapf(ErrInvalidVersion, "failed to unmarshal metadata from version: %s", version)`. From caller we can just `Wrap(err, "invalid counterparty version")`

modules/apps/29-fee/ibc_middleware_test.go
damiannolan: nice catch

modules/apps/29-fee/types/metadata.go
damiannolan: Will leave as is for now 👍

modules/apps/29-fee/types/metadata.go
damiannolan: Yea, was going to open an issue to use this in some other places! I can cherrypick the commits to main

modules/apps/29-fee/types/metadata.go
chatton: Happy to leave as is but I will throw out one more suggestion; `ExtractMetadataFromVersionString`

chatton (approved):
LGTM 🍏 

Implementation looks clean 👍

Issue #3497, opened 2023-04-20
Add Unit Tests for Channel Upgrade Event Emission


chatton:
## Summary 

Ref: [this comment](https://github.com/cosmos/ibc-go/pull/3456#discussion_r1170917426)
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

DimitrisJim:
Guess this can stay open to keep track of `ChanUpgradeConfirm`? (Or we could just open an additional issue tbh)

chatton:
Un-assigning myself for now since the original PR was merged, can leave this issue open until all the events are tested.

charleenfei:
I think the only one left is `UpgradeError`? can update with this one

charleenfei:
on hold until [this pr](https://github.com/cosmos/ibc-go/pull/5264) is merged

colin-axner:
There's commented out tests in `TestWriteUpgradeOpenChannel_UnorderedToOrdered()`

charleenfei:
Yea, I think all of them need the event testing now after the events were moved, will work on all of them in one go now

PR #3883, opened 2023-06-17
deps: sdk v0.50-alpha

faddat:
## Description

Upgrade to sdk 50

Related PRs in ibc-go:

* #4057 
* https://github.com/cosmos/ibc-go/pull/4159
* #4009 
* #3874 
* https://github.com/cosmos/ibc-go/pull/3811
* #3835 
* #3887 
* #3809 
* #3808 
* #3888 

closes: #3707 

Related PRs outside of ibc-go:

* https://github.com/CosmWasm/wasmd/pull/1534
* https://github.com/CosmWasm/wasmd/pull/1532
* https://github.com/strangelove-ventures/interchaintest/pull/674

### Commit Message / Changelog Entry

```bash
chore: upgrade to sdk 50
```

see the [guidelines](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) for commit messages. (view raw markdown for examples)

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#pull-request-targeting)).
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/11-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [commit message](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) to be used for the changelog entry in the PR description for review.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.
- [ ] Review `Codecov Report` in the comment section below once CI passes.

PR #4020, opened 2023-07-05
Add test for event emission ChanUpgradeInit

chatton:
## Description

closes: #3497 

N/A Feature branch

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade_test.go
+++ b/modules/core/04-channel/keeper/upgrade_test.go
@@ -774,6 +774,300 @@ func (suite *KeeperTestSuite) TestChanUpgradeOpen() {
 	}
 }
 
+// TestChanUpgradeOpenCounterPartyStates tests the handshake in the cases where
+// the counterparty is in a state other than OPEN.
+func (suite *KeeperTestSuite) TestChanUpgradeOpenCounterpartyStates() {
+	var path *ibctesting.Path
+	testCases := []struct {
+		name     string
+		malleate func()
+		expError error
+	}{
+		{
+			"success, counterparty in OPEN",
+			func() {
+				err := path.EndpointB.ChanUpgradeInit()
+				suite.Require().NoError(err)
+
+				err = path.EndpointA.ChanUpgradeTry()
+				suite.Require().NoError(err)
+
+				err = path.EndpointB.ChanUpgradeAck()
+				suite.Require().NoError(err)
+
+				suite.coordinator.CommitBlock(suite.chainA, suite.chainB)
+				suite.Require().NoError(path.EndpointA.UpdateClient())
+			},
+			nil,
+		},
+		{
+			"success, counterparty in TRYUPGRADE",
+			func() {
+				// Need to create a packet commitment on A so as to keep it from going to OPEN if no inflight packets exist.
+				sequence, err := path.EndpointA.SendPacket(defaultTimeoutHeight, disabledTimeoutTimestamp, ibctesting.MockPacketData)
+				suite.Require().NoError(err)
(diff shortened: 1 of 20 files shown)


faddat:
I have a sneaking suspicion that this may be relevant to or assist with #3883

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -117,13 +117,33 @@ func (suite *KeeperTestSuite) TestChanUpgradeInit() {
...
+						types.AttributeKeyUpgradeConnectionHops:     upgrade.Fields.ConnectionHops[0],
+						types.AttributeKeyUpgradeVersion:            upgrade.Fields.Version,
+						types.AttributeKeyUpgradeOrdering:           upgrade.Fields.Ordering.String(),
+						types.AttributeKeyUpgradeSequence:           fmt.Sprintf("%d", channel.UpgradeSequence),
+						types.AttributeKeyUpgradeChannelFlushStatus: channel.FlushStatus.String(),
+					},
+					sdk.EventTypeMessage: {
+						sdk.AttributeKeyModule: types.AttributeValueCategory,
+					},
+				}
```
chatton, line 140: This doesn't seem to be a super clean looking test, I'm open to suggestions for cleaning this up.

Maybe we could create a function that returns the expected events for `OnChanUpgradeInit` and re-use that both in this test and where the events are emitted.

This might look a little cleaner, but the current version is more explicit.

DimitrisJim: > Maybe we could create a function that returns the expected events for OnChanUpgradeInit and re-use that both in this test and where the events are emitted.

that does also sound better to me? Split the emitting from the creating in `events.go`. Addtional hurdle is adding all creating of events in `export_test` to have them be testable.

crodriguezvega: The only issue I see of using here the function that creates the event is that we are using the code that we want to test to actually test it, so if in the future we remove by mistake an attribute, the test will still pass... The same thing as if we use the attribute key constants here: if at some point we changed by mistake the string, we will not catch that here...

Another thing to take into account is that with SDK v0.50 an extra attribute is appended to the events, so we need to take that one into account when doing the check. [This was my hacky way of quickly fixing it in the SDK v0.50 branch](https://github.com/cosmos/ibc-go/pull/3883/commits/72796c495d7368e4d5a36b760d92dd28ac13889e).

DimitrisJim: ah, I thought we were testing for contents, not structure really. that's a valid point tho.

chatton: Maybe we should be less strict with the `AssertEvents` fn in that case, and just verify that the provided events are a subset of the actual events, I figure as far as this test is concerned, we should only need to worry about the upgradability related events, and not any additional events that sdk emits

Or even 

```golang
sdk.EventTypeMessage: {
    sdk.AttributeKeyModule: types.AttributeValueCategory,
},
```

I feel like our current [check](https://github.com/cosmos/ibc-go/blob/dffc7b7d12f644fad4b8f1b8b882b35653322ed8/testing/events.go#L153) for length is actually going to be quite fragile, and break all event tests in the case of a new sdk message that we are not concerned with.

What do you think?

colin-axner: The referenced length check should be fine because it references event attributes not event types, so a new sdk event shouldn't affect our events length of attributes. I believe we are searching a list of actual events for a subset of expected events.

I'm fine with how the code is currently situated. I agree it'd be nice to improve our approach to event testing so it feels natural. One problem we have already seen with the map approach is that it doesn't account for multiple emissions of the same event type

chatton: okay sounds good to me, we can merge this one as is and iterate upon it later.

damiannolan (approved):

Issue #3910, opened 2023-06-21
Consolidate channel upgrade message responses


chatton:
## Summary

Currently, we make use of a ResultType to indicate if a response was successful or failure as a way of making sure an error receipt is written to state and to inform the relayers that an upgrade should be cancelled.

However some other message responses don't have this field such as [MsgChannelUpgradeInitResponse](https://github.com/cosmos/ibc-go/blob/26daed9319de44494da753dff26985cdeae82043/modules/core/04-channel/types/tx.pb.go#L949) and [MsgChannelUpgradeCancelResponse](https://github.com/cosmos/ibc-go/blob/26daed9319de44494da753dff26985cdeae82043/modules/core/04-channel/types/tx.pb.go#L1366).

While they do not require the result to work, it may be a good idea to make the API consistent across all messages.

DimitrisJim:
adding discussion label back, missed to talk about it in previous eng. call (see https://github.com/cosmos/ibc-go/pull/4022#discussion_r1258369909)

crodriguezvega:
Use `ResponseResultType` in Ack and Confirm responses.

colin-axner:
Yea, I think the response result can only be success in the other upgrade handlers as a failure indicates a successful tx but a failed handshake step. If we find it useful to return information about the upgrade in the resposne, I'd recommend changing the result type to string and using enums which can be updated in post release to make the other handshake steps more useful:
```
type MsgChannelUpgradeTryResponse struct {
	ChannelId       string             `protobuf:"bytes,1,opt,name=channel_id,json=channelId,proto3" json:"channel_id,omitempty"`
	Upgrade         Upgrade            `protobuf:"bytes,2,opt,name=upgrade,proto3" json:"upgrade"`
	UpgradeSequence uint64             `protobuf:"varint,3,opt,name=upgrade_sequence,json=upgradeSequence,proto3" json:"upgrade_sequence,omitempty"`
	Status          string `protobuf:"varint,4,opt,name=status,proto3,json:"result,omitempty"`
}
```

```go
var (
    SUCCESS_UPGRADE_INIT = "SUCCESS_UPGRADE_INIT"
    SUCCESS_UPGRADE_FLUSHING = "SUCCESS_UPGRADE_FLUSHING"
    SUCCESS_UPGRADE_FLUSHCOMPLETE = "SUCCESS_UPGRADE_FLUSHCOMPLETE"
[...]

PR #4022, opened 2023-07-06
Add result field for upgrade responses

DimitrisJim:
## Description

Add result to all responses and tests for them where applicable.

closes: #3910

```bash
chore: consolidate channel upgrade message responses
```

---

write a little note why.


proto/ibc/core/channel/v1/tx.proto
```diff
@@ -266,9 +266,10 @@ message MsgChannelUpgradeInit {
 message MsgChannelUpgradeInitResponse {
   option (gogoproto.goproto_getters) = false;
 
-  string  channel_id       = 1;
-  Upgrade upgrade          = 2 [(gogoproto.nullable) = false];
-  uint64  upgrade_sequence = 3;
+  string             channel_id       = 1;
+  Upgrade            upgrade          = 2 [(gogoproto.nullable) = false];
+  uint64             upgrade_sequence = 3;
+  ResponseResultType result           = 4;
```
crodriguezvega, line 272: Wondering if we should make `result` the first field, just in case in the future we want to add more fields to the response and `result` is not in between.

crodriguezvega (approved):
Sweet! Left just one question; do not feel very strongly about either way.

proto/ibc/core/channel/v1/tx.proto
DimitrisJim: yeah, I thought about this myself. I think we should definitely raise the point during final walk-through/audit. willing to make the change here if anyone else shares the opinion.

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -375,4 +380,8 @@ message MsgChannelUpgradeCancel {
 }
 
 // MsgChannelUpgradeCancelResponse defines the MsgChannelUpgradeCancel response type
-message MsgChannelUpgradeCancelResponse {}
+message MsgChannelUpgradeCancelResponse {
+  option (gogoproto.goproto_getters) = false;
+
+  ResponseResultType result = 1;
```
colin-axner, line 386: It's a bit confusing to me that the result type `Success` means the handshake wasn't aborted or it was aborted, depending on the message type. I wonder if it'd be simpler to add a bool field for message types where the handshake could be aborted

DimitrisJim: We should discuss in engineering since this is part of the API and would be nice to get it right. I can see the point raised and how it might lead to some confusion (but also like the idea of going with `result` types everywhere if at least one handler uses it).

Issue #3975, opened 2023-06-27
Make separate go.mod for 08-wasm 


crodriguezvega:
To reduce dependencies of ibc-go and decouple the release of 08-wasm from ibc-go, the module should have its own go.mod.

crodriguezvega:
Closed by #4026

PR #4026, opened 2023-07-06
imp: make separate go.mod for 08-wasm

crodriguezvega:
## Description

closes: #3975 

```bash
imp: make separate go.mod for 08-wasm
```

---

write a little note why.


crodriguezvega:
The [error here](https://github.com/cosmos/ibc-go/actions/runs/5475897647/jobs/9972649441?pr=4026) I've seen it happening already for some time, but I haven't figure out what the problem is, because the type is correctly imported...

DimitrisJim:
I think wasmvm ain't playing nicely with arm, the relevant code is not generated in this case. It builds just fine locally (amd). Trying to think of a way to get this to be skipped when dealing with the wasm client.

charleenfei:
ahhh, this makes sense! i was really confused why it was working for me locally as well.

damiannolan:
can we try this [here](https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/.github/workflows/e2e.yaml#L79) to see if we can compile? There is no e2e tests using the 08-wasm features atm right?

```diff
- go test -c "$dir"
+ CGO_ENABLED=0 go test -c "$dir"
```

Has anyone managed to run our simd docker image and use a wasm client yet? I suspect the build setup in the Dockerfile will require some configuration judging by the [wasmvm readme](https://github.com/CosmWasm/wasmvm/blob/main/README.md). 

Does running wasm clients work out of the box on macos? I imagine it requires some rust libraries etc for simd to instantiate the vm?

DimitrisJim:
> Oh I just saw https://github.com/cosmos/ibc-go/pull/4029, maybe CI needs to be re-run after a merge

This should fix it for the `amd` case. For arm, I'm afraid the builds might need to use the `mac-os` runner instead of `ubuntu-latest` with a `go-arch` arg, cross compiling with `cgo` seems to just not work right. In this case, it's not entirely sure to me how I could replicate the full matrix (i.e `arm64` and `arm`, IIRC `arm` is 32bit?).

> Has anyone managed to run our simd docker image and use a wasm client yet?

The current error when building the simd image with the dockerfile stems from the fact that we add the modules after trying to download dependencies (since we pin 08-wasm to a local directory, it fails since we add the code in a later step). After reorganizing, the build succeeds but we do indeed fail again since we don't have shared library for libwasmvm. I think we could take inspiration from [here](https://github.com/CosmWasm/wasmd/blob/main/Dockerfile#L18) but I haven't attempted it. yet.

crodriguezvega:
> edit: Oh I just saw https://github.com/cosmos/ibc-go/pull/4029, maybe CI needs to be re-run after a merge

I merged and re-run, but the error was still happening... :(

crodriguezvega:
All checks have passed, @DimitrisJim and @charleenfei! 🥳

Issue #2442, opened 2022-09-29
Change simd base image to reduce image size


chatton:
## Summary of Bug

Currently we are using `ubuntu:20.04` as the base image for simd/icad. We could switch to an alpine based image to reduce image size and result in a minor speed improvement in e2e. (more so for forks as we only build the image once when running e2e )

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

DimitrisJim:
side comment: we build simd for fork pr's using `ubuntu-latest`, we might want to keep them in-sync unless we switch `e2e-fork.yml` to re-use the workflow `e2e.yaml` uses.

PR #4031, opened 2023-07-07
test: fix e2e and build issues (DO NOT REVIEW)

charleenfei:
## Description

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.


.github/workflows/test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-arch: ['amd64', 'arm', 'arm64']
+        go-arch: ['amd64', 'arm']
```
DimitrisJim, line 39: I think this matrix just needs to change unfortunately. Instead of cross compiling, compile using a native machine. Didn't have time to check yesterday if `mac-os` runner actually has `arm64` arch.

DimitrisJim: No linux arm support apparently: https://github.com/actions/runner-images/issues/5631
Similarly, mac-os doesn't seem to be using the new M1 arm64 chip: https://github.com/github/roadmap/issues/528

Dockerfile
```diff
@@ -11,25 +11,27 @@ RUN test -n "${IBC_GO_VERSION}"
...
 
+RUN go mod download
+
 COPY contrib/devtools/Makefile contrib/devtools/Makefile
 COPY Makefile .
 
-
 RUN make build
+# Grab the wasmvm library in order to copy it into the final image.
+RUN cp $(ldd build/simd | grep "libwasmvm" | awk '{ print $3 }') libwasmvm.so
```
DimitrisJim, line 26: A question. Do we want to keep using multi-stage builds? This (and the copy in the next image) is required because `libwasmvm` is included in the `wasmvm` go lib (which we lose access to in the next stage).

Are the space savings worth the hassle?

.github/workflows/test.yml
DimitrisJim: Update: can build on arm64, wasn't using cross compiler correctly. Can't on arm32 since [wasmvm can't support it](https://github.com/CosmWasm/wasmvm#builds-of-libwasmvm) due to Wasmer not supporting it.

The previous does apply if we'd like to run the tests on arm64 via github runners tho.

Dockerfile
DimitrisJim: Note, they are worth it. Looking into https://github.com/cosmos/ibc-go/issues/2442 though since that might offer benefits in reducing issues like this.

Issue #3959, opened 2023-06-26
Remove `code_id_key` from `GenesisContract`


crodriguezvega:
_Surfaced from Interchain IBC team's review of 08-wasm at hash https://github.com/strangelove-ventures/ibc-go/commit/96e9930ab87a171a3ae6ac9717ee25bfda77212b_

## Summary

Remove [`code_id_key` from `GenesisContract`](https://github.com/strangelove-ventures/ibc-go/blob/96e9930ab87a171a3ae6ac9717ee25bfda77212b/proto/ibc/lightclients/wasm/v1/genesis.proto#L19).

## Problem Definition

The [code ID key is used as the key where to store the contract's byte code](https://github.com/strangelove-ventures/ibc-go/blob/96e9930ab87a171a3ae6ac9717ee25bfda77212b/modules/light-clients/08-wasm/keeper/keeper.go#L127) during genesis import. However, the code ID key can be generated automatically: hash the contract byte code + prepend `codeId` key prefix. Therefore it's not really needed in `GenesisContract`, which the contract byte code is enough.

## Proposal

- Remove `code_id_key` from `GenesisContract`.
- Generate the code ID key automatically when importing the contracts.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged/assigned

crodriguezvega:
Closed by #4032

PR #4032, opened 2023-07-07
imp: remove code id key from genesis contract

crodriguezvega:
## Description

closes: #3959 

```bash
type: commit message
```

---

write a little note why.


proto/ibc/lightclients/wasm/v1/genesis.proto
```diff
@@ -15,8 +15,6 @@ message GenesisState {
 // A contract's store key and code
```
damiannolan, line 15: nit: outside the scope of this PR but proto docs should start with the message type name for both `GenesisContract` and `GenesisState`. 

Also, is the wrapper type `GenesisContract` needed? I think it would also be valid to use `repeated bytes` in `GenesisState`, but either works for me, it might be nice to keep the wrapper.

modules/light-clients/08-wasm/keeper/genesis_test.go
```diff
@@ -99,17 +70,18 @@ func (suite *KeeperTestSuite) TestExportGenesis() {
 	suite.SetupTest()
 	ctx := suite.chainA.GetContext()
 
+	expCodeID := "c64f75091a6195b036f472cd8c9f19a56780b9eac3c3de7ced0ec2e29e985b64"
```
damiannolan, line 73: I guess there is no way to avoid the magic string?

damiannolan (approved):
Ignoring CI failures. LGTM! 

side note: I spotted this part of the code in an unchanged diff:

```go
wasmCode, err = types.Uncompress(wasmCode, types.MaxWasmByteSize())
if err != nil {
	return sdkerrors.Wrap(err, "failed to store contract")
}
```

The `MaxWasmByteSize` is defined as `3 * 1024 * 1024`, something around 3MB or so? Do we have any info or data on what size we could expect? Is this a reasonable limit?

crodriguezvega:
> The MaxWasmByteSize is defined as 3 * 1024 * 1024, something around 3MB or so? Do we have any info or data on what size we could expect? Is this a reasonable limit?

The current contracts are less than 300 KB, so I guess 10 times that size should be more than enough.

proto/ibc/lightclients/wasm/v1/genesis.proto
crodriguezvega: Good point. Maybe just in case we can keep the wrapper for now, In case we add other fields in the future...

modules/light-clients/08-wasm/keeper/genesis_test.go
crodriguezvega: One way would be to make an exported wrapper for `generateWasmCodeHash` only for testing. I can look into this in a separate PR.

Issue #3957, opened 2023-06-26
Remove `timestamp` from `ConsensusState` and query contract to retrieve it


crodriguezvega:
_Surfaced from Interchain IBC team's review of 08-wasm at hash https://github.com/strangelove-ventures/ibc-go/commit/96e9930ab87a171a3ae6ac9717ee25bfda77212b_

## Summary

Remove [`timestamp`](https://github.com/strangelove-ventures/ibc-go/blob/96e9930ab87a171a3ae6ac9717ee25bfda77212b/proto/ibc/lightclients/wasm/v1/wasm.proto#L26) from `ConsensusState`.

## Problem Definition

The 08-wasm module is not completely a proxy client, since  in the implementation of [`GetTimestampAtHeight` function of the `ClientState` interface](https://github.com/strangelove-ventures/ibc-go/blob/feat/wasm-clients-main/modules/light-clients/08-wasm/types/client_state.go#L95) 08-wasm fetches the timestamp from the stored consensus state, instead of delegating to the light client contract the responsibility to provide that information.

The `timestamp` should also already be encoded in the `data` field of `ConsensusState` (and also in the `data` of `Header` when updating the client). Having a separate `timestamp` field in `ConsensusState` opens the possibility that a relayer constructs a `MsgCreateClient` with a consensus state where the timestamp does not match the timestamp encoded in the `data` field.

## Proposal

We would like light contracts to implement the [`GetTimestampAtHeight` function of the `ClientState` interface](https://github.com/strangelove-ventures/ibc-go/blob/96e9930ab87a171a3ae6ac9717ee25bfda77212b/modules/core/exported/client.go#L67), so that 08-wasm's implementation simply queries the contract to retrieve the timestamp, instead of retrieving it from the stored consensus state. The implementation of `GetTimestamp` function of the `ConsensusState` interface should return 0 (the function will be marked deprecated).

This change impacts existing implementation of light contracts, since they will need to implement an extra query to return the timestamp for the consensus state at a particular height.

Tagging @misko9 @blasrodri to hear their thoughts, since this change impacts the contracts.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged/assigned

crodriguezvega:
[...]

PR #4034, opened 2023-07-07
imp: remove `timestamp` from `consensusState` and query contract to retrieve it

crodriguezvega:
## Description

Still work in progress, since we need new versions of the contracts supporting the query to get the timestamp at a certain height. Tests are expected to fail.

closes: #3957 

```bash
type: commit message
```

---

write a little note why.


modules/light-clients/08-wasm/types/client_state.go
```diff
@@ -250,17 +257,37 @@ func call[T ContractResult](ctx sdk.Context, clientStore sdk.KVStore, cs *Client
...
+		return output, sdkerrors.Wrapf(err, "failed to unmarshal result of wasm execution")
+	}
+	if !output.Validate() {
+		return output, sdkerrors.Wrapf(errors.New(output.Error()), "error occurred while executing contract with code ID %s", hex.EncodeToString(cs.CodeId))
+	}
+	return output, nil
+}
+
+// wasmQuery queries the contract with the given payload and writes the result to output.
+func wasmQuery[T ContractResult](ctx sdk.Context, clientStore sdk.KVStore, cs *ClientState, payload any) (T, error) {
```
crodriguezvega, line 276: I added this function (which is very similar to `call` above) to wrap the query call to the contract. Since the query (still to be added) to return the timestamp at a certain height (with payload of type `timestampAtHeightQueryResponse`) might return an error if there's no consensus at that given height, I thought that we can also check the result and wrap if the error if there's any.

crodriguezvega: I had to named this function like this and not `query`, because `query` conflicted with another function defined in a package in the SDK that is also in scope. I will probably rename `call` above to `wasmExecute`, for consistency.

modules/light-clients/08-wasm/types/consensus_state.go
```diff
@@ -24,15 +22,11 @@ func (cs ConsensusState) ClientType() string {
 
 // GetTimestamp returns block time in nanoseconds of the header that created consensus state.
 func (cs ConsensusState) GetTimestamp() uint64 {
-	return cs.Timestamp
+	return 0
```
vmarkushin, line 25: Shouldn't we query the contract in this case?

modules/light-clients/08-wasm/types/client_state.go
```diff
@@ -97,18 +88,34 @@ func (cs ClientState) ZeroCustomFields() exported.ClientState {
...
+	payload := timestampAtHeightPayload{
+		TimestampAtHeight: timestampAtHeightInnerPayload{
+			Height: height,
+		},
+	}
+
+	result, err := wasmQuery[timestampAtHeightQueryResponse](ctx, clientStore, &cs, payload)
 	if err != nil {
-		return 0, sdkerrors.Wrapf(err, "height (%s)", height)
+		return 0, nil
```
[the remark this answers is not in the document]
crodriguezvega: I think this was my mistake. Thanks for catching it! I will put it back.

modules/light-clients/08-wasm/types/consensus_state.go
crodriguezvega: No, the idea is that we will mark this function as deprecated, since it is only used in the light clients, so it doesn't need to be part of the `ConsensusState` interface. In the meantime, until we remove it, we will just return 0.

modules/light-clients/08-wasm/types/client_state.go
```diff
@@ -249,20 +256,40 @@ func call[T ContractResult](ctx sdk.Context, clientStore sdk.KVStore, cs *Client
...
+// wasmQuery queries the contract with the given payload and writes the result to output.
+func wasmQuery[T ContractResult](ctx sdk.Context, clientStore sdk.KVStore, cs *ClientState, payload any) (T, error) {
+	var output T
+	encodedData, err := json.Marshal(payload)
+	if err != nil {
+		return output, errorsmod.Wrapf(err, "failed to marshal payload for wasm query")
+	}
+	out, err := queryContract(ctx, clientStore, cs.CodeHash, encodedData)
+	if err != nil {
+		return output, errorsmod.Wrapf(err, "call to wasm contract failed")
```
DimitrisJim, line 286: to distinguish between the two:

```suggestion
		return output, errorsmod.Wrapf(err, "query to wasm contract failed")
```

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -40,6 +34,26 @@ func (r contractResult) Error() string {
...
+	contractResult
+	GenesisMetadata []clienttypes.GenesisMetadata `json:"genesis_metadata"`
+}
+
+type timestampAtHeightQueryResponse struct {
+	contractResult
+	Timestamp uint64 `json:"timestamp"`
+}
+
+type CheckForMisbehaviourExecuteResult struct {
```
DimitrisJim, line 52: can be unexported from what I can tell

```suggestion
type checkForMisbehaviourExecuteResult struct {
```

modules/light-clients/08-wasm/types/client_state.go
```diff
@@ -96,18 +87,34 @@ func (cs ClientState) ZeroCustomFields() exported.ClientState {
...
 }
 
+type (
+	timestampAtHeightInnerPayload struct {
+		Height exported.Height `json:"height"`
+	}
+	timestampAtHeightPayload struct {
+		TimestampAtHeight timestampAtHeightInnerPayload `json:"timestamp_at_height"`
+	}
+)
```
damiannolan, line 97: this is also superseded by #4133, right?

I'd like to not use `type` blocks, e.g.:

```go
type (
    someType struct {}

    someOtherType struct {}
)
```

instead:

```go
type someType struct {}

type someOtherType struct {}
```

modules/light-clients/08-wasm/types/client_state.go
```diff
@@ -169,9 +176,9 @@ func (cs ClientState) VerifyMembership(
 		return errorsmod.Wrapf(ibcerrors.ErrInvalidType, "expected %T, got %T", commitmenttypes.MerklePath{}, path)
 	}
 
-	_, err := GetConsensusState(clientStore, cdc, height)
-	if err != nil {
-		return errorsmod.Wrap(err, "please ensure the proof was constructed against a height that exists on the client")
+	found := hasConsensusState(clientStore, height)
+	if !found {
```
damiannolan, line 180: can we do?

```go
if !hasConsensusState()
```

modules/light-clients/08-wasm/types/client_state.go
```diff
@@ -226,9 +233,9 @@ func (cs ClientState) VerifyNonMembership(
...
```
damiannolan, line 237: ditto

modules/light-clients/08-wasm/types/client_state.go
```diff
@@ -249,20 +256,40 @@ func call[T ContractResult](ctx sdk.Context, clientStore sdk.KVStore, cs *Client
...
 	}
 	if len(out.Messages) > 0 {
 		return output, errorsmod.Wrapf(ErrWasmSubMessagesNotAllowed, "code hash (%s)", hex.EncodeToString(cs.CodeHash))
 	}
 	return output, nil
 }
+
+// wasmQuery queries the contract with the given payload and writes the result to output.
+func wasmQuery[T ContractResult](ctx sdk.Context, clientStore sdk.KVStore, cs *ClientState, payload any) (T, error) {
+	var output T
```
damiannolan, line 279: is this `output`, `result`, or `response`?

I see these three different terms being used interchangeably in a lot of places, we be good to standardise on one. 
I'm happy enough with `result` / `ContractResult`, `response` or `ContractResponse` is also fine imo! :)

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -40,6 +34,26 @@ func (r contractResult) Error() string {
...
+	contractResult
+	GenesisMetadata []clienttypes.GenesisMetadata `json:"genesis_metadata"`
+}
+
+type timestampAtHeightQueryResponse struct {
+	contractResult
+	Timestamp uint64 `json:"timestamp"`
+}
+
+type checkForMisbehaviourExecuteResult struct {
```
damiannolan, line 52: this should be `checkForMisbehaviourReponse` or? how come the different naming?

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -40,6 +34,26 @@ func (r contractResult) Error() string {
...
+
+type metadataQueryResponse struct {
+	contractResult
+	GenesisMetadata []clienttypes.GenesisMetadata `json:"genesis_metadata"`
+}
+
+type timestampAtHeightQueryResponse struct {
+	contractResult
+	Timestamp uint64 `json:"timestamp"`
+}
```
damiannolan, line 50: In #4133 all the contract payload types are in msgs.go right? Should these be included alongside them? 

Even tho they are unexported, would be nice to have basic godocs.

modules/light-clients/08-wasm/types/store.go
```diff
@@ -110,25 +108,9 @@ func setConsensusState(clientStore sdk.KVStore, cdc codec.BinaryCodec, consensus
...
-
-	consensusState, ok := consensusStateI.(*ConsensusState)
-	if !ok {
-		return nil, errorsmod.Wrapf(clienttypes.ErrInvalidConsensus, "invalid consensus type. expected %T, got %T", &ConsensusState{}, consensusState)
-	}
-
-	return consensusState, nil
+// hasConsensusState returns true if a consensus state at the given height exists in the store.
+func hasConsensusState(clientStore sdk.KVStore, height exported.Height) bool {
+	return clientStore.Has(host.ConsensusStateKey(height))
```
damiannolan, line 113: mega nit: this function seems small enough to possibly warrant using `clientStore.Has(host.ConsensusStateKey(height))` directly in place, but happy to keep it around if people want it.

damiannolan (approved):
Leaving my approval, left some suggestions 👍

modules/light-clients/08-wasm/types/store.go
DimitrisJim: Ah, this should just be removed as we talked about in the call yesterday, isn't correct to assume light-client stores consensus states at specific heights.

modules/light-clients/08-wasm/types/client_state.go
crodriguezvega: Just for my own understanding: what's wrong with `type` blocks?

damiannolan: @chatton doesn't like them! 😆 

I think I have similar feelings!

modules/light-clients/08-wasm/types/client_state.go
crodriguezvega: This was removed.

modules/light-clients/08-wasm/types/client_state.go
crodriguezvega: Did some renaming

modules/light-clients/08-wasm/types/vm.go
crodriguezvega: Renamed as well.

modules/light-clients/08-wasm/types/vm.go
crodriguezvega: I will add them in #4133.

modules/light-clients/08-wasm/types/client_state.go
crodriguezvega: Well, I don't want to get @chatton angry, so...

PR #4052, opened 2023-07-10
Pass in counterparty portID, channelID when verifying channel in ChanUpgradeOpen.

DimitrisJim:
## Description

Fixes a bug raised by Luca Joss in the shared ibc-go-rs channel.

Rename to `proofCounterpartyChannel` to align with the rest.

closes: #XXXX

```bash
fix: pass counterparty portid, channelid.
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -319,7 +319,7 @@ func (k Keeper) ChanUpgradeOpen(
 	portID,
 	channelID string,
 	counterpartyChannelState types.State,
-	proofChannel []byte,
+	proofCounterpartyChannel []byte,
 	proofHeight clienttypes.Height,
 ) error {
 	if k.hasInflightPackets(ctx, portID, channelID) {
@@ -397,7 +397,14 @@ func (k Keeper) ChanUpgradeOpen(
 		panic(fmt.Sprintf("counterparty channel state should be in one of [%s, %s, %s]; got %s", types.TRYUPGRADE, types.ACKUPGRADE, types.OPEN, counterpartyChannelState))
 	}
 
-	err = k.connectionKeeper.VerifyChannelState(ctx, connection, proofHeight, proofChannel, portID, channelID, counterpartyChannel)
+	err = k.connectionKeeper.VerifyChannelState(
+		ctx,
+		connection,
+		proofHeight, proofCounterpartyChannel,
+		channel.Counterparty.PortId,
+		channel.Counterparty.ChannelId,
+		counterpartyChannel,
+	)
 	if err != nil {
 		return errorsmod.Wrapf(err, "failed to verify counterparty channel, expected counterparty channel state: %s", counterpartyChannel.String())
 	}
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -397,15 +397,14 @@ func (k Keeper) ChanUpgradeOpen(
 		panic(fmt.Sprintf("counterparty channel state should be in one of [%s, %s, %s]; got %s", types.TRYUPGRADE, types.ACKUPGRADE, types.OPEN, counterpartyChannelState))
 	}
 
-	err = k.connectionKeeper.VerifyChannelState(
+	if err = k.connectionKeeper.VerifyChannelState(
 		ctx,
 		connection,
 		proofHeight, proofCounterpartyChannel,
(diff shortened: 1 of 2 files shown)


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -319,7 +319,7 @@ func (k Keeper) ChanUpgradeOpen(
 	portID,
 	channelID string,
 	counterpartyChannelState types.State,
-	proofChannel []byte,
+	proofCounterpartyChannel []byte,
```
damiannolan, line 322: I think I'd prefer to keep this as `proofChannel` as we are always verify proofs of a counterparty (its implicit)

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -397,7 +397,14 @@ func (k Keeper) ChanUpgradeOpen(
...
+		ctx,
+		connection,
+		proofHeight, proofCounterpartyChannel,
+		channel.Counterparty.PortId,
+		channel.Counterparty.ChannelId,
+		counterpartyChannel,
+	)
 	if err != nil {
 		return errorsmod.Wrapf(err, "failed to verify counterparty channel, expected counterparty channel state: %s", counterpartyChannel.String())
 	}
```
damiannolan, line 410: nit: should we consolidate the error check to within the scope of the func call like it is in `startFlushUpgrade`

```go
if err := k.connectionKeeper.VerifyChannelState(...); err != nil {
    return errorsmod.Wrapf(err, "...")
}
```

damiannolan (approved):
Nice one! Just left nits, but thanks for the fix! 🚀

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: ah yea, definitely.

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: I do think `proofChannel` caused my initial mis-step on this. Other handles also prefer explicit. Change later all together if we decide so?

DimitrisJim: think there's some areas where different approaches have been taken (e.g also how we do connection look-up) deviating from the old handlers, definitely would prefer to hammer those differences out and be consistent. Opened https://github.com/cosmos/ibc-go/issues/4053 to keep track of it and not forget

colin-axner (approved):
Nice catch!

colin-axner:
We should add a test case that would catch this?

DimitrisJim:
> We should add a test case that would catch this?

I tried. This issue was a result of both chains having the same values for port/channel. It's tricky since we create the mock IBC app using a portID of `mock` and channel ids by default result in `channel-0` for both. Unsure if I'm missing some other option here.

crodriguezvega (approved):
Thank you for the quick fix, @DimitrisJim!

colin-axner:
> I tried. This issue was a result of both chains having the same values for port/channel. It's tricky since we create the mock IBC app using a portID of mock and channel ids by default result in channel-0 for both. Unsure if I'm missing some other option here.

One quick hack is to make a second path and call `endpoint.ChanOpenInit` on one side. This will generate channel-0 on that end. Then when you create the path for testing, one side will have `channel-0` and the other will have `channel-1`

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -712,6 +712,14 @@ func (suite *KeeperTestSuite) TestChanUpgradeOpen() {
 		suite.Run(tc.name, func() {
 			suite.SetupTest()
 
+			// Create an initial path used only to invoke a ChanOpenInit handshake.
+			// This bumps the channel identifier generated for chain A on the
+			// next path used to run the upgrade handshake.
+			// See issue 4062.
```
DimitrisJim, line 718: linking #4062

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -707,6 +707,14 @@ func (suite *KeeperTestSuite) TestChanUpgradeOpen() {
 		},
 	}
 
+	// Create an initial path used only to invoke a ChanOpenInit handshake.
+	// This bumps the channel identifier generated for chain A on the
+	// next path used to run the upgrade handshake.
+	// See issue 4062.
```
DimitrisJim, line 713: linking #4062

PR #4059, opened 2023-07-11
fix: update e2e test config to use simapp codec

charleenfei:
## Description

we currently do not set a codec in our ChainConfig for `interchaintest`, which causes problems when we try to run the e2e tests in PRs containing the `08-wasm`. this pr updates our `ChainConfig` to use the codec from simapp.

---

write a little note why.

### DIFF
--- a/e2e/testsuite/diagnostics/diagnostics.go
+++ b/e2e/testsuite/diagnostics/diagnostics.go
@@ -11,10 +11,8 @@ import (
 
 	dockertypes "github.com/docker/docker/api/types"
 	dockerclient "github.com/docker/docker/client"
-	"github.com/strangelove-ventures/interchaintest/v7/ibc"
 
 	"github.com/cosmos/ibc-go/e2e/dockerutil"
-	"github.com/cosmos/ibc-go/e2e/testconfig"
 )
 
 const (
@@ -25,11 +23,10 @@ const (
 
 // Collect can be used in `t.Cleanup` and will copy all the of the container logs and relevant files
 // into e2e/<test-suite>/<test-name>.log. These log files will be uploaded to GH upon test failure.
-func Collect(t *testing.T, dc *dockerclient.Client, cfg testconfig.ChainOptions) {
+func Collect(t *testing.T, dc *dockerclient.Client, chainNameA, chainNameB string, debugModeEnabled bool) {
 	t.Helper()
 
-	debugCfg := testconfig.LoadConfig().DebugConfig
-	if !debugCfg.DumpLogs {
+	if !debugModeEnabled {
 		// when we are not forcing log collection, we only upload upon test failing.
 		if !t.Failed() {
 			t.Logf("test passed, not uploading logs")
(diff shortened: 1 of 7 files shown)


e2e/testsuite/diagnostics/diagnostics.go
```diff
@@ -25,11 +23,10 @@ const (
 
 // Collect can be used in `t.Cleanup` and will copy all the of the container logs and relevant files
 // into e2e/<test-suite>/<test-name>.log. These log files will be uploaded to GH upon test failure.
-func Collect(t *testing.T, dc *dockerclient.Client, cfg testconfig.ChainOptions) {
+func Collect(t *testing.T, dc *dockerclient.Client, chainNameA, chainNameB string, debugModeEnabled bool) {
```
chatton, line 26: small nit, I know we already make the assumption that we have 2 chains in 99% of our tests, but might be nice to do

```golang
func Collect(t *testing.T, dc *dockerclient.Client, debugModeEnabled bool, chainNames ...string) {}
```

to allow to support a variable number of chains.

chatton (approved):
LGTM! I think the difference between testsuite and testconfig has become kind of arbitrary anyway, testconfig.go makes sense and just a file in the testsuite package!

Left one small suggestion but LGTM

Thanks for making a small PR for this change! 🥇

e2e/testsuite/testsuite.go
```diff
@@ -47,22 +46,22 @@ type E2ETestSuite struct {
 	pathNameIndex int64
 }
 
-// path is a pairing of two chains which will be used in a test.
-type path struct {
+// pathPair is a pairing of two chains which will be used in a test.
+type pathPair struct {
```
colin-axner, line 50: nit: this change feels a bit arbitrary, out of scope, if the e2e test suite struct still has `paths` as a field. Generally would like unrelated changes to be proposed in separate prs

e2e/testsuite/diagnostics/diagnostics.go
colin-axner: What's the reasoning for the API change? Seems like it might be useful for `Collect` to be able to reference the `DebugConfig`

colin-axner (approved):
LGTM! See comments/questions

e2e/testsuite/diagnostics/diagnostics.go
charleenfei: we had to change this also due to circular import issues: otherwise the `diagnostics` package would have to import this `testsuite` package (previously `testconfig` package`), which would then lead to circular import when `testsuite` imports this function

e2e/testsuite/testsuite.go
colin-axner: I see. That makes sense. Would be good to add justifications on the pr for reviewers to understand the reasoning behind changes

e2e/testsuite/diagnostics/diagnostics.go
colin-axner: I believe the reasoning here is also to avoid a circular dependency

e2e/testsuite/testsuite.go
charleenfei: i will definitely do in future, thanks <3

colin-axner:
Just curious, was there a reason to move the testconfig out of its own package? I agree it was a bit arbitrary, but it seems useful for the testconfig struct to be importable by packages that testsuite imports as well

DimitrisJim (approved):
:partying_face:

PR #4066, opened 2023-07-11
Fix Dockerfile build issues.

DimitrisJim:
## Description

Mirrors the Dockerfile currently used for `wasmd`. Based on `alpine`, grabs the static library for libwasm and builds ibc-go with it. Ideally most of these can be plucked and used in `main` beforehand but opening this first.

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
-FROM golang:1.20 as builder
-
+FROM golang:1.20-alpine3.18 as builder
 ARG IBC_GO_VERSION
+
+RUN set -eux; apk add --no-cache git libusb-dev linux-headers gcc musl-dev make;
 
 ENV GOPATH=""
 ENV GOMODULE="on"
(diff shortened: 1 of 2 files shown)


DimitrisJim:
ideally #4067 goes first to main, then can rebase this once it is included in base branch.

Dockerfile
```diff
@@ -21,11 +25,14 @@ ADD LICENSE LICENSE
...
 COPY Makefile .
 
+COPY go.mod .
+COPY go.sum .
 
-RUN make build
+RUN go mod download
 
-FROM ubuntu:22.04
+RUN BUILD_TAGS=muslc LINK_STATICALLY=true make build
```
DimitrisJim, line 33: Some more context. Statically linking is not necessary, we could also do dynamic linking here and just grab the `.so` libs from wasmvm's `internal/api` folder (this is how I did it in the now closed PR). That is generally a bit cleaner since we don't need to reference the specific version of `libwasmvm` to download (i.e, we download `1.2.1` here to match what we have in our `go.mod`).

The downside is that, since we use multi-stage builds, any shared libraries we reference in the first stage of the build (on the `builder` image) must be copied over. We could also have a similar situations with other shared libs in the future. (I.e we'd need an additional `COPY --from=builder shared/lib/path` to reference them).

I don't particularly think it makes sense to use shared libraries in a docker image that won't have many components actually sharing the libraries but I don't have that strong of an opinion on it.

crodriguezvega (approved):
🥇

Issue #4055, opened 2023-07-10
Remove dependency on ibc-go from capability module


damiannolan:
## Summary

Currently the capability module maintained under `modules/capability` depends on ibc-go in order to make use of simapp under `testing/simapp`. 

The capability module does NOT make use of any ibctesting features such as:
- `TestChain`
- `Path`
- `Endpoint`
- `Coordinator`

The capability module uses the ibc-go dependency in testing ONLY. It uses `simapp` for instantiating a new test app and making use of `baseapp` as well as key functionality such as `GetKey` and `GetMemKey` within test funcs.

## Problem Definition

While diving into #3978 I noticed how problematic this becomes when performing dependency updates, particularly with the sdk.
The PR in progress currently relies on a fork containing sdk v0.50 code in ibc-go in order to service the capability module's testing.

Without the fork we run into compiler errors, for example, such as store types being now imported from `cosmossdk.io/store` rather than the previously used `github.com/cosmos/cosmos-sdk/store/types`.

This becomes a real headache when trying to iterate quickly and upgrade go modules in isolation.

## Proposal

Add a lightweight testing app to `modules/capability` in order to provide testing functionality. 
It should be desirable to keep the testing app as barebones as possible - even importing and bootstrapping the app with baseapp and `x/capability` if at all possible. 

This will alleviate issues with the Eden upgrade in progress and allow us to iterate quickly and in isolation without having to rely on ibc-go or any other simulation app as a dependency.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
See https://github.com/cosmos/ibc-go/pull/4057#issuecomment-1631983864

I've opened #4068 to address the upgrade and remove the ibc-go dependency, while not adding any simapp to the capability module.

PR #4068, opened 2023-07-12
deps: upgrade capability to sdk v0.50 and remove ibc-go dependency

damiannolan:
## Description

This PR upgrades the capability module to sdk v0.50 eden release. 
It removes the dependency on ibc-go `testing/simapp` and refactors the tests of interest to not depend on a simulation app.
Instead it uses a test a context and in-memory db in order to accomplish the same level of testing provided by simapp.

Note, I have tried to clean up and add inline code comments to `capability_test.go` to make it easier to understand.

closes: #4055 

---

write a little note why.


modules/capability/genesis_test.go
```diff
@@ -1,21 +1,18 @@
...
-	"github.com/cosmos/ibc-go/v7/testing/simapp"
 
 	"github.com/cosmos/ibc-go/modules/capability"
 	"github.com/cosmos/ibc-go/modules/capability/keeper"
 	"github.com/cosmos/ibc-go/modules/capability/types"
 )
 
 func (suite *CapabilityTestSuite) TestGenesis() {
+	// InitGenesis must be called in order to set the intial index to 1.
+	capability.InitGenesis(suite.ctx, *suite.keeper, *types.DefaultGenesis())
```
damiannolan, line 14: `InitGenesis` has to be called directly here in order to set the initial capability index to 1. 

This used to be happening handled implicitly when testing with a simapp. This caught me off guard for a while as after removing simapp this test would fail below when trying to lookup the first capability after genesis is reinitialised. 

The reasoning for this is that without setting the initial index to 1 - `cap1` created on line 19 would be at index 0, and `ExportGenesis` starts the loop index [here](https://github.com/cosmos/ibc-go/blob/main/modules/capability/genesis.go#L31) at 1. So after exporting and then reinitialising the gen state we'd lose first capabiliity.

modules/capability/capability_test.go
```diff
@@ -3,101 +3,119 @@ package capability_test
...
-
-	// create new keeper so we can define custom scoping before init and seal
-	keeper := keeper.NewKeeper(cdc, app.GetKey(types.StoreKey), app.GetMemKey(types.MemStoreKey))
-
-	suite.app = app
-	suite.ctx = app.BaseApp.NewContext(checkTx, tmproto.Header{Height: 1})
-	suite.keeper = keeper
-	suite.cdc = cdc
-	suite.module = capability.NewAppModule(cdc, *keeper, false)
+	encodingCfg := moduletestutil.MakeTestEncodingConfig(capability.AppModuleBasic{})
```
damiannolan, line 42: Pretty nice way to create a test encoding config 👍

modules/capability/capability_test.go
```diff
@@ -3,101 +3,119 @@ package capability_test
...
+	cms := store.NewCommitMultiStore(db, log.NewNopLogger(), metrics.NewNoOpMetrics())
+	cms.MountStoreWithDB(suite.storeKey, storetypes.StoreTypeIAVL, db)
+	cms.MountStoreWithDB(suite.memStoreKey, storetypes.StoreTypeMemory, db)
+	cms.MountStoreWithDB(suite.mockMemStoreKey, storetypes.StoreTypeMemory, db)
+
+	err := cms.LoadLatestVersion()
+	suite.Require().NoError(err)
+
+	return sdk.NewContext(cms, tmproto.Header{}, false, log.NewNopLogger())
 }
```
damiannolan, line 64: This has been adapted from [here](https://github.com/cosmos/cosmos-sdk/blob/v0.50.0-alpha.1/testutil/context.go).

I'm unsure why the test helpers in the sdk do not support adding mem store keys, or even mounting multiple keys (I guess nobody has run into an issue with the current code yet). 

I was going to make a PR to the sdk to add some extra support for this, but I felt it was cleaner and quicker to adapt it and maintain it here.

modules/capability/capability_test.go
```diff
@@ -3,101 +3,119 @@ package capability_test
...
 
 // The following test case mocks a specific bug discovered in https://github.com/cosmos/cosmos-sdk/issues/9800
 // and ensures that the current code successfully fixes the issue.
+// This test emulates statesync by firstly populating persisted state by creating a new scoped keeper and capability.
+// In-memory storage is then discarded by creating a new capability keeper and app module using a mock memstore key.
+// BeginBlock is then called to populate the new in-memory store using the persisted state.
 func (suite *CapabilityTestSuite) TestInitializeMemStore() {
```
damiannolan, line 71: I spent a good bit of time trying to parse this testcase. So I've added a bunch of in-line comments as well as extra info to the godoc above. This was initially added by @AdityaSripal in https://github.com/cosmos/cosmos-sdk/pull/9835 but has since seen a whole bunch of devs chopping into the code. 

Hopefully its now clear what its doing and is easier to maintain.

modules/capability/capability_test.go
```diff
@@ -3,101 +3,119 @@ package capability_test
...
+	suite.Require().False(newKeeper.IsInitialized(suite.ctx), "memstore initialized flag set before BeginBlock")
 
-	// Mock app beginblock and ensure that no gas has been consumed and memstore is initialized
-	ctx = suite.app.BaseApp.NewContext(false, tmproto.Header{}).WithBlockGasMeter(sdk.NewGasMeter(50))
+	// add a new block gas meter to the context
+	ctx := suite.ctx.WithBlockGasMeter(storetypes.NewGasMeter(50))
 
+	prevGas := ctx.GasMeter().GasConsumed()
 	prevBlockGas := ctx.BlockGasMeter().GasConsumed()
-	prevGas := ctx.BlockGasMeter().GasConsumed()
```
damiannolan, line 69: `prevGas` here was incorrect, it was being set to the same value as `prevBlockGas`

damiannolan (commented):
Added some context for review. More testcases could be cleaned up but I left as is to reduce diffs. The main one of concern was `TestInitializeMemstore` which I've cleaned up

crodriguezvega:
Thank you so much for handling the upgrade of the capability module, @damiannolan. This PR will supersede #3978, right? So that one can be closed?

damiannolan:
> Thank you so much for handling the upgrade of the capability module, @damiannolan. This PR will supersede #3978, right? So that one can be closed?

Yep, indeed we can close #3978

chatton (approved):
LGTM nice investigation ❤️

modules/capability/capability_test.go
colin-axner: Thank you for this!

modules/capability/capability_test.go
```diff
@@ -3,101 +3,119 @@ package capability_test
...
-
-	// Ensure that the second transaction can still receive capability even if first tx fails.
-	ctx = suite.app.BaseApp.NewContext(false, tmproto.Header{})
+	// assert that the in-memory store is now initialized
+	suite.Require().True(newKeeper.IsInitialized(ctx), "memstore initialized flag not set")
 
-	cap1, ok = newSk1.GetCapability(ctx, "transfer")
+	// ensure that BeginBlock has populated the new in-memory store (using the mock memstore key) and initialized capabilities
+	cap1, ok := scopedKeeper.GetCapability(ctx, "transfer")
 	suite.Require().True(ok)
```
colin-axner, line 112: Is there a place in the code we can assert this would have returned false before calling begin block?

colin-axner (approved):
Damn, excellent work! This is a pretty epic pr if I have ever seen one :fire:

modules/capability/capability_test.go
damiannolan: Aye! Good call, added above :)

damiannolan:
cc @faddat 👀 

should be able to pick this up in #3883 using a `replace` directive with a local pin if you sync the latest from main, will also tag this soon and then we can move to that!

modules/capability/capability_test.go
damiannolan: Opened https://github.com/cosmos/cosmos-sdk/issues/16998

PR #4097, opened 2023-07-15
nit: error formatting improvements for channel upgrades

crodriguezvega:
## Description

Some small improvements to error strings.

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -667,7 +667,7 @@ func (k Keeper) startFlushUpgradeHandshake(
 
 	connection, err := k.GetConnection(ctx, channel.ConnectionHops[0])
 	if err != nil {
-		errorsmod.Wrapf(err, "failed to retrieve connection using the channel connection hops (%s)", channel.ConnectionHops[0])
+		return errorsmod.Wrapf(err, "failed to retrieve connection using the channel connection hops (%s)", channel.ConnectionHops[0])
 	}
 
 	if connection.GetState() != int32(connectiontypes.OPEN) {
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -56,7 +56,7 @@ func (k Keeper) WriteUpgradeInitChannel(ctx sdk.Context, portID, channelID strin
 
 	currentChannel, found := k.GetChannel(ctx, portID, channelID)
 	if !found {
-		panic(fmt.Sprintf("could not find existing channel when updating channel state in successful ChanUpgradeInit step, channelID: %s, portID: %s", channelID, portID))
+		panic(fmt.Sprintf("could not find existing channel when updating channel state in successful ChanUpgradeInit step,  portID: %s, channelID: %s", portID, channelID))
 	}
 
 	currentChannel.State = types.INITUPGRADE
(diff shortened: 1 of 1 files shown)


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -95,7 +95,7 @@ func (k Keeper) ChanUpgradeTry(
 
 	connection, err := k.GetConnection(ctx, channel.ConnectionHops[0])
 	if err != nil {
-		return types.Upgrade{}, errorsmod.Wrap(err, "failed to retrieve connection using the channel connection hops")
+		return types.Upgrade{}, errorsmod.Wrapf(err, "failed to retrieve connection using the channel connection hops (%s)", channel.ConnectionHops[0])
```
DimitrisJim, line 98: I think the idea here was that the `err` already contains the connection identifier.

crodriguezvega: Ah, yeah, true, because we're using the interface function. I will revert this one.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -760,7 +760,7 @@ func (k Keeper) validateSelfUpgradeFields(ctx sdk.Context, proposedUpgrade types
 	currentFields := extractUpgradeFields(currentChannel)
 
 	if reflect.DeepEqual(proposedUpgrade, currentFields) {
-		return errorsmod.Wrapf(types.ErrChannelExists, "existing channel end is identical to proposed upgrade channel end: got %s", proposedUpgrade)
+		return errorsmod.Wrapf(types.ErrChannelExists, "existing channel end (%s) is identical to proposed upgrade channel end (%s)", currentFields, proposedUpgrade)
```
chatton, line 763: this string makes it sound like we are displaying the channel end as opposed to the upgrade fields. 

Also due to the nature of the check, both values will always be the same, I don't think we need to display it twice.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -731,7 +731,7 @@ func (k Keeper) checkForUpgradeCompatibility(ctx sdk.Context, proposedUpgradeFie
 	if err != nil {
 		// NOTE: this error is expected to be unreachable as the proposed upgrade connectionID should have been
 		// validated in the upgrade INIT and TRY handlers
-		return errorsmod.Wrap(err, "expected proposed connection to be found")
```
charleenfei, line 734: i kind of like this error message better tbh because we should actually never reach this error case.

crodriguezvega: I changed to make the error more similar to the one that is returned in other places in `upgrade.go`, like [this one](https://github.com/cosmos/ibc-go/blob/f731de5d9c5f6668049ca9293b1fda46915f0a98/modules/core/04-channel/keeper/upgrade.go#L98). But I can change it back, if you want.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -56,7 +56,7 @@ func (k Keeper) WriteUpgradeInitChannel(ctx sdk.Context, portID, channelID strin
 
 	currentChannel, found := k.GetChannel(ctx, portID, channelID)
 	if !found {
-		panic(fmt.Sprintf("could not find existing channel when updating channel state in successful ChanUpgradeInit step, channelID: %s, portID: %s", channelID, portID))
+		panic(fmt.Sprintf("could not find existing channel when updating channel state in successful ChanUpgradeInit step,  portID: %s, channelID: %s", portID, channelID))
```
damiannolan, line 59: the Write funcs in `handhshake.go` also put channel ID first

DimitrisJim (approved):

PR #4109, opened 2023-07-17
add 08-wasm Keeper constructor that accepts pointer to Wasm VM instance

crodriguezvega:
## Description

Item 6.9 of Confio's report of 08-wasm audit.

I started writing the issue and sketching some code, but I decided to implement it directly. :)

closes: #XXXX

```bash
type: commit message
```

---

write a little note why.


testing/simapp/app.go
```diff
@@ -465,7 +466,31 @@ func NewSimApp(
...
+	// 2. If the chain does not use x/wasm:
+	// Even though it is still possible to use method 1 above
+	// (e.g. instantiating a Wasm VM in app.go an pass it in 08-wasm NewKeeper),
+	// since there is no need to share the Wasm VM instance with another module
+	// you can use NewKeeperWithConfig constructor function and provide
+	// the Wasm VM configuration parameters of your choice.
+	// Check out the WasmConfig type definition for more information on
+	// each parameter. Some parameters allow node-leve configurations.
+	//
+	// In the code below we use the second method because we are not using x/wasm in this app.go.
```
crodriguezvega, line 486: I plan to add similar documentation to the official docs, but since we all know that developers don't read documentation, at least, hopefully, they will see this when the copy our app.go. :)

modules/light-clients/08-wasm/types/config.go
```diff
@@ -0,0 +1,20 @@
+package types
+
+// contractMemoryLimit is the memory limit of each contract execution (in MiB)
+// constant value so all nodes run with the same limit.
+const ContractMemoryLimit = 32
+
+type WasmConfig struct {
```
crodriguezvega, line 7: I added this type to pass to the Keeper constructor the parameters to instantiate the VM instead of passing each parameter individually. I thought that in case in the future we add more parameters, then we can extend this type, but do not break the API. I can add a `NewWasmConfig` function if deemed necessary. There is a [similar type used in `x/wasm`](https://github.com/CosmWasm/wasmd/blob/e5049ba686ab71164a01f6e71e54347710a1f740/x/wasm/types/types.go#L316), so I will move this to folder (in a follow-up PR), as Ethan suggested, to keep all the copied code in one place.

modules/light-clients/08-wasm/types/config.go
```diff
@@ -0,0 +1,20 @@
+package types
+
+// contractMemoryLimit is the memory limit of each contract execution (in MiB)
+// constant value so all nodes run with the same limit.
+const ContractMemoryLimit = 32
```
crodriguezvega, line 5: [Following here also `x/wasm`](https://github.com/CosmWasm/wasmd/blob/e5049ba686ab71164a01f6e71e54347710a1f740/x/wasm/keeper/keeper.go#L34).

testing/simapp/app.go
```diff
@@ -465,7 +466,31 @@ func NewSimApp(
...
+	wasmConfig := wasmtypes.WasmConfig{
+		DataDir:           "ibc_08-wasm_client_data",
+		SupportedFeatures: "iterator",
```
DimitrisJim, line 489: haven't had time to look into it atm but leaving as a Q. How important are these features? What if a chain that uses `x/wasm` has a keeper set up w/o this supported feature?

crodriguezvega: It's a good question, @DimitrisJim. I also haven't looked extensively into this, but I have seen that the VM in `x/wasm` is configured using all available capabilities (see [here](https://github.com/CosmWasm/wasmd/blob/5d05b85875ee4f6befc69c843f6915df5bd68ab9/app/app.go#L586) and [here](https://github.com/CosmWasm/wasmd/blob/main/app/wasm.go#L6)). Looks like 08-wasm only needs the `iterator` (at least that's the one that was used in the code we got from Strangelove; there was another one -`storage`-, that apparently does not exist, so I removed it).

modules/light-clients/08-wasm/types/config.go
DimitrisJim: would be nice to also get the `DefaultWasmConfig` that's used there.

testing/simapp/app.go
```diff
@@ -465,7 +466,31 @@ func NewSimApp(
...
+	// Even though it is still possible to use method 1 above
+	// (e.g. instantiating a Wasm VM in app.go an pass it in 08-wasm NewKeeper),
+	// since there is no need to share the Wasm VM instance with another module
+	// you can use NewKeeperWithConfig constructor function and provide
+	// the Wasm VM configuration parameters of your choice.
+	// Check out the WasmConfig type definition for more information on
+	// each parameter. Some parameters allow node-leve configurations.
+	//
+	// In the code below we use the second method because we are not using x/wasm in this app.go.
+	wasmConfig := wasmtypes.WasmConfig{
```
DimitrisJim, line 487: and use the default wasm config here.

testing/simapp/app.go
```diff
@@ -465,7 +466,31 @@ func NewSimApp(
 		),
 	)
 
-	app.WasmClientKeeper = wasmkeeper.NewKeeper(appCodec, keys[wasmtypes.StoreKey], authtypes.NewModuleAddress(govtypes.ModuleName).String())
+	// 08-wasm's Keeper can be instantiated in two different ways:
+	// 1. If the chain uses x/wasm:
+	// Both x/wasm's Keeper and 08-wasm Keeper should share the same Wasm VM instante.
+	// - Instantiate the Wasm VM in app.go with the parameters of your choice.
+	// - Create an Option with this Wasm VM instance (see https: //github.com/CosmWasm/wasmd/blob/db93d7b6c7bb6f4a340d74b96a02cec885729b59/x/wasm/keeper/options.go#L21-L25).
```
chatton, line 473: is there a tag we can link to instead?

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -29,19 +27,15 @@ type Keeper struct {
...
-	wasmSupportedFeatures := strings.Join([]string{"storage", "iterator"}, ",")
-	wasmMemoryLimitMb := uint32(math.Pow(2, 12))
-	wasmPrintDebug := true
-	wasmCacheSizeMb := uint32(math.Pow(2, 8))
-
-	vm, err := cosmwasm.NewVM(wasmDataDir, wasmSupportedFeatures, wasmMemoryLimitMb, wasmPrintDebug, wasmCacheSizeMb)
-	if err != nil {
-		panic(err)
-	}
+// NewKeeper creates a new NewKeeper instance with the provided Wasm VM.
```
chatton, line 30:
```suggestion
// NewKeeper creates a new Keeper instance with the provided Wasm VM.
```

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -29,19 +27,15 @@ type Keeper struct {
...
+// This constructor function is meant to be used when the chain uses x/wasm
+// and the same Wasm VM instance should be shared with it.
+func NewKeeper(
+	cdc codec.BinaryCodec,
+	key storetypes.StoreKey,
+	authority string,
+	vm *cosmwasm.VM,
+) Keeper {
 	types.WasmVM = vm
```
chatton, line 39: Because this function needs to be used in a very specific way, I think it might be beneficial to deviate from the standard "NewKeeper" naming convention. Maybe `NewKeeperWithSharedVM` to be extra explicit about it.

I think it would be nice to do a check for

```golang
if types.WasmVM != nil {
   panic("meaningful message")
}
```

This at least prevents accidental re-assignment of the global vm when this use case.

modules/light-clients/08-wasm/types/config.go
chatton: since this is in the `wasm/types` package, I think `Config` is a better name "wasmtypes.Config". However, I think we often refer to the import path as just `types` in which case WasmConfig is better. I will leave the choice to you. (my personal preference is use the `wasmtypes` alias, but am fine with either 👍 )

modules/light-clients/08-wasm/keeper/keeper.go
crodriguezvega: I could rename it to `NewKeeperWithVM`, since the VM could be or not shared.

Regarding the check, why wouldn't we want to re-assign the global VM? If the global VM is set, that means that either it was set in the keeper constructor function or outside of it, and this should not be standard behaviour, so wouldn't it be better to just re-assign?

modules/light-clients/08-wasm/types/config.go
crodriguezvega: I think I will leave it as is. :)

modules/light-clients/08-wasm/keeper/keeper.go
DimitrisJim: I'd like `NewKeeperWithVM`, with `Shared` sounds scarier than it should be in my mind (and would also make the call to it in `NewKeeperWithConfig` slightly confusing :smile:

> why wouldn't we want to re-assign the global VM?

I'm unsure of the implications of allowing re-assigning, it seems like it would mess all the internals of the VM up? Definitely sounds like something we should at least look into (i.e ask confio or look at what wasmd does, if anything)

chatton: My thinking is that presumably if it is creating and set outside of the function, it will be the expectation of the dev that the configure vm will be used by the module. This VM would potentially be ignored if a different one is passed in. While not likely to happen, I think it makes sense to explicitly fail if the VM is configured in a non standard way. This way the only way the VM can be configured is via this function when there is not a VM already configured.

My understanding of the doc string in this PR is that this function should only be used when a shared VM is what is required. I think this check would slightly reduce the surface area for possible issues.

Please correct me if my understanding is incorrect!

modules/light-clients/08-wasm/types/config.go
damiannolan: Nice, I like this config struct idea. We should add some godoc to the top of struct also! :)

modules/light-clients/08-wasm/types/config.go
```diff
@@ -0,0 +1,37 @@
+package types
+
+const (
+	// contractMemoryLimit is the memory limit of each contract execution (in MiB)
+	// constant value so all nodes run with the same limit.
+	ContractMemoryLimit = 32
+
+	defaultDataDir           string = "ibc_08-wasm_client_data"
```
damiannolan, line 8: This is the default data dir if 08-wasm is used standalone, correct?

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -52,6 +46,22 @@ func NewKeeper(cdc codec.BinaryCodec, key storetypes.StoreKey, authority string)
...
+	cdc codec.BinaryCodec,
+	key storetypes.StoreKey,
+	authority string,
+	wasmConfig types.WasmConfig,
+) Keeper {
+	vm, err := cosmwasm.NewVM(wasmConfig.DataDir, wasmConfig.SupportedFeatures, types.ContractMemoryLimit, wasmConfig.ContractDebugMode, wasmConfig.MemoryCacheSize)
+	if err != nil {
+		panic(err)
+	}
+	return NewKeeperWithVM(cdc, key, authority, vm)
```
damiannolan, line 62: nit: breathing room 🧘‍♂️ 

```suggestion
	}
	
	return NewKeeperWithVM(cdc, key, authority, vm)
```

testing/simapp/app.go
damiannolan: agree!

testing/simapp/app.go
```diff
@@ -465,7 +466,32 @@ func NewSimApp(
...
+	// Check out the WasmConfig type definition for more information on
+	// each parameter. Some parameters allow node-leve configurations.
+	// Function DefaultWasmConfig can also be used to use default values.
+	//
+	// In the code below we use the second method because we are not using x/wasm in this app.go.
+	wasmConfig := wasmtypes.WasmConfig{
+		DataDir:           "ibc_08-wasm_client_data",
+		SupportedFeatures: "iterator",
+		MemoryCacheSize:   uint32(math.Pow(2, 8)),
+		ContractDebugMode: true,
```
damiannolan, line 492: the default is false, right?

damiannolan (approved):
Nice work! I left a couple of small comments, nothing major! Would be nice to see how this would be wired up alongside `x/wasm` at some point! :)

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -52,6 +46,22 @@ func NewKeeper(cdc codec.BinaryCodec, key storetypes.StoreKey, authority string)
...
+// and a Wasm VM needs to be instantiated using the provided parameters.
+func NewKeeperWithConfig(
+	cdc codec.BinaryCodec,
+	key storetypes.StoreKey,
+	authority string,
+	wasmConfig types.WasmConfig,
+) Keeper {
+	vm, err := cosmwasm.NewVM(wasmConfig.DataDir, wasmConfig.SupportedFeatures, types.ContractMemoryLimit, wasmConfig.ContractDebugMode, wasmConfig.MemoryCacheSize)
+	if err != nil {
+		panic(err)
```
damiannolan, line 60: should we add more context to the panic trace?

maybe something like:
`fmt.Sprintf("failed to instantiate new cosmwasm VM instance: %v", err)`

modules/light-clients/08-wasm/types/config.go
crodriguezvega: When are we going to have chatGPT writing our godocs?

modules/light-clients/08-wasm/types/config.go
crodriguezvega: yeah.

testing/simapp/app.go
crodriguezvega: true!

modules/light-clients/08-wasm/keeper/keeper.go
DimitrisJim: Panicking mucks up our tests which create a bunch of test chains. I think this should be changed to only panic if the VM that we're trying to set (if set) is different than what we already have. I'll push that change and open an issue to keep track of this and discuss at later point.

Issue #4145, opened 2023-07-20
Use error assertion in ica host execution tests


colin-axner:
## Summary

Currently missing code coverage for a message failing validate basic, see [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/keeper/relay.go#L71). A test case should be added to ensure that check remains there

Edit: `ValidateBasic` is being called twice. Once by SDK and once by ibc-go, so adding a test case to fail in ibc-go isn't possible. Instead we can change the [`OnRecvPacket`](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/keeper/relay_test.go#L26) test to use error assertion, so if both checks are removed, the test fails error is assertion

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #4161, opened 2023-07-24
Change host relay tests to check error returned

DimitrisJim:
## Description

closes: #4145 

```bash
chore: change host relay tests to check error returned
```

---

write a little note why.


modules/apps/27-interchain-accounts/host/keeper/relay_test.go
```diff
@@ -330,6 +330,40 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
 			},
 			true,
 		},
+		{
+			"Msg fails its ValidateBasic: MsgTransfer has an empty receiver",
```
colin-axner, line 334: tested locally by removing `ValidateBasic`, this doesn't actually fix the issue, because I believe it will fail to execute. We need a message which fails validate basic but succeeds in execution

colin-axner: I wonder if we should first open an issue to convert this test to assert the error returned

DimitrisJim: gotcha, let me look into it.

DimitrisJim: ya looked into this a bit, we get the same error even if we convert to assert the error. `executeMsg` just goes through the normal Msg processing flow and will eventually also call `ValidateBasic` on the faulty message we create:

https://github.com/cosmos/cosmos-sdk/blob/095a64195aba917187bbb33570063bc675494bcd/baseapp/msg_service_router.go#L130-L134

colin-axner: But wouldn't the returned error change if the validate basic check was removed?

modules/apps/27-interchain-accounts/types/codec.go
```diff
@@ -84,7 +84,7 @@ func DeserializeCosmosTx(cdc codec.Codec, data []byte, encoding string) ([]sdk.M
 	switch encoding {
 	case EncodingProtobuf:
 		if err := cdc.Unmarshal(data, &cosmosTx); err != nil {
-			return nil, errorsmod.Wrapf(err, "cannot unmarshal CosmosTx with protobuf")
+			return nil, errorsmod.Wrapf(ErrUnknownDataType, "cannot unmarshal CosmosTx with protobuf")
```
DimitrisJim, line 87: I can't see a reason why this was returning the raw unmarshal error? Lmk if I'm missing something here.

modules/apps/27-interchain-accounts/host/keeper/relay_test.go
DimitrisJim: resolved in slack discussion: see parent issue edit for resolution.

DimitrisJim:
hm, I'm also going to change this for the JSON tests real quick.

modules/apps/27-interchain-accounts/host/keeper/relay_test.go
```diff
@@ -494,11 +529,12 @@ func (suite *KeeperTestSuite) TestOnRecvPacket() {
 
 				txResponse, err := suite.chainB.GetSimApp().ICAHostKeeper.OnRecvPacket(suite.chainB.GetContext(), packet)
 
-				if tc.expPass {
-					suite.Require().NoError(err)
+				expPass := tc.expErr == nil
+				if expPass {
+					suite.Require().Nil(err)
```
colin-axner, line 534: just curious, why the change from `NoError` to `Nil`?

modules/apps/27-interchain-accounts/types/codec.go
colin-axner: `ErrUnknownDataType` is somewhat confusing to me here. It failed to unmarshal which could be from malformed data

> I can't see a reason why this was returning the raw unmarshal error?

Probably just based on the pattern of, "if you see an error, wrap it". I'm okay using our own error as a base, but we should be emitting the unmarshal information in the err msg

modules/apps/27-interchain-accounts/host/keeper/relay_test.go
DimitrisJim: I'm pretty sure `tc.expErr == nil` moved me in that direction as I was reading it. Looking at others I do see `NoError` preferred so I can swap to that.

modules/apps/27-interchain-accounts/types/codec.go
DimitrisJim: yea, that sounds sensible to me. 

was also slightly confused but went with what was currently used for json unmarhalling, this error seems to be slightly overused :smiley: could add a new one in a follow up

colin-axner (approved):
Thanks @DimitrisJim!!

chatton (approved):
LGTM nice cleanup 👍

damiannolan (approved):
Good stuff! Thanks @DimitrisJim 🚀

charleenfei (approved):
thanks for the cleanup!

PR #4174, opened 2023-07-25
testing: e2e support for sdkv50

faddat:
This PR begins work to make the e2e tests support sdk 50, 
based on this PR to interchaintest:

https://github.com/strangelove-ventures/interchaintest/pull/674


e2e/testsuite/grpc_query.go
```diff
@@ -271,8 +271,8 @@ func (s *E2ETestSuite) QueryProposalV1(ctx context.Context, chain ibc.Chain, pro
 
 // GetBlockHeaderByHeight fetches the block header at a given height.
 func (s *E2ETestSuite) GetBlockHeaderByHeight(ctx context.Context, chain ibc.Chain, height uint64) (Header, error) {
-	cmtservice := s.GetChainGRCPClients(chain).ConsensusServiceClient
-	res, err := cmtservice.GetBlockByHeight(ctx, &ccmtservice.GetBlockByHeightRequest{
+	ccmtservice := s.GetChainGRCPClients(chain).ConsensusServiceClient
```
DimitrisJim, line 274: are the diffs here just for the rename? any reason why we do them?

DimitrisJim: ah, conflicts with the import for the client I'd guess `"github.com/cosmos/cosmos-sdk/client/grpc/cmtservice"` and linter is muy triste about it, leaving comment here so others don't ask same thing

modules/apps/callbacks/callbacks_test.go
```diff
@@ -290,26 +287,3 @@ func (s *CallbacksTestSuite) AssertHasExecutedExpectedCallbackWithFee(
 	}
 	s.AssertHasExecutedExpectedCallback(callbackType, isSuccessful)
 }
-
-// GetExpectedEvent returns the expected event for a callback.
-func GetExpectedEvent(
```
colin-axner, line 295: I think this should be added back, looks like maybe it got removed accidentally?

charleenfei: is it actually being used anywhere?

colin-axner: Yea, I will add back. The usage was deleted :)

Issue #4048, opened 2023-07-09
Add `WasmTendermintConfig` struct implementing `ClientConfig` interface


crodriguezvega:
_Thank you @damiannolan for this suggestion._

In order to be able to write tests with the Tendermint wasm light client, a change in the testing package was made to specify where the light client is a regular Tendermint light client or a Tendermint wasm light client. A [bool parameter was added to `NewTendermintConfig`](https://github.com/strangelove-ventures/ibc-go/blob/d46298f6a7e51651bb16074086d0811a0eb2dbaf/testing/config.go#L25) to indicate this.

As proposed by @damiannolan, instead of using a bool to add behaviour to an existing type, add a new type (as that is what the `ClientConfig` interface is intended for):

- Remove [`UseWasmClient` bool from `TestChain`](https://github.com/strangelove-ventures/ibc-go/blob/d46298f6a7e51651bb16074086d0811a0eb2dbaf/testing/chain.go#L83), and [setter `SetWasm(bool)`](https://github.com/strangelove-ventures/ibc-go/blob/d46298f6a7e51651bb16074086d0811a0eb2dbaf/testing/chain.go#L197).
- The [`CodeID` field being added to the `Coordinator` type](https://github.com/strangelove-ventures/ibc-go/blob/d46298f6a7e51651bb16074086d0811a0eb2dbaf/testing/coordinator.go#L29) can be moved to the `WasmTendermintConfig` type as a new field. Add a new `ClientConfig` implementation to satisfy the interface:

```golang
type WasmTendermintConfig struct {
	TendermintConfig
	CodeID []byte // or string??
}

func NewWasmTendermintConfig(codeID []byte) *WasmTendermintConfig {
	return &WasmTendermintConfig{
		CodeID:                 codeID,
		TendermintConfig: TendermintConfig{
			TrustLevel:      DefaultTrustLevel,
			TrustingPeriod:  TrustingPeriod,
			UnbondingPeriod: UnbondingPeriod,
			MaxClockDrift:   MaxClockDrift,
		},
	}
}

func (wasmcfg *WasmTendermintConfig) GetClientType() string {
	// return exported.Wasm
	return "08-wasm"
}
```

- Add a `WithClientConfig` method to the testing Endpoint:
 
```golang
// WithClientConfig sets the client config on the associated Endpoint and returns it.
func (endpoint *Endpoint) WithClientConfig(clientCfg ClientConfig) *Endpoint {
[...]

PR #4209, opened 2023-07-31
testing: add wasm tendermint client config

crodriguezvega:
## Description

closes: #4048

```bash
type: commit message
```

---

write a little note why.


testing/endpoint.go
```diff
@@ -117,7 +123,12 @@ func (endpoint *Endpoint) CreateClient() (err error) {
 		if err != nil {
 			return err
 		}
-		clientState = wasmtypes.NewClientState(wasmClientState, endpoint.Chain.Coordinator.CodeHash, height)
+
+		wasmTendermintConfig, ok := endpoint.ClientConfig.(*WasmTendermintConfig)
```
damiannolan, line 127: I think this fails before we reach this code. On lines 114-115:

```go
tmConfig, ok := endpoint.ClientConfig.(*TendermintConfig)
require.True(endpoint.Chain.TB, ok)
```

maybe we move this to the top of the swtich case?

testing/coordinator.go
```diff
@@ -47,11 +47,6 @@ func NewCoordinator(t *testing.T, n int) *Coordinator {
 	return coord
 }
 
-// SetCodeHash sets the code hash of a wasm ligh client contract
-func (coord *Coordinator) SetCodeHash(codeHash []byte) {
-	coord.CodeHash = codeHash
```
damiannolan, line 52: we can remove `CodeHash` from `Coordinator` as well :)

testing/config.go
```diff
@@ -19,26 +19,42 @@ type TendermintConfig struct {
...
+
+func NewWasmTendermintConfig(codeHash []byte) *WasmTendermintConfig {
+	return &WasmTendermintConfig{
+		CodeHash: codeHash,
+		TendermintConfig: TendermintConfig{
+			TrustLevel:      DefaultTrustLevel,
+			TrustingPeriod:  TrustingPeriod,
+			UnbondingPeriod: UnbondingPeriod,
+			MaxClockDrift:   MaxClockDrift,
+		},
```
damiannolan, line 50: maybe add a godoc saying that this is assuming the same values are used within the contract code? we could add extra godoc to the struct type definition with some basic info. 

Are we also able to use `NewTendermintConfig` or no? because it returns a pointer as opposed to value?

modules/light-clients/08-wasm/types/types_test.go
```diff
@@ -160,3 +157,15 @@ func TestWasmTestSuite(t *testing.T) {
 func getAltSigners(altVal *tmtypes.Validator, altPrivVal tmtypes.PrivValidator) map[string]tmtypes.PrivValidator {
 	return map[string]tmtypes.PrivValidator{altVal.Address.String(): altPrivVal}
 }
+
+// NewPathWithWasmTendermint constructs an endpoint for each chain using
+// the default values for the endpoints, expect for the client config, for
+// which it uses the Wasm Tendermint config, so that both endpoints use the
+// Tendermint Wasm light client contract.
+// Each endpoint is updated to have a pointer to the counterparty endpoint.
+func NewPathWithWasmTendermint(chainA, chainB *ibctesting.TestChain, codeHash []byte) *ibctesting.Path {
```
damiannolan, line 166: I don't feel too strongly, but we use `NewTransferPath` in the transfer test suite. We could use `NewWasmTendermintPath` here to align

damiannolan (commented):
Looks great I just left a couple of small comments/suggestions :)

modules/light-clients/08-wasm/types/types_test.go
colin-axner: If it's defined in a test file in 08-wasm. I think it's fine to simply call it `NewPath` as well

testing/endpoint.go
```diff
@@ -117,7 +123,12 @@ func (endpoint *Endpoint) CreateClient() (err error) {
...
+		if !ok {
+			return fmt.Errorf("invalid client config type, expected %T, got: %T", (*WasmTendermintConfig)(nil), endpoint.ClientConfig)
+		}
+		clientState = wasmtypes.NewClientState(wasmClientState, wasmTendermintConfig.CodeHash, height)
```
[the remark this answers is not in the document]
damiannolan: That's a really good point. We should really try to avoid that dependency issue.

testing/endpoint.go
DimitrisJim: Wasn't sure on ^ point too but yeah, it should fail [as per the spec](https://go.dev/ref/spec#Type_assertions):

> More precisely, if T is not an interface type, x.(T) asserts that the dynamic type of x is [identical](https://go.dev/ref/spec#Type_identity) to the type T

testing/config.go
DimitrisJim: you should still be able to deref it

testing/endpoint.go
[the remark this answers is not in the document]
DimitrisJim: isn't this reminiscent of the 02-client refactor? Couldn't ClientConfig (or some other interface) just have each specific client implement its separate client creation/updating logic for testing purposes? This might just be your suggestion and I'm just being overtly verbose with it.

modules/light-clients/08-wasm/types/types_test.go
chatton: It is specifically a wasm tendermint path though, it might be useful to have that clarification in the name (as opposed to some other wasm client implementation )

Maybe

```golang
path := wasm.NewTendermintPath(...)
```

testing/endpoint.go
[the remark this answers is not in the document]
chatton: Yeah I think the testing package should probably only have dependencies on things like `exported.ClientState` and `exported.ConsensusState`.

ideally additional client types should be able to be added without changes to the testing package.

It might be nice to have something like

```golang
type ClientCreator interface {
	CreateClientState() exported.ClientState
	CreateConsensusState() exported.ConsensusState
}

func (endpoint *Endpoint) CreateClient(cc ClientCreator) (err error) {
	// ...
	
	clientState := cc.CreateClientState(...)
	consensusState := cc.CreateConsensusState(...)

	msg, err := clienttypes.NewMsgCreateClient(
		clientState, consensusState, endpoint.Chain.SenderAccount.GetAddress().String(),
	)
	require.NoError(endpoint.Chain.TB, err)

	// ...
	
	return nil
}

```

I am a big fan of defining small interfaces near where they are used and passing them as arguments, but this could be achieved in a few different ways. The actual cc could be configured at endpoint creation or passed in like in the code above.

testing/coordinator.go
chatton: 🪓

testing/endpoint.go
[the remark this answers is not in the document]
chatton: will also add that I think this could happen in a follow up, I think the current changes are still an improvement but are also in line with how we treat other clients (e.g. solomachine).

Would be happy to see this as a follow up pr 👍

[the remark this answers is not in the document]
DimitrisJim: and I thought I was verbose :laughing:

crodriguezvega:
Not needed anymore after https://github.com/cosmos/ibc-go/pull/4801.

Issue #4127, opened 2023-07-19
Implement `transfer` channel upgrades callbacks


damiannolan:
## Summary

Implement the channel upgrade callback handlers for `transfer`.
Separate PRs can be opened for each callback handler including tests.

The implementation should ensure that the upgrade parameters are compatible with the transfer app.
This includes ordering enforced in `ValidateTransferChannelParams` as well as version checking, i.e. the upgrade version must be `ics20-1`.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
The `OnChanUpgradeAck` should also validate the order/version as init/try do

damiannolan:
Upgrade fields are validated in INIT, TRY and ACK. Now that fields are added to `OnChanUpgradeOpen` I don't see why we would need to do any extra validation on them. Also, the `OnChanUpgradeOpen` callback does not have an error return.

Do you think its fine to no-op on `OnChanUpgradeOpen` in transfer? @colin-axner

damiannolan:
Closing this as I think all handlers are more or less completed for channel upgradability in transfer. There is working integration test with fee middleware in #4533. 

We can open follow ups post-audit if needs be.

PR #4129, opened 2023-07-19
[wip]: adding transfer fee upgrade integration test

damiannolan:
## Description

Adding the scaffolding here in draft. Please do not review yet.

closes: #3993 

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#pull-request-targeting)).
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/11-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [commit message](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) to be used for the changelog entry in the PR description for review.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.
- [ ] Review `Codecov Report` in the comment section below once CI passes.

PR #4215, opened 2023-08-01
feat: implement `OnChanUpgradeAck` for transfer

damiannolan:
## Description

- Adding `OnChanUpgradeAck` implementation to transfer.

This should complete the below issue, I will add additional tests for no-op methods (restore, open) in a follow-up PR.

ref: #4127 

---

write a little note why.

### DIFF
--- a/modules/apps/transfer/ibc_module_test.go
+++ b/modules/apps/transfer/ibc_module_test.go
@@ -420,7 +420,7 @@ func (suite *TransferTestSuite) TestOnChanUpgradeAck() {
 		{
 			"invalid upgrade version",
 			func() {
-				// NOTE: counterpartyUpgrade is queried by Endpoint.ChanUpgradeTry() so retrieve and mutate fields here appropriately to force failure
+				// NOTE: counterpartyUpgrade is queried by Endpoint.ChanUpgradeAck() so retrieve and mutate fields here appropriately to force failure
 				counterpartyUpgrade := path.EndpointB.GetChannelUpgrade()
 				counterpartyUpgrade.Fields.Version = "invalid-version"
 				path.EndpointB.SetChannelUpgrade(counterpartyUpgrade)
--- a/modules/apps/transfer/ibc_module_test.go
+++ b/modules/apps/transfer/ibc_module_test.go
@@ -420,14 +420,9 @@ func (suite *TransferTestSuite) TestOnChanUpgradeAck() {
 		{
 			"invalid upgrade version",
 			func() {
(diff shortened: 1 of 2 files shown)


modules/apps/transfer/ibc_module_test.go
```diff
@@ -403,3 +403,76 @@ func (suite *TransferTestSuite) TestOnChanUpgradeTry() {
...
+	}{
+		{
+			"success",
+			func() {}, // successful happy path for a standalone transfer app is swapping out the underlying connection
+			nil,
+		},
+		{
+			"invalid upgrade version",
+			func() {
+				// NOTE: counterpartyUpgrade is queried by Endpoint.ChanUpgradeTry() so retrieve and mutate fields here appropriately to force failure
```
damiannolan, line 423:
```suggestion
				// NOTE: counterpartyUpgrade is queried by Endpoint.ChanUpgradeAck() so retrieve and mutate fields here appropriately to force failure
```

modules/apps/transfer/ibc_module_test.go
```diff
@@ -403,3 +403,76 @@ func (suite *TransferTestSuite) TestOnChanUpgradeTry() {
...
+
+			err := path.EndpointA.ChanUpgradeInit()
+			suite.Require().NoError(err)
+
+			err = path.EndpointB.ChanUpgradeTry()
+			suite.Require().NoError(err)
+
+			tc.malleate()
+
+			err = path.EndpointA.ChanUpgradeAck()
```
colin-axner, line 457: interesting that you call the endpoint to obtain the error. Is this going to be problematic if we make changes to other module in the middleware stack? I wonder if we should just call the module's `OnChanUpgradeAck` directly, it would simplify this test code as well as we could remove assertions on what core IBC should be doing

damiannolan: I've updated to invoke the transfer callback directly on the method under test. `Endpoint` methods are still used for setup (INIT, TRY). 

I can see arguments for having both ways, direct call vs full stack with `Endpoint` call. 

We can keep using the full app stack calls in https://github.com/cosmos/ibc-go/pull/4129 but sure, it might be more maintainable to test callbacks directly here.

damiannolan: Refactored `OnChanUpgradeTry` tests here - https://github.com/cosmos/ibc-go/pull/4221

Testing for `OnChanUpgradeInit` can remain the same I think!

chatton (approved):
LGTM 👍

Issue #4252, opened 2023-08-02
Remove checkUpgradeForCompatibility in startFlushUpgradeHandshake function


colin-axner:
## Summary

Requires: 
- [x] #4250 
- [x] #4251 

Remove this [call](https://github.com/cosmos/ibc-go/blob/e00fb0fda1c3fce67a2e18656bef6f4f67577aed/modules/core/04-channel/keeper/upgrade.go#L727)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

Issue #4250, opened 2023-08-02
Add checkForUpgradeCompatibility check in try handler


colin-axner:
## Summary

[Spec change](https://github.com/cosmos/ibc/pull/999/files#diff-69d7c4ac1d4cdeeabcc757d3a34e85945ebc5613a034a52ebd0769cd8a0fc560R526)

Add a call to [`checkForUpgradeCompatibility`](https://github.com/cosmos/ibc-go/blob/e00fb0fda1c3fce67a2e18656bef6f4f67577aed/modules/core/04-channel/keeper/upgrade.go#L735) [here](https://github.com/cosmos/ibc-go/blob/e00fb0fda1c3fce67a2e18656bef6f4f67577aed/modules/core/04-channel/keeper/upgrade.go#L156). If an error is returned,  return the normal error so the transaction fails. There will now be a duplicate check (another one in startFlushUpgradeHandshake, the latter check will be removed in a followup pr)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chatton:
There is actually an issue with the implementation in this PR and the one that has already been merged in #4285 

Currently, just moving the `checkForUpgradeCompatibility` is not enough, since there is actually some [logic](https://github.com/cosmos/ibc-go/blob/ce03b9953f25f6a3b14bd6af8c6bd509f2d1c2d6/modules/core/04-channel/keeper/upgrade.go#L737-L750) which potentially modifies the `channel.UpgradeSequence`. 

The actual effect this has is that on chain A, the channel upgrade sequence is written in `WriteChanUpgradeInit`, however by the time the new `checkForUpgradeCompatibility` happens on chain B, the upgrade sequence is not fast forwarded and is still zero.

I'm going to put this on hold in favour of some of the other ongoing refactors, but we need to reconsider this when we pick up work on this again.

Issue #3825, opened 2023-06-12
Response formatting for channel upgrades msg server handlers


damiannolan:
## Summary

Discuss alternatives for formatting response information in channel upgrades msg server handlers.

[See thread](https://github.com/cosmos/ibc-go/pull/3791#discussion_r1225792629) and [thread](https://github.com/cosmos/ibc-go/pull/3791#discussion_r1226424744).

We currently chose to opt for returning `(Channel, Upgrade)` from `WriteUpgradeTryChannel` in order to have the response values up to date. However, ideally we would keep `WriteX` funcs void, with zero return args. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chatton:
Discussing with @damiannolan , we can keep the `WriteUpgradeTryChannel` fn the same, but update `WriteUpgradeInitChannel` to use the same method signature (return Channel, Upgrade) to stay consistent.

crodriguezvega:
The [signature of `WriteUpgradeInitChannel` changed already](https://github.com/cosmos/ibc-go/pull/5264/files#diff-0fca3dc84b5a25a4a314bceed2fdee673da45a5381aed16d164bb84cd891642eR49) in #5264, therefore the issue can be closed.

PR #4352, opened 2023-08-15
Add functions checkForUpgradeCompatibility and syncUpgradeSequence

chatton:
## Description

closes: #4250
closes: #4252

This PR removes `checkUpgradeForCompatibility` from `startFlushUpgradeHandshake` as well as moving the upgrade sequence fast forwarding logic to its own function. This logic is also executed in the ACK keeper function as it was missed in #4285

The tests which are being removed tested the now removed logic from `startFlushUpgradeHandshake`

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -167,7 +167,7 @@ func (k Keeper) ChanUpgradeTry(
 		return types.Upgrade{}, errorsmod.Wrap(err, "failed to verify counterparty channel state")
 	}
 
-	if err := k.checkForUpgradeSequences(ctx, portID, channelID, channel, counterpartyUpgradeSequence); err != nil {
+	if err := k.syncUpgradeSequence(ctx, portID, channelID, channel, counterpartyUpgradeSequence); err != nil {
 		return types.Upgrade{}, err
 	}
 
@@ -308,7 +308,7 @@ func (k Keeper) ChanUpgradeAck(
 		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "failed to retrieve channel upgrade: port ID (%s) channel ID (%s)", portID, channelID)
 	}
 
-	if err := k.checkForUpgradeSequences(ctx, portID, channelID, channel, counterpartyChannel.UpgradeSequence); err != nil {
+	if err := k.syncUpgradeSequence(ctx, portID, channelID, channel, counterpartyChannel.UpgradeSequence); err != nil {
 		return err
 	}
 
@@ -746,9 +746,9 @@ func (k Keeper) startFlushUpgradeHandshake(
 	return nil
 }
 
-// checkForUpgradeSequences ensures current upgrade handshake only continues if both channels are using the same upgrade sequence,
+// syncUpgradeSequence ensures current upgrade handshake only continues if both channels are using the same upgrade sequence,
 // otherwise an upgrade error is returned so that an error receipt will be written so that the upgrade handshake may be attempted again with synchronized sequences.
-func (k Keeper) checkForUpgradeSequences(ctx sdk.Context, portID, channelID string, channel types.Channel, counterpartyUpgradeSequence uint64) error {
+func (k Keeper) syncUpgradeSequence(ctx sdk.Context, portID, channelID string, channel types.Channel, counterpartyUpgradeSequence uint64) error {
 	// save the previous upgrade sequence for the error message
 	prevUpgradeSequence := channel.UpgradeSequence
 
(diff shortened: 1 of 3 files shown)


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -122,7 +123,7 @@ func (k Keeper) ChanUpgradeTry(
 
 		// NOTE: OnChanUpgradeInit will not be executed by the application
 
-		k.WriteUpgradeInitChannel(ctx, portID, channelID, upgrade)
+		channel = k.WriteUpgradeInitChannel(ctx, portID, channelID, upgrade)
```
chatton, line 126: this is required as the upgrade sequence is modified here, without this the `channel` variable defined above becomes stale.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -730,25 +743,25 @@ func (k Keeper) startFlushUpgradeHandshake(
...
-	// otherwise an error receipt must be written so that the upgrade handshake may be attempted again with synchronized sequences
-	if counterpartyChannel.UpgradeSequence != channel.UpgradeSequence {
-		// save the previous upgrade sequence for the error message
-		prevUpgradeSequence := channel.UpgradeSequence
+	return nil
+}
+
+// checkForUpgradeSequences ensures current upgrade handshake only continues if both channels are using the same upgrade sequence,
+// otherwise an upgrade error is returned so that an error receipt will be written so that the upgrade handshake may be attempted again with synchronized sequences.
+func (k Keeper) checkForUpgradeSequences(ctx sdk.Context, portID, channelID string, channel types.Channel, counterpartyUpgradeSequence uint64) error {
```
crodriguezvega, line 751: I think the word `check` in the function name squeaks a bit for me: I would expect that the function perform only checks and does not mutate state (which does by updating the channel's upgrade sequence). Would you consider `syncUpgradeSequence` or `updateUpgradeSequence` instead?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -60,6 +60,7 @@ func (k Keeper) WriteUpgradeInitChannel(ctx sdk.Context, portID, channelID strin
 	k.Logger(ctx).Info("channel state updated", "port-id", portID, "channel-id", channelID, "previous-state", types.OPEN.String(), "new-state", types.INITUPGRADE.String())
 
 	emitChannelUpgradeInitEvent(ctx, portID, channelID, channel, upgrade)
+	return channel
```
crodriguezvega, line 63: I am a bit of a `command-query separation` purist :) so I do like the convention that state-mutating functions should not return a value.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -299,6 +308,10 @@ func (k Keeper) ChanUpgradeAck(
 		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "failed to retrieve channel upgrade: port ID (%s) channel ID (%s)", portID, channelID)
 	}
 
+	if err := k.checkForUpgradeSequences(ctx, portID, channelID, channel, counterpartyChannel.UpgradeSequence); err != nil {
+		return err
+	}
```
crodriguezvega, line 313: Do we actually need this in `ChanUpgradeAck`? I see that the spec does not have this check, and I guess that if sequences would be different, the upgrade handshake would have already been stopped in `ChanUpgradeTry`?

chatton: this logic was happening in `startFlushUpgradeHandshake` so was already happening in `ChanUpgradeAck`. The current code should be equivalent to that. I think it is probably safe to remove it though since it does happen in the try step!

modules/core/04-channel/keeper/upgrade.go
chatton: I like `syncUpgradeSequence`!

damiannolan (approved):
Dropping my approval. We reviewed this together on a call and I left some suggestions, caught one issue and refactored tests. 

Happy to merge this now and make more progress on the refactor

modules/core/04-channel/keeper/upgrade.go
damiannolan: I think we can definitely come back to this later, I do agree with @crodriguezvega's preferred approach, but I think for now its fine so that we can make progress quickly on the refactor.

chatton:
@crodriguezvega I'll merge this PR to unblock continuing work on the other channel upgrades issues, if you think anything else needs to be changed let me know and I can create a follow up!

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: linking relevant issue opened for the try version #3825 (since I randomly bumped into it)

Issue #4080, opened 2023-07-13
Amend TimeoutPacket to allow timeouts during an upgrade.


DimitrisJim:
## Summary

Allow packets to be timed out on a channel undergoing an upgrade.

## Problem Definition

It currently isn't possible to timeout a packet due to the following condition:

https://github.com/cosmos/ibc-go/blob/49cdfc5ffadff9b8eef09d6a962201f0a4574e52/modules/core/04-channel/keeper/timeout.go#L87-L92

Spec doesn't have this condition, _probably_ due to the initial support for optimistic sends. 

Based on my understanding, this condition [should match that for `AcknowledgePacket`](https://github.com/cosmos/ibc-go/blob/8e53c775cdb915695538723f6ca75f4087bd54c7/modules/core/04-channel/keeper/packet.go#L383-L388) which is _disallowing_ it to timeout if `channel.State != OPEN && channel.FlushStatus != FLUSHING` (i.e only `OPEN` channels or channels in `TRYUPGRADE`/`ACKUPGRADE` that are currently flushing are actually allowed).
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #4367, opened 2023-08-16
Amend timeoutPacket to allow timeouts on any channel state.

DimitrisJim:
## Description

as discussed in the eng. call, this drops the check on channel state during a timeout. 

closes: #4080

```bash
chore: amend timeoutPacket to allow timeouts on any channel state.
```

---

write a little note why.


modules/core/04-channel/keeper/timeout.go
```diff
@@ -84,13 +84,6 @@ func (k Keeper) TimeoutPacket(
 		return types.ErrNoOpMsg
 	}
 
-	if channel.State != types.OPEN {
```
damiannolan, line 87: couldn't we also just extend this to allow timeouts on `OPEN` and `FLUSHING` channels?

DimitrisJim: that was my initial thought pre-discussion. This isn't spec'ed out in any way (i.e the check for `State != types.OPEN` also isn't there), so when we talked about it I remember us coming to the conclusion that it could be removed. Happy to wait for a review from @colin-axner when he gets back

colin-axner: Why would you want to restrict when packets can be timed out?

colin-axner (approved):

modules/core/04-channel/keeper/timeout.go
colin-axner: In theory that makes sense, but in practice it is problematic. There's a potential state where chainA has a closed channel and chainB doesn't. Until someone calls ChanCloseConfirm, all timeouts are valid, but with a restriction, after calling close confirm, timeouts are no longer valid, but that's a very loose assumption and I don't think apps are built to handle channel closes like that

My preference would be to open an issue on the spec if there's a concern. I don't see why ibc-go should be opinionated in this matter

colin-axner: There's also no check on TimeoutOnClose, so this restriction can be bypassed unless you added it there as well

damiannolan: Sounds fair enough to me!

damiannolan (approved):

modules/core/04-channel/keeper/timeout.go
colin-axner: I think the other point I'd make is that IBC only specifies optimistic sends as unsafe (potentially not recoverable). I think it's important to build the assumption that all sends can be reverted regardless of what occurs

Issue #4385, opened 2023-08-17
Amend RecvPacket to remove references to FlushStatus


DimitrisJim:
## Summary

Receive packet needs to be amended to match the updated spec:

https://github.com/cosmos/ibc/blob/256205c409a4bbef1bad9f3f2e4d8b444e4c5e17/spec/core/ics-004-channel-and-packet-semantics/README.md?plain=1#L665

Mainly involves dropping the references to in the channel state check (and adding a check for `FLUSHING`) and amending the check for
the packet sequence.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #4386, opened 2023-08-17
Amend recvPacket as per spec.

DimitrisJim:
## Description

closes: #4385 

```bash
chore: amend recvPacket to reflect updated spec
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/packet.go
+++ b/modules/core/04-channel/keeper/packet.go
@@ -135,14 +135,14 @@ func (k Keeper) RecvPacket(
 		return errorsmod.Wrap(types.ErrChannelNotFound, packet.GetDestChannel())
 	}
 
-	if !collections.Contains(channel.State, []types.State{types.OPEN, types.TRYUPGRADE, types.ACKUPGRADE}) {
-		return errorsmod.Wrapf(types.ErrInvalidChannelState, "expected channel state to be one of [%s, %s, %s], but got %s", types.OPEN.String(), types.TRYUPGRADE.String(), types.ACKUPGRADE.String(), channel.State.String())
+	if !collections.Contains(channel.State, []types.State{types.OPEN, types.STATE_FLUSHING}) {
+		return errorsmod.Wrapf(types.ErrInvalidChannelState, "expected channel state to be one of [%s, %s], but got %s", types.OPEN, types.STATE_FLUSHING, channel.State)
 	}
 
-	// in the case of the channel being in TRYUPGRADE or ACKUPGRADE we need to ensure that the channel is not in flushing,
-	// and that the counterparty last sequence send is less than or equal to the packet sequence.
+	// in the case of the channel being in FLUSHING we need to ensure that the the counterparty last sequence send
+	// is less than or equal to the packet sequence.
 	if counterpartyLastSequenceSend, found := k.GetCounterpartyLastPacketSequence(ctx, packet.GetDestPort(), packet.GetDestChannel()); found {
-		if channel.FlushStatus != types.FLUSHING || packet.GetSequence() > counterpartyLastSequenceSend {
+		if channel.State != types.STATE_FLUSHING || packet.GetSequence() > counterpartyLastSequenceSend {
 			return errorsmod.Wrapf(types.ErrInvalidFlushStatus, "expected channel flush status to be (%s) when counterparty last sequence send (%d) is set, failed to recv packet (%d)", types.FLUSHING, counterpartyLastSequenceSend, packet.GetSequence())
 		}
 	}
--- a/modules/core/04-channel/keeper/packet_test.go
+++ b/modules/core/04-channel/keeper/packet_test.go
@@ -330,35 +330,35 @@ func (suite *KeeperTestSuite) TestRecvPacket() {
 			// attempts to receive packet 2 without receiving packet 1
 			channelCap = suite.chainB.GetChannelCapability(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
 		}, true},
-		// {
-		// 	"success with channel in ACKUPGRADE: FLUSHING status",
-		// 	func() {
-		// 		suite.coordinator.Setup(path)
-		// 		sequence, err := path.EndpointA.SendPacket(defaultTimeoutHeight, disabledTimeoutTimestamp, ibctesting.MockPacketData)
-		// 		suite.Require().NoError(err)
-		// 		packet = types.NewPacket(ibctesting.MockPacketData, sequence, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, defaultTimeoutHeight, disabledTimeoutTimestamp)
-		// 		channelCap = suite.chainB.GetChannelCapability(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
+		{
+			"success with channel in Flushing status",
+			func() {
+				suite.coordinator.Setup(path)
+				sequence, err := path.EndpointA.SendPacket(defaultTimeoutHeight, disabledTimeoutTimestamp, ibctesting.MockPacketData)
+				suite.Require().NoError(err)
+				packet = types.NewPacket(ibctesting.MockPacketData, sequence, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, defaultTimeoutHeight, disabledTimeoutTimestamp)
+				channelCap = suite.chainB.GetChannelCapability(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
 
-		// 		// Move channel to correct state.
-		// 		path.EndpointB.ChannelConfig.ProposedUpgrade.Fields.Version = ibcmock.UpgradeVersion
+				// Move channel to correct state.
+				path.EndpointB.ChannelConfig.ProposedUpgrade.Fields.Version = ibcmock.UpgradeVersion
 
-		// 		err = path.EndpointB.ChanUpgradeInit()
-		// 		suite.Require().NoError(err)
+				err = path.EndpointB.ChanUpgradeInit()
+				suite.Require().NoError(err)
 
-		// 		err = path.EndpointA.ChanUpgradeTry()
-		// 		suite.Require().NoError(err)
+				err = path.EndpointA.ChanUpgradeTry()
+				suite.Require().NoError(err)
 
-		// 		err = path.EndpointB.ChanUpgradeAck()
-		// 		suite.Require().NoError(err)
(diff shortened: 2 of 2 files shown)


modules/core/04-channel/keeper/packet.go
```diff
@@ -128,14 +128,14 @@ func (k Keeper) RecvPacket(
...
 	}
 
-	// in the case of the channel being in TRYUPGRADE or ACKUPGRADE we need to ensure that the channel is not in flushing,
-	// and that the counterparty last sequence send is less than or equal to the packet sequence.
+	// in the case of the channel being in FLUSHING we need to ensure that the the counterparty last sequence send
+	// is less than or equal to the packet sequence.
 	if counterpartyLastSequenceSend, found := k.GetCounterpartyLastPacketSequence(ctx, packet.GetDestPort(), packet.GetDestChannel()); found {
-		if channel.FlushStatus != types.FLUSHING || packet.GetSequence() > counterpartyLastSequenceSend {
+		if channel.State != types.STATE_FLUSHING || packet.GetSequence() > counterpartyLastSequenceSend {
 			return errorsmod.Wrapf(types.ErrInvalidFlushStatus, "expected channel flush status to be (%s) when counterparty last sequence send (%d) is set, failed to recv packet (%d)", types.FLUSHING, counterpartyLastSequenceSend, packet.GetSequence())
```
damiannolan, line 139: We'll no longer be storing the counterparty last packet sequence send under its own key in state as we now store the entire counterparty upgrade.

I think this code will need to be replaced with something like this:

```golang
if channel.State == types.STATE_FLUSHING {
	counterpartyUpgrade, found := k.GetCounterpartyUpgrade(ctx, packet.GetDestPort(), packet.GetDestChannel())
	if !found {
		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "counterparty upgrade not found for channel: %s", packet.GetDestChannel())
	}

	if packet.GetSequence() > counterpartyUpgrade.LatestSequenceSend {
		return errorsmod.Wrapf(
			types.ErrInvalidPacket,
			"failed to receive packet (%d), cannot flush packet at sequence greater than counterparty last sequence send (%d)", packet.GetSequence(), counterpartyUpgrade.LatestSequenceSend,
		)
	}
}
```

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -322,35 +322,35 @@ func (suite *KeeperTestSuite) TestRecvPacket() {
...
-		// 		channel := path.EndpointB.GetChannel()
-		// 		channel.FlushStatus = types.FLUSHING
-		// 		path.EndpointB.SetChannel(channel)
-
-		// 		suite.chainB.GetSimApp().IBCKeeper.ChannelKeeper.SetCounterpartyLastPacketSequence(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, sequence+1)
-		// 	},
-		// 	true,
-		// },
+		{
+			"success with channel in Flushing status",
```
damiannolan, line 326:
```suggestion
			"success with channel in flushing state",
```

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -322,35 +322,35 @@ func (suite *KeeperTestSuite) TestRecvPacket() {
...
+				suite.Require().NoError(err)
+
+				err = path.EndpointB.ChanUpgradeAck()
+				suite.Require().NoError(err)
+
+				channel := path.EndpointB.GetChannel()
+				channel.State = types.STATE_FLUSHING
+				path.EndpointB.SetChannel(channel)
+
+				suite.chainB.GetSimApp().IBCKeeper.ChannelKeeper.SetCounterpartyLastPacketSequence(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, sequence+1)
```
damiannolan, line 350: aha yea! based on the refactor of the code I dumped in the other comment this could be removed, right?

Right now this manual setting is required to get tests to pass?

damiannolan: Just some food for thought - It might be more maintainable to directly set the chainB endpoint channel's state to `FLUSHING` and set the counterparty upgrade in its store with an appropriate packet sequence, rather than relying on doing the full upgrade flow with the endpoint helpers.

modules/core/04-channel/keeper/packet.go
damiannolan: Yeah! So now that we will store the full counterparty upgrade we no longer need to set the counterparty last sequence send under its own key. 

There is a catch here but I don't see it being a big concern really. Basically, the counterparty timeout and last packet send will only be fulfilled in ack or confirm handshake steps. So you might have a relayer attempt a `recvPacket` after the try handler, when the channel is in `FLUSHING` but an error is returned as `packet.GetSequence() > 0` - zero being the `counterpartyUpgrade.LastSequenceSend` as it hasn't been set yet. 

But, once the relayer does the next step of the upgrade handshake and sets the value, then the `recvPacket` should be able to be retried successfully.

damiannolan: I think I recall just advising to punt removing the state key and getter/setter for counterparty last send sequence until we had some more progress. I think we are probably very close to that point now

DimitrisJim: ah, let me ping ya on the point you raised in slack

modules/core/04-channel/keeper/packet_test.go
DimitrisJim: Went with setting it directly for now. I do recall a comment on another PR mentioning it would be nicer to use handshake to go to right state but we can decide on that after all changes have been finalized.

damiannolan: Yeah, I mean, there's pros and cons to both. I personally prefer setting directly as its easier to maintain the tests in isolation. A lot of the existing tests used the endpoint helpers doing the full handshake and when we came to start this refactor I found that it ended up breaking so many tests because they relied on the full handshake

modules/core/04-channel/keeper/packet.go
```diff
@@ -128,15 +128,23 @@ func (k Keeper) RecvPacket(
...
+	if channel.State == types.STATE_FLUSHING {
+		counterpartyUpgrade, found := k.GetCounterpartyUpgrade(ctx, packet.GetDestPort(), packet.GetDestChannel())
+		if !found {
+			return errorsmod.Wrapf(types.ErrUpgradeNotFound, "counterparty upgrade not found for channel: %s", packet.GetDestChannel())
+		}
+
+		if packet.GetSequence() > counterpartyUpgrade.LatestSequenceSend {
+			return errorsmod.Wrapf(
+				types.ErrInvalidPacket,
+				"failed to receive packet (%d), cannot flush packet at sequence greater than counterparty last sequence send (%d)", packet.GetSequence(), counterpartyUpgrade.LatestSequenceSend,
```
damiannolan, line 146: I wonder if this reads better? I'm not quite sure myself! Feel free to ignore the suggestion or apply it 

```suggestion
				"failed to receive packet, cannot flush packet at sequence greater than counterparty last sequence send (%d) > (%d)", packet.GetSequence(), counterpartyUpgrade.LatestSequenceSend,
```

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -361,45 +332,67 @@ func (suite *KeeperTestSuite) TestRecvPacket() {
...
-				suite.chainB.GetSimApp().IBCKeeper.ChannelKeeper.SetCounterpartyLastPacketSequence(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, sequence)
+				// set last packet sent sequence to sequence - 1
+				counterpartyUpgrade := types.NewUpgrade(types.UpgradeFields{}, types.Timeout{}, sequence-1)
+				path.EndpointB.SetChannelCounterpartyUpgrade(counterpartyUpgrade)
 			},
 			false,
 		},
 		{
-			"failure while upgrading channel, channel flushing, packet sequence > counterparty last send sequence",
+			"failure while upgrading channel, channel in FLUSH_COMPLETE",
```
damiannolan, line 383:
```suggestion
			"failure while upgrading channel, channel in flush complete state",
```

modules/core/04-channel/keeper/packet.go
```diff
@@ -128,15 +128,23 @@ func (k Keeper) RecvPacket(
...
-			return errorsmod.Wrapf(types.ErrInvalidFlushStatus, "expected channel flush status to be (%s) when counterparty last sequence send (%d) is set, failed to recv packet (%d)", types.FLUSHING, counterpartyLastSequenceSend, packet.GetSequence())
+	// in the case of the channel being in FLUSHING we need to ensure that the the counterparty last sequence send
+	// is less than or equal to the packet sequence.
+	if channel.State == types.STATE_FLUSHING {
+		counterpartyUpgrade, found := k.GetCounterpartyUpgrade(ctx, packet.GetDestPort(), packet.GetDestChannel())
+		if !found {
+			return errorsmod.Wrapf(types.ErrUpgradeNotFound, "counterparty upgrade not found for channel: %s", packet.GetDestChannel())
+		}
+
+		if packet.GetSequence() > counterpartyUpgrade.LatestSequenceSend {
```
damiannolan, line 143: There is one gotcha here! 

After TRY handler executes, we have `counterpartyUpgrade` and are in `FLUSHING` channel state **BUT** `LastSequenceSend` is not set yet, until after `ACK` and `CONFIRM` succeed. 
Right now, that means relayer will have `recvPacket` rejected until it goes further in the upgrade handshake and then it would need to retry packets. 

I wonder if wrapping this check in:
```go
if counterpartyUpgrade.LastSequenceSend != 0 { 
    // do greater than checking 
}
```

This way we know that if `LastSequenceSend` is still 0 that we can safely continue to receive packets while in `FLUSHING` state.

Happy to leave this as a future improvement, I'll open an issue based off this comment.

damiannolan (approved):
Awesome, I left one comment which I'll open a follow up issue with. Happy to keep the ball rolling and get this merged.

Would love to see the `recvPacket` tests refactored similarly to https://github.com/cosmos/ibc-go/pull/4439 - but can be handled later.

modules/core/04-channel/keeper/packet.go
DimitrisJim: lol, damn nice catch. Can pick it up after this PR is merged, along with the recv test refactor.

modules/core/04-channel/keeper/packet.go
DimitrisJim: batching this in, I prefer juicy information to be visually closer.

modules/core/04-channel/keeper/packet.go
damiannolan: I added the needs discussion label to that one and put it into the beta milestone since its not really the spec behaviour

DimitrisJim: sgtm

Issue #4313, opened 2023-08-08
Add `getUpgradeTimeout` function to calculate timeout for packet flushing in upgrades


damiannolan:
## Summary

Determine how upgrade timeouts for packet processing will be calculated. E.g. via on-chain parameter?

As stated by the spec:

> `getUpgradeTimeout` will return the upgrade timeout specified for the given channel. This may be a chain-wide parameter, or it can be a parameter chosen per channel. This is an implementation-level detail, so only the function signature is specified here. Note this should retrieve some stored timeout delta for the channel and add it to the current height and time to get the absolute timeout values.

```
function getUpgradeTimeout(portIdentifier: string, channelIdentifier: string) UpgradeTimeout {
}
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

damiannolan:
See #4441 for more context, should be linked to this issue

PR #4411, opened 2023-08-21
Add UpgradeTimeout as a params field in the channel submodule

chatton:
## Description

closes: #4313

This PR adds params to the channel submodule and sets a default upgrade timeout. 

If we decide to go with this approach, I can split this up into a separate PR for proto/messages if it makes it more readable.

---

write a little note why.


modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -483,3 +485,59 @@ func (suite *KeeperTestSuite) TestSetUpgradeErrorReceipt() {
...
+// TestParams tests that Param setting and retrieval works properly
+func (suite *KeeperTestSuite) TestParams() {
+	testCases := []struct {
+		name    string
+		input   types.Params
+		expPass bool
+	}{
+		{"success: set default params", types.DefaultParams(), true},
+		{"success: zero timeout timestamp", types.NewParams(types.NewTimeout(clienttypes.ZeroHeight(), 10000)), true},
+		{"success: zero timestamp height", types.NewParams(types.NewTimeout(types.DefaultTimeout.Height, 0)), true},
```
damiannolan, line 509: names should be flipped?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -782,15 +782,15 @@ func (k Keeper) startFlushing(ctx sdk.Context, portID, channelID string, upgrade
...
 
 	return nil
 }
 
-// TODO: use a hard coded value for now. Will be resolved in https://github.com/cosmos/ibc-go/issues/4313
-func getUpgradeTimeout() types.Timeout {
-	return types.NewTimeout(clienttypes.NewHeight(1, 1000), 0)
+// getUpgradeTimeout returns the timeout for the given upgrade.
+func (k Keeper) getUpgradeTimeout(ctx sdk.Context) types.Timeout {
+	return k.GetParams(ctx).UpgradeTimeout
```
damiannolan, line 793: nice, I think we maybe want to create a new timeout which is offset by the values in the timeout stored in params. 

The block height and timestamp values in the timeout stored in params are absolute values - maybe we should be adding the timeout height to the current block height from `ctx` as well as adding the timeout timestamp to the current block time from `ctx`.

what do you think?

proto/ibc/core/channel/v1/channel.proto
```diff
@@ -199,3 +199,9 @@ message Timeout {
   // block timestamp (in nanoseconds) after which the packet or upgrade times out
   uint64 timestamp = 2;
 }
+
+// Params defines the set of IBC channel parameters.
+message Params {
+  // timeout after which channel upgrades will time out.
```
damiannolan, line 205: if we do the the calc of adding timeout height to current block height as well as adding timeout timestamp to current block time then we should document it here that's its basically treated as a relative timeout (we'd want to document this in the our markdown docs as well I guess)

damiannolan (commented):
Looks fantastic! Super clean PR! 

All params stuff looks good to me. I think treating the timeout stored in params as relative values is probably best, and that they are just used to calculate the upgrade timeout during the handshake.

Issue #4128, opened 2023-07-19
Wrong URL encoded path used in solomachine client.


en:
## Summary of Bug

https://github.com/cosmos/ibc-go/blob/db25dc5f48fba7165dd31be422d3320b890f813b/modules/light-clients/06-solomachine/client_state.go#L139
The String() method returns a URL encoded path (like this: "/path%2Fto%2Fleaf"), which is not consistent with spec and other implementations (e.g. ibc-rs)

## Expected Behaviour
Should use path.Pretty() here.

## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Hi @en. Thanks for opening the issue. I just have a couple of questions:

- `which is not consistent with spec`:  do you mean with ICS 06? I briefly checked ICS 06 and also ICS 23 and I cannot find mention of the expected string formatting of a `CommimentPath`.
- `and other implementations (e.g. ibc-rs)`: As far as I know ibc-rs doesn't have an implementation of ICS 06? Or you mean the solo machine implemented [here](https://github.com/devashishdxt/stag)? The paths used in `stag` are the standard [paths specified in ICS 04](https://github.com/cosmos/ibc/blob/main/spec/core/ics-024-host-requirements/README.md#path-space), so there shouldn't be any spaces in the them. Those are the paths that ibc-go construct when verifying that the solo machine has commitment the data.
- To understand better your request: why would you need to have a path with spaces or other "unsafe" characters?

en:
Hi @crodriguezvega,

I mean the paths specified in ICS 04.
These paths are used as storage keys in ICS 07, and as part of the signed data in ICS 06. (https://github.com/cosmos/ibc-go/blob/db25dc5f48fba7165dd31be422d3320b890f813b/modules/light-clients/06-solomachine/client_state.go#L135-L140)

We(Octopus Network) are implementing ICS 06 based on ibc-rs. These paths have to be consistent on both sides so they can be validated by ibc-go. Although there are some definitions of paths in ibc-rs( https://github.com/cosmos/ibc-rs/blob/v0.28.0/crates/ibc/src/core/ics24_host/path.rs#L49-L54 ), we construct them manually into the escaped form(/path%2Fto%2Fleaf) to pass the validation.

So my question is:
Is ibc-go intentionally using the escaped form paths? Or ibc-go used the wrong path.

[...]

PR #4429, opened 2023-08-22
fix(statemachine)!: use path within IBC store without escaping characters in solomachine

crodriguezvega:
## Description

As discussed with @colin-axner, I implemented changes so that:

- the bytes the solo machine signs over are in a key that does not include the IBC store prefix, since the IBC store prefix only makes send in a multistore, and the diversifier already satisfy the function of namespacing the data.
- both `String` and `Pretty` functions become unused and can be removed later on. They make assumptions about formatting a merkle path in a multi store into a string that we shouldn't be making.

ref: #4128 

```bash
type: commit message
```

---

write a little note why.


modules/core/23-commitment/types/merkle.go
```diff
@@ -81,6 +81,12 @@ func NewMerklePath(keyPath ...string) MerklePath {
 // This represents the path in the same way the tendermint KeyPath will
 // represent a key path. The backslashes partition the key path into
 // the respective stores they belong to.
+//
+// Deprecated: This function makes assumptions about the way a merkle path
```
crodriguezvega, line 85: I added a deprecated notice to these functions instead of removing them completely because I think removing them is API breaking and if I remove them in this PR the backport to v7.3.x will be more difficult. I will remove them in a follow up PR once this one gets merged.

modules/core/exported/commitment.go
```diff
@@ -28,7 +28,7 @@ type Prefix interface {
 // Path implements spec:CommitmentPath.
 // A path is the additional information provided to the verification function.
 type Path interface {
-	String() string
+	String() string // deprecated
```
crodriguezvega, line 31: If we remove this function, I am not sure if this interface makes much sense anymore...

testing/solomachine.go
```diff
@@ -516,11 +516,14 @@ func (solo *Solomachine) GenerateClientStateProof(clientState exported.ClientSta
 	data, err := clienttypes.MarshalClientState(solo.cdc, clientState)
 	require.NoError(solo.t, err)
 
+	merklePath := solo.GetClientStatePath(clientIDSolomachine)
```
crodriguezvega, line 519: This testing functions in this file return a merle path with 2 elements (first one for the IBC store), I could also change those to return only the second element. So something like this for `GetClientStatePath`:

```
func (solo *Solomachine) GetClientStatePath(counterpartyClientIdentifier string) commitmenttypes.MerklePath {
  return commitmenttypes.NewMerklePath(host.FullClientStatePath(counterpartyClientIdentifier))
}
```

Although the function the becomes a one-liner, so maybe we don't even to have a separate function to construct the path. Then in the test I would get key 0, instead of 1 as the path over which the solo machine signs the bytes.

colin-axner (approved):
LGTM

testing/solomachine.go
crodriguezvega: Yes, I will do that better in a follow-up PR, otherwise if I remove these functions now that's also API breaking.

modules/light-clients/06-solomachine/client_state.go
```diff
@@ -128,15 +128,21 @@ func (cs *ClientState) VerifyMembership(
...
 	}
 
-	if merklePath.Empty() {
-		return errorsmod.Wrap(commitmenttypes.ErrInvalidProof, "path is empty")
+	if len(merklePath.GetKeyPath()) != 2 {
+		return errorsmod.Wrapf(host.ErrInvalidPath, "path must be of length 2: %s", merklePath.GetKeyPath())
+	}
+
+	// in a multistore context: index 0 is the key for the IBC store in the multistore, index 1 is the key in the IBC store
+	key, err := merklePath.GetKey(1)
```
damiannolan, line 136: `GetKey` seems to do some unescaping on the string also, is this okay here?

from `GetKey`:

```
key, err := url.PathUnescape(mp.KeyPath[i])
```

damiannolan (approved):
Looks good to me! LGTM, left one question about the `GetKey` method

modules/light-clients/06-solomachine/client_state.go
colin-axner: Maybe we should just use `merklePath.KeyPath[1]`?

DimitrisJim: wouldn't this be fine considering escaping calls have been removed? It seems like it would just be best to tweak `GetKey`.

crodriguezvega: That's a good observation @damiannolan! 

I think it is still fine, since it is doing unescaping. And also according to ICS 24 `%` is not a valid character in identifiers and paths, so it should not happen that a path contains an escaped character and then `GetKey` would return it unescaped.

crodriguezvega: What do you guys prefer to do?
-

damiannolan: Happy with opt 2 or 3 I think!

DimitrisJim: if we're having a vote I'd go towards 3.

DimitrisJim (approved):

modules/light-clients/06-solomachine/client_state.go
crodriguezvega: Ok, I will go for 2 and implement 3 in the next PR.

crodriguezvega: Done. I will merge when tests pass.

Issue #4433, opened 2023-08-23
Refactor acknowledgePacket tests to use expResult func


damiannolan:
The `AcknowledgePacket` tests in 04-channel currently expose an `expError` var and in #4412 an additional `assertFn` has been added. 

We can refactor the test case structure to use an `expResult` func and perform explicit error checking and posterior state checks after tests are run.

_Originally posted by @damiannolan in https://github.com/cosmos/ibc-go/pull/4412#discussion_r1301437559_

DimitrisJim:
would definitely be nice to do this for others, `recv` has this `expError` variable that is set in malleate, easy to miss (as I did). Would be nice to also amend it to use `expResult` instead. Can pick it up.

PR #4439, opened 2023-08-23
refactor: use `expResult` in acknowledgePacket tests

damiannolan:
## Description

This PR refactors a large number of tests - specifically `AcknowledgePacket` tests.

The previous code relied on exposing vars for `assertFn` as well as `expError`. This PR replaces those vars with an `expResult` function on the testcase structure.

There is no rush on this PR as it should not be affected by other changes to channel upgradability. I highly recommend reviewing using split-diffs for this PR.

closes: #4433 

---

write a little note why.


modules/core/04-channel/keeper/packet_test.go
```diff
@@ -730,379 +730,514 @@ func (suite *KeeperTestSuite) TestAcknowledgePacket() {
...
+				suite.Require().Error(err)
+				suite.Require().ErrorIs(err, types.ErrPacketSequenceOutOfOrder)
+			},
+		},
 	}
 
 	for i, tc := range testCases {
 		tc := tc
-		suite.Run(fmt.Sprintf("Case %s, %d/%d tests", tc.msg, i, len(testCases)), func() {
+		suite.Run(fmt.Sprintf("Case %s, %d/%d tests", tc.name, i, len(testCases)), func() {
```
chatton, line 1238: not strictly relevant to the refactor, but I think we can just pass in tc.name directly

```golang
suite.Run(tc.name, func(){})
```

The extra formatting produces extra noise in the test output that I don't think is useful. e.g. `TestAcknowledgePacket/Case_success_on_channel_in_flushing_state_with_valid_timeout,_3/20_tests`

modules/core/04-channel/keeper/packet_test.go
```diff
@@ -730,379 +730,514 @@ func (suite *KeeperTestSuite) TestAcknowledgePacket() {
...
-			suite.coordinator.Setup(path)
+				// create packet commitment
+				sequence, err := path.EndpointA.SendPacket(defaultTimeoutHeight, disabledTimeoutTimestamp, ibctesting.MockPacketData)
+				suite.Require().NoError(err)
+				packet = types.NewPacket(ibctesting.MockPacketData, sequence, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, defaultTimeoutHeight, disabledTimeoutTimestamp)
+				channelCap = suite.chainA.GetChannelCapability(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+			},
+			expResult: func(commitment []byte, err error) {
+				suite.Require().Error(err)
+				// suite.Require().ErrorIs(err, connectiontypes.ErrConnectionNotFound)
```
chatton, line 1146: looks like this case didn't assert a specific error so we can remove this?

chatton (approved):
Refactor looks great, much easier to reason about. One thing that might actually be a useful addition to the existing tests is a state check on the channel.

Now that this function is possibly modifying the channel state to OPEN/FLUSHCOMPLETE or leaving at FLUSHING, it might be nice to add a check for OPEN/CLOSED in all the existing tests WDYT?

modules/core/04-channel/keeper/packet_test.go
damiannolan: oh my bad, I remember commenting this out and meaning to come back to it. 

I can push a commit to fix this

DimitrisJim (approved):
sweet

modules/core/04-channel/keeper/packet_test.go
damiannolan: Updated to correct assert `commitmenttypes.ErrInvalidProof`

Issue #1614, opened 2022-06-29
Add upgrade callbacks `IBCModule` interface


damiannolan:
## Summary

Update the [`IBCModule`](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/types/module.go#L13) interface to include application callback handlers for the `04-channel` upgrade API. 

As adapted from [ICS004 Channel Upgrades](https://github.com/cosmos/ibc/blob/master/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#Upgrade-Handshake):

```go
func OnChanUpgradeInit(
        ctx sdk.Context,
        portID,
        channelID string,
        version string,
        order channeltypes.Order,
        connectionHops []string,
        previousVersion string,
) (string, error)

func OnChanUpgradeTry(
        ctx sdk.Context,
        portID,
        channelID string,
        order channeltypes.Order,
        connectionHops []string,
        counterpartyVersion string,
) (string, error)

func OnChanUpgradeAck(
        ctx sdk.Context,
        portID,
        channelID, 
        counterpartyChannelID,
        counterpartyVersion string
) error

func OnChanUpgradeConfirm(
        ctx sdk.Context,
        portID,
        channelID string,
) error

func OnChanUpgradeRestore(
        ctx sdk.Context,
        portID,
        channelID string,
) error
```
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

AdityaSripal:
Transfer and ICA will be trivial since upgrades are not supported at the moment.

We will want to provide a path for non-fee enabled channels to become fee enabled by using the upgrade callbacks in ics-29

damiannolan:
Check the spec for any updates to callbacks. (e.g. previous channel version should be included)

damiannolan:
Scaffolding with no-op callback handlers has been added in #3323

AdityaSripal:
I think this should be a separate interface so it isn't absolutely necessary to implement for IBC functionality.

Similar to how AppModule in the SDK actually has multiple interfaces that each extend the previous with more functionality

damiannolan:
I agree it would be nice to not enforce it.

So essentially remove this https://github.com/cosmos/ibc-go/blob/04-channel-upgrades/modules/core/05-port/types/module.go#L15. When we retrieve `cbs` using the Router we will need to do a conversion then in order to invoke the upgrade callbacks.

[...]

PR #4478, opened 2023-08-28
refactor: adding additional upgrade fields args to onChanUpgradeOpen

damiannolan:
## Description

Note, unfortunately this requires an additional store lookup in order to pass the upgrade fields to the `onChanUpgradeOpen` application callback.

ref: #1614 

---

write a little note why.


modules/core/05-port/types/module.go
```diff
@@ -140,6 +140,9 @@ type UpgradableModule interface {
 		ctx sdk.Context,
 		portID,
 		channelID string,
+		order channeltypes.Order,
+		connectionHops []string,
+		version string,
```
crodriguezvega, line 145: The spec doesn't include these parameters in the callback. What's the reason to add them? What checksI  we want to do?

damiannolan: In the initial handshake app callbacks, the app is expected to do validation of upgrade params only. So adding them to this callback allows them to get access the fields once again when the upgrade is finalized. 

E.g. in fee middleware, we can only validate version info in the upgrade handshake, and can't set fee enabled or disabled until the upgrade is finalized - otherwise packets recv/ack logic within ics29 would start running in the app stack. 
On upgrade open, the only way we can infer whether to set fee enabled or disabled is by what the final upgrade version looks like.

DimitrisJim: wouldn't it we be able to achieve same effect by having the write for upgrade open be performed before the callback?

damiannolan: Yes, in theory, I think suggested that we should do that before as it doesn't really make sense to call the app callback first - same with the `onChanUpgradeRestore` callback. 

However, if you take that approach and remove the arguments, the app would have to manually lookup fields if it needs to do any action which seems like poor application dev ux.

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
```diff
@@ -245,7 +245,8 @@ func (IBCMiddleware) OnChanUpgradeAck(ctx sdk.Context, portID, channelID, counte
 }
 
 // OnChanUpgradeOpen implements the IBCModule interface
-func (IBCMiddleware) OnChanUpgradeOpen(ctx sdk.Context, portID, channelID string) {}
+func (IBCMiddleware) OnChanUpgradeOpen(ctx sdk.Context, portID, channelID string, order channeltypes.Order, connectionHops []string, version string) {
+}
```
chatton, line 249: it bothers me that this is on a separate line when the other ones aren't, but the linter is happy with this so 🤷

chatton (approved):
LGTM!

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
damiannolan: yeah for some reason my editor kept doing this on save :D

damiannolan:
Going to merge this PR so we can continue to make progress in apps. We can comeback to this during the audit and change them if necessary

Issue #4670, opened 2023-09-15
Move `legacySubspace` key table initialisation to migration code


damiannolan:
## Summary

Move `keyTable` initialisation for `legacySubspace` params to the migration handler code which performs the upgrade to self contained params.

Ensure the e2e upgrade test for v8 is re-run as a regression.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #4730, opened 2023-09-20
move legacy keytable init to migration handler code

charleenfei:
## Description

closes: #4670 

```text
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/controller/keeper/migrations.go
+++ b/modules/apps/27-interchain-accounts/controller/keeper/migrations.go
@@ -8,6 +8,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	capabilitytypes "github.com/cosmos/ibc-go/modules/capability/types"
+	"github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/controller/types"
 	controllertypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/controller/types"
 	icatypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/types"
 	host "github.com/cosmos/ibc-go/v8/modules/core/24-host"
@@ -55,6 +56,11 @@ func (m Migrator) AssertChannelCapabilityMigrations(ctx sdk.Context) error {
 // MigrateParams migrates the controller submodule's parameters from the x/params to self store.
 func (m Migrator) MigrateParams(ctx sdk.Context) error {
 	if m.keeper != nil {
+		// set KeyTable if it has not already been set
+		if !m.keeper.legacySubspace.HasKeyTable() {
+			m.keeper.legacySubspace = m.keeper.legacySubspace.WithKeyTable(types.ParamKeyTable())
+		}
+
 		var params controllertypes.Params
 		m.keeper.legacySubspace.GetParamSet(ctx, &params)
 
--- a/modules/apps/27-interchain-accounts/controller/keeper/keeper.go
+++ b/modules/apps/27-interchain-accounts/controller/keeper/keeper.go
@@ -47,11 +47,6 @@ func NewKeeper(
 	ics4Wrapper porttypes.ICS4Wrapper, channelKeeper icatypes.ChannelKeeper, portKeeper icatypes.PortKeeper,
 	scopedKeeper exported.ScopedKeeper, msgRouter icatypes.MessageRouter, authority string,
 ) Keeper {
-	// set KeyTable if it has not already been set
-	if !legacySubspace.HasKeyTable() {
-		legacySubspace = legacySubspace.WithKeyTable(types.ParamKeyTable())
-	}
-
 	if strings.TrimSpace(authority) == "" {
 		panic(fmt.Errorf("authority must be non-empty"))
 	}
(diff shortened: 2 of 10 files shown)


modules/apps/27-interchain-accounts/controller/keeper/migrations_test.go
```diff
@@ -86,7 +86,10 @@ func (suite *KeeperTestSuite) TestMigratorMigrateParams() {
 			func() {
 				params := icacontrollertypes.DefaultParams()
 				subspace := suite.chainA.GetSimApp().GetSubspace(icacontrollertypes.SubModuleName) // get subspace
-				subspace.SetParamSet(suite.chainA.GetContext(), &params)                           // set params
+				if !subspace.HasKeyTable() {
```
DimitrisJim, line 89: will it ever be the case that this is now initialized? I.e can't the `if` be removed completely for testing purposes?

colin-axner: yes this is true. It should never be initialized in the testing setup

modules/apps/27-interchain-accounts/controller/keeper/migrations.go
```diff
@@ -55,6 +55,11 @@ func (m Migrator) AssertChannelCapabilityMigrations(ctx sdk.Context) error {
 // MigrateParams migrates the controller submodule's parameters from the x/params to self store.
 func (m Migrator) MigrateParams(ctx sdk.Context) error {
 	if m.keeper != nil {
+		// set KeyTable if it has not already been set
+		if !m.keeper.legacySubspace.HasKeyTable() {
```
colin-axner, line 59: @damiannolan do we know if this is necessary? I was trying to pickup on why we have this. It seems to me the key table doesn't set the values. When I modify the tests, if the key table isn't set, this will still panic as the param set would need to be set as well

I feel like it is safe to assume that this is run during an in-place migration which means that any chain which has run InitGenesis on a binary which did not use the param set in the keeper store, would have already initialized and set the key table in the param subspace. This should be done on a per chain execution basis, not per node?

colin-axner: Or maybe the issue is that the binary would be restarted, so the key table gets wiped and our tests have a hard time mocking this behaviour because they don't have the pre-set kv store values?

colin-axner: yea, the codecov doesn't seem to be hitting the if statement

colin-axner: Looks like the best way to test this is to set the param keys in the kv store instead of initializing the key table. I think it's just:
```go
store := prefix.NewStore(ctx.KVStore(app.GetKey(paramstypes.StoreKey)), append(icacontrollertypes.SubModuleName, '/'))
store.Set(param.Key, param.Value)
```

damiannolan (approved):

testing/simapp/app.go
```diff
@@ -1005,12 +1006,13 @@ func BlockedAddresses() map[string]bool {
...
 
-	// TODO: ibc module subspaces can be removed after migration of params
-	// https://github.com/cosmos/ibc-go/issues/2010
-	paramsKeeper.Subspace(ibctransfertypes.ModuleName)
-	paramsKeeper.Subspace(ibcexported.ModuleName)
-	paramsKeeper.Subspace(icacontrollertypes.SubModuleName)
-	paramsKeeper.Subspace(icahosttypes.SubModuleName)
+	// register the key tables for legacy param subspaces
+	keyTable := ibcclienttypes.ParamKeyTable()
+	keyTable.RegisterParamSet(&ibcconnectiontypes.Params{})
```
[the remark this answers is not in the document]
charleenfei: it looks like you dont really need the returned key table as the `RegisterParamSet` method also manipulates the table itself?  

```// RegisterParamSet registers multiple ParamSetPairs from a ParamSet in a KeyTable.
func (t KeyTable) RegisterParamSet(ps ParamSet) KeyTable {
	for _, psp := range ps.ParamSetPairs() {
		t = t.RegisterType(psp)
	}
	return t
}
```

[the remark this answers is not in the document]
damiannolan: yeah, its operating on a map in `RegisterType` ...seems pointless to return the `KeyTable`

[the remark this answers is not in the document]
charleenfei: yea, i actually found the function signatures on these params keytables to be a bit deceiving, they all just manipulate the table itself but all also return the table :(

colin-axner (approved):
Thanks for all the work on this @charleenfei! :pray: Everything LGTM!

testing/simapp/app.go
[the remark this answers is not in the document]
DimitrisJim: ah, this is because `KeyTable` is wrapping a map which are basically hidden pointers in go. I'd go with the re-assignment to be on the safe side though I doubt they'll change the internals of `KeyTable` anytime soon.

charleenfei:
> Thanks for all the work on this @charleenfei! 🙏 Everything LGTM!

thank YOU -- another bug found and squished

PR #4749, opened 2023-09-22
update NewErrorAcknowledgement

mdanny1209:
practically, it is hard to debug the issue e.g. on ICA tx failure, as it is only returning see the event logs, and I recommend to update this function to precisely show the issue - for normal transactions, such issue is exactly shown.

### DIFF
--- a/modules/apps/27-interchain-accounts/controller/ibc_middleware.go
+++ b/modules/apps/27-interchain-accounts/controller/ibc_middleware.go
@@ -178,8 +178,9 @@ func (im IBCMiddleware) OnRecvPacket(
 ) ibcexported.Acknowledgement {
 	err := errorsmod.Wrapf(icatypes.ErrInvalidChannelFlow, "cannot receive packet on controller chain")
 
-	logger := im.keeper.Logger(ctx)
-	logger.Error(fmt.Sprintf("ibc middleware: %s sequence %d", err.Error(), packet.Sequence))
+	if err != nil {
+	    im.keeper.Logger(ctx).Error(fmt.Sprintf("27-interchain-accounts: recv packet handling error: %s", err.Error()))
+	}
 
 	ack := channeltypes.NewErrorAcknowledgement(err)
 	keeper.EmitAcknowledgementEvent(ctx, packet, ack, err)
--- a/modules/apps/27-interchain-accounts/controller/ibc_middleware.go
+++ b/modules/apps/27-interchain-accounts/controller/ibc_middleware.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"errors"
+	"fmt"
 
 	errorsmod "cosmossdk.io/errors"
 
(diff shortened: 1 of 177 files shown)


chatton (changes requested):
Hi @mdanny1209, unfortunately this change would lead to non-deterministic error messages if the error strings ever change. We intentionally use a hard coded string and only include the (deterministic) ABCI code.

This is only an issue since Acknowledgements are written to state. (see the note in the docstring for more details)

Perhaps instead to aid debugging we could add additional logging with the concrete error?

mdanny1209:
> Hi @mdanny1209, unfortunately this change would lead to non-deterministic error messages if the error strings ever change. We intentionally use a hard coded string and only include the (deterministic) ABCI code.
> 
> This is only an issue since Acknowledgements are written to state. (see the note in the docstring for more details)
> 
> Perhaps instead to aid debugging we could add additional logging with the concrete error?

 yes, so that the error log could be found on host chain daemon log.

damiannolan:
Hi, thanks for the PR! The error is already logged [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/ibc_module.go#L128), please let us know if that helps.

mdanny1209:
> Hi, thanks for the PR! The error is already logged [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/ibc_module.go#L128), please let us know if that helps.

It would be good to add special message like 27-interchain-accounts: recv packet handling error  so that it can be easy to check the log.
On ICA product development, debugging the failure takes a lot of time, and this simple log change would reduce a lot of time debugging.

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
```diff
@@ -170,12 +171,17 @@ func (im IBCMiddleware) OnChanCloseConfirm(
...
 // OnRecvPacket implements the IBCMiddleware interface
-func (IBCMiddleware) OnRecvPacket(
+func (im IBCMiddleware) OnRecvPacket(
 	ctx sdk.Context,
 	packet channeltypes.Packet,
 	_ sdk.AccAddress,
 ) ibcexported.Acknowledgement {
 	err := errorsmod.Wrapf(icatypes.ErrInvalidChannelFlow, "cannot receive packet on controller chain")
+
+	if err != nil {
```
DimitrisJim, line 181: err will always be non-nil here?

DimitrisJim: we create the err on line 179, the check to see if it isn't nil is not needed is what I'm trying to say :smile:

mdanny1209: I know this PR is not needed but I added a comment.
```It would be good to add special message like 27-interchain-accounts: recv packet handling error so that it can be easy to check the log.
On ICA product development, debugging the failure takes a lot of time, and this simple log change would reduce a lot of time debugging.```

Comment (mdanny1209) [reply modules/apps/27-interchain-accounts/controller/ibc_middleware.go:181]: ```It would be good to add special message like 27-interchain-accounts: recv packet handling error so that it can be easy to check the log.```

Comment (mdanny1209) [reply modules/apps/27-interchain-accounts/controller/ibc_middleware.go:181]: ```On ICA product development, debugging the failure takes a lot of time, and this simple log change would reduce a lot of time debugging.```

Comment (mdanny1209) [reply modules/apps/27-interchain-accounts/controller/ibc_middleware.go:181]: You want me to close this PR?

Comment (DimitrisJim) [reply modules/apps/27-interchain-accounts/controller/ibc_middleware.go:181]: sorry, wasn't clear! the point I was trying to make is that `if err != nil` is superfluous here, i.e, it could just be:

```
im.keeper.Logger(ctx).Error(fmt.Sprintf("27-interchain-accounts: recv packet handling error: %s", err.Error()))
```

without guarding it with the `if`.

mdanny1209: But damiannolan said it's already implemented in [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/ibc_module.go#L128)

damiannolan: > It would be good to add special message like 27-interchain-accounts: recv packet handling error so that it can be easy to check the log.

You should be able to grep the logs pretty easily for `27-interchain-accounts` logs.

The logger used includes this within its context. See here -> https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/host/keeper/keeper.go#L85. 

Look for: `"module": "x/ibc-interchainaccounts"`

PR #4765, opened 2023-09-26
chore: mergify task and e2e tests for `release/v8.0.x`

crodriguezvega:
## Description

closes: #XXXX

```text
type: commit message
```

---

write a little note why.


.github/compatibility-test-matrices/main/client-chain-a.json
```diff
@@ -18,6 +18,7 @@
     "TestClientTestSuite"
   ],
   "test": [
+    "TestClientUpdateProposal_Succeeds",
```
crodriguezvega, line 21: I believe we needed to add [this test](https://github.com/cosmos/ibc-go/pull/4758) back, right?

colin-axner: I'm don't think so. It only really needs to be run on the version we are releasing. This is single chain functionality, so there's no reason to expect changes on main to affect an already released version for client recovery. I think it was unnecessary to have it in compatibility test. We should ensure it is run before new releases though

.github/compatibility-test-matrices/release-v8.0.x/client-chain-a.json
```diff
@@ -0,0 +1,28 @@
...
+    "release-v8.0.x"
+  ],
+  "entrypoint": [
+    "TestClientTestSuite"
+  ],
+  "test": [
+    "TestClientUpdateProposal_Succeeds",
+    "TestRecoverClient_Succeeds",
+    "TestClient_Update_Misbehaviour",
+    "TestAllowedClientsParam"
```
colin-axner, line 23: I don't think proposal tests need to be run in compatibility tests?

.github/compatibility-test-matrices/main/client-chain-a.json
crodriguezvega: I remove from `main` and leave it in the others then?

colin-axner: If we want to limit the amount of tests run, then I think we can:
- make a new file `proposal-chain-a.json` which has the proposal types that primarily only need to be tested on one chain, maybe `single-chain-test.json` works as well (could include genesis restart/upgrades)
- set the matrix to test chain-a version as the release branch and chain-b version as the release branch (no need to set many versions for chain-b)

Is this mostly correct, @chatton? Does this make sense @crodriguezvega?

I expect the following tests to be single chain:
- TestClientUpdateProposal_Succeeds 
- TestRecoverClient_Succeeds (>= v8)
- TestClient_Update_Misbehaviour
- TestAllowedClientsParam
- TestScheduleIBCUpgrade_Succeeds
- TestMaxExpectedTimePerBlockParam
- UpgradeTests
- TestIBCGenesis

chatton: yep we can remove the `chain-b` json file, and keep a single item under the `chain-b` in `proposal-chain-a.json`/`single-chain-test.json` (I prefer `single-chain-test.json` as a name).

crodriguezvega: I updated the PR. I removed the `...-chain-b.json` files where needed, but didn't rename the `...-chain-a.json` files or created a new `single-chain-test.json` file. The reason is that we already had a `localhost-transfer-chain-a.json` file to run the localhost tests and instead of grouping all the single chain tests in the same file, I thought it would still be nice to group them by topic (client, connection, etc). If you don't agree with this, I can still change it.

.github/compatibility-test-matrices/release-v6.1.x/client-chain-a.json
```diff
@@ -3,21 +3,13 @@
...
-    "v5.3.1",
-    "v5.2.1",
-    "v4.4.2",
     "release-v6.1.x"
   ],
   "entrypoint": [
     "TestClientTestSuite"
   ],
   "test": [
-    "TestRecoverClient_Succeeds",
```
crodriguezvega, line 20: I remove `TestRecoverClient_Succeeds` from all releases < than v8.0.0, since the gov v1 message will not be available in those releases.

.github/compatibility-test-matrices/release-v8.0.x/genesis-chain-a.json
```diff
@@ -0,0 +1,17 @@
...
+    "release-v8.0.x"
+  ],
+  "chain-b": [
+    "release-v8.0.x"
+  ],
+  "entrypoint": [
+    "TestGenesisTestSuite"
+  ],
+  "test": [
+    "TestIBCGenesis"
```
crodriguezvega, line 12: I added this file with this test for v8.0.x. I can added for earlier releases, if you think it's worth.

.github/compatibility-test-matrices/unreleased/client.json
```diff
@@ -23,6 +25,7 @@
     "TestClientTestSuite"
   ],
   "test": [
+    "TestClientUpdateProposal_Succeeds",
     "TestRecoverClient_Succeeds",
```
crodriguezvega, line 29: I think we will need to do something special for `TestRecoverClient_Succeeds` in this scenario, because the message will only be available for >= v8.0.x.

colin-axner: same for scheduling IBC upgrade, which I think needs to be added here

.github/compatibility-test-matrices/release-v8.0.x/transfer-chain-a.json
```diff
@@ -0,0 +1,30 @@
...
+  "entrypoint": [
+    "TestTransferTestSuite"
+  ],
+  "test": [
+    "TestMsgTransfer_Succeeds_Nonincentivized",
+    "TestMsgTransfer_Fails_InvalidAddress",
+    "TestMsgTransfer_Timeout_Nonincentivized",
+    "TestMsgTransfer_WithMemo",
+    "TestSendEnabledParam",
+    "TestReceiveEnabledParam"
```
colin-axner, line 25: just for what it is worth, there are 16 unnecessary tests run here, but I guess it is okay

colin-axner (approved):
Mostly LGTM, see comments. Thanks for the followup changes

.github/workflows/e2e-compatibility.yaml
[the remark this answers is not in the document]
crodriguezvega: I deleted them.

.github/compatibility-test-matrices/release-v8.0.x/ica-gov-chain-a.json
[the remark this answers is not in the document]
crodriguezvega: Good catch! I fixed that now.

.github/compatibility-test-matrices/unreleased/client.json
crodriguezvega: True, I added the test as well here now.

.github/compatibility-test-matrices/release-v8.0.x/genesis-chain-a.json
chatton: I think there's no harm adding this to all the active release lines actually. This test has the potential to catch some pretty nasty bugs.

.github/compatibility-test-matrices/main/genesis-chain-a.json
```diff
@@ -0,0 +1,17 @@
...
+    "main"
+  ],
+  "entrypoint": [
+    "TestGenesisTestSuite"
+  ],
+  "test": [
+    "TestIBCGenesis"
+  ],
+  "relayer-type": [
+    "rly"
```
chatton, line 15: it looks like all of the new files are using `rly` but we've updated to use `hermes` as the default everywhere else. I think we should make sure we're consistent. (down the line we can maybe run both here?)

chatton (commented):
Looks great @crodriguezvega, just had one comment regarding using `rly` as the default.

.github/compatibility-test-matrices/main/genesis-chain-a.json
crodriguezvega: oh, yes. Should I change to `hermes` only for `release/v8.0.x` or also for all the previous release lines?

chatton: I think we can do it for everything now, and then re-add rly later on.

crodriguezvega: I switched to hermes everywhere.

chatton (approved):
LGTM!

Issue #4186, opened 2023-07-27
Check `MsgChannelUpgradeInit` is signed by authority


crodriguezvega:
_Reference: https://github.com/cosmos/ibc/pull/999_

Ideally we work on this after https://github.com/cosmos/ibc-go/pull/4160 is merged, since that PR will bring [the addition of `authority` to the core IBC keeper](https://github.com/cosmos/ibc-go/blob/0b746b68c58450d2e60c7a01ac349da8563b42b0/modules/core/keeper/keeper.go#L37), and then the change would be quite simple:

- In [`ChanUpgradeInit` rpc handler](https://github.com/cosmos/ibc-go/blob/e00fb0fda1c3fce67a2e18656bef6f4f67577aed/modules/core/keeper/msg_server.go#L699), add at the top:

```golang
if k.GetAuthority() != msg.Signer {
  return nil, errorsmod.Wrapf(ibcerrors.ErrUnauthorized, "expected %s, got %s", k.GetAuthority(), msg.Signer)
}
```

charleenfei:
I started working on this issue, but ran into complications while testing.

Basically, the way we currently run tests for the `msg_server` code relies on us passing the msg into the `suite.chainA.SendMsgs(msg)` function, which will call `simapp.SignAndDeliver`. However, the simapp function requires the private key of the signer in order to generate the mock signed tx. We typically grab from the `TestChain` struct because we have access to it as the sender account is [generated from a private key which is custom generated for this use case and then saved in the `TestChain` struct. ](https://github.com/cosmos/ibc-go/blob/main/testing/chain.go#L110)

However, in this testing case we set up the IBC keeper authority in the `app.go` to the gov module, but we do not have access to the private key for this account. Therefore, the test will fail when at the `SignAndDeliver` step because the private key will not match the sender account address. 

[...]

PR #4773, opened 2023-09-26
fix: check that `MsgChannelUpgradeInit` is signed by authority

charleenfei:
## Description

closes: #4186 

---

write a little note why.


modules/core/keeper/msg_server.go
```diff
@@ -737,6 +737,10 @@ func (k Keeper) Acknowledgement(goCtx context.Context, msg *channeltypes.MsgAckn
 func (k Keeper) ChannelUpgradeInit(goCtx context.Context, msg *channeltypes.MsgChannelUpgradeInit) (*channeltypes.MsgChannelUpgradeInitResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	if k.GetAuthority() != msg.Signer {
+		return nil, errorsmod.Wrapf(ibcerrors.ErrUnauthorized, "expected signer address to be the same as the authority address %s, got %s", k.GetAuthority(), msg.Signer)
```
colin-axner, line 741: Maybe we should have a consistent error message for this check? https://github.com/cosmos/ibc-go/blob/c694852e20e8191fd49ad3cfd4b25208b78e5b1c/modules/core/keeper/msg_server.go#L106

testing/endpoint.go
```diff
@@ -587,18 +591,68 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+
+	events := res.Events
+	for _, event := range events {
+		for _, attribute := range event.Attributes {
+			if attribute.Key == "proposal_id" {
+				proposalID, err = strconv.Atoi(attribute.Value)
+				require.NoError(endpoint.Chain.TB, err)
+			}
+		}
+	}
```
colin-axner, line 631: we can use the proposal id in the msg response https://github.com/cosmos/cosmos-sdk/blob/main/x/gov/types/v1/tx.pb.go#L151

testing/endpoint.go
```diff
@@ -587,18 +591,68 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+
+	// vote on proposal
+	ctx := endpoint.Chain.GetContext()
+	require.NoError(endpoint.Chain.TB, endpoint.Chain.GetSimApp().GovKeeper.AddVote(ctx, uint64(proposalID), endpoint.Chain.SenderAccount.GetAddress(), govtypesv1.NewNonSplitVoteOption(govtypesv1.OptionYes), ""))
+	require.NoError(endpoint.Chain.TB, endpoint.Chain.GetSimApp().GovKeeper.AddVote(ctx, uint64(proposalID), endpoint.Chain.SenderAccount.GetAddress(), govtypesv1.NewNonSplitVoteOption(govtypesv1.OptionYes), ""))
```
colin-axner, line 636: duplicate lines of code?

testing/endpoint.go
```diff
@@ -587,18 +591,68 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+
+	// fast forward the chain context to end the voting period
+	params, err := endpoint.Chain.GetSimApp().GovKeeper.Params.Get(ctx)
+	require.NoError(endpoint.Chain.TB, err)
+
+	newHeader := endpoint.Chain.GetContext().BlockHeader()
+	newHeader.Time = endpoint.Chain.GetContext().BlockHeader().Time.Add(*params.MaxDepositPeriod).Add(*params.VotingPeriod)
+	ctx = ctx.WithBlockHeader(newHeader)
+	require.NoError(endpoint.Chain.TB, govtypes.EndBlocker(ctx, &endpoint.Chain.GetSimApp().GovKeeper))
```
colin-axner, line 645: use [this](https://github.com/cosmos/ibc-go/blob/main/testing/coordinator.go#L57) function, you aren't incrementing the global time, so after executing this function, the chain is still in a state of pre proposal processing, it may lead to confusing side affects, you should be able to increment a block after incrementing the time and the proposal should be processed

testing/endpoint.go
```diff
@@ -587,18 +591,68 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+
+	// check if proposal passed or failed on msg execution
+	p, err := endpoint.Chain.GetSimApp().GovKeeper.Proposals.Get(ctx, 1)
+	require.NoError(endpoint.Chain.TB, err)
+	if p.Status != govtypesv1.StatusPassed {
+		return fmt.Errorf("proposal failed: %s", p.FailedReason)
+	}
+
+	endpoint.Chain.Coordinator.CommitBlock(endpoint.Chain)
```
colin-axner, line 654: this can be removed if you increment the block instead of calling end block directly

colin-axner (commented):
Mostly LGTM! See suggestions and let me know if you have any questions

testing/endpoint.go
chatton: With this approach we would need to convert the any into a `MsgSubmitProposalResponse` there is [some code](https://github.com/cosmos/ibc-go/blob/07ad1aa70c4ed258c3ea2f5b12b058bc4fbdc73b/e2e/testsuite/codec.go#L89-L113) in the e2e tests that does this already and might be useful.

testing/endpoint.go
```diff
@@ -587,18 +591,68 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+	res, err := endpoint.Chain.SendMsgs(proposal)
+	if err != nil {
+		return err
+	}
+
+	events := res.Events
+	for _, event := range events {
+		for _, attribute := range event.Attributes {
+			if attribute.Key == "proposal_id" {
+				proposalID, err = strconv.Atoi(attribute.Value)
```
crodriguezvega, line 627: Since the proposal ID is used later as a `uint64`, maybe it can be parsed already here as a `uint64` with something like `strconv.ParseUint(attribute.Value, 10, 64)`.

testing/endpoint.go
```diff
@@ -587,18 +591,68 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+	params, err := endpoint.Chain.GetSimApp().GovKeeper.Params.Get(ctx)
+	require.NoError(endpoint.Chain.TB, err)
+
+	newHeader := endpoint.Chain.GetContext().BlockHeader()
+	newHeader.Time = endpoint.Chain.GetContext().BlockHeader().Time.Add(*params.MaxDepositPeriod).Add(*params.VotingPeriod)
+	ctx = ctx.WithBlockHeader(newHeader)
+	require.NoError(endpoint.Chain.TB, govtypes.EndBlocker(ctx, &endpoint.Chain.GetSimApp().GovKeeper))
+
+	// check if proposal passed or failed on msg execution
+	p, err := endpoint.Chain.GetSimApp().GovKeeper.Proposals.Get(ctx, 1)
```
crodriguezvega, line 648:
```suggestion
	p, err := endpoint.Chain.GetSimApp().GovKeeper.Proposals.Get(ctx, proposalID)
```

testing/endpoint.go
```diff
@@ -587,18 +591,68 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+	for _, event := range events {
+		for _, attribute := range event.Attributes {
+			if attribute.Key == "proposal_id" {
+				proposalID, err = strconv.Atoi(attribute.Value)
+				require.NoError(endpoint.Chain.TB, err)
+			}
+		}
+	}
+
+	// vote on proposal
```
crodriguezvega, line 633: Maybe it becomes handy to make a helper function out of this code to vote for the proposal, end voting period and check proposal passed.

crodriguezvega (commented):
Thank you, @charleenfei. Left a couple of comments to maybe create a couple of helper functions.

testing/endpoint.go
```diff
@@ -587,18 +591,68 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+	newHeader.Time = endpoint.Chain.GetContext().BlockHeader().Time.Add(*params.MaxDepositPeriod).Add(*params.VotingPeriod)
+	ctx = ctx.WithBlockHeader(newHeader)
+	require.NoError(endpoint.Chain.TB, govtypes.EndBlocker(ctx, &endpoint.Chain.GetSimApp().GovKeeper))
+
+	// check if proposal passed or failed on msg execution
+	p, err := endpoint.Chain.GetSimApp().GovKeeper.Proposals.Get(ctx, 1)
+	require.NoError(endpoint.Chain.TB, err)
+	if p.Status != govtypesv1.StatusPassed {
+		return fmt.Errorf("proposal failed: %s", p.FailedReason)
+	}
```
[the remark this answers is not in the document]
charleenfei: I wanted to return the error here so that I could do an assertion in the test itself, if that makes sense?

modules/core/keeper/msg_server.go
charleenfei: actually, it looks like theres an invalid signer error in `govtypes`. maybe we can update all of these to use that error return, looks like that's what the 08-wasm client returns in `StoreCode`

colin-axner: sounds good, maybe we open an issue and address once all the feature branches are merged to main?

colin-axner (approved):
ACK, thanks @charleenfei!

testing/endpoint.go
```diff
@@ -587,18 +590,46 @@ func (endpoint *Endpoint) QueryChannelUpgradeProof() ([]byte, []byte, clienttype
...
+
+	events := res.Events
+	for _, event := range events {
+		for _, attribute := range event.Attributes {
+			if attribute.Key == "proposal_id" {
+				proposalID, err = strconv.ParseUint(attribute.Value, 10, 64)
+				require.NoError(endpoint.Chain.TB, err)
+			}
+		}
+	}
```
damiannolan, line 630: we could sweep this under a `ibctesting.ParseProposalIDFromEvents()` func but can be a future improvement 👍

testing/utils.go
```diff
@@ -23,3 +26,27 @@ func ApplyValSetChanges(tb testing.TB, valSet *tmtypes.ValidatorSet, valUpdates
...
+	require.NoError(endpoint.Chain.TB, endpoint.Chain.GetSimApp().GovKeeper.AddVote(ctx, proposalID, endpoint.Chain.SenderAccount.GetAddress(), govtypesv1.NewNonSplitVoteOption(govtypesv1.OptionYes), ""))
+
+	// fast forward the chain context to end the voting period
+	params, err := endpoint.Chain.GetSimApp().GovKeeper.Params.Get(ctx)
+	require.NoError(endpoint.Chain.TB, err)
+
+	endpoint.Chain.Coordinator.IncrementTimeBy(*params.VotingPeriod + *params.MaxDepositPeriod)
+	newHeader := endpoint.Chain.GetContext().BlockHeader()
+	ctx = ctx.WithBlockHeader(newHeader)
+	endpoint.Chain.NextBlock()
```
damiannolan, line 43: won't `NextBlock()` update the context? I think the `newHeader` and `ctx.WithBlockHeader` is not needed here? 

You'd have to use `endpoint.Chain.GetContext()` again below, as the `ctx` var would be stale. 

Might be worth making this change

damiannolan (approved):
Nice work! Left two suggestions, lgtm! 🚀

testing/endpoint.go
damiannolan: created https://github.com/cosmos/ibc-go/issues/4836

testing/utils.go
colin-axner: yea I saw the same thing. The `NextBlock` is likely redundant, but I think this is an instance where it is nicer to just use `endpoint.Chain.GetContext` instead of assinging to a var

testing/endpoint.go
colin-axner: I think there's like 3 issues now for this :smile: #4803, #4809

damiannolan: LOL, I guess we should do it then :D

crodriguezvega (approved):
Thank you for handling the review comments!

PR #4788, opened 2023-09-29
imp: internal package for global variables

crodriguezvega:
## Description

closes: #XXXX

```text
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/light-clients/08-wasm/internal/globals/globals.go
+++ b/modules/light-clients/08-wasm/internal/globals/globals.go
@@ -1,0 +1,16 @@
+package globals
+
+import (
+	wasmvm "github.com/CosmWasm/wasmvm"
+
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
+)
+
+var (
+	WasmVM *wasmvm.VM
+	// Store key for 08-wasm module, required as a global so that the KV store can be retrieved
+	// in the ClientState Initialize function which doesn't have access to the keeper.
+	// The storeKey is used to check the code hash of the contract and determine if the light client
+	// is allowed to be instantiated.
+	WasmStoreKey storetypes.StoreKey
+)
--- a/modules/light-clients/08-wasm/keeper/keeper.go
+++ b/modules/light-clients/08-wasm/keeper/keeper.go
@@ -17,6 +17,7 @@ import (
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
+	"github.com/cosmos/ibc-go/modules/light-clients/08-wasm/internal/globals"
 	"github.com/cosmos/ibc-go/modules/light-clients/08-wasm/types"
 )
 
@@ -44,7 +45,7 @@ func NewKeeperWithVM(
 		panic(errors.New("wasm VM must be not nil"))
 	}
 
-	if types.WasmVM != nil && !reflect.DeepEqual(types.WasmVM, vm) {
+	if globals.WasmVM != nil && !reflect.DeepEqual(globals.WasmVM, vm) {
 		panic(errors.New("global Wasm VM instance should not be set to a different instance"))
 	}
 
@@ -52,8 +53,8 @@ func NewKeeperWithVM(
 		panic(errors.New("authority must be non-empty"))
 	}
 
-	types.WasmVM = vm
-	types.WasmStoreKey = key
+	globals.WasmVM = vm
+	globals.WasmStoreKey = key
 
 	return Keeper{
 		cdc:       cdc,
--- a/modules/light-clients/08-wasm/types/vm.go
+++ b/modules/light-clients/08-wasm/types/vm.go
@@ -10,19 +10,12 @@ import (
 
 	errorsmod "cosmossdk.io/errors"
 
-	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/cosmos/ibc-go/modules/light-clients/08-wasm/internal/globals"
 )
 
-var (
-	WasmVM *wasmvm.VM
-	// Store key for 08-wasm module, required as a global so that the KV store can be retrieved
-	// in the ClientState Initialize function which doesn't have access to the keeper.
-	// The storeKey is used to check the code hash of the contract and determine if the light client
-	// is allowed to be instantiated.
-	WasmStoreKey  storetypes.StoreKey
-	VMGasRegister = NewDefaultWasmGasRegister()
-)
+var VMGasRegister = NewDefaultWasmGasRegister()
 
 // initContract calls vm.Init with appropriate arguments.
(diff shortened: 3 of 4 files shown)


modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -44,16 +45,16 @@ func NewKeeperWithVM(
 		panic(errors.New("wasm VM must be not nil"))
 	}
 
-	if types.WasmVM != nil && !reflect.DeepEqual(types.WasmVM, vm) {
+	if globals.WasmVM != nil && !reflect.DeepEqual(globals.WasmVM, vm) {
```
crodriguezvega, line 48: I wanted to remove the check `!reflect.DeepEqual(globals.WasmVM, vm)` but if I do that, then tests fail because more than one chain is setup and then the condition `globals.WasmVM != nil` will fail. If `WasmVM` is now internal and there is no risk of misusage outside of 08-wasm package, then maybe we can get rid of both checks?

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -10,20 +10,13 @@ import (
...
-)
 
-var (
-	WasmVM *wasmvm.VM
-	// Store key for 08-wasm module, required as a global so that the KV store can be retrieved
-	// in the ClientState Initialize function which doesn't have access to the keeper.
-	// The storeKey is used to check the code hash of the contract and determine if the light client
-	// is allowed to be instantiated.
-	WasmStoreKey  storetypes.StoreKey
-	VMGasRegister = NewDefaultWasmGasRegister()
```
crodriguezvega, line 24: I could also move `VMGasRegister` to the internal package, but that requires reorganizing more code. Let me know if you think it's worth.

modules/light-clients/08-wasm/internal/globals/globals.go
```diff
@@ -0,0 +1,34 @@
...
+	"errors"
+
+	wasmvm "github.com/CosmWasm/wasmvm"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
+)
+
+var (
+	WasmVM *wasmvm.VM
```
crodriguezvega, line 13: Can make this unexported and have also get/set functions, if y'all prefer that.

modules/light-clients/08-wasm/internal/globals/globals.go
```diff
@@ -0,0 +1,34 @@
+package globals
```
crodriguezvega, line 1: Open to better names.

modules/light-clients/08-wasm/keeper/events.go
```diff
@@ -5,19 +5,19 @@ import (
...
 func emitStoreWasmCodeEvent(ctx sdk.Context, codeHash []byte) {
 	ctx.EventManager().EmitEvents(sdk.Events{
 		sdk.NewEvent(
-			clienttypes.EventTypeStoreWasmCode,
-			sdk.NewAttribute(clienttypes.AttributeKeyWasmCodeHash, hex.EncodeToString(codeHash)),
+			types.EventTypeStoreWasmCode,
+			sdk.NewAttribute(types.AttributeKeyWasmCodeHash, hex.EncodeToString(codeHash)),
 		),
 		sdk.NewEvent(
 			sdk.EventTypeMessage,
```
charleenfei, line 19: i guess per @colin-axner 's comment in the call, this section might not be necessary, because this is not a submodule?

modules/light-clients/08-wasm/internal/globals/globals.go
damiannolan: We could just `wasm`? And then rename `wasmVM` to `vm`, as it would stutter otherwise

damiannolan:
Definitely think the `internal` pkg is a good call. Will review more soon :)

modules/light-clients/08-wasm/keeper/events.go
damiannolan: I think we may actually need to keep this. 

It's not necessarily to do with submodules, but a msg's type url. And because `MsgStoreCode` has a type url of [ibc.lightclients.wasm.v1.MsgStoreCode](https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/proto/ibc/lightclients/wasm/v1/tx.proto#L2), when [sdk events are created](https://github.com/cosmos/cosmos-sdk/blob/13c66e12b84c611162041d04a71fc92f9362c9f3/baseapp/baseapp.go#L1049-L1051) - [it will parses index 1](https://github.com/cosmos/cosmos-sdk/blob/main/types/tx_msg.go#L131) from it and then we would end up with module name as `lightclients`.

colin-axner: ah yea that's correct

modules/light-clients/08-wasm/internal/globals/globals.go
colin-axner: I'd be happy with `08-wasm/internal/globals.go` and import as `internal.WasmVM` or `internal.GetVM()`, either is fine by me. Don't really know if we need to create a whole directory for the globals.. but I'm okay with whatever y'all decide

DimitrisJim: isn't it convention to have a file named wasm that corresponds and adds funcs for the proto with a similar name?

I'm fine with globals. It can chill there and be a permanent reminder of how horrible globals are.

chatton: I think as long as the import in the go code is not `globals` I'm fine with a globals.go file. It's not really global sinne the scope is limited to the wasm directory.

crodriguezvega: > Don't really know if we need to create a whole directory for the globals.. but I'm okay with whatever y'all decide

I tried to do it without the subdirectory, but then the package was not internal and I was still able to import it from outside of the module...

modules/light-clients/08-wasm/types/events.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// IBC 08-wasm events
+const (
+	EventTypeStoreWasmCode = "store_wasm_code"
+
+	AttributeKeyWasmCodeHash = "wasm_code_hash"
+)
+
+var AttributeValueCategory = fmt.Sprintf("%s_%s", ibcexported.ModuleName, ModuleName)
```
colin-axner, line 16: should we remove the ibc module prefix?

modules/light-clients/08-wasm/types/vm.go
colin-axner: in a followup pr makes sense to me, then we could swap it out for the wasmd maintained one if a separate go.mod is ever made there (without breaking api)

modules/light-clients/08-wasm/internal/ibcwasm/expected_interfaces.go
```diff
@@ -1,4 +1,4 @@
-package types
+package ibcwasm
```
damiannolan, line 1: 👍

damiannolan (approved):
Nice! I like the `ibcwasm` pkg :)

Issue #4873, opened 2023-10-16
Improve mock WasmVM Query implementation


chatton:
## Summary

Currently, the mock vm has a [Query](https://github.com/cosmos/ibc-go/blob/7032548cdfa2c0f0633ddf2a701cfaddda39c2c6/modules/light-clients/08-wasm/testing/mock_engine.go#L43-L48) function which only has a single override.

In order to ensure that the correct payload is being passed, we should provide a mechanism to register a callback function per message type.

We can create a type constraint, and a mapping of callbacks

```golang
type queryMessage interface {
  exportMetadataMsg | statusMsg | timestampAtHeightMsg | verifyClientMessageMsg | checkForMisbehaviourMsg
}

map[queryMessage]CallbackFn
```

And then the mock engine could expose something like

```golang
mockVm.RegisterQueryCallback(ExportMetadataMsg{}, func(){...})
```

where `ExportMetadataMsg` is a type alias of the unexported `exportMetadataMsg`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #4879, opened 2023-10-16
Improve mock wasmvm query implementation

chatton:
## Description

closes: #4873

The implementation wasn't as straightforward as I was expecting.

It seems it is not possible to use type constraints as keys to a map, because of this I opted to use the type name and pre-populate the map with all allowed types so it is only possible to register types for the expected values.

I needed to move the mock vm to the same package as the types to be able to actually do this.

I'm not too happy about that, and I think we can actually revert that if we just opt to use a set of string constants instead of types. Open to suggestions!

I also looked into the possibility of using an `internal` directory, but types aren't importable from types_test from a separate package.

---

write a little note why.

### DIFF
--- a/modules/light-clients/08-wasm/testing/wasm_endpoint.go
+++ b/modules/light-clients/08-wasm/testing/wasm_endpoint.go
@@ -1,9 +1,9 @@
-package wasmtesting
+package testing
 
 import (
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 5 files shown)


charleenfei (approved):
i think this is a good base to start migrating the tests with so we can check that the correct q methods are called, happy for you to merge it in first and then we iterate after as needed!

damiannolan (approved):

Issue #4825, opened 2023-10-04
Add TestInitialize using mock vm for 08-wasm


colin-axner:
## Summary

Add `TestInitialize` using mock vm, see [commented out test](https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/modules/light-clients/08-wasm/types/client_state_test.go#L265)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #4894, opened 2023-10-17
test(08-wasm): added 'TestInitialize' using mockVM

srdtrk:
## Description

This PR:
- implements `TestInitialize` using mockVM
- renames `grandpaClientID` to `wasmClientID`

closes: #4825 

```text
test(08-wasm): added 'TestInitialize' using mockVM
```

---

write a little note why.

### DIFF
--- a/modules/light-clients/08-wasm/types/client_state_test.go
+++ b/modules/light-clients/08-wasm/types/client_state_test.go
@@ -47,7 +47,7 @@ func (suite *TypesTestSuite) TestStatusGrandpa() {
 
 				clientState = types.NewClientState(clientStateData, suite.codeHash, clienttypes.NewHeight(2000, 5))
 
-				suite.chainA.App.GetIBCKeeper().ClientKeeper.SetClientState(suite.ctx, grandpaClientID, clientState)
+				suite.chainA.App.GetIBCKeeper().ClientKeeper.SetClientState(suite.ctx, wasmClientID, clientState)
 			},
 			exported.Frozen,
 		},
@@ -59,7 +59,7 @@ func (suite *TypesTestSuite) TestStatusGrandpa() {
 
 				clientState = types.NewClientState(clientStateData, suite.codeHash, clienttypes.NewHeight(2000, 36))
 
-				suite.chainA.App.GetIBCKeeper().ClientKeeper.SetClientState(suite.ctx, grandpaClientID, clientState)
+				suite.chainA.App.GetIBCKeeper().ClientKeeper.SetClientState(suite.ctx, wasmClientID, clientState)
 			},
 			exported.Expired,
 		},
@@ -80,8 +80,8 @@ func (suite *TypesTestSuite) TestStatusGrandpa() {
 		suite.Run(tc.name, func() {
 			suite.SetupWasmGrandpaWithChannel()
 
-			clientStore := suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(suite.ctx, grandpaClientID)
-			clientState, ok = suite.chainA.App.GetIBCKeeper().ClientKeeper.GetClientState(suite.ctx, grandpaClientID)
+			clientStore := suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(suite.ctx, wasmClientID)
+			clientState, ok = suite.chainA.App.GetIBCKeeper().ClientKeeper.GetClientState(suite.ctx, wasmClientID)
 			suite.Require().True(ok)
 
 			tc.malleate()
@@ -246,7 +246,7 @@ func (suite *TypesTestSuite) TestInitializeGrandpa() {
 			consensusState = types.NewConsensusState(consensusStateData, 0)
 
 			tc.malleate()
-			clientStore := suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(suite.ctx, grandpaClientID)
+			clientStore := suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(suite.ctx, wasmClientID)
 			err = clientState.Initialize(suite.ctx, suite.chainA.Codec, clientStore, consensusState)
 
 			if tc.expPass {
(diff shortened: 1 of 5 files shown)


modules/light-clients/08-wasm/types/types_test.go
```diff
@@ -34,8 +34,8 @@ import (
 )
 
 const (
-	tmClientID      = "07-tendermint-0"
-	grandpaClientID = "08-wasm-0"
+	tmClientID   = "07-tendermint-0"
+	wasmClientID = "08-wasm-0"
 )
```
srdtrk, line 39: At one point, I was using grandpaClientID in my test (I'm not anymore). So I had renamed it to wasmClientID. I can revert this if requested.

srdtrk: reverted

modules/light-clients/08-wasm/types/client_state_test.go
```diff
@@ -262,66 +262,83 @@ func (suite *TypesTestSuite) TestInitializeGrandpa() {
...
+func (suite *TypesTestSuite) TestInitialize() {
+	panicMsg := errors.New("panic in InstantiateFn")
+	testCases := []struct {
+		name     string
+		malleate func()
+		expError error
+		expPanic interface{}
+	}{
+		{
+			name:     "success: new mock client",
```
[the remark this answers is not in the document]
srdtrk: removed keys!

modules/light-clients/08-wasm/types/client_state_test.go
```diff
@@ -276,66 +276,83 @@ func (suite *TypesTestSuite) TestInitializeGrandpa() {
...
-// 		expPass  bool
-// 	}{
-// 		{
-// 			name: "valid consensus",
-// 			malleate: func() {
-// 				tmConsensusState := tmtypes.NewConsensusState(time.Now(), commitmenttypes.NewMerkleRoot([]byte{0}), []byte(codeHash))
-// 				tmConsensusStateData, err := suite.chainA.Codec.MarshalInterface(tmConsensusState)
-// 				suite.Require().NoError(err)
+func (suite *TypesTestSuite) TestInitialize() {
+	panicMsg := errors.New("panic in InstantiateFn")
```
damiannolan, line 280: I'm unsure if we should bother testing for a panic... what do you think? We should never have to handle a panic from the vm inside of our code, right? Aren't all panics caught [here](https://github.com/CosmWasm/wasmvm/blob/main/internal/api/callbacks.go#L56) within the actual vm and propagated as errors?

Would probably simplify the test structure if we didn't have to include this, but either way test looks good!

modules/light-clients/08-wasm/types/types_test.go
damiannolan: Curious, how come you reverted? I think `wasmClientID` makes more sense, doesn't it?

modules/light-clients/08-wasm/types/types_test.go
```diff
@@ -82,6 +82,9 @@ func (suite *TypesTestSuite) SetupWasmWithMockVM() {
 	suite.coordinator = ibctesting.NewCoordinator(suite.T(), 1)
 	suite.chainA = suite.coordinator.GetChain(ibctesting.GetChainID(1))
 
+	suite.ctx = suite.chainA.GetContext().WithBlockGasMeter(storetypes.NewInfiniteGasMeter())
+	suite.store = suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(suite.ctx, grandpaClientID)
```
damiannolan, line 86: I'd be in favour of switching var name or having another const `DefaultWasmClientID` or similar

damiannolan (approved):
Thanks @srdtrk, nice work. LGTM! 🙏

modules/light-clients/08-wasm/types/types_test.go
srdtrk: It made this PR harder to review due to large diff. I want to make a new PR for this

modules/light-clients/08-wasm/types/types_test.go
srdtrk: Makes sense, will be making another PR for this

modules/light-clients/08-wasm/types/client_state_test.go
srdtrk: Since wasm engine is an interface now, I think it should be in our scope to include basic tests that might break certain assumptions about wasmvm. Also, I'm not sure if they panic later again somewhere in their code

DimitrisJim: I'd agree with Damian here. From what I can understand, the idea is to test that panics are propagated across the boundary (which might not be FFI like wasmvm)? I'm trying to understand the benefit of testing for this, especially given that wasmvm appears to translate em all into err's i.e if we plugged wasmvm here, it would fail.

W/e the case, happy to keep on this path if there's consensus!

damiannolan: cc. @chatton @charleenfei @colin-axner any thoughts/preference? I didn't implement panic tests in #4923 yet but if there's consensus to do it I will push an update. 

I'd see more of an argument to testing for panics if we were actually recovering from them in client state methods, but we aren't, if a panic occurs across the api boundary here its just going to propagated to baseapp anyways. 
I'd be happy to skip the panic tests as it simplifies the structure and is easier to read and maintain in the long run

srdtrk: Yeah, I guess the only benefit I see to implementing panic tests is making sure that there are no panic recoveries anywhere in the stack between callbackFn and 02-client cc. @damiannolan

charleenfei: sorry late to this thread, but i told @damiannolan my thoughts in person (and [in my original discussion on this topic](https://github.com/cosmos/ibc-go/pull/4901#discussion_r1365545002) with @srdtrk)

Issue #4824, opened 2023-10-04
Add TestCheckForMisbehaviour using mock vm for 08-wasm


colin-axner:
## Summary

Add TestCheckForMisbehaviour. See this [commented out test](https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/modules/light-clients/08-wasm/types/misbehaviour_handle_test.go#L579). There is currently no equivalent test with the grandpa client

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Closed by #4901

PR #4901, opened 2023-10-18
test: add test for `CheckForMisbehaviour` payload using mock vm

charleenfei:
## Description

closes: #4824 

---

write a little note why.


modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +483,131 @@ package types_test
...
+// 				clientMessage,
+// 			)
+
+// 			suite.Require().Equal(tc.foundMisbehaviour, foundMisbehaviour)
+// 		})
+// 	}
 // }
 
-/* func (suite *TypesTestSuite) TestCheckForMisbehaviourGrandpa() {
+func (suite *TypesTestSuite) TestCheckForMisbehaviourTendermint() {
```
srdtrk, line 513: Change name to `TestCheckForMisbehaviour`

srdtrk (commented):
Thanks for the work. Left some comments

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +483,131 @@ package types_test
...
+			clientStore = suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(suite.chainA.GetContext(), endpoint.ClientID)
+			clientState = endpoint.GetClientState()
 
 			if tc.expPass {
-				suite.Require().True(foundMisbehaviour)
+				suite.Require().NotPanics(tc.checkMisbehaviour, "unexpected panic")
+				suite.Require().Equal(tc.foundMisbehaviour, foundMisbehaviour)
 			} else {
-				suite.Require().False(foundMisbehaviour)
+				suite.Require().Panicsf(tc.checkMisbehaviour, "failed to unmarshal result of wasm query")
```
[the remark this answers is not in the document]
charleenfei: i think the reason we panic is because the return value of this function is `bool`, so we can't really throw an error (because we are mocking the tendermint vm/functions, i guess we wouldnt change this function signature)

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +483,131 @@ package types_test
...
+			"query failed, vm panic", func() {
+				data, err := base64.StdEncoding.DecodeString(suite.testData["header"])
+				suite.Require().NoError(err)
+				clientMessage = &types.ClientMessage{
+					Data: data,
+				}
+
+				suite.mockVM.RegisterQueryCallback(types.CheckForMisbehaviourMsg{}, func(codeID wasmvm.Checksum, env wasmvmtypes.Env, queryMsg []byte, store wasmvm.KVStore, goapi wasmvm.GoAPI, querier wasmvm.Querier, gasMeter wasmvm.GasMeter, gasLimit uint64, deserCost wasmvmtypes.UFraction) ([]byte, uint64, error) {
+					resp := "cannot be unmarshalled"
+					return []byte(resp), types.DefaultGasUsed, nil
```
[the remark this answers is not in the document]
charleenfei: see above comment about function signature

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
charleenfei: I thought about this too actually :) but the vm functions we have are specific to the Tendermint ClientState so I opted to leave it like this. happy to change it if other people have strong opinions!

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +483,131 @@ package types_test
...
-				data, err := base64.StdEncoding.DecodeString(suite.testData["header"])
-				suite.Require().NoError(err)
-				clientMsg = &types.ClientMessage{
-					Data: data,
-				}
-
-				err = clientState.VerifyClientMessage(suite.ctx, suite.chainA.Codec, suite.store, clientMsg)
-				suite.Require().NoError(err)
+				suite.mockVM.RegisterQueryCallback(types.CheckForMisbehaviourMsg{}, func(codeID wasmvm.Checksum, env wasmvmtypes.Env, queryMsg []byte, store wasmvm.KVStore, goapi wasmvm.GoAPI, querier wasmvm.Querier, gasMeter wasmvm.GasMeter, gasLimit uint64, deserCost wasmvmtypes.UFraction) ([]byte, uint64, error) {
+					resp := fmt.Sprintf(`{"found_misbehaviour":%t}`, false)
```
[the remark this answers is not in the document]
charleenfei: good call! i approved that PR. I think it's less error prone too

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
srdtrk: Sure, I will add here for the record that I have a strong opinion XD

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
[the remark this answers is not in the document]
srdtrk: I was talking about returning an error/panic in the callback function, not `CheckForMisbehaviour`. So you could still assert that it panics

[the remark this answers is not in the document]
charleenfei: you mean testing if the right callback was submitted to `RegisterQueryCallback?`?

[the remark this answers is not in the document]
srdtrk: I thought we should test all possible outcomes of `CheckForMisbehaviour` based on the contract callback result. In general, the contract can:
- Return a valid response -> `CheckForMisbehaviour` = true/false based on response
- Return a malformed response -> `CheckForMisbehaviour` = false
- Return an error -> `CheckForMisbehaviour` = false
- Panic (including out of gas) -> `CheckForMisbehaviour` propagates panic

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +483,125 @@ package types_test
...
+// 	}
 // }
 
-/* func (suite *TypesTestSuite) TestCheckForMisbehaviourGrandpa() {
+func (suite *TypesTestSuite) TestCheckForMisbehaviourTendermint() {
 	var (
-		ok          bool
-		clientMsg   exported.ClientMessage
-		clientState exported.ClientState
+		clientState       exported.ClientState
```
srdtrk, line 515: `clientState` doesn't need to be defined at the top right?

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +483,125 @@ package types_test
...
+		foundMisbehaviour bool
 	)
 
 	testCases := []struct {
-		name     string
-		malleate func()
-		expPass  bool
+		name              string
+		checkMisbehaviour func()
+		foundMisbehaviour bool
```
srdtrk, line 524: nit: maybe rename to `expFoundMisbehaviour`

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +483,125 @@ package types_test
...
+					suite.Assert().NoError(err)
+					return resp, types.DefaultGasUsed, nil
+				})
+
+				foundMisbehaviour = clientState.CheckForMisbehaviour(
+					suite.chainA.GetContext(),
+					suite.chainA.App.AppCodec(),
+					clientStore,
+					clientMessage,
+				)
```
srdtrk, line 540: It seems like these are duplicated in every `tc.checkMisbehaviour`. Is it possible to move this inside `suite.Run`? This would also let you remove `clientStore` and `clientMessage` from the top and allow you to define them in `suite.Run`

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +483,125 @@ package types_test
...
+					resp, err := json.Marshal(types.CheckForMisbehaviourResult{FoundMisbehaviour: false})
+					suite.Assert().NoError(err)
+					return resp, types.DefaultGasUsed, nil
+				})
+
+				foundMisbehaviour = clientState.CheckForMisbehaviour(
+					suite.chainA.GetContext(),
+					suite.chainA.App.AppCodec(),
+					clientStore,
+					clientMessage,
```
srdtrk, line 539: Isn't `clientMessage` nil in this test case? And this is why you're getting `false`. Not because of valid response. This test case would fail if your valid response was true. I think you should add `true` this as a test case.

srdtrk (changes requested):
Thanks for the fixes! I found some other issues that might need fixing.

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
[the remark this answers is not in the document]
charleenfei: i agree with first two test cases, the third and fourth i find a bit redundant bc in both cases what check for misbehaviour actually receives is an error from `wasmQuery`, this is why i just have the test for a malformed resp, but sure, i can add another error test. i can add a test case for a panic propagation as well!

modules/light-clients/08-wasm/types/__debug_bin3281163392
srdtrk, line 1: This file should not be in the commit

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +481,128 @@ package types_test
...
+			suite.Require().NoError(err)
+
+			clientStore := suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(suite.chainA.GetContext(), endpoint.ClientID)
+			clientState := endpoint.GetClientState()
 
 			tc.malleate()
 
-			foundMisbehaviour := clientState.CheckForMisbehaviour(suite.ctx, suite.chainA.Codec, suite.store, clientMsg)
+			data, err := base64.StdEncoding.DecodeString(suite.testData["header"])
+			suite.Require().NoError(err)
```
srdtrk, line 591:
```suggestion
			data := []byte{1}
```
You're using test data for grandpa client here right? So shouldn't this suffice.

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +481,128 @@ package types_test
...
 	)
 
 	testCases := []struct {
-		name     string
-		malleate func()
-		expPass  bool
+		name                 string
+		malleate             func()
+		expFoundMisbehaviour bool
+		expPanic             bool
```
srdtrk, line 521:
```suggestion
		expPanic             error
```
This way, you can remove `panicErr` from the top and check `tc.expPanic != nil` instead.

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +481,128 @@ package types_test
...
 	}{
 		{
-			"valid update no misbehaviour",
+			"no misbehaviour",
```
srdtrk, line 524:
```suggestion
			"success: no misbehaviour",
```
Someone mentioned that we were trying to use `success` and `failure` keywords in test names when I first joined. Can't remember who it was now

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +481,128 @@ package types_test
...
-				clientMsg = &types.ClientMessage{
-					Data: data,
-				}
-
-				err = clientState.VerifyClientMessage(suite.ctx, suite.chainA.Codec, suite.store, clientMsg)
-				suite.Require().NoError(err)
+			"misbehaviour found", func() {
+				suite.mockVM.RegisterQueryCallback(types.CheckForMisbehaviourMsg{}, func(codeID wasmvm.Checksum, env wasmvmtypes.Env, queryMsg []byte, store wasmvm.KVStore, goapi wasmvm.GoAPI, querier wasmvm.Querier, gasMeter wasmvm.GasMeter, gasLimit uint64, deserCost wasmvmtypes.UFraction) ([]byte, uint64, error) {
+					resp, err := json.Marshal(types.CheckForMisbehaviourResult{FoundMisbehaviour: true})
+					suite.Assert().NoError(err)
```
srdtrk, line 539:
```suggestion
					suite.Require().NoError(err)
```
I don't really know the difference but I think we always use `Require`

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +481,128 @@ package types_test
...
-				suite.Require().NoError(err)
-				clientMsg = &types.ClientMessage{
-					Data: data,
-				}
-
-				err = clientState.VerifyClientMessage(suite.ctx, suite.chainA.Codec, suite.store, clientMsg)
-				suite.Require().NoError(err)
+				suite.mockVM.RegisterQueryCallback(types.CheckForMisbehaviourMsg{}, func(codeID wasmvm.Checksum, env wasmvmtypes.Env, queryMsg []byte, store wasmvm.KVStore, goapi wasmvm.GoAPI, querier wasmvm.Querier, gasMeter wasmvm.GasMeter, gasLimit uint64, deserCost wasmvmtypes.UFraction) ([]byte, uint64, error) {
+					resp, err := json.Marshal(types.CheckForMisbehaviourResult{FoundMisbehaviour: false})
+					suite.Assert().NoError(err)
```
srdtrk, line 528:
```suggestion
					suite.Require().NoError(err)
```
Same for all Asserts.

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +481,128 @@ package types_test
...
+			data, err := base64.StdEncoding.DecodeString(suite.testData["header"])
+			suite.Require().NoError(err)
+			clientMessage := &types.ClientMessage{
+				Data: data,
+			}
 
-			if tc.expPass {
-				suite.Require().True(foundMisbehaviour)
+			if !tc.expPanic {
+				foundMisbehaviour = clientState.CheckForMisbehaviour(suite.chainA.GetContext(), suite.chainA.App.AppCodec(), clientStore, clientMessage)
```
srdtrk, line 597:
```suggestion
				foundMisbehaviour = clientState.CheckForMisbehaviour(suite.ctx, suite.chainA.App.AppCodec(), suite.store, clientMessage)
```
You can drop clientStore and `GetContext` entirely. [This was added in a recent commit.](https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/modules/light-clients/08-wasm/types/types_test.go#L84-L85)

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +481,128 @@ package types_test
...
+				suite.Require().Equal(tc.expFoundMisbehaviour, foundMisbehaviour)
 			} else {
-				suite.Require().False(foundMisbehaviour)
+				suite.PanicsWithError(
+					panicErr.Error(),
+					func() {
+						clientState.CheckForMisbehaviour(suite.chainA.GetContext(), suite.chainA.App.AppCodec(), clientStore, clientMessage)
```
srdtrk, line 603:
```suggestion
						clientState.CheckForMisbehaviour(suite.ctx, suite.chainA.App.AppCodec(), suite.store, clientMessage)
```
same here

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +481,128 @@ package types_test
...
-			suite.SetupWasmGrandpaWithChannel()
-			clientState, ok = suite.chainA.App.GetIBCKeeper().ClientKeeper.GetClientState(suite.ctx, grandpaClientID)
-			suite.Require().True(ok)
+			// reset suite to create fresh application state
+			suite.SetupWasmWithMockVM()
+			endpoint := wasmtesting.NewWasmEndpoint(suite.chainA)
+			err := endpoint.CreateClient()
+			suite.Require().NoError(err)
+
+			clientStore := suite.chainA.App.GetIBCKeeper().ClientKeeper.ClientStore(suite.chainA.GetContext(), endpoint.ClientID)
```
srdtrk, line 585:
```suggestion
```
`clientStore` can be replaced by `suite.store`. [This was added in a recent commit.](https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/modules/light-clients/08-wasm/types/types_test.go#L84-L85)

srdtrk (approved):
Thanks so much for the fixes. Just left some nits.

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
charleenfei: actually this was the mockvm based on a tm wasm client iirc, that's why i had header in there [(see data ref)](https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/modules/light-clients/08-wasm/test_data/data.json)

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
charleenfei: i dont think if it makes sense to put success or failure in this case as the test case is for if `checkformisbehaviour` is working as expected right?

srdtrk: You could put `failure` for panic case and `success` otherwise. This is basically what I did in my tests. But totally up to you

crodriguezvega (approved):
Thank you, @charleenfei!

modules/light-clients/08-wasm/types/misbehaviour_handle_test.go
```diff
@@ -469,372 +480,120 @@ package types_test
...
+		{
+			"failure: contract panics, panic propogated", func() {
+				suite.mockVM.RegisterQueryCallback(types.CheckForMisbehaviourMsg{}, func(codeID wasmvm.Checksum, env wasmvmtypes.Env, queryMsg []byte, store wasmvm.KVStore, goapi wasmvm.GoAPI, querier wasmvm.Querier, gasMeter wasmvm.GasMeter, gasLimit uint64, deserCost wasmvmtypes.UFraction) ([]byte, uint64, error) {
+					panic(errors.New("panic in query to contract"))
+				})
 			},
 			false,
+			errors.New("panic in query to contract"),
 		},
 	}
```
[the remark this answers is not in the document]
charleenfei: i think the parameter for this function already dictates it's a `ClientMessage` type so it should fail to compile at all if it's not a `ClientMessage`.

PR #4944, opened 2023-10-25
fix(msg): register proto interface

emidev98:
## Description

Register the ICA msgs to the interface registry allowing any BaseAccount (externally owned account) to interact with the ICA module from offchain without the necessity to be through the CLI. 

Not registering module Msgs can be an issue when using the LCD to submit signed messages due the following error:
```
data: {
  code: 2,
  message: 'unable to resolve type URL /ibc.applications.interchain_accounts.controller.v1.MsgRegisterInterchainAccountResponse',
  details: []
}
```
> reproduced here https://github.com/terra-money/core/pull/196

```text
fix(msg): register proto interface
```

Credits to @javiersuweijie !

### DIFF
--- a/modules/apps/27-interchain-accounts/controller/types/codec.go
+++ b/modules/apps/27-interchain-accounts/controller/types/codec.go
@@ -3,6 +3,7 @@ package types
 import (
 	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/msgservice"
 )
 
 // RegisterInterfaces registers the interchain accounts controller message types using the provided InterfaceRegistry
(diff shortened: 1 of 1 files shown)


colin-axner (approved):
Excellent find!! Sorry for the trouble :heart: 

Would it be useful to backport this fix? It appears to me that this should be non state machine breaking as it looks like it is just adding the ICA [Msg Response](https://github.com/cosmos/cosmos-sdk/blob/main/types/msgservice/msg_service.go#L54) types to the registry? And to the best of my knowledge the state machine doesn't need to unpack response types (please correct me if I am missing something)

damiannolan (approved):
Thank you! ❤️  LGTM

crodriguezvega:
@emidev98 Since you opened #4945 I guess you would like to have this back ported to v7.3.x, right? If you confirm that and if you don't mind, then I will close that PR and add the back port label here, so that mergify opens the PR automatically. It helps me keeping track what things are back ported.

emidev98:
Hey @crodriguezvega yes please. You can proceed closing this PR. Thanks!

PR #4945, opened 2023-10-25
fix(msg): register proto interface

emidev98:
## Description

Register the ICA msgs to the interface registry allowing any BaseAccount (externally owned account) to interact with the ICA module from offchain without the necessity to be through the CLI. 

Not registering module Msgs can be an issue when using the LCD to submit signed messages due the following error:
```
data: {
  code: 2,
  message: 'unable to resolve type URL /ibc.applications.interchain_accounts.controller.v1.MsgRegisterInterchainAccountResponse',
  details: []
}
```
> reproduced here https://github.com/terra-money/core/pull/196

```text
fix(msg): register proto interface
```

Credits to @javiersuweijie !


crodriguezvega:
Mergify will open the backport PR.

Issue #4287, opened 2023-08-03
ibctesting: func ParsePacketFromEvents only returns one packet from events


puneet2019:
## Problem Definition
- ibctesting package has a function ParsePacketFromEvents which helps to parse events so they can be relayed.
- but when I try to send multiple packets. this function only parses first packet and returns.

- inorder to relay all the packets that are part of events while testing my ibc app, a function is required which parses all packets which then can be relayed.
## Proposal
- create a function which rather returns array of Packets instead of first found.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

tbruyelle:
I came to the same conclusion, I don't understand why `ParsePacketFromEvents` returns a single packet, while it takes a slice of events. On my side I had to write a very similar function to overcome that : 

```go
func ParsePacketsFromEvents(events []abci.Event) ([]channeltypes.Packet, error) {
	var packets []channeltypes.Packet
	for i, ev := range events {
		if ev.Type == channeltypes.EventTypeSendPacket {
			packet, err := ibctesting.ParsePacketFromEvents(events[i:])
			if err != nil {
				return nil, err
			}
			packets = append(packets, packet)
		}
	}
	return packets, nil
}
```

Maybe this is just for convenience, in that case do you have any objection if I open a PR to add `ParsePacketsFromEvents` ?

puneet2019:
I have none, all yours!

tbruyelle:
Thanks @puneet2019, any input from the IBC team ? @crodriguezvega ?

PR #4975, opened 2023-10-27
test: add ParsePacketsFromEvents

tbruyelle:
## Description

Returns all the packets found in the events. Comes in addition with ParsePacketFromEvents (singular version), which only returns the first packet found in the events.

closes: #4287

```text
type: commit message
```

---

write a little note why.

### DIFF
--- a/testing/events.go
+++ b/testing/events.go
@@ -64,10 +64,26 @@ func ParseChannelIDFromEvents(events []abci.Event) (string, error) {
 
 // ParsePacketFromEvents parses events emitted from a MsgRecvPacket and returns
 // the first packet found.
+// Returns an error if no packet is found.
 func ParsePacketFromEvents(events []abci.Event) (channeltypes.Packet, error) {
+	packets, err := ParsePacketsFromEvents(events)
+	if err != nil {
+		return channeltypes.Packet{}, err
+	}
+	return packets[0], nil
+}
+
+// ParsePacketsFromEvents parses events emitted from a MsgRecvPacket and returns
+// all the packets found.
+// Returns an error if no packet is found.
+func ParsePacketsFromEvents(events []abci.Event) ([]channeltypes.Packet, error) {
+	ferr := func(err error) ([]channeltypes.Packet, error) {
+		return nil, fmt.Errorf("ibctesting.ParsePacketsFromEvents: %w", err)
+	}
+	var packets []channeltypes.Packet
 	for _, ev := range events {
 		if ev.Type == channeltypes.EventTypeSendPacket {
-			packet := channeltypes.Packet{}
+			var packet channeltypes.Packet
 			for _, attr := range ev.Attributes {
 				switch attr.Key {
 				case channeltypes.AttributeKeyData: //nolint:staticcheck // DEPRECATED
@@ -76,7 +92,7 @@ func ParsePacketFromEvents(events []abci.Event) (channeltypes.Packet, error) {
 				case channeltypes.AttributeKeySequence:
 					seq, err := strconv.ParseUint(attr.Value, 10, 64)
 					if err != nil {
-						return channeltypes.Packet{}, err
+						return ferr(err)
 					}
 
 					packet.Sequence = seq
@@ -96,7 +112,7 @@ func ParsePacketFromEvents(events []abci.Event) (channeltypes.Packet, error) {
 				case channeltypes.AttributeKeyTimeoutHeight:
 					height, err := clienttypes.ParseHeight(attr.Value)
 					if err != nil {
-						return channeltypes.Packet{}, err
+						return ferr(err)
 					}
 
 					packet.TimeoutHeight = height
@@ -104,7 +120,7 @@ func ParsePacketFromEvents(events []abci.Event) (channeltypes.Packet, error) {
 				case channeltypes.AttributeKeyTimeoutTimestamp:
 					timestamp, err := strconv.ParseUint(attr.Value, 10, 64)
 					if err != nil {
-						return channeltypes.Packet{}, err
+						return ferr(err)
 					}
 
 					packet.TimeoutTimestamp = timestamp
@@ -114,27 +130,11 @@ func ParsePacketFromEvents(events []abci.Event) (channeltypes.Packet, error) {
 				}
 			}
 
-			return packet, nil
-		}
-	}
-	return channeltypes.Packet{}, fmt.Errorf("acknowledgement event attribute not found")
-}
-
-// ParsePacketsFromEvents parses events emitted from a MsgRecvPacket and returns
-// all the packets found.
-func ParsePacketsFromEvents(events []abci.Event) ([]channeltypes.Packet, error) {
-	var packets []channeltypes.Packet
-	for i, ev := range events {
-		if ev.Type == channeltypes.EventTypeSendPacket {
-			packet, err := ParsePacketFromEvents(events[i:])
-			if err != nil {
-				return nil, err
-			}
 			packets = append(packets, packet)
 		}
 	}
 	if len(packets) == 0 {
-		return nil, fmt.Errorf("acknowledgement event attribute not found")
+		return ferr(fmt.Errorf("acknowledgement event attribute not found"))
 	}
 	return packets, nil
 }
--- a/testing/events.go
+++ b/testing/events.go
@@ -133,6 +133,9 @@ func ParsePacketsFromEvents(events []abci.Event) ([]channeltypes.Packet, error) 
 			packets = append(packets, packet)
 		}
 	}
+	if len(packets) == 0 {
+		return nil, fmt.Errorf("acknowledgement event attribute not found")
+	}
 	return packets, nil
 }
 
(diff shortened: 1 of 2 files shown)


charleenfei (approved):
thanks @tbruyelle 🙏

testing/events.go
```diff
@@ -120,6 +120,22 @@ func ParsePacketFromEvents(events []abci.Event) (channeltypes.Packet, error) {
 	return channeltypes.Packet{}, fmt.Errorf("acknowledgement event attribute not found")
 }
 
+// ParsePacketsFromEvents parses events emitted from a MsgRecvPacket and returns
+// all the packets found.
+func ParsePacketsFromEvents(events []abci.Event) ([]channeltypes.Packet, error) {
+	var packets []channeltypes.Packet
+	for i, ev := range events {
+		if ev.Type == channeltypes.EventTypeSendPacket {
+			packet, err := ParsePacketFromEvents(events[i:])
```
srdtrk, line 129:
```suggestion
			packet, err := ParsePacketFromEvents(events[i:i+1])
```
Is there any reason why we are not doing this?

testing/events.go
```diff
@@ -120,6 +120,22 @@ func ParsePacketFromEvents(events []abci.Event) (channeltypes.Packet, error) {
 	return channeltypes.Packet{}, fmt.Errorf("acknowledgement event attribute not found")
 }
 
+// ParsePacketsFromEvents parses events emitted from a MsgRecvPacket and returns
+// all the packets found.
+func ParsePacketsFromEvents(events []abci.Event) ([]channeltypes.Packet, error) {
```
srdtrk, line 125: I'd actually prefer if the main logic was happening in this function and have `ParsePacketFromEvents` would call into this and pick the first element from the result. I'm ok either way though.

srdtrk:
Thanks for this :)

testing/events.go
```diff
@@ -120,6 +120,22 @@ func ParsePacketFromEvents(events []abci.Event) (channeltypes.Packet, error) {
...
+	var packets []channeltypes.Packet
+	for i, ev := range events {
+		if ev.Type == channeltypes.EventTypeSendPacket {
```
crodriguezvega, line 128: I think we could also remove this `if` statement, because the same one is in `ParsePacketFromEvents`.

testing/events.go
crodriguezvega: True, because it you use `packet, err := ParsePacketFromEvents(events[i:])` and you have a send packet event in index, let's say 4, then you would parse the same packet multiple times, right?

testing/events.go
crodriguezvega: Yeah, the implementation might become easier to reason like that.

crodriguezvega (commented):
Thank you, @tbruyelle.

crodriguezvega:
@tbruyelle if you agree with the review comments, would you be able to work on them? Otherwise, if you're busy, we are happy to take over the PR and address the comments. If you give me permissions to push to your branch, I am happy to do that.

tbruyelle:
> @tbruyelle if you agree with the review comments, would you be able to work on them? Otherwise, if you're busy, we are happy to take over the PR and address the comments. If you give me permissions to push to your branch, I am happy to do that.

@crodriguezvega oups indeed I was super busy the past weeks, but today is quieter, I'll address the comments, thanks for the ping!

testing/events.go
tbruyelle: Good idea, so I added a test and then moved the logic from `ParsePacketFromEvents` to `ParsePacketsFromEvents`.

testing/events.go
tbruyelle: No because of the condition just above. Basically here, `events[i]` is always of type `EventTypeSendPacket`. Anyway I resolved this conversation because the code is outdated: the logic has been moved to an other function.

testing/events.go
tbruyelle: done with the logic moved from one function to the other.

tbruyelle:
@crodriguezvega I addressed all the comments. As requested I moved the logic from one function to the other, and also added a test before that to ensure I didn't break anything during the process.

testing/events_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+	abci "github.com/cometbft/cometbft/abci/types"
+
+	"github.com/cosmos/ibc-go/v8/modules/core/02-client/types"
+	channeltypes "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+	ibctesting "github.com/cosmos/ibc-go/v8/testing"
+)
+
+func TestParsePacketsFromEvents(t *testing.T) {
+	tests := []struct {
```
crodriguezvega, line 17: nit: this is the naming we use in the majority of other tests.

```suggestion
	testCases := []struct {
```

testing/events_test.go
```diff
@@ -0,0 +1,223 @@
...
+			firstPacket, err := ibctesting.ParsePacketFromEvents(tt.events)
+
+			if tt.expectedError != "" {
+				assert.ErrorContains(err, tt.expectedError)
+			} else {
+				require.NoError(err)
+				assert.Equal(tt.expectedPackets[0], firstPacket)
+			}
+		})
+	}
```
crodriguezvega, line 222: I think this could written like this (assuming that `tests` is renamed to `testCases`):


```golang
  t.Run(tc.name, func(t *testing.T) {
    allPackets, err := ibctesting.ParsePacketsFromEvents(tc.events)

    if tc.expectedError == "" {
      require.NoError(t, err)
      require.Equal(t, tc.expectedPackets, allPackets)
    } else {
      require.ErrorContains(t, err, tc.expectedError)
    }

    firstPacket, err := ibctesting.ParsePacketFromEvents(tc.events)

    if tc.expectedError == "" {
      require.NoError(t, err)
      require.Equal(t, tc.expectedPackets[0], firstPacket)
    } else {
      require.ErrorContains(t, err, tc.expectedError)
    }
   })
}
```

So that:
- We don't need to import `assert`.
- We execute first the checks for the success case.

testing/events_test.go
```diff
@@ -0,0 +1,223 @@
...
+							Key:   channeltypes.AttributeKeyTimeoutTimestamp,
+							Value: "x",
+						},
+					},
+				},
+			},
+			expectedError: "ibctesting.ParsePacketsFromEvents: strconv.ParseUint: parsing \"x\": invalid syntax",
+		},
+		{
+			name: "ok: events with packets",
```
crodriguezvega, line 86: We usually have the success case the first on of the slice.

```suggestion
			name: "success",
```

testing/events_test.go
```diff
@@ -0,0 +1,223 @@
...
+	tests := []struct {
+		name            string
+		events          []abci.Event
+		expectedPackets []channeltypes.Packet
+		expectedError   string
+	}{
+		{
+			name:          "fail: no events",
+			events:        []abci.Event{},
+			expectedError: "ibctesting.ParsePacketsFromEvents: acknowledgement event attribute not found",
```
crodriguezvega, line 26: Since we are using `ErrorContains` it could be enough to just have this (same for the other failure test cases):

```suggestion
			expectedError: "acknowledgement event attribute not found",
```

crodriguezvega (commented):
Thank you for addressing the comments and adding a test, @tbruyelle. I have left a few more comments now. :)

testing/events.go
```diff
@@ -62,12 +62,28 @@ func ParseChannelIDFromEvents(events []abci.Event) (string, error) {
...
+	}
+	return packets[0], nil
+}
+
+// ParsePacketsFromEvents parses events emitted from a MsgRecvPacket and returns
+// all the packets found.
+// Returns an error if no packet is found.
+func ParsePacketsFromEvents(events []abci.Event) ([]channeltypes.Packet, error) {
+	ferr := func(err error) ([]channeltypes.Packet, error) {
+		return nil, fmt.Errorf("ibctesting.ParsePacketsFromEvents: %w", err)
```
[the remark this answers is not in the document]
tbruyelle: As far as I know, the stdlib error package doesn't embed the callstack (unlike other more sophisticated error packages, for example).

This is an attempt to make it easier to find the source of the error. But I can understand if this doesn't suit the code style of the repo. Let me know if you want me to just return unlabelled errors.

tbruyelle:
@crodriguezvega comments addressed here https://github.com/cosmos/ibc-go/pull/4975/commits/8e06417a36c5eb0b78bbaba42e0ba27cfc6253b5

- rename tests to testCases, tt to tc
- remove assert package usage
- revert `if tc.expectedError` condition
- rename success case to "success", move it on top
- remove `expectedError` common prefix

crodriguezvega (approved):
Nice work, @tbruyelle. Thank you for addressing all the comments!

If you rebase with main, I will try to merge this before your branch gets out of sync again. 🤞

tbruyelle:
> Nice work, @tbruyelle. Thank you for addressing all the comments!
> 
> If you rebase with main, I will try to merge this before your branch gets out of sync again. 🤞

Thank you Carlos, I've just rebased on main!

chatton:
ah, those 2 failures are not real failures, feel free to disregard them in this PR

Issue #4925, opened 2023-10-23
Implement `MigrateContract` RPC endpoint


crodriguezvega:
Pre-requisite:
- [x] https://github.com/cosmos/ibc-go/issues/4928

This endpoint will be used for migrating a light client contract to a new wasm byte code.

We will have a new message (e.g. `MsgMigrateContract`) and RPC endpoint (e.g. `MigrateContract`). `MsgMigrateContract` could have the following fields:

- Client ID (for the contract instance that that should be migrated to the new wasm code).
- Code hash (of the wasm code the contract will be migrated to).
- [A `msg` field with the bytes that can be passed to the contract at migration](https://github.com/CosmWasm/wasmd/blob/894076a25cb2ff5219486e3d3628a5dd982e7ac6/proto/cosmwasm/wasm/v1/tx.proto#L217), even if we don't use it right now.

We don't need to have a field for the code hash of the contract we migrate from, since we have the client ID. 

Depending on https://github.com/cosmos/ibc-go/issues/4928 we use the client ID as the contract address, or we use the client ID to access the client store and retrieve the contract address. hat we need to use in the call to `wasmVM.Migrate`.

In `MigrateContract`:

- We check that the new code hash is in the list of allowed code hashes. Since this RPC endpoint is in 08-wasm, we have access to the module store and we can do this check.
- If the old and new code hashes are the same, then we can exit early.
- We call `wasmVM.Migrate` with the contract address ([either the client ID or the contract address stored in the client-prefixed store](https://github.com/cosmos/ibc-go/issues/4928)) and the new code hash.
- We update the code hash in the client state.

We need to update the contracts to add the migrate endpoint to them, even if they no-op at the moment.

crodriguezvega:
I just opened this (draft) PR to add the migrate entry point to the contracts. I kept it simple for now and the entry point just no ops. However, I have a question: I had to add a `MigrateMsg` struct, which is also empty for now, but I guess whenever we want to call `wasmVM.Migrate` with the json-encoding bytes of message, they should deserialise into the `MigrateMsg` struct, right? Does that mean that we need to define already want information we would need to include in `MigrateMsg` in case we want to migrate a contract? Otherwise, if the contract is deployed but we realise later on that we need specific information, then we cannot pass it because it will not serialised correctly, right? 

Any thoughts, @srdtrk?

srdtrk:
[...]

PR #4723, opened 2023-09-19
Add snapshotter extension 

trinitys7:
## Description

This PR is to add the snapshotter extension to sync wasm directory :

- Move default wasm directory from runtime dictionary into chain home folder, `ibc_08-wasm_client_data` will be subfolder of ~/.simapp
- Add IteratorCode function to iterate wasmcode that saved in chain state.
- Testing:  setup testing app with snapshot extension enable.

closes: #4275

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#pull-request-targeting)).
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/build/building-modules/11-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [commit message](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) to be used for the changelog entry in the PR description for review.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.
- [ ] Review `Codecov Report` in the comment section below once CI passes.

PR #5000, opened 2023-11-01
feat(08-wasm): implement 'MigrateContract' RPC endpoint

srdtrk:
## Description

closes: #4925 

```text
feat(08-wasm): implement 'MigrateContract' RPC endpoint
```

---

write a little note why.


modules/light-clients/08-wasm/keeper/events.go
```diff
@@ -21,3 +21,19 @@ func emitStoreWasmCodeEvent(ctx sdk.Context, codeHash []byte) {
...
+			sdk.NewAttribute(types.AttributeKeyClientID, clientID),
+			sdk.NewAttribute(types.AttributeKeyWasmCodeHash, hex.EncodeToString(codeHash)),
+			sdk.NewAttribute(types.AttributeKeyNewCodeHash, hex.EncodeToString(newCodeHash)),
+		),
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+		),
+	})
+}
```
srdtrk, line 39: Events are not tested. This was the case with previous events functions too so maybe this should be its own PR?

modules/light-clients/08-wasm/types/migrate_contract.go
```diff
@@ -0,0 +1,34 @@
...
+	ctx sdk.Context, cdc codec.BinaryCodec, clientID string,
+	clientStore storetypes.KVStore, newCodeHash, migrateMsg []byte,
+) error {
+	if !HasCodeHash(ctx, newCodeHash) {
+		return ErrWasmCodeHashNotFound
+	}
+
+	if bytes.Equal(cs.CodeHash, newCodeHash) {
+		return nil
+	}
```
srdtrk, line 26: https://github.com/cosmos/ibc-go/issues/4925 explains that we should return early if code hashes are the same. I think we should error. I'd like to hear other's opinions on this

modules/light-clients/08-wasm/internal/ibcwasm/expected_interfaces.go
```diff
@@ -51,6 +51,24 @@ type WasmEngine interface {
 		deserCost wasmvmtypes.UFraction,
 	) ([]byte, uint64, error)
 
+	// Migrate will migrate an existing contract to a new code binary.
+	// This takes storage of the data from the original contract and the CodeID of the new contract that should
+	// replace it. This allows it to run a migration step if needed, or return an error if unable to migrate
+	// the given data.
+	//
+	// MigrateMsg has some data on how to perform the migration.
```
srdtrk, line 59: Maybe I should improve the godocs here including mentioning explicitly that:

1. Migrate should run on the new code.
2. CodeHash change in client state is performed by the go code already (but wasm code can't do this currently as the code doesn't know its own code hash unless we pass it).

Open to more suggestions on what I should add

modules/light-clients/08-wasm/keeper/msg_server.go
```diff
@@ -31,3 +31,20 @@ func (k Keeper) StoreCode(goCtx context.Context, msg *types.MsgStoreCode) (*type
 		Checksum: codeHash,
 	}, nil
 }
+
+func (k Keeper) MigrateContract(goCtx context.Context, msg *types.MsgMigrateContract) (*types.MsgMigrateContractResponse, error) {
```
srdtrk, line 35: add godocs

proto/ibc/lightclients/wasm/v1/tx.proto
```diff
@@ -27,3 +30,19 @@ message MsgStoreCodeResponse {
...
+
+// MsgMigrateContract defines the request type for the MigrateContract rpc.
+message MsgMigrateContract {
+  option (cosmos.msg.v1.signer) = "signer";
+
+  string signer = 1;
+  // the client id of the contract
+  string client_id = 2;
+  // the code hash of the new wasm byte code for the contract
+  bytes code_hash = 3;
```
srdtrk, line 42: Should this be `bytes` or hex encoded string? I wasn't sure which is more appropriate.

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -134,3 +144,51 @@ func (k Keeper) storeWasmCode(ctx sdk.Context, code []byte) ([]byte, error) {
...
+// GetWasmClientState returns the 08-wasm client state for the given client identifier.
+func (k Keeper) GetWasmClientState(ctx sdk.Context, clientID string) (*types.ClientState, error) {
+	clientState, found := k.clientKeeper.GetClientState(ctx, clientID)
+	if !found {
+		return nil, errorsmod.Wrapf(types.ErrClientNotFound, "wasm client with the given identifier (%s) not found", clientID)
+	}
+
+	wasmClientState, ok := clientState.(*types.ClientState)
+	if !ok {
+		return nil, errorsmod.Wrapf(types.ErrInvalidData, "client state is not a wasm client state")
```
DimitrisJim, line 190:
```suggestion
                return nil, errorsmod.Wrapf(clienttypes.ErrInvalidClient, "expected type %T, got %T", &ClientState{}, wasmClientState)
```

modules/light-clients/08-wasm/keeper/keeper_test.go
```diff
@@ -58,6 +73,63 @@ func (suite *KeeperTestSuite) SetupTest() {
...
+	suite.mockVM = wasmtesting.NewMockWasmEngine()
+	// TODO: move default functionality required for wasm client testing to the mock VM
+	suite.mockVM.StoreCodeFn = func(code wasmvm.WasmCode) (wasmvm.Checksum, error) {
+		hash := sha256.Sum256(code)
+		return wasmvm.Checksum(hash[:]), nil
+	}
+
+	suite.mockVM.PinFn = func(codeID wasmvm.Checksum) error {
+		return nil
+	}
```
DimitrisJim, line 96: note: can be removed if #4723 lands first

modules/light-clients/08-wasm/internal/ibcwasm/expected_interfaces.go
DimitrisJim: I'd be perfectly happy leaving way wasmd has it

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -36,9 +39,14 @@ type Keeper struct {
 func NewKeeperWithVM(
 	cdc codec.BinaryCodec,
 	storeService storetypes.KVStoreService,
+	clientKeeper types.ClientKeeper,
 	authority string,
 	vm ibcwasm.WasmEngine,
 ) Keeper {
+	if clientKeeper == nil {
```
DimitrisJim, line 46: I don't think we typically do this but can't hurt.

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -88,11 +97,146 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+			},
+			types.ErrClientNotFound,
+		},
+		{
+			"failure: contract returns error",
+			func() {
+				msg = types.NewMsgMigrateContract(govAcc, defaultWasmClientID, newCodeHash, []byte("{}"))
+
+				suite.mockVM.MigrateFn = func(_ wasmvm.Checksum, _ wasmvmtypes.Env, _ []byte, _ wasmvm.KVStore, _ wasmvm.GoAPI, _ wasmvm.Querier, _ wasmvm.GasMeter, _ uint64, _ wasmvmtypes.UFraction) (*wasmvmtypes.Response, uint64, error) {
+					return nil, wasmtesting.DefaultGasUsed, types.ErrInvalid
```
DimitrisJim, line 197: can use `wasmtesting.ErrMockContract`

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -88,11 +97,146 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
 				suite.Require().Empty(events)
 			}
 		})
 	}
 }
+
+func (suite *KeeperTestSuite) TestMsgMigrateContract() {
+	oldCodeHash := sha256.Sum256(wasmtesting.Code)
+
+	mockByteCode := []byte("MockByteCode-TestMsgMigrateContract")
```
DimitrisJim, line 111: ditto `wwasmtesting.Code`

modules/light-clients/08-wasm/types/migrate_contract.go
DimitrisJim: gov should pay for their mistakes :money_with_wings:

modules/light-clients/08-wasm/types/msgs.go
```diff
@@ -36,3 +36,35 @@ func (m MsgStoreCode) ValidateBasic() error {
...
+
+	if err := ValidateClientID(m.ClientId); err != nil {
+		return err
+	}
+
+	if len(m.Msg) == 0 {
+		return errorsmod.Wrap(ibcerrors.ErrInvalidRequest, "migrate message cannot be empty")
+	}
+
+	return nil
```
DimitrisJim, line 69: could also see a cheeky refactor like this

```suggestion
	_, err := sdk.AccAddressFromBech32(m.Signer)
	if err != nil {
		return errorsmod.Wrapf(ibcerrors.ErrInvalidAddress, "string could not be parsed as address: %v", err)
	}

	if len(m.Msg) == 0 {
		return errorsmod.Wrap(ibcerrors.ErrInvalidRequest, "migrate message cannot be empty")
	}

	if err := ValidateWasmCodeHash(m.CodeHash); err != nil {
		return err
	}

        return ValidateClientID(m.ClientId)
```

modules/light-clients/08-wasm/types/validation.go
```diff
@@ -19,3 +28,31 @@ func ValidateWasmCode(code []byte) error {
...
+func ValidateClientID(clientID string) error {
+	isClientID := clienttypes.IsValidClientID(clientID)
+	if !isClientID {
+		return errorsmod.Wrapf(ErrInvalidWasmClientID, "invalid client identifier %s", clientID)
+	}
+
+	hasWasmPrefix := strings.HasPrefix(clientID, exported.Wasm)
+	if !hasWasmPrefix {
+		return errorsmod.Wrapf(ErrInvalidWasmClientID, "client identifier %s does not contain %s prefix", clientID, exported.Wasm)
+	}
```
DimitrisJim, line 55: nit: think in these cases we can make do w/o vars assigned, directly plop the function call in the `if`.

DimitrisJim (commented):
dope! left a couple of comments/nits after a quick scan

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: `wasmtesting.Code` is the oldCode, so the `mockByteCode` is the new Code

modules/light-clients/08-wasm/types/migrate_contract.go
srdtrk: they'd pay either way right?

modules/light-clients/08-wasm/keeper/msg_server_test.go
DimitrisJim: ah, maybe a rename then to indicate that

modules/light-clients/08-wasm/types/migrate_contract.go
crodriguezvega: Either works for me.

proto/ibc/lightclients/wasm/v1/tx.proto
crodriguezvega: We use `bytes` in `ClientState`, so we could also use `bytes` here.

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -134,3 +144,51 @@ func (k Keeper) storeWasmCode(ctx sdk.Context, code []byte) ([]byte, error) {
...
+	emitMigrateContractEvent(ctx, clientID, oldCodeHash, newCodeHash)
+
+	return nil
+}
+
+// GetWasmClientState returns the 08-wasm client state for the given client identifier.
+func (k Keeper) GetWasmClientState(ctx sdk.Context, clientID string) (*types.ClientState, error) {
+	clientState, found := k.clientKeeper.GetClientState(ctx, clientID)
+	if !found {
+		return nil, errorsmod.Wrapf(types.ErrClientNotFound, "wasm client with the given identifier (%s) not found", clientID)
```
crodriguezvega, line 185: Just to keep it simple and consistent with same error messages in other parts of the codebase.

```suggestion
		return nil, errorsmod.Wrapf(types.ErrClientNotFound, "clientID %s", clientID)
```

modules/light-clients/08-wasm/types/migrate_contract.go
```diff
@@ -0,0 +1,34 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// MigrateContract calls the migrate entry point on the contract with the given
+// migrateMsg. The contract must exist and the code hash must be found in the
+// store. If the code hash is the same as the current code hash, return nil.
+// This does not update the code hash in the client state.
+func (cs ClientState) MigrateContract(
+	ctx sdk.Context, cdc codec.BinaryCodec, clientID string,
+	clientStore storetypes.KVStore, newCodeHash, migrateMsg []byte,
```
crodriguezvega, line 18: Nit about ordering.

```suggestion
	ctx sdk.Context, cdc codec.BinaryCodec, clientStore storetypes.KVStore,
	clientID string, newCodeHash, migrateMsg []byte,
```

crodriguezvega (commented):
Great work, @srdtrk. I also did a quick review (didn't look at the tests yet) and left a bunch of minor comments.

modules/light-clients/08-wasm/keeper/events.go
srdtrk: I'll add to this PR

proto/ibc/lightclients/wasm/v1/tx.proto
srdtrk: But the code hash query results are returned in a hex encoded string

modules/light-clients/08-wasm/types/msgs.go
srdtrk: Yeah I noticed this too. I remember at some point, @damiannolan said he prefers explicit nil returns in a different context.

modules/light-clients/08-wasm/types/migrate_contract.go
srdtrk: turned into error

modules/light-clients/08-wasm/keeper/keeper.go
srdtrk: done

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: done

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: done

modules/light-clients/08-wasm/types/validation.go
srdtrk: done

modules/light-clients/08-wasm/internal/ibcwasm/expected_interfaces.go
```diff
@@ -51,6 +51,24 @@ type WasmEngine interface {
 		deserCost wasmvmtypes.UFraction,
 	) ([]byte, uint64, error)
 
+	// Migrate will migrate an existing contract to a new code binary.
```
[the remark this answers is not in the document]
srdtrk: done

modules/light-clients/08-wasm/internal/ibcwasm/expected_interfaces.go
```diff
@@ -51,6 +51,24 @@ type WasmEngine interface {
...
+	// This takes storage of the data from the original contract and the CodeID of the new contract that should
```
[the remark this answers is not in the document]
srdtrk: done

modules/light-clients/08-wasm/types/errors.go
```diff
@@ -8,14 +8,16 @@ var (
 	ErrInvalidCodeHash      = errorsmod.Register(ModuleName, 4, "invalid code hash")
 	ErrInvalidClientMessage = errorsmod.Register(ModuleName, 5, "invalid client message")
 	ErrRetrieveClientID     = errorsmod.Register(ModuleName, 6, "failed to retrieve client id")
+	ErrInvalidWasmClientID  = errorsmod.Register(ModuleName, 7, "invalid wasm client id")
```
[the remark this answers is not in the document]
srdtrk: done

modules/light-clients/08-wasm/keeper/keeper.go
srdtrk: done

modules/light-clients/08-wasm/types/migrate_contract.go
srdtrk: done

modules/light-clients/08-wasm/types/errors.go
```diff
@@ -8,14 +8,16 @@ var (
...
+	ErrClientNotFound       = errorsmod.Register(ModuleName, 8, "client not found")
```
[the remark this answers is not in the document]
srdtrk: done

modules/light-clients/08-wasm/keeper/keeper_test.go
srdtrk: removed

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -88,11 +97,146 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+
+				suite.Require().NoError(err)
+				suite.Require().NotNil(res)
+
+				suite.Require().Equal(expClientState, clientState)
+			} else {
+				suite.Require().ErrorIs(err, tc.expError)
+				suite.Require().Nil(res)
+
+				suite.Require().Equal(expClientState, clientState)
```
colin-axner, line 238: state checks are unnecessary for error conditions as you can assume baseapp will revert any potential state changes

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -88,11 +97,146 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+				}
+			},
+			nil,
+		},
+		{
+			"failure: same code hash",
+			func() {
+				newCodeHash = oldCodeHash[:]
+				msg = types.NewMsgMigrateContract(govAcc, defaultWasmClientID, newCodeHash, []byte("{}"))
+
```
colin-axner, line 163: does `newCodeHash` need to be reassigned? I find it easier to read if oldCodeHash is provided as arg to msg migrate

modules/light-clients/08-wasm/types/migrate_contract.go
```diff
@@ -0,0 +1,36 @@
...
+	errorsmod "cosmossdk.io/errors"
+	storetypes "cosmossdk.io/store/types"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// MigrateContract calls the migrate entry point on the contract with the given
+// migrateMsg. The contract must exist and the code hash must be found in the
+// store. If the code hash is the same as the current code hash, return nil.
```
colin-axner, line 16:
```suggestion
// store. If the code hash is the same as the current code hash, an error is returned.
```

modules/light-clients/08-wasm/types/migrate_contract_test.go
```diff
@@ -0,0 +1,140 @@
...
+		{
+			"success: update client state",
+			func() {
+				newHash = sha256.Sum256([]byte{1, 2, 3})
+				err := ibcwasm.CodeHashes.Set(suite.chainA.GetContext(), newHash[:])
+				suite.Require().NoError(err)
+
+				suite.mockVM.MigrateFn = func(_ wasmvm.Checksum, _ wasmvmtypes.Env, _ []byte, store wasmvm.KVStore, _ wasmvm.GoAPI, _ wasmvm.Querier, _ wasmvm.GasMeter, _ uint64, _ wasmvmtypes.UFraction) (*wasmvmtypes.Response, uint64, error) {
+					expClientState = types.NewClientState([]byte{1}, newHash[:], clienttypes.NewHeight(2000, 2))
+					store.Set(host.ClientStateKey(), clienttypes.MustMarshalClientState(suite.chainA.App.AppCodec(), expClientState))
```
colin-axner, line 63: should this have the same assertions as the success case?

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -18,15 +18,19 @@ import (
 
 	"github.com/cosmos/ibc-go/modules/light-clients/08-wasm/internal/ibcwasm"
 	"github.com/cosmos/ibc-go/modules/light-clients/08-wasm/types"
+	clienttypes "github.com/cosmos/ibc-go/v8/modules/core/02-client/types"
```
damiannolan, line 21: we should be able to remove this import and rely solely on the expected keeper interface added in `types`

modules/light-clients/08-wasm/types/msgs.go
DimitrisJim: fair enough, I'd tend to agree with larger funcs where I wouldn't be able to see sig

modules/light-clients/08-wasm/types/msgs.go
```diff
@@ -36,3 +36,35 @@ func (m MsgStoreCode) ValidateBasic() error {
...
+		return err
+	}
+
+	if err := ValidateClientID(m.ClientId); err != nil {
+		return err
+	}
+
+	if len(m.Msg) == 0 {
+		return errorsmod.Wrap(ibcerrors.ErrInvalidRequest, "migrate message cannot be empty")
+	}
```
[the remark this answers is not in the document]
srdtrk: You need to send at least `[]byte("{}")`. So yes. But we could also do this for the user if they send empty bytes. What do you think?

modules/light-clients/08-wasm/keeper/keeper.go
srdtrk: This is for importing errors though?

proto/ibc/lightclients/wasm/v1/tx.proto
charleenfei: it seems like when we actually use the codehash we typically want a `string` format so my vote would be for string, otherwise the user has to convert from bytes to string every time anyway

modules/light-clients/08-wasm/keeper/keeper_test.go
```diff
@@ -5,23 +5,35 @@ import (
...
 	"github.com/cosmos/ibc-go/modules/light-clients/08-wasm/testing/simapp"
 	"github.com/cosmos/ibc-go/modules/light-clients/08-wasm/types"
+	clienttypes "github.com/cosmos/ibc-go/v8/modules/core/02-client/types"
+	host "github.com/cosmos/ibc-go/v8/modules/core/24-host"
+	"github.com/cosmos/ibc-go/v8/modules/core/exported"
 	ibctesting "github.com/cosmos/ibc-go/v8/testing"
 )
 
+const (
+	defaultWasmClientID = "08-wasm-0"
```
charleenfei, line 34: just a nit but could we move this to a general value? we also declare it in `types_test`

srdtrk: We cannot right? As test packages are not exported no matter where they are. So we wouldn't be able to export this constant unless we made it a part of our public package.

modules/light-clients/08-wasm/types/migrate_contract_test.go
srdtrk: I think this is what we did in other tests too. The point of sucess case is to assert right params were passed but I don't think we need to check env, migrateMsg and etc again in here

modules/light-clients/08-wasm/types/migrate_contract.go
srdtrk: done

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: done

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: removed

modules/light-clients/08-wasm/types/validation.go
```diff
@@ -19,3 +28,31 @@ func ValidateWasmCode(code []byte) error {
...
+		return errorsmod.Wrap(ErrInvalidCodeHash, "code hash cannot be empty")
+	}
+	if lenCodeHash != 32 { // sha256 output is 256 bits long
+		return errorsmod.Wrapf(ErrInvalidCodeHash, "expected length of 32 bytes, got %d", lenCodeHash)
+	}
+
+	return nil
+}
+
+// ValidateClientID validates the client identifier
```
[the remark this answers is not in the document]
srdtrk: done

modules/light-clients/08-wasm/types/msgs.go
```diff
@@ -36,3 +36,35 @@ func (m MsgStoreCode) ValidateBasic() error {
 
 	return nil
 }
+
+// MsgMigrateContract creates a new MsgMigrateContract instance
+func NewMsgMigrateContract(signer string, clientID string, codeHash []byte, migrateMsg []byte) *MsgMigrateContract {
```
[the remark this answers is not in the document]
srdtrk: I think I added now?

proto/ibc/lightclients/wasm/v1/tx.proto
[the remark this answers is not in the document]
srdtrk: changed field name

modules/light-clients/08-wasm/keeper/events.go
srdtrk: Added events tests

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -134,3 +144,51 @@ func (k Keeper) storeWasmCode(ctx sdk.Context, code []byte) ([]byte, error) {
...
+	oldCodeHash := wasmClientState.CodeHash
+
+	// client state may be updated by the contract migration
+	wasmClientState, err = k.GetWasmClientState(ctx, clientID)
+	if err != nil {
+		// note that this also ensures that the updated client state is
+		// still a wasm client state
+		return errorsmod.Wrap(err, "failed to retrieve the updated wasm client state")
+	}
+
```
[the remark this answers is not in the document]
srdtrk: I don't think so unless we add this check to every time sudo is called which I might be in favour of. What do you think?

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -128,6 +142,39 @@ func wasmSudo[T ContractResult](ctx sdk.Context, clientStore storetypes.KVStore,
...
 }
 
+// wasmMigrate migrate calls the migrate entry point of the contract with the given payload and returns the result.
+// wasmMigrate returns an error if:
+// - the contract migration returns an error
+// - the response of the contract call contains non-empty messages
+// - the response of the contract call contains non-empty events
+// - the response of the contract call contains non-empty attributes
+// - the data bytes of the response cannot be unmarshal into the result type
+func wasmMigrate[T ContractResult](ctx sdk.Context, clientID string, clientStore storetypes.KVStore, cs *ClientState, payload []byte) (T, error) {
```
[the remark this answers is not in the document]
srdtrk: Others such as `wasmSudo` don't have dedicated tests. Should be a separate issue at the very least imo.

proto/ibc/lightclients/wasm/v1/tx.proto
srdtrk: Can we get extra opinions here @colin-axner @damiannolan

modules/light-clients/08-wasm/keeper/keeper.go
damiannolan: my b 👍

modules/light-clients/08-wasm/types/validation_test.go
```diff
@@ -14,44 +21,140 @@ func TestValidateWasmCode(t *testing.T) {
...
+			"failure: empty clientID",
+			func() {
+				clientID = ""
+			},
+			errorsmod.Wrapf(types.ErrInvalidWasmClientID, "invalid client identifier %s", clientID),
+		},
+		{
+			"failure: clientID is not a wasm client identifier",
+			func() {
+			},
```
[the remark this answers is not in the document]
srdtrk: yes, done

modules/light-clients/08-wasm/types/vm.go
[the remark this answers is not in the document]
colin-axner: There's a test for wasmSudo, the test name just forgot to be changed as well. https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/modules/light-clients/08-wasm/types/vm_test.go#L131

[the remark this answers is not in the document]
srdtrk: Ah got it. I'll add this to the PR

modules/light-clients/08-wasm/keeper/keeper.go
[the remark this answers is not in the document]
colin-axner: It'd be an extra state read, but I think it could be a nice safety check to add after sudo/migrate. It would assert:
* wasm client isn't deleted from store
* it isn't changed to another type
* its code hash only changes via MsgMigrateContract

Happy to open an issue and tackle later. It should be non-API breaking check so no rush to include now. What do you think?

modules/light-clients/08-wasm/types/msgs.go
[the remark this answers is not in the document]
colin-axner: https://github.com/cosmos/ibc-go/pull/5000/commits/5b06718f33adeb4dbd6db6a7fe66665c174d47b3

modules/light-clients/08-wasm/types/msgs.go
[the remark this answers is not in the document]
colin-axner: I see, I'm fine with as is in that case. We can always change the behaviour if someone requests it

proto/ibc/lightclients/wasm/v1/tx.proto
colin-axner: The field for the old code hash and the new code hash should probably be of the same type. Generally it is nice if the users don't need to do any conversions between the type they query for and the type they submit as input. The hex encoding is simply for readability? Making this a hex encoded string doesn't solve user issues when creating the client and needing to provide the initial code hash in bytes

modules/light-clients/08-wasm/types/migrate_contract.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	if bytes.Equal(cs.CodeHash, newCodeHash) {
+		return errorsmod.Wrapf(ErrWasmCodeExists, "new code hash (%s) is the same as current code hash (%s)", hex.EncodeToString(newCodeHash), hex.EncodeToString(cs.CodeHash))
+	}
+
+	// update the code hash, this needs to be done before the contract migration
+	// so that wasmMigrate can call the right code. Note that this is not
+	// persisted to the client store.
+	cs.CodeHash = newCodeHash
```
colin-axner, line 33: Okay, so we call migrate on the new code hash and the new code hash is given access to the old code hashes client store? If the new contract needed the specific code hash of the old contract, it could query it via the client state?

modules/light-clients/08-wasm/keeper/keeper_test.go
colin-axner: yea or move it to `08-wasm/testing`. I'm happy with whatever, doesn't feel too much of a issue to me to define once per package

modules/light-clients/08-wasm/types/migrate_contract.go
srdtrk: Yes, it is given the access to clientStore so that it can perform any state migration if needed! The new code should have no reason to need the old code hash. If it is needed for any reason, it is already in the clientState. Or it could be a built in constant in the new code since it is a future iteration of the same code.

modules/light-clients/08-wasm/types/vm.go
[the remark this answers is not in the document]
srdtrk: Added. This was a good idea as it allowed me to improve the `wasmMigrate` api. I've modelled it after instantiate instead of sudo.

modules/light-clients/08-wasm/keeper/keeper_test.go
srdtrk: Yeah, I think it makes more sense to move all of mockvm logic there, and if we end up doing that refactor, then we can move the constant along with it.

modules/light-clients/08-wasm/types/msgs.go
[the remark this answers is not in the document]
srdtrk: Sounds good

modules/light-clients/08-wasm/keeper/keeper.go
```diff
@@ -134,3 +145,50 @@ func (k Keeper) storeWasmCode(ctx sdk.Context, code []byte) ([]byte, error) {
...
+	clientStore := k.clientKeeper.ClientStore(ctx, clientID)
+
+	err = wasmClientState.MigrateContract(ctx, k.cdc, clientStore, clientID, newCodeHash, migrateMsg)
+	if err != nil {
+		return errorsmod.Wrap(err, "contract migration failed")
+	}
+
+	// client state may be updated by the contract migration
+	wasmClientState, err = k.GetWasmClientState(ctx, clientID)
```
DimitrisJim, line 164: would be nice to have cov for the `err` case

modules/light-clients/08-wasm/types/validation_test.go
```diff
@@ -14,44 +22,141 @@ func TestValidateWasmCode(t *testing.T) {
...
 
-		err := ValidateWasmCode(code)
+		err := types.ValidateClientID(clientID)
 
-		if tc.expPass {
+		if tc.expError == nil {
 			require.NoError(t, err, tc.name)
 		} else {
-			require.Error(t, err, tc.name)
+			require.ErrorContains(t, err, tc.expError.Error(), tc.name)
```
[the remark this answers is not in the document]
DimitrisJim: same error different err msg is what is checked.

modules/light-clients/08-wasm/types/validation_test.go
```diff
@@ -14,44 +22,141 @@ func TestValidateWasmCode(t *testing.T) {
...
+
+	testCases := []struct {
+		name     string
+		malleate func()
+		expError error
+	}{
+		{
+			"success: valid wasm client identifier",
+			func() {
+				clientID = defaultWasmClientID
```
DimitrisJim, line 131: could also assign clientID before malleate and have empty func for success but not particularly bullish on that

modules/light-clients/08-wasm/types/msgs.go
```diff
@@ -36,3 +38,37 @@ func (m MsgStoreCode) ValidateBasic() error {
 
 	return nil
 }
+
+// MsgMigrateContract creates a new MsgMigrateContract instance
+//
+//nolint:interfacer
```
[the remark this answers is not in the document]
DimitrisJim: nope! sdk.Msg interface is implemented by proto generated code

modules/light-clients/08-wasm/types/migrate_contract_test.go
```diff
@@ -0,0 +1,140 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupWasmWithMockVM()
+
+			oldHash = sha256.Sum256(wasmtesting.Code)
+			newHash = oldHash
```
DimitrisJim, line 111: I think it makes more sense to actually have the setting of the new code hash before the malleate and have this in the failure case. wdyt?

modules/light-clients/08-wasm/testing/mock_engine.go
```diff
@@ -143,6 +144,14 @@ func (m *MockWasmEngine) Query(codeID wasmvm.Checksum, env wasmvmtypes.Env, quer
 	return callbackFn(codeID, env, queryMsg, store, goapi, querier, gasMeter, gasLimit, deserCost)
 }
 
+// Migrate implements the WasmEngine interface.
+func (m *MockWasmEngine) Migrate(codeID wasmvm.Checksum, env wasmvmtypes.Env, migrateMsg []byte, store wasmvm.KVStore, goapi wasmvm.GoAPI, querier wasmvm.Querier, gasMeter wasmvm.GasMeter, gasLimit uint64, deserCost wasmvmtypes.UFraction) (*wasmvmtypes.Response, uint64, error) {
+	if m.MigrateFn == nil {
+		panic("mock engine is not properly initialized")
```
DimitrisJim, line 150: unrelated but should probably make these more specific in the future.

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -79,20 +88,176 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+		name     string
+		malleate func()
+		expError error
+	}{
+		{
+			"success",
+			func() {
+				msg = types.NewMsgMigrateContract(govAcc, defaultWasmClientID, newCodeHash, []byte("{}"))
+
+				suite.mockVM.MigrateFn = func(_ wasmvm.Checksum, _ wasmvmtypes.Env, _ []byte, _ wasmvm.KVStore, _ wasmvm.GoAPI, _ wasmvm.Querier, _ wasmvm.GasMeter, _ uint64, _ wasmvmtypes.UFraction) (*wasmvmtypes.Response, uint64, error) {
```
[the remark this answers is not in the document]
DimitrisJim: w/o being certain on this, does the migrate entry point typically operate the same as `Sudo` and `Query`? I.e would we define a specific API with contracts allowing only certain types of migrations?

DimitrisJim (commented):
some more minor things after a second pass

proto/ibc/lightclients/wasm/v1/tx.proto
crodriguezvega: In general `x/wasm` uses `bytes` for the code hash (see [here](https://github.com/CosmWasm/wasmd/blob/a2bbd36364a78867319ad0a28f85203c8cdfbc8d/proto/cosmwasm/wasm/v1/tx.proto#L108) and [here](https://github.com/CosmWasm/wasmd/blob/a2bbd36364a78867319ad0a28f85203c8cdfbc8d/proto/cosmwasm/wasm/v1/tx.proto#L412)).

modules/light-clients/08-wasm/keeper/msg_server_test.go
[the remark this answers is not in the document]
crodriguezvega: I believe we discussed this in one of our calls: the `[]byte` passed here is the bytes of the JSON-encoded string in the `msg` field of `MsgMigrateContract`. This is the same behaviour as `x/wasm`. It should be up to the contracts to make sure that the bytes Marshall to one of the messages that they would expect. Each light client implementation might have different needs for what fields should be included in the message, so it's difficult to generalise in 08-wasm.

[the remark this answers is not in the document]
crodriguezvega: We could though at least check that the bytes are properly JSON encoded.

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -79,20 +88,176 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+
+				suite.Require().NoError(err)
+				suite.Require().NotNil(res)
+
+				suite.Require().Equal(expClientState, clientState)
+
+				// Verify events
+				expectedEvents := sdk.Events{
+					sdk.NewEvent(
+						types.EventTypeMigrateContract,
```
crodriguezvega, line 243: For the reason mentioned above I would use directly the string...

```suggestion
						"migrate_contract",
```

chatton (commented):
Looking great! Really nice test coverage thank you 🥇

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -79,20 +88,176 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+			newCodeHash = storeWasmCode(suite, newByteCode)
+
+			endpoint := wasmtesting.NewWasmEndpoint(suite.chainA)
+			err := endpoint.CreateClient()
+			suite.Require().NoError(err)
+
+			oldClientState, ok := endpoint.GetClientState().(*types.ClientState)
+			suite.Require().True(ok)
+
+			expClientState = oldClientState
```
crodriguezvega, line 220: Since `oldClientState` is a pointer and you assign it to `expClientState`, then when modifying `CodeHash` below, you modify it in `oldClientState`. It's not a big deal, because `oldClientState` is not used, but if in the future it becomes used, it could cause some head scratching for a different developer. So why not something like this?

```golang
expClientState := func() *types.ClientState {
  cs, ok := endpoint.GetClientState().(*types.ClientState)
  suite.Require().True(ok)
  cs.CodeHash = newCodeHash
  return cs
}()
```

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -79,20 +88,176 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+			nil,
+		},
+		{
+			"success: update state",
+			func() {
+				msg = types.NewMsgMigrateContract(govAcc, defaultWasmClientID, newCodeHash, []byte("{}"))
+
+				suite.mockVM.MigrateFn = func(_ wasmvm.Checksum, _ wasmvmtypes.Env, _ []byte, store wasmvm.KVStore, _ wasmvm.GoAPI, _ wasmvm.Querier, _ wasmvm.GasMeter, _ uint64, _ wasmvmtypes.UFraction) (*wasmvmtypes.Response, uint64, error) {
+					// the code hash written in here will be overwritten
+					expClientState = types.NewClientState([]byte{1}, []byte(ibctesting.InvalidID), clienttypes.NewHeight(2000, 2))
```
crodriguezvega, line 151: I personally found it a bit misleading that you are using `InvalidID` here as the code hash and then realised that you are changing it to the actually elected code hash...

modules/light-clients/08-wasm/types/migrate_contract_test.go
crodriguezvega: +1

modules/light-clients/08-wasm/types/migrate_contract_test.go
```diff
@@ -0,0 +1,140 @@
...
+	}{
+		{
+			"success: new and old code hash are different",
+			func() {
+				newHash = sha256.Sum256([]byte{1, 2, 3})
+				err := ibcwasm.CodeHashes.Set(suite.chainA.GetContext(), newHash[:])
+				suite.Require().NoError(err)
+
+				payload = []byte{1}
+				expChecksum := wasmvmtypes.ForceNewChecksum(hex.EncodeToString(newHash[:]))
```
crodriguezvega, line 39: Why do we need this? Cannot we use `newHash` directly in line 42?

modules/light-clients/08-wasm/types/migrate_contract_test.go
```diff
@@ -0,0 +1,140 @@
...
+		expClientState *types.ClientState
+	)
+
+	testCases := []struct {
+		name     string
+		malleate func()
+		expErr   error
+	}{
+		{
+			"success: new and old code hash are different",
```
crodriguezvega, line 32: I think this helps me to understand at a quick glance what the difference between this test and the one below is, because in both cases, for the test to succeed, the code hashes need to be different.

```suggestion
			"success: no update client state",
```

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -79,20 +88,176 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+					data, err := json.Marshal(types.EmptyResult{})
+					suite.Require().NoError(err)
+
+					return &wasmvmtypes.Response{Data: data}, wasmtesting.DefaultGasUsed, nil
+				}
+			},
+			nil,
+		},
+		{
+			"success: update state",
```
crodriguezvega, line 145:
```suggestion
			"success: update client state",
```

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -79,20 +88,176 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
+		expClientState *types.ClientState
+	)
+
+	testCases := []struct {
+		name     string
+		malleate func()
+		expError error
+	}{
+		{
+			"success",
```
crodriguezvega, line 131: ditto

```suggestion
			"success: no update client state",
```

crodriguezvega (commented):
Left a few more comments. Will sleep over it and approve tomorrow!

modules/light-clients/08-wasm/keeper/keeper.go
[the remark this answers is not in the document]
colin-axner: https://github.com/cosmos/ibc-go/issues/5011

modules/light-clients/08-wasm/keeper/keeper_test.go
```diff
@@ -61,6 +73,54 @@ func (suite *KeeperTestSuite) SetupTest() {
...
+	db := dbm.NewMemDB()
+	app := simapp.NewSimApp(log.NewNopLogger(), db, nil, true, simtestutil.EmptyAppOptions{}, suite.mockVM)
+
+	// reset DefaultTestingAppInit to its original value
+	ibctesting.DefaultTestingAppInit = setupTestingApp
+	return app, app.DefaultGenesis()
+}
+
+// storeWasmCode stores the wasm code on chain and returns the code hash.
+func storeWasmCode(suite *KeeperTestSuite, wasmCode []byte) []byte {
```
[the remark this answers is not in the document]
colin-axner: Don't think there is a reason we cannot, but I think this was copied from [here](https://github.com/cosmos/ibc-go/blob/feat/wasm-clients/modules/light-clients/08-wasm/types/types_test.go#L190), so maybe we can add to a list of post-merge testing improvements?

modules/light-clients/08-wasm/keeper/msg_server.go
```diff
@@ -31,3 +31,21 @@ func (k Keeper) StoreCode(goCtx context.Context, msg *types.MsgStoreCode) (*type
...
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	err := k.migrateContractCode(ctx, msg.ClientId, msg.NewCodeHash, msg.Msg)
+	if err != nil {
+		return nil, errorsmod.Wrap(err, "failed to migrate contract")
+	}
+
+	// event emission is handled in migrateContractCode
```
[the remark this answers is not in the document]
damiannolan: was also going to make this comment but I think it doesn't really make a difference. We have access to the same args here as within the `migrateContractCode` function. 

I'd vote for here, its nice to have them at the bottom of the msg server handler.

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -79,20 +88,176 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
...
 				suite.Require().Empty(events)
 			}
 		})
 	}
 }
+
+func (suite *KeeperTestSuite) TestMsgMigrateContract() {
+	oldCodeHash := sha256.Sum256(wasmtesting.Code)
+
+	newByteCode := []byte("MockByteCode-TestMsgMigrateContract")
```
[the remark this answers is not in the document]
colin-axner: Yes, but maybe we can add this to a followup issue?

colin-axner (approved):
NICE!!! Wahoo! Really cool feature :)

Leaving my approval. I think there are some good suggestions from the others left on this pr, but I don't have any to add

modules/light-clients/08-wasm/keeper/msg_server_test.go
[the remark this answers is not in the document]
damiannolan: Sounds good to me! 👍

modules/light-clients/08-wasm/keeper/msg_server_test.go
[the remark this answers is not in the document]
crodriguezvega: It's such a small change, that if @srdtrk agrees, I am happy to push it myself.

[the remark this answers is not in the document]
srdtrk: There is already mock code there `wasmtesting.Code`. This mock byte code is just to test upgrades and no other use case. I'd suggest we don't.

proto/ibc/lightclients/wasm/v1/tx.proto
```diff
@@ -27,3 +30,19 @@ message MsgStoreCodeResponse {
...
+
+// MsgMigrateContract defines the request type for the MigrateContract rpc.
+message MsgMigrateContract {
+  option (cosmos.msg.v1.signer) = "signer";
+
+  string signer = 1;
+  // the client id of the contract
+  string client_id = 2;
+  // the code hash of the new wasm byte code for the contract
+  bytes new_code_hash = 3;
```
crodriguezvega, line 42: I would just call this `code_hash` and skip `new`, just like [they do in `x/wasm`](https://github.com/CosmWasm/wasmd/blob/main/proto/cosmwasm/wasm/v1/tx.proto#L215).

srdtrk: I initially called it that but changed based on @colin-axner 's [suggestion here](https://github.com/cosmos/ibc-go/pull/5000#discussion_r1378669084).

modules/light-clients/08-wasm/types/validation_test.go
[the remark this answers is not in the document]
srdtrk: Yeah. `ErrorIs` apparently only checks the error type. For example
- `errorsmod.Wrap(types.ErrInvalidCodeHash, "code hash cannot be empty")`
- `errorsmod.Wrapf(types.ErrInvalidCodeHash, "expected length of 32 bytes, got %d", 1)`

Will both pass `ErrorIs` for the same error. But won't pass `ErrorContains`

modules/light-clients/08-wasm/internal/ibcwasm/expected_interfaces.go
```diff
@@ -51,6 +51,24 @@ type WasmEngine interface {
 		deserCost wasmvmtypes.UFraction,
 	) ([]byte, uint64, error)
 
+	// Migrate migrates an existing contract to a new code binary.
+	// This takes storage of the data from the original contract and the code hash of the new contract that should
+	// replace it. This allows it to run a migration step if needed, or return an error if unable to migrate
+	// the given data.
+	//
+	// MigrateMsg has some data on how to perform the migration.
```
charleenfei, line 59:
```suggestion
	// MigrateContract has some data on how to perform the migration.
```

is that what was intended?

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: Yes but should I change it?

modules/light-clients/08-wasm/types/migrate_contract_test.go
srdtrk: Because wasmvm wraps it in this way

modules/light-clients/08-wasm/internal/ibcwasm/expected_interfaces.go
srdtrk: yeah intended. It is talking about the payload

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: done

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: done

modules/light-clients/08-wasm/types/migrate_contract_test.go
srdtrk: done

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: `oldClientState` was unused. So I removed it

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: done

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -79,20 +88,176 @@ func (suite *KeeperTestSuite) TestMsgStoreCode() {
 				// Verify events
 				expectedEvents := sdk.Events{
 					sdk.NewEvent(
-						"store_wasm_code",
```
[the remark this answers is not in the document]
srdtrk: reverted

modules/light-clients/08-wasm/types/migrate_contract_test.go
srdtrk: done

modules/light-clients/08-wasm/types/validation_test.go
srdtrk: I don't think it is needed in this case as there are only 3 test cases. And it is good to see the clientID in each case since this is the only thing we are setting

modules/light-clients/08-wasm/keeper/keeper.go
srdtrk: done

proto/ibc/lightclients/wasm/v1/tx.proto
```diff
@@ -27,3 +30,19 @@ message MsgStoreCodeResponse {
   // the sha256 hash of the stored code
   bytes checksum = 1;
 }
+
+// MsgMigrateContract defines the request type for the MigrateContract rpc.
+message MsgMigrateContract {
+  option (cosmos.msg.v1.signer) = "signer";
+
+  string signer = 1;
```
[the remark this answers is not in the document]
srdtrk: done

proto/ibc/lightclients/wasm/v1/tx.proto
```diff
@@ -27,3 +30,19 @@ message MsgStoreCodeResponse {
...
+  // the client id of the contract
+  string client_id = 2;
+  // the code hash of the new wasm byte code for the contract
+  bytes new_code_hash = 3;
+  // the migrate msg to be passed to the contract
+  bytes msg = 4;
```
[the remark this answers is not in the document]
srdtrk: done

modules/light-clients/08-wasm/types/msgs.go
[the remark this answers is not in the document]
srdtrk: removed

proto/ibc/lightclients/wasm/v1/tx.proto
crodriguezvega: I also prefer `code_hash`, but I am ok to keep `new_code_hash` if Colin feels strongly about it. Fine merging with either and changing it in a follow-up PR if needed.

modules/light-clients/08-wasm/keeper/msg_server_test.go
crodriguezvega: I would change it, just on the principle of least surprise :) If I was initially confused, I guess another engineer might as well be, and it's nice if we can avoid that. Happy to change this my self, if you want!

modules/light-clients/08-wasm/keeper/msg_server_test.go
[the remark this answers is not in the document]
crodriguezvega: I am fine leaving it as is and see how people usage before making a decision. What would be the reason then for `x/wasm` to say that the `msg` should be JSON-encoded?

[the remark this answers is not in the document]
damiannolan: > What would be the reason then for x/wasm to say that the msg should be JSON-encoded?

I'd imagine to at least have some kind of encoding standardisation. IMO just general good practice

crodriguezvega (approved):
Great work, @srdtrk. Thank you for addressing the comments.

proto/ibc/lightclients/wasm/v1/tx.proto
srdtrk: reverted back to `code_hash`. Consensus by @srdtrk, @crodriguezvega, @charleenfei

Issue #4951, opened 2023-10-25
Implement `RemoveCodeHash` RPC endpoint


srdtrk:
## Summary

Currently, there is no way for governance to prevent users from instantiating a wasm client from a previously stored code.

## Problem Definition

You might need to do this if the previously stored code is buggy. The `MigrateContract` RPC  endpoint proposed in https://github.com/cosmos/ibc-go/issues/4925 does not cover this use case as it would only migrate the light client with the given clientID,

## Proposal
1. Add a `RemoveCodeHash` function to `types/wasm.go`
```go
// RemoveCodeHash adds a removes code hash from the list of stored code hashes.
func RemoveCodeHash(ctx sdk.Context, cdc codec.BinaryCodec, codeHash []byte) error {
	codeHashes, err := GetCodeHashes(ctx, cdc)
	if err != nil {
		return err
	}

	// TODO: remove code hash here

	wasmStoreKey := ibcwasm.GetWasmStoreKey()
	store := ctx.KVStore(wasmStoreKey)
	bz, err := cdc.Marshal(&codeHashes)
	if err != nil {
		return err
	}

	store.Set([]byte(KeyCodeHashes), bz)

	return nil
}
```

2. Add a new RPC endpoint called `RemoveCodeHash` to call this the function above. Only authority should be able to call this.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Good idea, @srdtrk! I think this is nice to have, but I don't see it as a blocker for v0.1.0 release. I could see this added in a future release, since I don't expect we will encounter so soon the situation that we need to migrate contracts and we should remove the code hash of a buggy contract. What do you think?

srdtrk:
Agreed. I think given that it is a relatively small addition. I think I can implement this easily before we mark `v0.1.0`. I'll mark it as a nice to have then, and see if we have time to implement it before `v0.1`

crodriguezvega:
Closed by #5006

PR #5006, opened 2023-11-02
feat(08-wasm): implement 'RemoveCodeHash' RPC endpoint

srdtrk:
## Description

closes: #4951

```text
feat(08-wasm): implement 'RemoveCodeHash' RPC endpoint
```

---

write a little note why.


modules/light-clients/08-wasm/keeper/msg_server.go
```diff
@@ -31,3 +32,19 @@ func (k Keeper) StoreCode(goCtx context.Context, msg *types.MsgStoreCode) (*type
...
 	}, nil
 }
+
+// RemoveCodeHash defines a rpc handler method for MsgRemoveCodeHash
+func (k Keeper) RemoveCodeHash(goCtx context.Context, msg *types.MsgRemoveCodeHash) (*types.MsgRemoveCodeHashResponse, error) {
+	if k.GetAuthority() != msg.Signer {
+		return nil, errorsmod.Wrapf(ibcerrors.ErrUnauthorized, "expected %s, got %s", k.GetAuthority(), msg.Signer)
+	}
+
+	found := types.HasCodeHash(goCtx, msg.CodeHash)
```
crodriguezvega, line 42: I though `HasCodeHash` takes an SDK context?

```suggestion
	if !types.HasCodeHash(goCtx, msg.CodeHash) {
		return nil, errorsmod.Wrap(types.ErrWasmCodeHashNotFound, msg.CodeHash)
	}
```

proto/ibc/lightclients/wasm/v1/tx.proto
```diff
@@ -27,3 +30,18 @@ message MsgStoreCodeResponse {
...
+
+  string signer = 1;
+  // code hash to be removed from the store
+  bytes code_hash = 2;
+}
+
+// MsgStoreCodeResponse defines the response type for the StoreCode rpc
+message MsgRemoveCodeHashResponse {
+  // true if the code hash was found in the store
+  bool found = 1;
```
crodriguezvega, line 46: What would be the benefit of signalling here if the code hash was not found instead of erroring in the RPC handler?

crodriguezvega (commented):
Had a quick look!

modules/light-clients/08-wasm/keeper/msg_server.go
srdtrk: I modified it. Collections take goCtx. SDK is planning to get rid of `sdk.Ctx` and use `context.Context` instead

proto/ibc/lightclients/wasm/v1/tx.proto
srdtrk: I think the goal of remove is to ensure the item is not on the list. I don't think it constitutes an error if the item was not in there, or reversion tx.

crodriguezvega: I see your point, but I could also argue that for the `MigrateContract` RPC if the code hash is the same instead of error, we could also just return with a successful response, since the outcome of the operation is the same as what the user asked for, which was to migrate to that other code hash (that turns out to be the same one that it already had).

modules/light-clients/08-wasm/types/msgs.go
```diff
@@ -37,6 +39,30 @@ func (m MsgStoreCode) ValidateBasic() error {
...
+//
+//nolint:interfacer
+func NewMsgRemoveCodeHash(signer string, codeHash []byte) *MsgRemoveCodeHash {
+	return &MsgRemoveCodeHash{
+		Signer:   signer,
+		CodeHash: codeHash,
+	}
+}
+
+// ValidateBasic implements sdk.Msg
```
DimitrisJim, line 52: I guess it doesn't hurt to tweak this to `HasValidateBasic`.

modules/light-clients/08-wasm/types/msgs_test.go
```diff
@@ -153,7 +153,52 @@ func TestMsgMigrateContractValidateBasic(t *testing.T) {
...
+			types.NewMsgRemoveCodeHash(ibctesting.InvalidID, codeHash[:]),
+			ibcerrors.ErrInvalidAddress,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+
+		err := tc.msg.ValidateBasic()
+		expPass := tc.expErr == nil
```
DimitrisJim, line 197: taking Damians nits and proliferating them, breathing space

```suggestion
		err := tc.msg.ValidateBasic()

		expPass := tc.expErr == nil
```

proto/ibc/lightclients/wasm/v1/tx.proto
DimitrisJim: yup, I'd agree with Carlos from POV it should be similar to what we do for migrate.

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -278,3 +279,102 @@ func (suite *KeeperTestSuite) TestMsgMigrateContract() {
...
+
+				codeHashes, err := types.GetAllCodeHashes(suite.chainA.GetContext())
+				suite.Require().NoError(err)
+
+				// Check equality of code hashes up to order
+				suite.Require().Subset(expCodeHashes, codeHashes)
+				suite.Require().Subset(codeHashes, expCodeHashes)
+
+				// Verify events
+				suite.Require().Len(events, 0)
```
DimitrisJim, line 373: I wouldn't add this necessarily but if we do we can atleast move getting the events in the successful case.

proto/ibc/lightclients/wasm/v1/tx.proto
srdtrk: I think part of the outcome of the `MigrateContract` is the calling of the migrate entry point in the new code. Since returning early wouldn't do that, I wanted to return an error. I could see a case for it if we called the migrate entry point in the new code.

DimitrisJim: Carlo's comment is w.r.t returning nil or err in migrate contract when new code hash is equal to old. In both cases we return early, unless your point just flew over my head.

modules/light-clients/08-wasm/keeper/msg_server.go
```diff
@@ -32,6 +33,22 @@ func (k Keeper) StoreCode(goCtx context.Context, msg *types.MsgStoreCode) (*type
...
 
+// RemoveCodeHash defines a rpc handler method for MsgRemoveCodeHash
+func (k Keeper) RemoveCodeHash(goCtx context.Context, msg *types.MsgRemoveCodeHash) (*types.MsgRemoveCodeHashResponse, error) {
+	if k.GetAuthority() != msg.Signer {
+		return nil, errorsmod.Wrapf(ibcerrors.ErrUnauthorized, "expected %s, got %s", k.GetAuthority(), msg.Signer)
+	}
+
+	found := types.HasCodeHash(goCtx, msg.CodeHash)
+
+	err := ibcwasm.CodeHashes.Remove(goCtx, msg.CodeHash)
```
chatton, line 44: looking at the implementation of `Remove` it appears errors if the key is not present, this should mean that we will never be able to return a `found` value of `false`. We would instead error out.

I think we can instead do an early return of `&types.MsgRemoveCodeHashResponse{Found: false}`

srdtrk: It doesn't return an error if the key is not present.

modules/light-clients/08-wasm/keeper/msg_server.go
```diff
@@ -32,6 +33,22 @@ func (k Keeper) StoreCode(goCtx context.Context, msg *types.MsgStoreCode) (*type
...
+	if err != nil {
+		return nil, errorsmod.Wrap(err, "failed to remove code hash")
```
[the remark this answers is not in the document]
srdtrk: I don't know how I can hit this case. I could maybe get a deserialization error if we weren't using bytes. Even nil didn't hit this case!

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: You think we should emit events?

modules/light-clients/08-wasm/types/msgs_test.go
srdtrk: improved

modules/light-clients/08-wasm/types/msgs_test.go
```diff
@@ -52,7 +52,7 @@ func TestMsgStoreCodeValidateBasic(t *testing.T) {
 		err := tc.msg.ValidateBasic()
 		expPass := tc.expErr == nil
 		if expPass {
-			require.NoError(t, err)
+			require.NoError(t, err, tc.name)
```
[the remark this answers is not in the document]
srdtrk: removed

modules/light-clients/08-wasm/types/msgs.go
srdtrk: done

modules/light-clients/08-wasm/types/msgs.go
```diff
@@ -37,6 +39,30 @@ func (m MsgStoreCode) ValidateBasic() error {
 	return nil
 }
 
+// NewMsgRemoveCodeHash creates a new MsgRemoveCodeHash instance
+//
+//nolint:interfacer
```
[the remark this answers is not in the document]
srdtrk: done

proto/ibc/lightclients/wasm/v1/tx.proto
srdtrk: Ok. If one more person agrees I'll change it XD. It is worth noting that in most APIs I've seen, `Remove` functions don't return an error if the key is not present, including keyset's remove api, and the KVStore's. But I think this might make sense in a cosmos tx I suppose.

damiannolan: I'd suggest removing the bool from the response. It seems unnecessary, if it's requested by users then we can add it later with no overhead.

modules/light-clients/08-wasm/keeper/msg_server.go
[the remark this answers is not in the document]
DimitrisJim: aaaah yea, I remember this was what was perplexing me when `MustUnmarshal` was changed due to fear of panicking, I couldn't get it to panic. Ignore this!

modules/light-clients/08-wasm/keeper/msg_server.go
damiannolan: isn't @crodriguezvega's suggestion still relevant here? I think an error should be returned `if !types.HasCodeHash()`

modules/light-clients/08-wasm/keeper/msg_server.go
damiannolan, line 44: This doesn't remove the code from the vm cache, should we also be calling https://github.com/CosmWasm/wasmvm/blob/main/lib.go#L69

srdtrk: I don't think we should remove it. Wouldn't this break already existing light clients using the removed code?

damiannolan: Yeah, I think you're right it would, good catch! As you could no longer instantiate the contract to call into right? You'd be kind of in no man's land. Maybe that could also be a good thing if this was being used as some sort of killswitch, but once its removed I don't see much of a path for recovery - probably upgrade the channel to a new connection to recover funds would be the only option. 

Happy to keep it as you have 👍

proto/ibc/lightclients/wasm/v1/tx.proto
srdtrk: Done. I have removed found, and now returning an error instead if not found

modules/light-clients/08-wasm/keeper/msg_server_test.go
damiannolan: is there actually 0 events emitted? hmm, would've thought sdk baseapp would emit some std ones

modules/light-clients/08-wasm/keeper/msg_server.go
chatton: ah looks like it would only error on a nil key, not a not found one 👍

srdtrk: Didn't even error on `nil` when I tried

modules/light-clients/08-wasm/keeper/msg_server_test.go
srdtrk: The context is not going through the baseapp. I'm directly calling the msg_server handler with the context

damiannolan: makes sense!

DimitrisJim: not sure re events. I dont think they'd hurt tho.

modules/light-clients/08-wasm/types/codec.go
```diff
@@ -27,6 +27,14 @@ func RegisterInterfaces(registry codectypes.InterfaceRegistry) {
 		(*sdk.Msg)(nil),
 		&MsgStoreCode{},
 	)
+	registry.RegisterImplementations(
```
DimitrisJim, line 30: sweet, remember to add test cases for the registraction in `codec_test`.

damiannolan (approved):
Looks clean to me. Just left one additional comment on test helper usage. 

Thank you! 🙏

modules/light-clients/08-wasm/keeper/msg_server_test.go
```diff
@@ -278,3 +279,95 @@ func (suite *KeeperTestSuite) TestMsgMigrateContract() {
...
+			if tc.expError == nil {
+				suite.Require().NoError(err)
+				suite.Require().NotNil(res)
+
+				codeHashes, err := types.GetAllCodeHashes(suite.chainA.GetContext())
+				suite.Require().NoError(err)
+
+				// Check equality of code hashes up to order
+				suite.Require().Subset(expCodeHashes, codeHashes)
+				suite.Require().Subset(codeHashes, expCodeHashes)
```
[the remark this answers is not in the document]
srdtrk: Lol makes sense. I just didn't know about elements match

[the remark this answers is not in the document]
DimitrisJim: `Equal` would also do, depends if we care for ordering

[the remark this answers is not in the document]
srdtrk: Ordering is the problem, that's why equal wouldn't work. I think keyset orders them randomly (from our perspective), not in the order of insertion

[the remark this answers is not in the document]
DimitrisJim: I would assume these should be ascending considering its an IAVL? Either way, not too nitpicky about this, ElementsMatch works perfect!

DimitrisJim (approved):
pending comment on testing codec's registered, lgtm!

modules/light-clients/08-wasm/types/codec.go
srdtrk: done

PR #5007, opened 2023-11-02
Sync gas_register.go with wasmd main

chatton:
## Description

Our `gas_register.go` was created pre sdk 0.50 and is out of sync with wasmd's.

This PR does the following:

- Do a direct copy paste of [gas_register.go](https://github.com/CosmWasm/wasmd/blob/main/x/wasm/types/gas_register.go) and paste it into ibc-go.
- Pull the additional constructs we had been making use of into `gas_register_custom.go`. Note, these seem to be primarily functions that exist as keeper functions in wasmd but have been refactored slightly to remove the dependency on the keeper. E.g. [this](https://github.com/CosmWasm/wasmd/blob/e65480838a1ded147ef53d35fa3bd9709a61226f/x/wasm/keeper/keeper.go#L1100) keeper fn now looks like [this](https://github.com/cosmos/ibc-go/blob/28f9ad7ef224242b389414f56eca4b809a0a099b/modules/light-clients/08-wasm/types/gas_register_custom.go#L27).
- I explicitly opted to copy more than we need, I would prefer that if we are copying a file, we may as well copy the entire thing, to make it easier to see what has changed when syncing in the future.
- the re-addition of the interface functions as is, introduce a new field `pinned`. For now I have passed `false` in, as this results in technically the same code path as before.

---

write a little note why.

### DIFF
--- a/modules/light-clients/08-wasm/types/gas_register.go
+++ b/modules/light-clients/08-wasm/types/gas_register.go
@@ -1,21 +1,15 @@
 package types
 
 import (
-	"math"
-
-	wasmvm "github.com/CosmWasm/wasmvm"
 	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
 
 	errorsmod "cosmossdk.io/errors"
+	sdkmath "cosmossdk.io/math"
 	storetypes "cosmossdk.io/store/types"
 
-	sdk "github.com/cosmos/cosmos-sdk/types"
-
-	ibcerrors "github.com/cosmos/ibc-go/v8/modules/core/errors"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-// Copied subset of gas features from wasmd
-// https://github.com/CosmWasm/wasmd/blob/v0.31.0/x/wasm/keeper/gas_register.go
 const (
 	// DefaultGasMultiplier is how many CosmWasm gas points = 1 Cosmos SDK gas point.
 	//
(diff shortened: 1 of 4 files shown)


modules/light-clients/08-wasm/types/vm.go
```diff
@@ -35,7 +35,7 @@ func initContract(ctx sdk.Context, clientStore storetypes.KVStore, codeHash []by
 		Funds:  nil,
 	}
 
-	ctx.GasMeter().ConsumeGas(VMGasRegister.NewContractInstanceCosts(len(msg)), "Loading CosmWasm module: instantiate")
+	ctx.GasMeter().ConsumeGas(VMGasRegister.NewContractInstanceCosts(false, len(msg)), "Loading CosmWasm module: instantiate")
```
DimitrisJim, line 38: I think this should be `true` considering we `Pin`? wasmd does conditionally pass this based on `IsPinnedCode` but we do it unconditionally.

DimitrisJim: we do save some gas from it apparently:

```
       dataCosts := storetypes.Gas(msgLen) * g.c.ContractMessageDataCost
	if pinned {
		return dataCosts
	}
	return g.c.InstanceCost + dataCosts
```

would be nice to ask Simon just for the verification.

chatton: I think you're right, it probably should be true across the board. This will result in different gas costs. E.g. `dataCosts` instead of `g.c.InstanceCost + dataCosts` (which I believe is actually correct)

Will update this!

DimitrisJim (approved):
lgtm, would add note in `custom_gas` that these are also technically copied but placed there due to our keeper issues.

modules/light-clients/08-wasm/types/gas_register.go
```diff
@@ -1,21 +1,15 @@
...
 	storetypes "cosmossdk.io/store/types"
 
-	sdk "github.com/cosmos/cosmos-sdk/types"
-
-	ibcerrors "github.com/cosmos/ibc-go/v8/modules/core/errors"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
-// Copied subset of gas features from wasmd
-// https://github.com/CosmWasm/wasmd/blob/v0.31.0/x/wasm/keeper/gas_register.go
```
colin-axner, line 18: should we leave this with updated link?

colin-axner (approved):
Nice work @chatton :raised_hands:

modules/light-clients/08-wasm/types/gas_register.go
colin-axner: also happy to just put this in a doc in the types directory if you want to keep the file line for line the same

chatton: We have a link to the `main` branch in `gas_register_custom.go`, do you think that is enough?

chatton:
merged with just the link in gas_register_custom.go, we can create an issue if we want to add more details.

PR #5154, opened 2023-11-21
imp: use bytes in wasm contract api instead of wrapped types

crodriguezvega:
## Description

Changes for:

- [x] `VerifyClientMessageMsg`
- [x] `InstantiateMessage`
- [x] `VerifyUpgradeAndUpdateStateMsg`
- [x] `CheckForMisbehaviourMsg`
- [x] `UpdateStateMsg`
- [x] `UpdateStateOnMisbehaviourMsg`

Contract changes [here](https://github.com/misko9/centauri/pull/7).

closes: #XXXX

```text
type: commit message
```

---

write a little note why.


modules/light-clients/08-wasm/types/contract_api.go
```diff
@@ -7,8 +7,9 @@ import (
 
 // InstantiateMessage is the message that is sent to the contract's instantiate entry point.
 type InstantiateMessage struct {
-	ClientState    *ClientState    `json:"client_state"`
-	ConsensusState *ConsensusState `json:"consensus_state"`
+	ClientState    []byte `json:"client_state"`
+	ConsensusState []byte `json:"consensus_state"`
+	Checksum       []byte `json:"checksum"`
```
crodriguezvega, line 12: I left the checksum here for now.

modules/light-clients/08-wasm/types/consensus_state.go
```diff
@@ -9,7 +9,7 @@ import (
 var _ exported.ConsensusState = (*ConsensusState)(nil)
 
 // NewConsensusState creates a new ConsensusState instance.
-func NewConsensusState(data []byte, timestamp uint64) *ConsensusState {
```
crodriguezvega, line 12: I removed the timestamp parameter because it's not used.

modules/light-clients/08-wasm/types/client_state.go
```diff
@@ -121,8 +121,9 @@ func (cs ClientState) Initialize(ctx sdk.Context, cdc codec.BinaryCodec, clientS
 	}
 
 	payload := InstantiateMessage{
-		ClientState:    &cs,
-		ConsensusState: consensusState,
+		ClientState:    cs.Data,
```
srdtrk, line 124: If we do this, how would the contract verify the initial latest height?

We could also use `clienttypes.MarshalClientState` in order to address this issue and pass the entire client state. I'd be in favour of this because in the previous api, we were passing the entire client state.

modules/light-clients/08-wasm/types/upgrade.go
```diff
@@ -36,8 +36,8 @@ func (cs ClientState) VerifyUpgradeAndUpdateState(
 
 	payload := SudoMsg{
 		VerifyUpgradeAndUpdateState: &VerifyUpgradeAndUpdateStateMsg{
-			UpgradeClientState:         *wasmUpgradeClientState,
-			UpgradeConsensusState:      *wasmUpgradeConsState,
+			UpgradeClientState:         wasmUpgradeClientState.Data,
+			UpgradeConsensusState:      wasmUpgradeConsState.Data,
```
srdtrk, line 40: I think we must pass the entire client state here not ` wasmUpgradeClientState.Data`. Because it carries the latest height. And this is a must have information.

modules/light-clients/08-wasm/types/contract_api.go
srdtrk: What about latest height. Also, I don't see why contracts would need their own checksum. I think we should remove this.

crodriguezvega: I didn't add it because the contract takes the [latest height from the underlying client state (the bytes that we are passing here) and uses that to set the latest height of the 08-wasm wrapping type](https://github.com/misko9/centauri/blob/master/light-clients/ics10-grandpa-cw/src/client.rs#L306).

Happy to remove the checksum if you do me the favour to figure out how the contract can calculate that itself. I don't think I will have the time to investigate that. 🙏

modules/light-clients/08-wasm/types/client_state.go
crodriguezvega: Looks like the contract doesn't verify the latest height that we pass at the moment, and it actually just [overrides it with the latest height in from the underlying client state](https://github.com/misko9/centauri/blob/master/light-clients/ics10-grandpa-cw/src/client.rs#L306). I believe the latest height field in the 08-wasm wrapping type is just there as a convenience, so that when you query for client states you can easily see what the latest height the chain has for that particular 08-wasm client.

modules/light-clients/08-wasm/types/upgrade.go
srdtrk: This doesn't make sense to me. The new client state must overwrite the old one. Including the latest height right? [We do this in tendermint client](https://github.com/cosmos/ibc-go/blob/c54ec070d0c4fb7e6b923d2c1d475fd2e6011254/modules/light-clients/07-tendermint/upgrade.go#L100).

And implicitly, isn't it used [here](https://github.com/misko9/centauri/blob/4a9a3313e1f925ebef5e6c4590b87cbf621679fc/light-clients/ics10-grandpa/src/client_def.rs#L419) in the grandpa client?

modules/light-clients/08-wasm/types/client_state.go
srdtrk: See my [other comment](https://github.com/cosmos/ibc-go/pull/5154#discussion_r1401805306)

modules/light-clients/08-wasm/types/upgrade.go
crodriguezvega: > And implicitly, isn't it used [here](https://github.com/misko9/centauri/blob/4a9a3313e1f925ebef5e6c4590b87cbf621679fc/light-clients/ics10-grandpa/src/client_def.rs#L419) in the grandpa client?

The `upgraded_client_state` that is used in this line is actually the underlying client state (so not the 08-wasm wrapping client state). You can see that it is retrieved [here](https://github.com/misko9/centauri/blob/4a9a3313e1f925ebef5e6c4590b87cbf621679fc/light-clients/ics10-grandpa-cw/src/contract.rs#L316) (the [`try_from` function will unwrap the underlying client state and return it](https://github.com/misko9/centauri/blob/4a9a3313e1f925ebef5e6c4590b87cbf621679fc/light-clients/ics10-grandpa-cw/src/msg.rs#L375)).

srdtrk: I see. But is this what they should be doing? Then why is the 07-tendermint client written this way?

colin-axner: Contracts should be writing a 08-wasm client with the updated latest height, but the latest height can absolutely be obtained from `wasmUpgradeClientState.Data` (it will be in the underlying client state)

modules/light-clients/08-wasm/types/contract_api.go
srdtrk: Happy to keep. Contracts can only get this through a querier which we are passing `nil`. I think this was the only way to get it for standard contracts. I still don't see the reason why the contract should be aware of its own codehash.

Also it can get its own code hash during other executions from the clientStore right? Eitherway. Happy to keep.

modules/light-clients/08-wasm/types/upgrade.go
srdtrk: Ah, this makes sense. It's good then

modules/light-clients/08-wasm/types/upgrade_test.go
```diff
@@ -101,7 +103,7 @@ func (suite *TypesTestSuite) TestVerifyClientMessage() {
 			clientState := endpoint.GetClientState()
 
 			clientMsg = &types.ClientMessage{
-				Data: []byte{1},
```
crodriguezvega, line 104: I could have left this as it was, but I like using tests as some sort of documentation as well, and thus being explicit that we expect a header here it's more helpful for the reader, I think.

modules/light-clients/08-wasm/testing/values.go
```diff
@@ -26,9 +32,24 @@ var (
 	ErrMockVM       = errors.New("mock vm error")
 )
 
+// CreateMockWrappedClientState returns valid wrapped client state for use in tests.
+func CreateMockWrappedClientState(height clienttypes.Height) *ibctm.ClientState {
```
srdtrk, line 36:
```suggestion
// CreateMockTmClientState returns valid wrapped client state for use in tests.
func CreateMockTmClientState(height clienttypes.Height) *ibctm.ClientState {
```

I understand the intention here. But the naming of these functions are confusing imo.

modules/light-clients/08-wasm/testing/wasm_endpoint.go
```diff
@@ -29,8 +29,11 @@ func (endpoint *WasmEndpoint) CreateClient() error {
 	checksum, err := types.CreateChecksum(Code)
 	require.NoError(endpoint.Chain.TB, err)
 
-	clientState := types.NewClientState(contractClientState, checksum, clienttypes.NewHeight(0, 1))
-	consensusState := types.NewConsensusState(contractConsensusState)
+	wrappedClientStateBz := clienttypes.MustMarshalClientState(endpoint.Chain.App.AppCodec(), CreateMockWrappedClientState(clienttypes.NewHeight(1, 5)))
```
srdtrk, line 32:
```suggestion
	wrappedClientStateBz := clienttypes.MustMarshalClientState(endpoint.Chain.App.AppCodec(), MockWrappedClientState)
```

Would this work?

modules/light-clients/08-wasm/types/contract_api.go
srdtrk: I think we should assume that the contract cannot get it's own codehash in instantiate if we remove this since we are not passing in a querier. But it can get it during other executions. I think the contract does not need it's own codeHash in instantiate and we could remove this. I'm in favour of removal but happy to hear what others think

damiannolan:
CI is now green 👍🏻 

I didn't modify the migrate contract files `migrate_success.wasm.gz` and `migrate_error.wasm.gz` but the tests are now passing. Is this expected @srdtrk?

crodriguezvega:
Note for myself: I need to update the docs.

damiannolan (approved):
LGTM, thanks everyone who contributed to this PR 🙏🏻

docs/docs/03-light-clients/04-wasm/07-contracts.md
```diff
@@ -15,12 +15,13 @@ The `08-wasm` light client proxy performs calls to the Wasm light client via the
 
 ## `InstantiateMessage`
 
-This is the message sent to the contract's `instantiate` entry point. It contains the client and consensus state provided in [`MsgCreateClient`](https://github.com/cosmos/ibc-go/blob/v7.2.0/proto/ibc/core/client/v1/tx.proto#L25-L37).
+This is the message sent to the contract's `instantiate` entry point. It contains the bytes of the protobuf-encoded client and consensus states of the underlying light client, both provided in [`MsgCreateClient`](https://github.com/cosmos/ibc-go/blob/v7.2.0/proto/ibc/core/client/v1/tx.proto#L25-L37).
```
damiannolan, line 18:
```suggestion
This is the message sent to the contract's `instantiate` entry point. It contains the bytes of the protobuf-encoded client and consensus states of the underlying light client, both provided in [`MsgCreateClient`](https://github.com/cosmos/ibc-go/blob/v7.2.0/proto/ibc/core/client/v1/tx.proto#L25-L37).
Please note that the bytes contained within the JSON message are represented as base64 encoded strings.
```

damiannolan (approved):
ACK docs, thanks @crodriguezvega 🙏🏻 

I left one suggestion for noting the base64 encoded byte strings

modules/light-clients/08-wasm/testing/wasm_endpoint.go
crodriguezvega: I wanted to use a different mock client state with a different height than `MockWrappedClientState` because there are some tests where the client state is updated with `MockWrappedClientState` and wanted to verify that the height actually changes.

modules/light-clients/08-wasm/testing/values.go
crodriguezvega: Renamed!

Issue #5145, opened 2023-11-21
Implement OnChanUpgradeAck on Controller Chain for interchain-accounts


chatton:

crodriguezvega:
Closed by #5160

PR #5160, opened 2023-11-22
Implement OnChanUpgradeAck on Controller Chain for interchain-accounts

chatton:
## Description

(this one is also r4r after `OnChanUpgradeTry` for host is merged)

Review after https://github.com/cosmos/ibc-go/issues/5140 and https://github.com/cosmos/ibc-go/issues/5142

closes: #5145

ref: [the spec](https://github.com/cosmos/ibc/blob/main/spec/app/ics-027-interchain-accounts/README.md#upgrade-handshake)

---

write a little note why.


modules/apps/27-interchain-accounts/types/metadata.go
```diff
@@ -120,13 +120,15 @@ func ValidateHostMetadata(ctx sdk.Context, channelKeeper ChannelKeeper, connecti
 		return errorsmod.Wrapf(ErrUnknownDataType, "unsupported transaction type %s", metadata.TxType)
 	}
 
-	connection, err := channelKeeper.GetConnection(ctx, connectionHops[0])
-	if err != nil {
-		return err
-	}
+	if len(connectionHops) > 0 {
```
damiannolan, line 123: `len(connectionHops)` should never zero in this case. 

Any reason this PR is changing `ValidateHostMetadata` when its not being used here? 

This would also make this implementation different form `ValidateControllerMetadata` above - I'd imagine they both should be consistent in what they're doing.

damiannolan (changes requested):
one requested change, otherwise lgtm

modules/apps/27-interchain-accounts/types/metadata.go
charleenfei: i'll revert for now -- can also simplify this file to have one `ValidateMetadata` function if they're doing the same thing then

damiannolan: they're not doing exactly the same thing, they pass different args to `validateConnectionParams` as the self/counterparty values will be different/opposing when called from either controller or host

damiannolan: we still need both functions at the moment unless a bigger refactor was done to accommodate it differently somehow, but I think we will always need to handle this one way or another

charleenfei: yes, i know, but can't we just add a boolean flag to the Validate function to indicate if the caller is a controller module or host?

charleenfei: i'll push the code in a sec so you can see what it looks like, if you don't like it we can revert and keep both functions.

damiannolan (approved):
Thanks @charleenfei for pushing this over the line! 🚀  And @chatton for getting this started!

damiannolan:
This PR will need type assertions on the `UpgradeableModule` interface to route callbacks after #5375. 

[See here for example](https://github.com/cosmos/ibc-go/pull/5375/files#diff-db995ef310109739e7fac5569466a6f1f9927a586ebcbdb981fa8844daca106f)

If this is merged first I can update my PR.

crodriguezvega (approved):
Looks great! Thank you again, @chatton and @charleenfei.

Issue #5168, opened 2023-11-24
Add attribute `header_hash` to update client event


crodriguezvega:
The [header has been removed from the update client event](https://github.com/cosmos/ibc-go/pull/5110), but we can add a new attribute with the header hash.

> @mattverse I checked with the relayers and at least the hermes relayer is depending at the moment on this attribute to perform misbehaviour detection, but they can change the relayer to fetch the header from the `/header` RPC endpoint. So we can merge this change, but we will not back port it and release it only in the next major release (v9).
> And adding a different attribute with the hash of the header is a good idea, so feel free to go ahead and add that. Thank you!

_Originally posted by @crodriguezvega in https://github.com/cosmos/ibc-go/issues/5110#issuecomment-1820489042_

cc @mattverse

crodriguezvega:
I will close the issue for now, since calculating the hash of a Tendermint header is not as simple as doing SHA 256 of the bytes. Also because we now support wasm clients and we cannot assume in 02-client that the header is of type Tendermint. It seems like relayers will be able to manage without this attribute, but we can revisit this decision in the future if needed.

PR #5172, opened 2023-11-26
feat: Add Attribute header_hash To Update Client Event

Pipello:
## Description

This PR adds the header_hash attribute to the update client evemt

closes: #5168 


crodriguezvega (commented):
Thank you, @Pipello! I pushed a couple of commits, one of them replacing `sha1` with `sha256`, since I think that's the hash algorithm we use in the rest of the codebase and I guess CometBFT header is also hashed with this algorithm. Having said this, I would be very nice if you could do a sanity check and run these changes locally to make sure that the hash emitted in the events matches the hash generated  by CometBFT and that can be used in the [RPC endpoint `header_by_hash`](https://docs.cometbft.com/main/rpc/#/Info/header_by_hash).

Pipello:
> I pushed a couple of commits, one of them replacing sha1 with sha256, since I think that's the hash algorithm we use in the rest of the codebase and I guess CometBFT header is also hashed with this algorithm.
Thank you @crodriguezvega :) 
> I would be very nice if you could do a sanity check and run these changes locally to make sure that the hash emitted in the events matches the hash generated by CometBFT and that can be used in the [RPC endpoint header_by_hash](https://docs.cometbft.com/main/rpc/#/Info/header_by_hash).
I will try to check that

Pipello:
@crodriguezvega I was able to run 1 test but not more, I am short in time to dedicate to open source work. If you think this single test is enough then that's good. Otherwise if you or somebody would like to be safer and run more tests no problems.

crodriguezvega:
> @crodriguezvega I was able to run 1 test but not more, I am short in time to dedicate to open source work. If you think this single test is enough then that's good. Otherwise if you or somebody would like to be safer and run more tests no problems.

Thank you, @Pipello! Just for confirmation: so you were able to run a chain and after the light client was updated, you could call `header_by_hash` with the hash in the event and retrieve the header?

crodriguezvega:
@Pipello I did a quick test and unfortunately the hash emitted in the event doesn't match the expected header hash. I also took a quick look at CometBFT's code and it seems [this is how the hash is calculated](https://github.com/cometbft/cometbft/blob/c88d5512350f6671c92051cf5ebbbe9841f644cb/types/block.go#L439C28) (so it's not just a sha 256 hash of the header bytes). Therefore I am not sure how useful it is to emit the sha 256 hash, since it will not be possible to use it to query CometBFT's RPC to retrieve the header...

crodriguezvega:
Thanks for the PR, @Pipello, but I think it's best for now to not add the attribute, since it's not as simple as we thought it would be to calculate the hash of the header that can be used with the `header_by_hash` RPC. Since it seems that relayers will be able to manage without the attribute, I think it's best just not spend time on this right now.

Issue #5218, opened 2023-11-29
In `ChanUpgradeTry` move fast forward sequence down below after the call to `ChanUpgradeInit`.


chatton:
In `ChanUpgradeTry` move [fast forward sequence](https://github.com/cosmos/ibc-go/blob/39df8d2b579fc623a1e748b4330738ff40adb682/modules/core/04-channel/keeper/upgrade.go#L118-L122) down below after the call to `ChanUpgradeInit`.

ThanhNhann:
plz assign me this ser @chatton

colin-axner:
closed by #5231

PR #5231, opened 2023-11-29
refactor: remove unnecessary decrement when fast forwarding upgrade sequence

ThanhNhann:
## Description

closes: #5218 
closes: #5299 

---

write a little note why.


modules/core/04-channel/keeper/upgrade.go
```diff
@@ -114,19 +114,19 @@ func (k Keeper) ChanUpgradeTry(
...
+		// NOTE: OnChanUpgradeInit will not be executed by the application
+		upgrade, err = k.ChanUpgradeInit(ctx, portID, channelID, proposedUpgradeFields)
+		if err != nil {
+			return types.Upgrade{}, errorsmod.Wrap(err, "failed to initialize upgrade")
+		}
+
 		// if the counterparty sequence is greater than the current sequence, we fast-forward to the counterparty sequence.
 		if counterpartyUpgradeSequence > channel.UpgradeSequence {
 			// using -1 as WriteUpgradeInitChannel increments the sequence
 			channel.UpgradeSequence = counterpartyUpgradeSequence - 1
```
AdityaSripal, line 126: Decrement should be removed here and a test that it still does the same behaviour

ThanhNhann: updated and it good

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -116,19 +116,17 @@ func (k Keeper) ChanUpgradeTry(
...
-
 		// NOTE: OnChanUpgradeInit will not be executed by the application
 		upgrade, err = k.ChanUpgradeInit(ctx, portID, channelID, proposedUpgradeFields)
 		if err != nil {
 			return types.Upgrade{}, errorsmod.Wrap(err, "failed to initialize upgrade")
 		}
 
+		// if the counterparty sequence is greater than the current sequence, we fast-forward to the counterparty sequence.
+		if counterpartyUpgradeSequence > channel.UpgradeSequence {
+			k.SetChannel(ctx, portID, channelID, channel)
```
charleenfei, line 127: i believe we still need to fastforward to the counterparty sequence, in case of the sequence being very much out of sync (timeouts).

ThanhNhann: cc @chatton

crodriguezvega: @charleenfei is right and I think you need to fast forward to the counterparty sequence before `SetChannel`. I think the reason the tests are passing at the moment is because we are missing a check to make sure that the upgrade sequences of both channel ends are in sync after `ChannelUpgradeTry` succeeds. If you add after [this line](https://github.com/cosmos/ibc-go/blob/63d7fc1eaec82c2af6f8ee4654abb0b5206d8f7f/modules/core/04-channel/keeper/upgrade_test.go#L304) a check like this `suite.Require().Equal(path.EndpointA.GetChannel().UpgradeSequence, path.EndpointB.GetChannel().UpgradeSequence)` you will see that [this test](https://github.com/cosmos/ibc-go/blob/63d7fc1eaec82c2af6f8ee4654abb0b5206d8f7f/modules/core/04-channel/keeper/upgrade_test.go#L170) will fail.

But I am thinking that probably we still need to do `channel.UpgradeSequence = counterpartyUpgradeSequence - 1` anyway, because `channel.UpgradeSequence` is actually incremented in `WriteUpgradeInitChannel` below... So if this is the case, not sure if it's worth moving the code from where it was originally...

crodriguezvega: Thank you, @ThanhNhann. To be honest, I don't know what would be best now: either leave the code where it was before or move it after `WriteUpgradeInitChannel` and then we probably don't need to do `channel.UpgradeSequence = counterpartyUpgradeSequence - 1` and we can do just `channel.UpgradeSequence = counterpartyUpgradeSequence`. @damiannolan any thoughts?

colin-axner: fixed

colin-axner (approved):
Thanks @ThanhNhann!

modules/core/04-channel/keeper/upgrade.go
damiannolan: Thanks @colin-axner

Issue #5226, opened 2023-11-29
Make sure that the error receipt we write always has a sequence greater than the existing one.


chatton:

crodriguezvega:
Closed by #5237

Issue #5207, opened 2023-11-29
Panic in `WriteErrorReceipt` instead of returning an error


crodriguezvega:
Panic in WriteErrorReceipt [instead of returning an error](https://github.com/cosmos/ibc-go/blob/39df8d2b579fc623a1e748b4330738ff40adb682/modules/core/04-channel/keeper/upgrade.go#L932), which is ignored.

hieuvubk:
I will take this one, thanks @crodriguezvega

PR #5237, opened 2023-11-29
Make sure that the error receipt we write always has a sequence greater than the existing one. 

chatton:
## Description

closes: #5226

~Will wait until #5207 is merged to mark as R4R as the tests will need a small refactor.~

---

write a little note why.


crodriguezvega (approved):
LGTM!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -944,3 +944,21 @@ func (k Keeper) restoreChannel(ctx sdk.Context, portID, channelID string, upgrad
 
 	return channel
 }
+
+// WriteErrorReceipt will write an error receipt from the provided UpgradeError.
+func (k Keeper) WriteErrorReceipt(ctx sdk.Context, portID, channelID string, upgradeError *types.UpgradeError) {
```
damiannolan, line 949: Question regarding this new method, is the intension to make `SetUpgradeErrorReceipt` private? seems a little off to have many functions exposed for this. 

I was looking at the cancellation handler and also thought that it is not super clear that an error receipt is set within the `restoreChannel` func above this method. It might be worth while opening another issue to see if we could improve that.

chatton: yeah I think we can probably make `WriteErrorReceipt` unexported and add a wrapper in `export_test.go`, and potentially just remove the writing of the error receipt from the restore fn, and just call it separately. the error receipt is kind of like a side effect of restoring the channel so I'd be happy with that, will create an issue for it.

PR #5262, opened 2023-11-30
fix: allow error receipt to be nil in case of the msg sender is the authority

charleenfei:
## Description

closes: #5260 

```text
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -560,20 +560,23 @@ func (k Keeper) ChanUpgradeCancel(ctx sdk.Context, portID, channelID string, err
 		return errorsmod.Wrapf(types.ErrUpgradeNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
 	}
 
-	if !isAuthority && errorReceipt.Sequence < channel.UpgradeSequence {
-		return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be greater than or equal to current upgrade sequence (%d)", errorReceipt.Sequence, channel.UpgradeSequence)
-	}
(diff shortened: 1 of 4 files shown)


crodriguezvega (approved):
Nice one; thanks, @charleenfei!

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -569,22 +569,25 @@ func (k Keeper) ChanUpgradeCancel(ctx sdk.Context, portID, channelID string, err
...
 	// if the msgSender is authorized to make and cancel upgrades AND the current channel has not already reached FLUSHCOMPLETE
 	// then we can restore immediately without any additional checks
 	// otherwise, we can only cancel if the counterparty wrote an error receipt during the upgrade handshake
 	if isAuthority && channel.State != types.FLUSHCOMPLETE {
 		return nil
 	}
 
+	// an error receipt proof must be provided if the sender is not authorized to cancel upgrades.
+	// the error receipt should also have a sequence greater than or equal to the current upgrade sequence.
+	if !isAuthority {
```
colin-axner, line 581: This is a bug? As per comment above:
```
// if the msgSender is authorized to make and cancel upgrades AND the current channel has not already reached FLUSHCOMPLETE
// then we can restore immediately without any additional checks
// otherwise, we can only cancel if the counterparty wrote an error receipt during the upgrade handshake
```

If the channel state is in FLUSHCOMPLETE, regardless of whether the authority called this function, we must prove that the counterparty wrote an error receipt for a valid upgrade sequence. Otherwise the authority could cancel an upgrade while the counterparty simultaneously is provided historical proof to upgrade to open. In this case, the authority would be harming itself, but no harm in ensuring this isn't possible

Issue #5169, opened 2023-11-24
Add list of allowed packet data keys to `Allocation` of `TransferAuthorization`


crodriguezvega:
_Thanks @colin-axner for the idea._

---

Currently a transfer grantee can submit a `MsgTransfer` with a non-empty memo field and any packet data. We should give the granter the possibility to specify what packet data keys they allow. 

Add [here](https://github.com/cosmos/ibc-go/blob/2bd29c08fd1fe50b461fc33a25735aa792dc896e/modules/apps/transfer/types/keys.go#L32) a constant string that holds the value that we expect if a granter wishes to allow any packet data key in the memo field of `MsgTransfer`:

```go
// AllowAllPacketDataKeys holds the string key that allows all packet data keys in authz transfer messages
AllowAllPacketDataKeys = "*"
```

Add an extra field to [message `Allocation` in `authz.proto`](https://github.com/cosmos/ibc-go/blob/main/proto/ibc/applications/transfer/v1/authz.proto#L12) that may hold the keys allowed to be used in the memo field:

```proto
// allow list of packet data keys, an empty list.
// an empty list prohibits all packet data keys; a list only with "*" permits any packet data key
repeated string allow_packet_data_list = 5; 
```

Implement function to check that the memo field contains only allowed packet data keys:

```go
func areAllowedPacketDataKeys(
  ctx sdk.Context,
  memo string,
  allowePacketDataList []string
) (bool, error) {
  // if the allow list is empty, then the memo must be an empty string
  if len(allowePacketDataList) == 0 {
    return len(strings.TrimSpace(memo)) == 0, nil
  }

  if len(allowePacketDataList) == 1 && allowePacketDataList[0] == AllowAllPacketDataKeys {
    return true, nil
  }

  jsonObject := make(map[string]interface{})
  err := json.Unmarshal([]byte(memo), &jsonObject)
  if err != nil {
    return false, err
  }

  gasCostPerIteration := ctx.KVGasConfig().IterNextCostFlat
  for k := range jsonObject {
    delete(jsonObject, k)
  }

  for key := range jsonObject {
    ctx.GasMeter().ConsumeGas(gasCostPerIteration, "transfer authorization")

    if !slices.Contains(allowePacketDataList, key) {
      return false, fmt.Errorf("not allowed packet data key: %s", key)
    }
  }

  return true, nil
}
```

[...]

PR #5280, opened 2023-11-30
Add list of allowed packet data keys to Allocation of TransferAuthorization

GNaD13:
## Description

closes: #5169

```text
type: commit message
```

---

write a little note why.


crodriguezvega (commented):
Thank you for quickly opening a PR for this issue, @GNaD13. I have left a bunch of minor, small nits.

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +158,39 @@
...
+
+	gasCostPerIteration := ctx.KVGasConfig().IterNextCostFlat
+
+	for key := range jsonObject {
+		ctx.GasMeter().ConsumeGas(gasCostPerIteration, "transfer authorization")
+
+		if !slices.Contains(allowedPacketDataList, key) {
+			return false, fmt.Errorf("not allowed packet data key: %s", key)
+		}
+	}
```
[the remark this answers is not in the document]
GNaD13: I think we no need to worry about this. The `jsonObject` only use here. It isn't change anything when we iteration over it

GNaD13:
> Thank you for quickly opening a PR for this issue, @GNaD13. I have left a bunch of minor, small nits.

Hi @crodriguezvega. Thanks for your reviewed. I've updated the code

modules/apps/transfer/types/transfer_authorization.go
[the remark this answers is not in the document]
colin-axner: The code will need to be readjusted. Map iteration in go is non-determinstic so if you have a key which is invalid, each node will take a different number of iterations to hit the invalid key. Since the gas is incremented for each iteration, each node could potentially have a different amount of gas used. Since the gas used is used to construct the block hash, nodes will diverge and create an app hash mismatch resulting in a chain halt

Someone from our team will take a look and come up with an alternative and safe approach

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +157,43 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
...
+		_, ok := jsonObject[key]
+		if ok {
+			ctx.GasMeter().ConsumeGas(gasCostPerIteration, "transfer authorization")
+			delete(jsonObject, key)
+		}
+	}
+
+	if len(jsonObject) != 0 {
+		return false, fmt.Errorf("not allowed packet data key")
+	}
```
crodriguezvega, line 192: Nice way of turning around the iteration and solving the issue, @GNaD13! :)

Maybe it would be nice if you could add to the error message the keys that are left in `jsonObject`, so that it's easier to identify what keys are not valid.

proto/ibc/applications/transfer/v1/authz.proto
```diff
@@ -19,6 +19,9 @@ message Allocation {
       [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
   // allow list of receivers, an empty allow list permits any receiver address
   repeated string allow_list = 4;
+  // allow list of packet data keys, an empty list.
+  // an empty list prohibits all packet data keys; a list only with "*" permits any packet data key
+  repeated string allow_packet_data_list = 5;
```
damiannolan, line 24: I would like to drop the `_list` suffix, its implicit given the `repeated` keyword

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +157,43 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
 	return false
 }
 
+// areAllowedPacketDataKeys returns a boolean indicating if the memo is valid for transfer. If it is not valid and non-nil error is also returned.
+func areAllowedPacketDataKeys(ctx sdk.Context, memo string, allowedPacketDataList []string) (bool, error) {
```
damiannolan, line 161:
```suggestion
func isAllowedPacketDataKeys(ctx sdk.Context, memo string, allowedPacketDataList []string) (bool, error) {
```

GNaD13:
Hi @damiannolan @DimitrisJim, I've updated the code. Thanks for your suggestion

modules/apps/transfer/types/transfer_authorization.go
GNaD13: Thanks for your suggestion @crodriguezvega. I've updated the code

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +158,48 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
 	return false
 }
 
+// isAllowedPacketDataKeys returns a boolean indicating if the memo is valid for transfer. If it is not valid and non-nil error is also returned.
```
colin-axner, line 161:
```suggestion
// isAllowedPacketDataKeys returns a boolean indicating if the memo is valid for transfer. If it is not valid a non-nil error is also returned.
```

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +158,48 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
...
+func isAllowedPacketDataKeys(ctx sdk.Context, memo string, allowedPacketDataList []string) (bool, error) {
+	// if the allow list is empty, then the memo must be an empty string
+	if len(allowedPacketDataList) == 0 {
+		return len(strings.TrimSpace(memo)) == 0, nil
```
colin-axner, line 165: should we be returning an error if this return false?

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +158,48 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
...
+	}
+
+	// if allowedPacketDataList has only 1 element and it equals AllowAllPacketDataKeys
+	// then accept all the packet data keys
+	if len(allowedPacketDataList) == 1 && allowedPacketDataList[0] == AllowAllPacketDataKeys {
+		return true, nil
+	}
+
+	// unmarshal memo
```
colin-axner, line 174:
```suggestion
```
super nit: I don't feel the comment adds meaning content

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +158,48 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
...
+	jsonObject := make(map[string]interface{})
+	err := json.Unmarshal([]byte(memo), &jsonObject)
+	if err != nil {
+		return false, err
+	}
+
+	if len(jsonObject) > len(allowedPacketDataList) {
+		return false, fmt.Errorf("packet type greater than packet allow list")
```
colin-axner, line 182: should this be a wrapped error? Once upon a time SDK discarded error message if they didn't contain a base error registered with the SDK (which fmt.Errorf is non-registered)

colin-axner, line 182:
```suggestion
		return false, fmt.Errorf("packet contains more packet data keys than packet allow list has")
```

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -2,9 +2,13 @@ package types
 
 import (
 	"context"
+	"encoding/json"
+	fmt "fmt"
 	"math/big"
+	"strings"
 
 	"github.com/cosmos/gogoproto/proto"
+	"golang.org/x/exp/maps"
```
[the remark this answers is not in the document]
colin-axner: In general I would avoid experimental features, especially since this is being used for an error message format

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +158,48 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
...
+		ctx.GasMeter().ConsumeGas(gasCostPerIteration, "transfer authorization")
+
+		_, ok := jsonObject[key]
+		if ok {
+			delete(jsonObject, key)
+		}
+	}
+
+	if len(jsonObject) != 0 {
+		return false, fmt.Errorf("not allowed packet data keys: %v", maps.Keys(jsonObject))
```
colin-axner, line 197: ditto on using registered error

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -50,6 +54,15 @@ func (a TransferAuthorization) Accept(ctx context.Context, msg proto.Message) (a
 			return authz.AcceptResponse{}, errorsmod.Wrap(ibcerrors.ErrInvalidAddress, "not allowed receiver address for transfer")
 		}
 
+		isAllowedPacket, err := isAllowedPacketDataKeys(sdk.UnwrapSDKContext(ctx), msgTransfer.Memo, allocation.AllowPacketData)
+		if err != nil {
+			return authz.AcceptResponse{}, errorsmod.Wrapf(ErrInvalidMemo, "error: %v", err)
```
colin-axner, line 59: typically I recommend wrapping the base error, especially when we produce the base error, this avoids switching up the order of the trace.

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -50,6 +54,15 @@ func (a TransferAuthorization) Accept(ctx context.Context, msg proto.Message) (a
...
+		}
+
+		if !isAllowedPacket {
+			return authz.AcceptResponse{}, ErrInvalidMemo
```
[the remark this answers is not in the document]
colin-axner: maybe we should just return an error rather than true/false and rename to something like `validateMemo`

proto/ibc/applications/transfer/v1/authz.proto
```diff
@@ -19,6 +19,9 @@ message Allocation {
       [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
   // allow list of receivers, an empty allow list permits any receiver address
   repeated string allow_list = 4;
+  // allow list of packet data keys, an empty list.
+  // an empty list prohibits all packet data keys; a list only with "*" permits any packet data key
+  repeated string allow_packet_data = 5;
```
colin-axner, line 24: super super nit that this could be named `allowed_packet_data` or `allow_packet_data_list`. My brain just stumbles over the current naming

colin-axner (commented):
Thank you so much for your contribution @GNaD13! I left several suggestions, but I think the implementation looks great overall :+1:

GNaD13:
Hi @colin-axner, Thanks for your reviewed. I've updated the code

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +152,51 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
 	return false
 }
 
+// validateMemo returns a nil error indicating if the memo is valid for transfer.
+func validateMemo(ctx sdk.Context, memo string, allowedPacketDataList []string) error {
+	// if the allow list is empty, then the memo must be an empty string
+	if len(allowedPacketDataList) == 0 {
+		if len(strings.TrimSpace(memo)) != 0 {
```
damiannolan, line 159: should these two conditions be included in a logical AND? 

The current code means you cannot include any memo unless there is something in `allowPacketDataList`. What if I want to add a random string memo "damian's packet" - it would fail.

modules/apps/transfer/types/errors.go
```diff
@@ -6,14 +6,15 @@ import (
...
+	ErrInvalidDenomForTransfer     = errorsmod.Register(ModuleName, 3, "invalid denomination for cross-chain transfer")
+	ErrInvalidVersion              = errorsmod.Register(ModuleName, 4, "invalid ICS20 version")
+	ErrInvalidAmount               = errorsmod.Register(ModuleName, 5, "invalid token amount")
+	ErrTraceNotFound               = errorsmod.Register(ModuleName, 6, "denomination trace not found")
+	ErrSendDisabled                = errorsmod.Register(ModuleName, 7, "fungible token transfers from this chain are disabled")
+	ErrReceiveDisabled             = errorsmod.Register(ModuleName, 8, "fungible token transfers to this chain are disabled")
+	ErrMaxTransferChannels         = errorsmod.Register(ModuleName, 9, "max transfer channels")
+	ErrInvalidAuthorization        = errorsmod.Register(ModuleName, 10, "invalid transfer authorization")
+	ErrInvalidMemo                 = errorsmod.Register(ModuleName, 11, "invalid memo")
+	ErrTransferAuthorizationPacket = errorsmod.Register(ModuleName, 12, "error TransferAuthorizationPacket not allowed")
```
damiannolan, line 19: I think we can probably reuse something like [`ErrUnauthorized`](https://github.com/cosmos/ibc-go/blob/main/modules/core/errors/errors.go#L18) rather than adding a new error here

damiannolan (changes requested):
Looks like there is a bug here. The `Allocation` at the current index is updated and is returned to x/authz in the `authz.AcceptReponse`. See L79-89.

The `AllowedPacketData` field is ignored when the allocation is recreated, it should be added on L84

modules/apps/transfer/types/transfer_authorization.go
GNaD13: yeah, according to the design if `allowedPacketDataList` is empty, we cannot include any memo. We will allow any memo if the `allowedPacketDataList` is "*"

GNaD13:
Thanks for your suggestion @damiannolan. I've updated the code.

modules/apps/transfer/types/transfer_authorization.go
damiannolan: Okay, so authz users must set the `allowedPacketDataList` to use any kind of memo at all. That's good I guess.. definitely explicit 😅

modules/apps/transfer/types/transfer_authorization.go
damiannolan: I think the errors returned from `validateMemo` look well formulated enough now to return directly here? wdyt?

```suggestion
			return authz.AcceptResponse{}, err
```

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +157,43 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
...
+
+	for _, key := range allowedPacketDataList {
+		_, ok := jsonObject[key]
+		if ok {
+			ctx.GasMeter().ConsumeGas(gasCostPerIteration, "transfer authorization")
+			delete(jsonObject, key)
+		}
+	}
+
+	if len(jsonObject) != 0 {
```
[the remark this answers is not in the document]
damiannolan: is this code path unreachable now?

damiannolan (commented):
Last two comments from me! Sorry for holding this up!

modules/apps/transfer/types/transfer_authorization.go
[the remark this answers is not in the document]
GNaD13: this condition will happend when the memo is not allowed by `allowedPacketDataList `

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -145,6 +153,51 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
...
+
+	jsonObject := make(map[string]interface{})
+	err := json.Unmarshal([]byte(memo), &jsonObject)
+	if err != nil {
+		return err
+	}
+
+	if len(jsonObject) > len(allowedPacketDataList) {
+		return errorsmod.Wrapf(ErrInvalidAuthorization, "packet contains more packet data keys than packet allow list has")
+	}
```
crodriguezvega, line 181: The issue I see of doing this is that we don't give any information to the user of what keys are the ones that are not allowed? cc @DimitrisJim

proto/ibc/applications/transfer/v1/authz.proto
```diff
@@ -19,6 +19,9 @@ message Allocation {
       [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"];
   // allow list of receivers, an empty allow list permits any receiver address
   repeated string allow_list = 4;
+  // allow list of packet data keys, an empty list.
+  // an empty list prohibits all packet data keys; a list only with "*" permits any packet data key
```
damiannolan, line 23:
```suggestion
  // allow list of packet data keys, an empty list prohibits all packet data keys;
  // a list only with "*" permits any packet data key
```

docs/docs/02-apps/01-transfer/08-authorizations.md
```diff
@@ -48,6 +51,9 @@ type Allocation struct {
   SpendLimit sdk.Coins  
   // allow list of receivers, an empty allow list permits any receiver address
   AllowList []string 
+  // allow list of packet data keys, an empty list.
+  // an empty list prohibits all packet data keys; a list only with "*" permits any packet data key
```
damiannolan, line 55:
```suggestion
  // allow list of packet data keys, an empty list prohibits all packet data keys;
  // a list only with "*" permits any packet data key
```

damiannolan (approved):

modules/apps/transfer/types/transfer_authorization.go
crodriguezvega: So you're fine if I drop this code as well?

crodriguezvega: >I mean the printing of the keys was dropped!

Ay, true! I can put that back and drop this for now then.

crodriguezvega: I will tackle this in a follow-up because the keys were shown before using the experimental feature, so I will try getting the keys in a different way.

Issue #5062, opened 2023-11-08
Restructure E2E tests to reduce number of runners used


chatton:
## Summary 

Currently, our E2E tests use one host per runner, each test spins up a new chain, we run a test, and tear down the chain.

In order to save on runners, we can instead try to run 1 test suite per runner. Implementation wise the following may work

1. Setup of test suite would create docker resources
2. Setup of test would create a new channel/connection
3. Perform test as usual with the addition of `t.Parallel()`

We would need to watch out for tests that have chain-wide effects like anything using gov proposals etc.

Ideally we can still run individual tests the way we do now locally. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

vuong177:
@chatton could I pick up this issue ?

chatton:
@vuong177 sure! It might be nice to try and PoC this on a single test suite, maybe [this one](https://github.com/cosmos/ibc-go/blob/08d36eb0735caa29ddf981a872d0a55d218be3f6/e2e/tests/transfer/base_test.go#L27)?

Let me know if you need a hand with any of the workflow files. Also if you're able to get it working locally, I'd be happy to help out with the CI integration.

PR #5317, opened 2023-12-05
e2e: Restructure to reduce number of runners used

anhductn2001:
## Description
Restructure to run 1 test suite per runner. And we can still run individual tests the way we do now locally.

Implementation the following:

1. Setup of test suite would create docker resources
2. Setup of test would create a new channel/connection
3. Perform test as usual with the addition of t.Parallel()

closes: #5062 

---

write a little note why.

### DIFF
--- a/cmd/build_test_suite_matrix/main.go
+++ b/cmd/build_test_suite_matrix/main.go
@@ -1,0 +1,157 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const (
+	testSuitePrefix    = "Test"
+	testSuiteSuffix    = "Suite"
+	testFileNameSuffix = "_test.go"
+	e2eTestDirectory   = "e2e"
+	// testEntryPointEnv specifes a single test function to run if provided.
+	testEntryPointEnv = "TEST_ENTRYPOINT"
+	// testExclusionsEnv is a comma separated list of test function names that will not be included
+	// in the results of this script.
+	testSuiteExclusionsEnv = "TEST_SUITE_EXCLUSIONS"
+	// testNameEnv if provided returns a single test entry so that only one test is actually run.
+)
+
+// GithubActionTestMatrix represents
+type GithubActionTestMatrix struct {
+	Include []TestSuite `json:"include"`
+}
+
+type TestSuite struct {
+	EntryPoint string `json:"entrypoint"`
+}
+
+func main() {
+	githubActionMatrix, err := getGithubActionMatrixForTests(e2eTestDirectory, getTestEntrypointToRun(), getExcludedTestSuiteFunctions())
+	if err != nil {
+		fmt.Printf("error generating github action json: %s", err)
+		os.Exit(1)
+	}
+
+	ghBytes, err := json.Marshal(githubActionMatrix)
+	if err != nil {
+		fmt.Printf("error marshalling github action json: %s", err)
+		os.Exit(1)
+	}
+	fmt.Println(string(ghBytes))
+}
+
+// getTestEntrypointToRun returns the specified test function to run if present, otherwise
+// it returns an empty string which will result in running all test suites.
+func getTestEntrypointToRun() string {
+	testSuite, ok := os.LookupEnv(testEntryPointEnv)
+	if !ok {
+		return ""
+	}
+	return testSuite
+}
+
+// getExcludedTestFunctions returns a list of test functions that we don't want to run.
+func getExcludedTestSuiteFunctions() []string {
+	exclusions, ok := os.LookupEnv(testSuiteExclusionsEnv)
+	if !ok {
+		return nil
+	}
+	return strings.Split(exclusions, ",")
+}
+
+func contains(s string, items []string) bool {
+	for _, elem := range items {
+		if elem == s {
+			return true
+		}
+	}
+	return false
+}
+
+// getGithubActionMatrixForTests returns a json string representing the contents that should go in the matrix
+// field in a github action workflow. This string can be used with `fromJSON(str)` to dynamically build
+// the workflow matrix to include all E2E tests under the e2eRootDirectory directory.
+func getGithubActionMatrixForTests(e2eRootDirectory, suite string, excludedItems []string) (GithubActionTestMatrix, error) {
+	testSuite := []string{}
+	fset := token.NewFileSet()
+	err := filepath.Walk(e2eRootDirectory, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return fmt.Errorf("error walking e2e directory: %s", err)
+		}
+
+		// only look at test files
+		if !strings.HasSuffix(path, testFileNameSuffix) {
+			return nil
+		}
+
+		f, err := parser.ParseFile(fset, path, nil, 0)
+		if err != nil {
+			return fmt.Errorf("failed parsing file: %s", err)
+		}
+
+		suiteNameForFile, err := extractSuite(f)
+		if err != nil {
+			return fmt.Errorf("failed extracting test suite name: %s", err)
+		}
+
+		if contains(suiteNameForFile, excludedItems) {
+			return nil
+		}
+
+		if suiteNameForFile != "" {
+			testSuite = append(testSuite, suiteNameForFile)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return GithubActionTestMatrix{}, err
+	}
+
+	gh := GithubActionTestMatrix{
+		Include: []TestSuite{},
+	}
+
+	for _, testSuiteName := range testSuite {
+		gh.Include = append(gh.Include, TestSuite{
+			EntryPoint: testSuiteName,
+		})
+	}
+
+	return gh, nil
+}
+
+// extractSuiteAndTestNames extracts the name of the test suite function as well
+// as all tests associated with it in the same file.
+func extractSuite(file *ast.File) (string, error) {
+	var suiteNameForFile string
+
+	for _, d := range file.Decls {
+		if f, ok := d.(*ast.FuncDecl); ok {
+			functionName := f.Name.Name
+			if isTestSuiteMethod(f) {
+				if suiteNameForFile != "" {
+					return "", fmt.Errorf("found a second test function: %s when %s was already found", f.Name.Name, suiteNameForFile)
+				}
+				suiteNameForFile = functionName
+				continue
+			}
+		}
+	}
+	return suiteNameForFile, nil
+}
+
+// isTestSuiteMethod returns true if the function is a test suite function.
+// e.g. func TestFeeMiddlewareTestSuite(t *testing.T) { ... }
+func isTestSuiteMethod(f *ast.FuncDecl) bool {
+	return strings.HasPrefix(f.Name.Name, testSuitePrefix) && strings.HasSuffix(f.Name.Name, testSuiteSuffix)
+}
--- a/e2e/Makefile
+++ b/e2e/Makefile
@@ -13,6 +13,9 @@ init:
 e2e-test: init cleanup-ibc-test-containers
 	./scripts/run-e2e.sh $(test) $(entrypoint)
 
+e2e-test-suite: init cleanup-ibc-test-containers
+	./scripts/run-e2e-suite.sh $(entrypoint)
+
 compatibility-tests:
 	./scripts/run-compatibility-tests.sh $(release_branch)
 
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -262,6 +262,7 @@ require (
 replace (
 	github.com/ChainSafe/go-schnorrkel => github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d
 	github.com/ChainSafe/go-schnorrkel/1 => github.com/ChainSafe/go-schnorrkel v1.0.0
+	github.com/strangelove-ventures/interchaintest/v8 => github.com/decentrio/interchaintest/v8 v8.0.1-0.20231204174445-4c0a01717be4
 	github.com/vedhavyas/go-subkey => github.com/strangelove-ventures/go-subkey v1.0.7
 )
 
--- a/e2e/scripts/run-e2e-suite.sh
+++ b/e2e/scripts/run-e2e-suite.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+set -eo pipefail
+
+ENTRY_POINT="${1}"
+
+function _verify_jq() {
+      if ! command -v jq > /dev/null ; then
+          echo "jq is required to extract test entrypoint."
+          exit 1
+      fi
+}
+
+
+function _verify_dependencies() {
+    # jq is always required to determine the entrypoint of the test.
+    _verify_jq
+}
+
+_verify_dependencies
+
+# if jq is installed, we can automatically determine the test entrypoint.
+# if command -v jq > /dev/null; then
+#    cd ..
+#    ENTRY_POINT="$(go run -mod=readonly cmd/build_test_matrix/main.go | jq -r --arg TEST "${TEST}" '.include[] | select( .test == $TEST)  | .entrypoint')"
+#    cd - > /dev/null
+# fi
+
+# find the name of the file that has this test in it.
+test_file="$(grep --recursive --files-with-matches './' -e "${ENTRY_POINT}(t")"
+
+# we run the test on the directory as specific files may reference types in other files but within the package.
+test_dir="$(dirname $test_file)"
+
+# run the test file directly, this allows log output to be streamed directly in the terminal sessions
+# without needed to wait for the test to finish.
+# it shouldn't take 30m, but the wasm test can be quite slow, so we can be generous.
+go test -v "${test_dir}" --run ${ENTRY_POINT} -timeout 150m
--- a/e2e/tests/core/02-client/client_test.go
+++ b/e2e/tests/core/02-client/client_test.go
@@ -50,6 +50,15 @@ func TestClientTestSuite(t *testing.T) {
 
 type ClientTestSuite struct {
 	testsuite.E2ETestSuite
+	chainA ibc.Chain
+	chainB ibc.Chain
+	rly    ibc.Relayer
+}
+
+func (s *ClientTestSuite) SetupTest() {
+	ctx := context.TODO()
+	s.chainA, s.chainB = s.GetChains()
+	s.rly = s.SetupRelayer(ctx, s.TransferChannelOptions(), s.chainA, s.chainB)
 }
 
 // Status queries the current status of the client
@@ -79,20 +88,21 @@ func (s *ClientTestSuite) TestScheduleIBCUpgrade_Succeeds() {
 	t := s.T()
 	ctx := context.TODO()
 
-	_, _ = s.SetupChainsRelayerAndChannel(ctx, nil)
-	chainA, chainB := s.GetChains()
-	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount, s.chainA)
 
 	const planHeight = int64(300)
 	const legacyPlanHeight = planHeight * 2
 	var newChainID string
 
 	t.Run("execute proposal for MsgIBCSoftwareUpgrade", func(t *testing.T) {
-		authority, err := s.QueryModuleAccountAddress(ctx, govtypes.ModuleName, chainA)
+		_, err := s.rly.GetChannels(ctx, s.GetRelayerExecReporter(), s.chainA.Config().ChainID)
+		s.Require().NoError(err)
+
+		authority, err := s.QueryModuleAccountAddress(ctx, govtypes.ModuleName, s.chainA)
 		s.Require().NoError(err)
 		s.Require().NotNil(authority)
 
-		clientState, err := s.QueryClientState(ctx, chainB, ibctesting.FirstClientID)
+		clientState, err := s.QueryClientState(ctx, s.chainB, ibctesting.FirstClientID)
 		s.Require().NoError(err)
 
 		originalChainID := clientState.(*ibctm.ClientState).ChainId
@@ -114,19 +124,19 @@ func (s *ClientTestSuite) TestScheduleIBCUpgrade_Succeeds() {
 			upgradedClientState,
 		)
 		s.Require().NoError(err)
-		s.ExecuteAndPassGovV1Proposal(ctx, scheduleUpgradeMsg, chainA, chainAWallet)
+		s.ExecuteAndPassGovV1Proposal(ctx, scheduleUpgradeMsg, s.chainA, chainAWallet)
 	})
 
 	t.Run("check that IBC software upgrade has been scheduled successfully on chainA", func(t *testing.T) {
 		// checks there is an upgraded client state stored
-		cs, err := s.QueryUpgradedClientState(ctx, chainA, ibctesting.FirstClientID)
+		cs, err := s.QueryUpgradedClientState(ctx, s.chainA, ibctesting.FirstClientID)
 		s.Require().NoError(err)
 
 		upgradedClientState, ok := cs.(*ibctm.ClientState)
 		s.Require().True(ok)
 		s.Require().Equal(upgradedClientState.ChainId, newChainID)
 
-		plan, err := s.QueryCurrentUpgradePlan(ctx, chainA)
+		plan, err := s.QueryCurrentUpgradePlan(ctx, s.chainA)
 		s.Require().NoError(err)
 
 		s.Require().Equal("upgrade-client", plan.Name)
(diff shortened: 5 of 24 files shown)


e2e/testsuite/tx.go
```diff
@@ -33,7 +33,48 @@ import (
...
+		resp, err = broadcastFunc()
+	}
+	s.Require().NoError(err)
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 2, chainA, chainB))
+
+	return resp
+}
+
+func (s *E2ETestSuite) BroadcastMessagesWithoutChainAB(ctx context.Context, chain ibc.Chain, user ibc.Wallet, msgs ...sdk.Msg) sdk.TxResponse {
```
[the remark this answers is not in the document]
anhductn2001: It means that the function does not need to pass both chainA and chainB parameters. Initially, my idea was to keep the existing functions because they were being used for UpgradeTest and GrandpaTest. I will consider bringing it together into a single function in the next work.

cmd/build_test_suite_matrix/main.go
```diff
@@ -0,0 +1,145 @@
+package main
```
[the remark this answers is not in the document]
anhductn2001: Totally agree. One thing I want to say is that I haven't really done anything related to optimizing the current code. I want us to look at the idea and then I'll finalize it.

e2e/Makefile
```diff
@@ -13,6 +13,9 @@ init:
 e2e-test: init cleanup-ibc-test-containers
 	./scripts/run-e2e.sh $(test) $(entrypoint)
 
+e2e-test-suite: init cleanup-ibc-test-containers
+	./scripts/run-e2e-suite.sh $(entrypoint)
```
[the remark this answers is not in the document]
anhductn2001: you need to give it a testsuite like `make e2e-test-suite entrypoint=TestClientTestSuite   `

e2e/tests/core/02-client/client_test.go
```diff
@@ -50,6 +50,15 @@ func TestClientTestSuite(t *testing.T) {
...
 	testsuite.E2ETestSuite
+	chainA ibc.Chain
+	chainB ibc.Chain
+	rly    ibc.Relayer
+}
+
+func (s *ClientTestSuite) SetupTest() {
+	ctx := context.TODO()
+	s.chainA, s.chainB = s.GetChains()
+	s.rly = s.SetupRelayer(ctx, s.TransferChannelOptions(), s.chainA, s.chainB)
```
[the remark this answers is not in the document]
anhductn2001: Ok, as far as I know it's not possible, but I'll try it and let you know the results.

e2e/scripts/run-e2e-suite.sh
```diff
@@ -0,0 +1,38 @@
...
+# find the name of the file that has this test in it.
+test_file="$(grep --recursive --files-with-matches './' -e "${ENTRY_POINT}(t")"
+
+# we run the test on the directory as specific files may reference types in other files but within the package.
+test_dir="$(dirname $test_file)"
+
+# run the test file directly, this allows log output to be streamed directly in the terminal sessions
+# without needed to wait for the test to finish.
+# it shouldn't take 30m, but the wasm test can be quite slow, so we can be generous.
+go test -v "${test_dir}" --run ${ENTRY_POINT} -timeout 150m
```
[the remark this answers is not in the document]
anhductn2001: Don't worry, this is a temporary number to ensure the test does not time out.

e2e/tests/core/02-client/client_test.go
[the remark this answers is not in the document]
anhductn2001: There is a problem with relayer setup, we have different channelOpts.

anhductn2001:
It seems like the Summary progress in the previous commit is running beyond the expected time causing the CI to be pending. Can you help me cancel it? @chatton

e2e/go.mod
```diff
@@ -263,6 +263,7 @@ require (
 replace (
 	github.com/ChainSafe/go-schnorrkel => github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d
 	github.com/ChainSafe/go-schnorrkel/1 => github.com/ChainSafe/go-schnorrkel v1.0.0
+	github.com/strangelove-ventures/interchaintest/v8 => github.com/decentrio/interchaintest/v8 v8.0.1-0.20231204174445-4c0a01717be4
```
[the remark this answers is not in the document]
anhductn2001: `DockerSetup` func is in the internal package of `interchaintest`. Do you have any idea how to configure it?

.github/workflows/e2e-test-suite-workflow-call.yml
```diff
@@ -0,0 +1,252 @@
+on:
```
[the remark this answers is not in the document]
anhductn2001: and use make `e2e-test-suite` instead `e2e-test`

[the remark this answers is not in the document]
chatton: if this is the only difference, I think we can instead just propagate a single field e.g. the `TEST_BY_SUITE: 'True'` env to the workflow.

e2e/go.mod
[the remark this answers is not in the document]
chatton: Ah, I mean we can create a PR against interchain test making this configurable. I'm happy to help out with this, I've worked a little bit on that repo before.

e2e/tests/core/02-client/client_test.go
```diff
@@ -52,6 +52,14 @@ type ClientTestSuite struct {
 	testsuite.E2ETestSuite
 }
 
+func (s *ClientTestSuite) SetupTest() {
+	ctx := context.TODO()
+	chainA, chainB := s.GetChains()
```
chatton, line 57: one thing I notice here, is that `s.GetChains()` allows for an optional function to be provided to modify the chain configurations, with these current `SetupTest` functions, the default options are used in all cases. I think this isn't the end of the world as custom test suites can provide their own `SetupTest` functions when needed, but worth mentioning I think.

e2e/testsuite/tx.go
```diff
@@ -33,7 +33,48 @@ import (
...
+	}
+	if s.relayers.ContainsRelayer(s.T().Name(), user) {
+		// Retry five times, the value of 5 chosen is arbitrary.
+		resp, err = s.retryNtimes(broadcastFunc, 5)
+	} else {
+		resp, err = broadcastFunc()
+	}
+	s.Require().NoError(err)
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 2, chainA, chainB))
```
chatton, line 72: it looks like the only reason we are adding chainB as an argument is to wait for blocks.

If this is true, I don't think we need to do this. If we should only need to wait for blocks on the chain we are broadcasting a message to.

If we do want to wait for the counterparty chain, we can just do that on a per test level.

I think adding chainB as an arg here makes usage less clear.

e2e/testsuite/tx.go
```diff
@@ -70,6 +111,7 @@ func (s *E2ETestSuite) BroadcastMessages(ctx context.Context, chain ibc.Chain, u
 	s.Require().NoError(err)
 
 	chainA, chainB := s.GetChains()
```
chatton, line 113: ah, I see we were already waiting for chainB , I don't think we even need to do this here.

e2e/testsuite/tx.go
```diff
@@ -201,7 +243,39 @@ func (s *E2ETestSuite) waitForGovV1ProposalToPass(ctx context.Context, chain ibc
 
 // ExecuteAndPassGovV1Beta1Proposal submits the given v1beta1 governance proposal using the provided user and uses all validators to vote yes on the proposal.
 // It ensures the proposal successfully passes.
-func (s *E2ETestSuite) ExecuteAndPassGovV1Beta1Proposal(ctx context.Context, chain ibc.Chain, user ibc.Wallet, content govtypesv1beta1.Content) {
+func (s *E2ETestSuite) ExecuteAndPassGovV1Beta1Proposal(ctx context.Context, chainA ibc.Chain, user ibc.Wallet, content govtypesv1beta1.Content, chainB ibc.Chain) {
```
chatton, line 246: I think chainB doesn't need to be an arg, this is just being added because of the `BroadcastMessages` method signature update right?

chatton (commented):
Leaving some more comments, thanks for all the work so far!

e2e/tests/interchain_accounts/base_test.go
```diff
@@ -34,34 +37,46 @@ type InterchainAccountsTestSuite struct {
 	testsuite.E2ETestSuite
 }
 
+func (s *InterchainAccountsTestSuite) SetupTest() {
+	ctx := context.TODO()
+	chainA, chainB := s.GetChains()
+	relayer := s.SetupRelayer(ctx, nil, chainA, chainB)
+	s.SetChainsIntoSuite(chainA, chainB)
+	s.SetRelayerIntoSuite(relayer)
```
chatton, line 45: we probably don't want to setup the relayer in every test. At the moment it looks like we have one test suite level relayer, but we are setting it up in every test. I think we should probably set up 1 relayer per test that is acting on a single channel. 

One issue that we might run into here, is that I think the default hermes config is watch all channels, so it might be easier for the first draft to do a single relayer set up that will watch all channels.

We can always tweak it in the future, if we go with that approach though, we'll want the relayer set up only a single time.

chatton: sorry, this is incorrect, I thought `SetupTest` ran before each test but it runs before the test suite.

I think it makes sense for chain creation to exist here, maybe we would want to move channel and relayer creation into `BeforeTest(suiteName, testName string)`.

e2e/Makefile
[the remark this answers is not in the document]
anhductn2001: I think it would be great. I will add error message to them.

e2e/go.mod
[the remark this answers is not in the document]
anhductn2001: Ah! Make sense! I will create an issue and PR against interchain test. This will probably take a little more time, but no matter, it's good!

e2e/tests/core/02-client/client_test.go
anhductn2001: Hmm, I think in special cases where we need to modify chainOption, we will have to move it to a separate suite. perhaps that's the trade-off with the current solution.

e2e/testsuite/testsuite.go
```diff
@@ -217,36 +238,36 @@ func (s *E2ETestSuite) UpdateClients(ctx context.Context, ibcrelayer ibc.Relayer
 // is unique to the current test being run. Note: this function does not create containers.
 func (s *E2ETestSuite) GetChains(chainOpts ...ChainOptionConfiguration) (ibc.Chain, ibc.Chain) {
 	if s.paths == nil {
+		testName = s.T().Name()
 		s.paths = map[string]pathPair{}
-	}
+		path, ok := s.paths[s.T().Name()]
+		if ok {
+			return path.chainA, path.chainB
+		}
```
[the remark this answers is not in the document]
anhductn2001: I just copied it from the previous func.

e2e/testsuite/tx.go
anhductn2001: Totally reasonable, I am going to fix it right away.

e2e/testsuite/testsuite.go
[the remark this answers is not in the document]
chatton: logically it isn't quite the same, is there a reason not to retain the previous code?

```golang
if s.paths == nil {
    s.paths = map[string]pathPair{}
}
```

cmd/build_test_matrix/main.go
```diff
@@ -31,24 +32,101 @@ type GithubActionTestMatrix struct {
...
 	Test       string `json:"test"`
 	EntryPoint string `json:"entrypoint"`
 }
 
 func main() {
-	githubActionMatrix, err := getGithubActionMatrixForTests(e2eTestDirectory, getTestToRun(), getTestEntrypointToRun(), getExcludedTestFunctions())
-	if err != nil {
-		fmt.Printf("error generating github action json: %s", err)
-		os.Exit(1)
+	testBySuite, _ := os.LookupEnv("TEST_BY_SUITE")
```
chatton, line 49: maybe we can add a function that returns this as a bool? 

return `"strings.ToLowerCase(testBySuite) == "true"` to be easier to work with.

To be consistent with the other functions maybe we can create a constant for `"TEST_BY_SUITE"` also. WDYT?

e2e/tests/transfer/authz_test.go
```diff
@@ -22,34 +21,31 @@ import (
...
 )
 
-func TestAuthzTransferTestSuite(t *testing.T) {
-	testifysuite.Run(t, new(AuthzTransferTestSuite))
-}
-
-type AuthzTransferTestSuite struct {
-	testsuite.E2ETestSuite
-}
+func (s *TransferTestSuite) TestAuthz_MsgTransfer_Succeeds() {
```
chatton, line 24: little confused as to the removal of the `TestAuthzTransferTestSuite` fn.

One limitation to the tests that you may not have been aware of is that we need each file to have a single entrypoint (e.g. TestAuthzTransferTestSuite) and the test suite defined in the same file along side it. 

Was this just an optimization to run more transfer tests using the same 2 chains? If this isn't required, we should revert this.

I tried to run `TestAuthz_MsgTransfer_Succeeds` locally and got `testing: warning: no tests to run` 

I believe this is because the test isn't being picked up correctly due to the limitation I mentioned.

e2e/tests/transfer/authz_test.go
```diff
@@ -22,34 +21,31 @@ import (
...
+	relayer := s.GetRelayerFromSuite()
+
+	channelA, err := relayer.GetChannels(ctx, s.GetRelayerExecReporter(), chainA.Config().ChainID)
+	s.Require().NoError(err)
+	chainAChannels := channelA[len(channelA)-1]
 
 	chainADenom := chainA.Config().Denom
 
-	granterWallet := suite.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+	granterWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
```
chatton, line 38: this PR is already quite large, can we avoid any cosmetic changes for now? Thank you!

e2e/tests/transfer/authz_test.go
```diff
@@ -156,54 +152,62 @@ func (suite *AuthzTransferTestSuite) TestAuthz_MsgTransfer_Succeeds() {
...
 	ctx := context.TODO()
 
-	relayer, channelA := suite.SetupChainsRelayerAndChannel(ctx, suite.TransferChannelOptions())
-	chainA, chainB := suite.GetChains()
+	chainA, chainB := s.GetChains()
+	relayer := s.GetRelayerFromSuite()
+
+	channelA, err := relayer.GetChannels(ctx, s.GetRelayerExecReporter(), chainA.Config().ChainID)
+	s.Require().NoError(err)
+	chainAChannels := channelA[len(channelA)-1]
```
chatton, line 195: this seems like a good candidate for a helper function, it seems it's now used in almost all the tests!  Maybe we can add this to the `E2ETestSuite`?

chatton: Maybe we could actually even make a helper that returns the relayer and the channel info.

e2e/tests/transfer/authz_test.go
```diff
@@ -233,107 +237,111 @@ func (suite *AuthzTransferTestSuite) TestAuthz_InvalidTransferAuthorizations() {
...
 			} else {
-				suite.AssertTxFailure(resp, sdkerrors.ErrInvalidAddress)
+				s.AssertTxFailure(resp, sdkerrors.ErrInvalidAddress)
 			}
 		})
 	})
+
+	t.Run("stop relayer", func(t *testing.T) {
+		s.StopRelayer(ctx, relayer)
+	})
```
chatton, line 333: do we need this addition?

e2e/tests/transfer/incentivized2_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+	"github.com/cosmos/ibc-go/e2e/testsuite"
+	"github.com/cosmos/ibc-go/e2e/testvalues"
+	feetypes "github.com/cosmos/ibc-go/v8/modules/apps/29-fee/types"
+	transfertypes "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	channeltypes "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+)
+
+func TestIncentivized2TransferTestSuite(t *testing.T) {
+	testifysuite.Run(t, new(IncentivizedTransferTestSuite))
```
chatton, line 25: this is not running the `Incentivized2TransferTestSuite` defined in this file

e2e/tests/transfer/incentivized2_test.go
```diff
@@ -0,0 +1,266 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/cosmos/ibc-go/e2e/testsuite"
+	"github.com/cosmos/ibc-go/e2e/testvalues"
+	feetypes "github.com/cosmos/ibc-go/v8/modules/apps/29-fee/types"
+	transfertypes "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	channeltypes "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+)
+
+func TestIncentivized2TransferTestSuite(t *testing.T) {
```
chatton, line 24: what's the reason we need a `TestIncentivized2TransferTestSuite`? Is it okay to stick with a single test suite for incenvitizes transfers? 

If it is a requirement to have a new test suite, I think we need to decide on a better name, as this doesn't really tell us anything.

e2e/tests/transfer/base_test.go
```diff
@@ -196,62 +92,24 @@ func (s *TransferTestSuite) TestMsgTransfer_Fails_InvalidAddress() {
...
-		s.Require().Equal(testvalues.StartingTokenAmount, bal)
+	t.Run("stop relayer", func(t *testing.T) {
+		s.StopRelayer(ctx, relayer)
 	})
 }
 
 // TestSendEnabledParam tests changing ics20 SendEnabled parameter
 func (s *TransferTestSuite) TestSendEnabledParam() {
 	t := s.T()
+
```
chatton, line 104: are we missing a `t.Parallel()`?

chatton (changes requested):
The tests are looking great, it looks like the tests are all passing on your PR, but I noticed a few things that need to be changed.

I think we're very close to be able to merge these changes, thank you for all the work so far!

cmd/build_test_matrix/main.go
anhductn2001: Nice, thanks for your reviews!

e2e/dockerutil/dockerutil.go
```diff
@@ -24,7 +24,8 @@ func GetTestContainers(ctx context.Context, t *testing.T, dc *dockerclient.Clien
 		Filters: filters.NewArgs(
 			// see: https://github.com/strangelove-ventures/interchaintest/blob/0bdc194c2aa11aa32479f32b19e1c50304301981/internal/dockerutil/setup.go#L31-L36
 			// for the label needed to identify test containers.
-			filters.Arg("label", testLabel+"="+t.Name()),
+			// TODO: Add handle test name
```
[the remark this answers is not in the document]
anhductn2001: Currently, because we are using a single container for each chain from start to finish of the test, I am fixing the label here. Maybe we'll have a better way to handle it in the future.

cmd/build_test_matrix/main.go
```diff
@@ -31,24 +32,101 @@ type GithubActionTestMatrix struct {
...
+			os.Exit(1)
+		}
+
+		ghBytes, err := json.Marshal(githubActionMatrix)
+		if err != nil {
+			fmt.Printf("error marshalling github action json: %s", err)
+			os.Exit(1)
+		}
+		fmt.Println(string(ghBytes))
+	} else {
```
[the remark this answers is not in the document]
anhductn2001: I have combined them into a single function. Wdyt?

e2e/tests/transfer/authz_test.go
anhductn2001: we had this problem if we didn't StopRelayer `tried to start relayer again without stopping first`

e2e/tests/transfer/base_test.go
anhductn2001: I tried it but seems it should only be used in certain situations. Like the problem you predicted when you created the issue, the tests interfere with each other so they don't work properly.

e2e/tests/transfer/incentivized2_test.go
anhductn2001: the problem here is that they will interact with each other if we put them together in a test suite, same reason why we are missing t.Parallel(). Do you have any ideas?

chatton: Ah, I think it is fine if they are in a different test suite, but I think we can maybe give them better names.

e2e/tests/transfer/base_test.go
chatton: Okay! In this case, maybe this params test can go in its own file, and it's own test suite, so we are not waiting on this test to run before starting others.

e2e/tests/transfer/authz_test.go
anhductn2001: We can specify entrypoints for them like: 
`make e2e-test test=TestAuthz_MsgTransfer_Succeeds entrypoint=TransferTestSuite `

anhductn2001: It's true that this is just to optimize the amount of resources generated.

anhductn2001: this is really limited by the entrypoint recognition functionality. I will revert them!

chatton (commented):
Things are looking great, but the one big thing that needs to be addressed before we can merge this into main is the one relayer per test. 

I think what we can do here, is I'll create a feature branch for this, we can merge your PR into that, then we can work on making the changes required for the relayers in a follow up.

Does that work for you?

There is no rush to immediately start working on the relayer issue, it might be a bit of work. Either way, we'll make sure that you are a co-author on the commit when the feature branch gets merged into main!

anhductn2001:
We are in big trouble adding t.Parallel(). I'm thinking of ways and trying to solve it. Thanks @chatton  , I find your comments absolutely correct.

e2e/tests/interchain_accounts/gov_test.go
```diff
@@ -117,5 +130,9 @@ func (s *InterchainAccountsGovTestSuite) TestInterchainAccountsGovIntegration()
 			expected := testvalues.IBCTransferAmount + testvalues.StartingTokenAmount
 			s.Require().Equal(expected, balance.Int64())
 		})
+
+		t.Run("stop relayer", func(t *testing.T) {
+			s.StopRelayer(ctx, relayer)
+		})
```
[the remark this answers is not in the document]
anhductn2001: The problem I'm having is creating a new relayer without creating a new chain.

anhductn2001:
Currently, with each initialization of a relayer instance, we must come with a new pair of chains. Do you have any ideas to fix this? @chatton

chatton:
@anhductn2001 once possible solution I can think of is to maintain a map of `chainPair -> relayerInstance`, this could require a refactor to the logic of the `startRelyerFn`, perhaps maybe a `map[pathName]Relayer`?

One thing that might be a problem is if the relayer is configured to watch all channels, we need to make sure that the relayer is configured to only relay packets we care about for the specific test.

For hermes  [this](https://github.com/informalsystems/hermes/blob/ab732666fe35de129ada98731280d03411f6375f/config.toml#L350) looks like a setting we will want to configure. The default behaviour will not filter any packets and so everything will be relayed. (interfering with other tests).

Currently, the relayer has a lot of default configuration options that are specified in interchaintest [here](https://github.com/strangelove-ventures/interchaintest/blob/main/relayer/hermes/hermes_config.go)

It looks like `chains.packet_filter` isn't even an option here, so the `Config` type may need an update to allow for it.


I think the following things need to happen before we can make this work.

1. Update the `Config` struct in interchaintest to allow for enabling these fields 
2. Enable support to make arbitrary changes to the `Config` struct when creating a hermes relayer instance in interchaintest. This may not be as simple as it seems, [this](https://github.com/strangelove-ventures/interchaintest/blob/02878ba0929563f146ecb6fe9bccf91b6567a986/relayer/hermes/hermes_relayer.go#L80) code may need to be updated to accept new types or functions which make modifications. - A slightly hackier alternative would be to provide some way to write arbitrary bytes to where the hermes config file will live, this way the logic could live in `ibc-go`, I think the arbitrary configuration of the config file would be really great feature to exist in interchaintest though, and if possible it would be great to add it there.
3. Update the `E2ETestSuite` in ibc-go to have a mapping of `pathName -> Relayer`
4. Update the `E2ETestSuite.StartRelayer` function to start the relevant relayer based on the pathName in the tests.


The interchaintest modifications I suggested will be the tricker thing to implement as it might end up requiring some non-trivial refactors.

anhductn2001:
Could you help me update branch `feat/run-full-test....`. I'm trying to run the tests but seem like they require ics27, ics29, etc

chatton:
Hi @anhductn2001, I tried running tests from `decentrio:restructure_e2e` and they seem to be running fine, which commands are you using to run tests that aren't working for you so I can try them

anhductn2001:
command: ` make e2e-test-suite entrypoint=TestInterchainAccountsTestSuite`

anhductn2001:
` raw_log: 'failed to execute message; message index: 0: ORDER_NONE_UNSPECIFIED: invalid`

chatton:
ah! @anhductn2001 that test specifically has an issue on main and is being worked on, can you try running a different test, maybe the transfer successful case `TestMsgTransfer_Succeeds_Nonincentivized` ?

anhductn2001:
Thank you @DimitrisJim , sounds good, I'm looking forward to it

PR #5343, opened 2023-12-08
feat: checked if controller enabled or not

vishal-kanna:


crodriguezvega (commented):
Thank you, @vishal-kanna. Would it be possible for you to add a test case? Something similar to the [test case we have in the host submodule for `OnRecvPacket`](https://github.com/cosmos/ibc-go/blob/908a0983c73bb85f418a2edb8f067d80fd7de259/modules/apps/27-interchain-accounts/host/ibc_module_test.go#L406-L410).

crodriguezvega (approved):
Thanks for fixing this, @vishal-kanna!

PR #5389, opened 2023-12-12
feat : reused the error 

vishal-kanna:


damiannolan:
Hi, it looks like there's some compiler errors here from a redeclared function.

I would also suggest naming the errors `err` as is standard in golang apps.

```diff
- invalidupgraderoute := errorsmod.Wrapf(porttypes.ErrInvalidRoute, "upgrade route not found to module: %s", module)
+ err = errorsmod.Wrapf(porttypes.ErrInvalidRoute, "upgrade route not found to module: %s", module)
```

chatton (approved):
LGTM!

modules/core/keeper/msg_server.go
```diff
@@ -856,19 +856,22 @@ func (k Keeper) ChannelUpgradeAck(goCtx context.Context, msg *channeltypes.MsgCh
 
 	app, ok := k.Router.GetRoute(module)
 	if !ok {
-		ctx.Logger().Error("channel upgrade ack failed", "port-id", msg.PortId, "error", errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module))
-		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
+		err = errorsmod.Wrapf(porttypes.ErrInvalidRoute, "route not found to module: %s", module)
```
damiannolan, line 859: we could reinitialise new errors here as they'd be within the scope of the `if`, but I don't really feel super strong on it and happy to reuse `err`

modules/core/keeper/msg_server.go
```diff
@@ -856,19 +856,22 @@ func (k Keeper) ChannelUpgradeAck(goCtx context.Context, msg *channeltypes.MsgCh
...
-		return nil, errorsmod.Wrapf(porttypes.ErrInvalidRoute, "upgrade route not found to module: %s", module)
+		err = errorsmod.Wrapf(porttypes.ErrInvalidRoute, "upgrade route not found to module: %s", module)
+		ctx.Logger().Error("channel upgrade ack failed", "port-id", msg.PortId, "error", err)
+		return nil, err
 	}
 
 	err = k.ChannelKeeper.ChanUpgradeAck(ctx, msg.PortId, msg.ChannelId, msg.CounterpartyUpgrade, msg.ProofChannel, msg.ProofUpgrade, msg.ProofHeight)
 	if err != nil {
-		ctx.Logger().Error("channel upgrade ack failed", "error", errorsmod.Wrap(err, "channel upgrade ack failed"))
+		errChanUpgradeFailed := errorsmod.Wrap(err, "channel upgrade ack failed")
```
damiannolan, line 873: any reason not to reuse `err` here instead of creating a new var with a different name? 

`err = errorsmod.Wrap(err, "channel upgrade ack failed")` should work fine too, right?

damiannolan (approved):
LGTM! Thanks @vishal-kanna

modules/core/keeper/msg_server.go
vishal-kanna: @damiannolan Using the err works fine.

DimitrisJim: ah it won't. `IsUpgradeError` is a direct cast to `UpgradeError`, wrapping fails. (I think Cian mentioned this in a call at some point).

modules/core/keeper/msg_server.go
DimitrisJim: will push it in as is for now, can't say I have a strong pref either way too (slight preference for yours since it scopes values nicely but not enough pref to make the change)

DimitrisJim:
do we backport to 8.1? cc @crodriguezvega

Issue #3999, opened 2023-07-04
Add coordinator `Setup` functions to the Path type


colin-axner:
## Summary

The coordinator has the following functions:
```go
type Coordinator interface {
    IncrementTime() 
    IncrementTimeBy(increment time.Duration)
    UpdateTime()
    UpdateTimeForChain(chain *TestChain)
    
    Setup(path *Path) 
    SetupClients(path *Path)
    SetupConnections(path *Path)

    CreateConnections(path *Path)
    CreateMockChannels(path *Path)
    CreateTransferChannels(path *Path)
    CreateChannels(path *Path)
    
    GetChain(chainID string) *TestChain
    
    CommitBlock(chains ...*TestChain)
    CommitNBlocks(chain *TestChain, n uint64)
}
```

The following functions should be moved to be under the path type:
```go
type Path interface {
    Setup() 
    SetupClients()
    SetupConnections()
    
    CreateConnections()
    CreateChannels()
}
```

I think we could consider a way to reduce this API further (definitely keeping `Setup()`, but maybe reducing the redundancy between `SetupConnections` and `CreateConnections`)

## Problem Definition

The coordinator is an unnecessary type in the setup of the path. It is more natural to simply reference the path type directly. This clears the way to reduce the usage of coordinator which in the long run may be an unnecessary layer of the testing pkg. 

## Proposal

Change:
```go
    path := ibctesting.NewPath(suite.chainA, suite.chainB)
    suite.coordinator.Setup(path)
```

to
```go
    path := ibctesting.NewPath(suite.chainA, suite.chainB)
    path.Setup()
```

**NOTE**: I recommend we leave the coordinator functions (deprecation) and add the functions to the path and replace our internal usage. After release of v8, then I think we can remove the unnecessary coordinator functions. Because v8 is a massive release that is intermixed with a SDK and cometbft update, I think we should limit unnecessary breaking changes

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

colin-axner:
Lets add the proposed functions to path and deprecate the legacy functions on the coordinator. I will open a separate issue to remove the deprecated functions. That way we can make this change and choose when to have users update at a reasonable time period

hoangdv2429:
Can I work on this sir ?

PR #5397, opened 2023-12-13
Add coordinator Setup functions to the Path type

chandiniv1:
closes: #3999


chatton:
Thanks @chandiniv1, it looks like there are still some compilation errors, specifically uses of `suite.coordinator.Setup` (which is not defined)

chandiniv1:
> Thanks @chandiniv1, it looks like there are still some compilation errors, specifically uses of `suite.coordinator.Setup` (which is not defined)

@chatton I have fixed the compilation errors

testing/path.go
```diff
@@ -115,3 +115,97 @@ func (path *Path) RelayPacketWithResults(packet channeltypes.Packet) (*abci.Exec
...
+
+	err = path.EndpointB.CreateClient()
+	if err != nil {
+		panic(err)
+	}
+}
+
+// SetupClientConnections is a helper function to create clients and the appropriate
+// connections on both the source and counterparty chain. It assumes the caller does not
+// anticipate any errors.
```
chatton, line 145:
```suggestion
// SetupConnections is a helper function to create clients and the appropriate
// connections on both the source and counterparty chain. It assumes the caller does not
// anticipate any errors.
```

testing/path.go
```diff
@@ -115,3 +115,97 @@ func (path *Path) RelayPacketWithResults(packet channeltypes.Packet) (*abci.Exec
...
+func (path *Path) SetupConnections() {
+	path.SetupClients()
+
+	path.CreateConnections()
+}
+
+// CreateConnection constructs and executes connection handshake messages in order to create
+// OPEN channels on chainA and chainB. The connection information of for chainA and chainB
+// are returned within a TestConnection struct. The function expects the connections to be
+// successfully opened otherwise testing will fail.
```
chatton, line 155:
```suggestion
// CreateConnections constructs and executes connection handshake messages in order to create
// OPEN connections on chainA and chainB. The function expects the connections to be
// successfully opened otherwise testing will fail.
```

Seems this docstring was out of date

testing/path.go
```diff
@@ -115,3 +115,97 @@ func (path *Path) RelayPacketWithResults(packet channeltypes.Packet) (*abci.Exec
...
+	// ensure counterparty is up to date
+	err = path.EndpointA.UpdateClient()
+	if err != nil {
+		panic(err)
+	}
+}
+
+// CreateChannel constructs and executes channel handshake messages in order to create
+// OPEN channels on chainA and chainB. The function expects the channels to be successfully
+// opened otherwise testing will fail.
```
chatton, line 185:
```suggestion
// CreateChannels constructs and executes channel handshake messages in order to create
// OPEN channels on chainA and chainB. The function expects the channels to be successfully
// opened otherwise testing will fail.
```

testing/path.go
```diff
@@ -115,3 +115,97 @@ func (path *Path) RelayPacketWithResults(packet channeltypes.Packet) (*abci.Exec
 
 	return nil, nil, fmt.Errorf("packet commitment does not exist on either endpoint for provided packet")
 }
+
+// Setup constructs a TM client, connection, and channel on both chains provided. It will
+// fail if any error occurs. The clientID's, TestConnections, and TestChannels are returned
+// for both chains. The channels created are connected to the ibc-transfer application.
```
chatton, line 121:
```suggestion
// Setup constructs a TM client, connection, and channel on both chains provided. It will
// fail if any error occurs. The channels created are connected to the ibc-transfer application.
```

chatton (commented):
Looks great! The only thing is that the docstrings seem to be a bit stale, I've proposed some suggestions for changes.

After this I think it looks good to merge, thank you!

AdityaSripal (approved):

chatton (approved):
LGTM!

testing/coordinator.go
```diff
@@ -71,57 +71,6 @@ func (coord *Coordinator) UpdateTimeForChain(chain *TestChain) {
 	chain.CurrentHeader.Time = coord.CurrentTime.UTC()
 }
 
-// Setup constructs a TM client, connection, and channel on both chains provided. It will
-// fail if any error occurs. The clientID's, TestConnections, and TestChannels are returned
-// for both chains. The channels created are connected to the ibc-transfer application.
-func (coord *Coordinator) Setup(path *Path) {
```
colin-axner, line 77: I think maybe we should deprecate these functions? It's a quick change, but will cause enormous diffs when updating to v9 for our users. So might be worth giving them to the option to separate this diff between v9 and v10

We can just add these back calling into path.Setup()

testing/coordinator.go
```diff
@@ -146,22 +126,9 @@ func (coord *Coordinator) CreateTransferChannels(path *Path) {
 // CreateChannel constructs and executes channel handshake messages in order to create
 // OPEN channels on chainA and chainB. The function expects the channels to be successfully
 // opened otherwise testing will fail.
+// DEPRECATED: please use path.SetupChannels(), this function will be removed in v10
```
colin-axner, line 129:
```suggestion
// DEPRECATED: please use path.CreateChannels(), this function will be removed in v10
```

colin-axner (approved):
Thanks @chandiniv1!

Issue #41, opened 2020-10-30
Return errors instead of panicking in router


cwgoes:
Instead, panic in the app.

crodriguezvega:
So this is about turning the panics in this [file](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/types/router.go) into errors that are returned?

colin-axner:
This applies to all of 05-port, not just the router file in types

gjermundgaraba:
Closing this for now. We might reopen this in the future, but this is not currently planned.

PR #5401, opened 2023-12-13
fix(api)!: Return errors instead of panicking in 05-port

chandiniv1:
closes: #41


damiannolan:
Briefly checking in on this PR. I'm wondering if we should go ahead with this or not at the moment as I believe @AdityaSripal was looking into a new structure for app/middleware routing and wiring. 

This change would be API breaking and require a new major release, at which point it could be superseded by the changes mentioned above. It may minimise disruptions by leaving the current code in-place for the time being but I'm not 100% sure how others feel. 

cc. @colin-axner for any additional thoughts.

colin-axner:
I am in agreement with @damiannolan. Especially since it feels the error returned would likely just panic in some other part of the code trace

Thanks @chandiniv1 for your contribution! I would recommend close this pr and wait to see the outcome of the app/middleware routing/wiring designs before addressing this issue

Issue #3937, opened 2023-06-22
Prune stale acknowledgements after successful channel upgrade


colin-axner:
## Summary

Prune stale acknowledgements after successful channel upgrade.

## Problem Definition

Acknowledgements currently live forever in IBC state despite no longer being necessary once a packet has completed its lifecycle. Overtime this can add up to quite a bit as outlined in #2869

Once a channel successfully upgrades, all old packets using previous channel parameters have been flushed and completed their lifecycle. The acknowledgement on the counterparty can then be safely removed. 

## Proposal

To avoid potential gas issues in block processing due to an unpredictable amount of acknowledgements to prune, I suggest the channel keeper store a mapping from the port/channel ID to the highest stale sequence. By stale sequence, I mean a sequence which refers to packets which have fully completed their lifecycle and thus all their information may be safely deleted. 

The channel upgrade will already receive the counterparty latest sequence send, so on an upgrade it simply needs to store it:

```go
// naming can be changed
k.setAcknowledgementPruningSequence(portID, channelID, counterpartyLatestSequenceSend)
```

A new msg type will need to be added which may prune any acknowledgement with a sequence <= to the highest stale packet sequence

```proto
message MsgPruneAcknowledgements {
  string port_id = 1;
  string channel_id    = 2;
  // optionally add a parameter to control the amount of acknowledgements to prune to avoid overrunning gas (we can alternatively account for this in the handler by checking gas left after each deletion)
}
```

```go
// PruneAcknowledgements defines a rpc handler method for MsgCreateClient.
func (k Keeper) PruneAcknowledgements(goCtx context.Context, msg *clienttypes.MsgPruneAcknowledgements) (*channeltypes.MsgPruneAcknowledgementsResponse,
    ctx := sdk.UnwrapSDKContext(goCtx)

    pruningSequence, found := k.channelKeeper.GetAcknowledgementPruningSequence(ctx, msg.PortId, msg.ChannelId)
    if !found {
        return nil, err 
    }   

    // reverse iterate from earliest packet sequence, delete each acknowledgement until the acknowledgement sequence > pruningSequence 
    k.channelKeeper.pruneAcknowledegments(ctx, msg.PortId, msg.ChannelId, pruningSequence)
   

    return &channeltypes.MsgPruneAcknowledgementResponse{}, nil 
}
```

We can also add some result value to the `MsgPruneAcknowledgementResponse` to indicate if all possible acknowledgements were pruned, or if there are some left

[...]

PR #5444, opened 2023-12-19
Add pruning logic, implementation in message server.

DimitrisJim:
## Description

closes: #3937 

```text
type: commit message
```

---

write a little note why.


modules/core/keeper/msg_server.go
```diff
@@ -1131,6 +1131,18 @@ func (k Keeper) UpdateChannelParams(goCtx context.Context, msg *channeltypes.Msg
...
+	if !found {
+		return nil, errorsmod.Wrapf(channeltypes.ErrPruningSequenceNotFound, "port ID (%s) channel ID (%s)", msg.PortId, msg.ChannelId)
+	}
+
+	sequence = k.ChannelKeeper.PruneAcknowledgements(ctx, msg.PortId, msg.ChannelId, msg.Limit, sequence)
+
+	// update pruning sequence in store
+	k.ChannelKeeper.SetAcknowledgementPruningSequence(ctx, msg.PortId, msg.ChannelId, sequence)
+
 	return &channeltypes.MsgPruneAcknowledgementsResponse{}, nil
```
crodriguezvega, line 1147: Should we update the message to return the number of acks and receipts that were pruned?

modules/core/04-channel/keeper/keeper.go
```diff
@@ -629,8 +647,31 @@ func (k Keeper) SetAcknowledgementPruningSequence(ctx sdk.Context, portID, chann
...
 
-// PruneAcknowledgements prunes packet acknowledgements from the store that have a sequence number less than or equal to the pruning sequence.
+// PruneAcknowledgements prunes packet acknowledgements from the store that have a sequence number less than or equal to the last packet send.
 // The number of packet acknowledgements pruned is equal to limit. Pruning only occurs after a channel has been upgraded.
-func (Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit, pruningSequence uint64) error {
-	return nil
+//
+// The pruningSequence keeps track of the packet acknowledgement that can be pruned next. When the pruning sequence reaches
+// the last send sequence, pruning is complete.
+func (k Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit, pruningSequence uint64) uint64 {
```
crodriguezvega, line 655: Should the naming reflect that it also prunes packet receipts?

crodriguezvega (commented):
I know this is still in draft, but left just a couple of comments. :)

modules/core/04-channel/keeper/keeper.go
```diff
@@ -629,8 +647,31 @@ func (k Keeper) SetAcknowledgementPruningSequence(ctx sdk.Context, portID, chann
...
+	// TODO(jim): Update to use last sequence send.
+	lastSequenceSend := uint64(0)
+
+	limit = pruningSequence + limit
+	for ; pruningSequence < limit; pruningSequence++ {
```
crodriguezvega, line 660: Should we be charging some gas in this loop?

crodriguezvega: True, true. That's probably fine then.

modules/core/04-channel/keeper/keeper.go
```diff
@@ -653,8 +671,33 @@ func (k Keeper) HasPruningSequenceStart(ctx sdk.Context, portID, channelID strin
...
 
-// PruneAcknowledgements prunes packet acknowledgements from the store that have a sequence number less than or equal to the pruning sequence.
+// PruneAcknowledgements prunes packet acknowledgements from the store that have a sequence number less than or equal to the last packet send.
 // The number of packet acknowledgements pruned is equal to limit. Pruning only occurs after a channel has been upgraded.
-func (Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit, pruningSequence uint64) error {
-	return nil
+//
+// The pruningSequence keeps track of the packet acknowledgement that can be pruned next. When the pruning sequence reaches
+// the last send sequence, pruning is complete.
+func (k Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit, pruningSequenceStart uint64) (uint64, error) {
```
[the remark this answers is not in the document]
colin-axner: yes it does. I like having self contained logic

modules/core/04-channel/keeper/keeper.go
```diff
@@ -653,8 +671,33 @@ func (k Keeper) HasPruningSequenceStart(ctx sdk.Context, portID, channelID strin
...
+	pruningSequenceEnd, found := k.GetPruningSequenceEnd(ctx, portID, channelID)
+	if !found {
+		return pruningSequenceStart, errorsmod.Wrapf(types.ErrPruningSequenceStartNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	limit = pruningSequenceStart + limit
+	for ; pruningSequenceStart < limit; pruningSequenceStart++ {
+		// break if the pruning sequence reaches the last send sequence
+		if pruningSequenceStart > pruningSequenceEnd {
```
colin-axner, line 688:
```suggestion
		if pruningSequenceStart >= pruningSequenceEnd {
```
We don't want to prune the next sequence send ack https://github.com/cosmos/ibc-go/blob/04-channel-upgrades/modules/core/04-channel/keeper/upgrade.go#L586

modules/core/04-channel/keeper/keeper.go
```diff
@@ -653,8 +671,41 @@ func (k Keeper) HasPruningSequenceStart(ctx sdk.Context, portID, channelID strin
...
+// The pruningSequence keeps track of the packet acknowledgement that can be pruned next. When the pruning sequence reaches
+// the last send sequence, pruning is complete.
+func (k Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit uint64) error {
+	if !k.HasPruningSequenceStart(ctx, portID, channelID) {
+		// indicates that a channel upgrade has not been performed
+		return errorsmod.Wrapf(types.ErrPruningSequenceStartNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+	pruningSequenceStart := k.GetPruningSequenceStart(ctx, portID, channelID)
+
+	pruningSequenceEnd, found := k.GetPruningSequenceEnd(ctx, portID, channelID)
```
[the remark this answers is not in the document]
charleenfei: i guess it also depends on what we want to do with an error if ie: found is `false`/if we want to panic on nil key using `Has`?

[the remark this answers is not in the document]
AdityaSripal: But you are using `pruningSequenceEnd` later. I think if `value` is actually being used then we should use `Get`

If we only need to know if the key exists or not but we don't care what the value is, then use `Has`.

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,260 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+func (suite *KeeperTestSuite) TestPruneStalePacketData() {
+	var (
+		path          *ibctesting.Path
+		limit         uint64
+		upgradeFields types.UpgradeFields
+
+		// postPruneExpState is a helper function to verify the expected state after pruning. Argument expLeft
+		// denotes the expected amount of packet acks and receipts left after pruning. Argument expSequenceStart
+		// denotes the expected value of PruneSequenceStart.
+		postPruneExpState = func(expAcksLen, expReceiptsLen, expSequenceStart uint64) {
```
crodriguezvega, line 555:
```suggestion
		postPruneExpState = func(expAcksLen, expReceiptsLen, expPruningSeqStart uint64) {
```

modules/core/keeper/msg_server.go
crodriguezvega: Let's do logging at least then.

modules/core/keeper/msg_server_test.go
```diff
@@ -2016,3 +2016,77 @@ func (suite *KeeperTestSuite) TestUpdateConnectionParams() {
...
+			suite.Require().NoError(err)
+
+			err = path.EndpointB.ChanUpgradeConfirm()
+			suite.Require().NoError(err)
+
+			err = path.EndpointA.ChanUpgradeOpen()
+			suite.Require().NoError(err)
+
+			err = path.EndpointA.UpdateClient()
+			suite.Require().NoError(err)
```
crodriguezvega, line 2070: Can you use here the `UpgradeChannel` function you added in `keeper_test`?

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,260 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+				sequenceEnd, found := suite.chainA.App.GetIBCKeeper().ChannelKeeper.GetPruningSequenceEnd(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+				suite.Require().True(found)
+
+				// We expect nothing to be left and sequenceStart == sequenceEnd.
+				postPruneExpState(0, 0, sequenceEnd)
+			},
+			nil,
+		},
+		{
+			"success: stale packet data partially pruned, update, prune again",
```
crodriguezvega, line 650:
```suggestion
			"success: stale packet data partially pruned, upgrade, prune again",
```

modules/core/04-channel/keeper/keeper.go
```diff
@@ -653,8 +671,41 @@ func (k Keeper) HasPruningSequenceStart(ctx sdk.Context, portID, channelID strin
...
-func (Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit, pruningSequence uint64) error {
+//
+// The pruningSequence keeps track of the packet acknowledgement that can be pruned next. When the pruning sequence reaches
+// the last send sequence, pruning is complete.
+func (k Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit uint64) error {
+	if !k.HasPruningSequenceStart(ctx, portID, channelID) {
+		// indicates that a channel upgrade has not been performed
+		return errorsmod.Wrapf(types.ErrPruningSequenceStartNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+	pruningSequenceStart := k.GetPruningSequenceStart(ctx, portID, channelID)
```
crodriguezvega: Should we also do here `pruningSequenceStart, found := k.GetPruningSequenceStart(...`?

crodriguezvega (commented):
I think this is looking very good. Nice work, @charleenfei.

modules/core/04-channel/keeper/keeper.go
```diff
@@ -653,8 +671,46 @@ func (k Keeper) HasPruningSequenceStart(ctx sdk.Context, portID, channelID strin
...
-// PruneAcknowledgements prunes packet acknowledgements from the store that have a sequence number less than or equal to the pruning sequence.
-// The number of packet acknowledgements pruned is equal to limit. Pruning only occurs after a channel has been upgraded.
-func (Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit, pruningSequence uint64) error {
-	return nil
+// PruneStalePacketData prunes packet acknowledgements and receipts that have a sequence number less than pruning sequence end.
+// The number of packet acks/receipts pruned is equal to limit. Pruning can only occur after a channel has been upgraded.
+//
+// Pruning sequence start keeps track of the packet ack/receipt that can be pruned next. When it reaches pruningSequenceEnd,
+// pruning is complete.
+func (k Keeper) PruneStalePacketData(ctx sdk.Context, portID, channelID string, limit uint64) (uint64, uint64, error) {
```
AdityaSripal, line 679: Sorry for all the conflicting names, but PacketData already has a different semantic, its the opaque bytes sent to the application. How about `PruneStalePacketState`?

AdityaSripal: Though honestly was fine with PruneAcknowledgements, since receipts are really just a tiny acknowledgement for core while the full ack is the ack for the app as well

modules/core/keeper/msg_server_test.go
chatton: happy with a follow up for this refactor

modules/core/keeper/msg_server_test.go
```diff
@@ -2016,3 +2016,77 @@ func (suite *KeeperTestSuite) TestUpdateConnectionParams() {
...
+
+			msg = channeltypes.NewMsgPruneStalePacketState(
+				path.EndpointA.ChannelConfig.PortID,
+				path.EndpointA.ChannelID,
+				10,
```
chatton, line 2075: should we add a test case that modifies the limit?

EDIT: nvm we test limit in `TestPruneStalePacketState`, I think I'm happy enough with that!

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -455,4 +455,9 @@ message MsgPruneAcknowledgements {
 }
 
 // MsgPruneAcknowledgementsResponse defines the response type for the PruneAcknowledgements rpc.
-message MsgPruneAcknowledgementsResponse {}
\ No newline at end of file
+message MsgPruneAcknowledgementsResponse {
+  // Number of sequences pruned (might include both packet acknowledgements and packet receipts).
+  uint64 pruned = 1;
+  // Number of sequences left after pruning.
+  uint64 left = 2;
```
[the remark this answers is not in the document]
chatton: I think `pruned` and `left` are fine. I think I ever so slightly prefer `prunedSequences` / `remainingSequences` but not enough that I will fight anyone over it!

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,323 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+		},
+		{
+			"success: stale packet state partially pruned",
+			func() {
+				// Send 10 packets from B -> A, creating 10 packet receipts and 10 packet acks on A.
+				suite.sendMockPackets(path.EndpointB, path.EndpointA, 10)
+			},
+			func() {
+				// Prune only 5 packet acks.
+				limit = 5
```
chatton, line 621: I think this test should set a limit of anything other than 5 so pruned/left are not both also equal to 5.

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,323 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+			},
+			func(pruned, left uint64) {
+				// We expect 5 to be left and sequenceStart == 6.
+				postPruneExpState(5, 5, 6)
+
+				// We expect 5 to be pruned and 5 left.
+				suite.Require().Equal(uint64(5), pruned)
+				suite.Require().Equal(uint64(5), left)
```
chatton, line 629: maybe set a limit of 6 so we would have 6 pruned and 4 left to make sure nothing funky is going on

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,323 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+			},
+			nil,
+		},
+		{
+			"success: stale packet state pruned, two upgrades",
```
chatton, line 634: nice consideration!

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,323 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+				// This is _before_ the first upgrade.
+				suite.sendMockPackets(path.EndpointB, path.EndpointA, 10)
+			},
+			func() {
+				// Prune 5 on A.
+				pruned, left, err := suite.chainA.App.GetIBCKeeper().ChannelKeeper.PruneStalePacketState(
+					suite.chainA.GetContext(),
+					path.EndpointA.ChannelConfig.PortID,
+					path.EndpointA.ChannelID,
+					5, // limit == 5
```
chatton, line 678: same comment as before, should probably use different values

chatton (approved):
LGTM, I absolutely love the test coverage, really nice job! Had just a few small suggestions for tweaks in the tests but nothing major

proto/ibc/core/channel/v1/tx.proto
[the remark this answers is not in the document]
charleenfei: i actually really like pruneSequences/remainingSequences, happy to push the commit to change

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -69,8 +69,8 @@ service Msg {
   // UpdateChannelParams defines a rpc handler method for MsgUpdateParams.
   rpc UpdateChannelParams(MsgUpdateParams) returns (MsgUpdateParamsResponse);
 
-  // PruneAcknowledgements defines a rpc handler method for MsgPruneAcknowledgements.
-  rpc PruneAcknowledgements(MsgPruneAcknowledgements) returns (MsgPruneAcknowledgementsResponse);
+  // PruneStalePacketState defines a rpc handler method for MsgPruneStalePacketState.
+  rpc PruneStalePacketState(MsgPruneStalePacketState) returns (MsgPruneStalePacketStateResponse);
```
damiannolan, line 73: My vote would be for `PruneAcknowledgements` based on @AdityaSripal's comment about receipts being kind of like a subset of acks used within ibc core. Or if not going with `PruneAcknowledgements` I would drop the word "stale" as its implicit that anything that can be pruned must be stale or not required anymore.

proto/ibc/core/channel/v1/tx.proto
```diff
@@ -454,5 +454,10 @@ message MsgPruneAcknowledgements {
...
 
-// MsgPruneAcknowledgementsResponse defines the response type for the PruneAcknowledgements rpc.
-message MsgPruneAcknowledgementsResponse {}
\ No newline at end of file
+// MsgPruneStalePacketStateResponse defines the response type for the PruneStalePacketState rpc.
+message MsgPruneStalePacketStateResponse {
+  // Number of sequences pruned (might include both packet acknowledgements and packet receipts).
+  uint64 pruned_sequences = 1;
+  // Number of sequences left after pruning.
+  uint64 remaining_sequences = 2;
```
damiannolan, line 462: I think we should add `total_` prefix to these fields. 

`pruned_sequences` to me suggests a `repeated` field whereas with the "total" prefix its explicit

damiannolan (commented):
Just leaving two initial comments on the proto api. 

Will review fully after lunch and I can also help push any follow up commits to the branch!

modules/core/04-channel/keeper/keeper.go
```diff
@@ -653,8 +671,46 @@ func (k Keeper) HasPruningSequenceStart(ctx sdk.Context, portID, channelID strin
...
+	end := pruningSequenceStart + limit
+	for ; start < end; start++ {
+		// Stop pruning if pruningSequenceStart has reached pruningSequenceEnd, pruningSequenceEnd is
+		// set to be equal to the _next_ sequence to be sent by the counterparty.
+		if start >= pruningSequenceEnd {
+			break
+		}
+
+		if k.HasPacketAcknowledgement(ctx, portID, channelID, start) {
+			k.deletePacketAcknowledgement(ctx, portID, channelID, start)
```
AdityaSripal, line 700: Nit, you can just delete here. Delete should be a no-op if it doesn't exist so you don't have to do the prior check has

modules/core/04-channel/keeper/keeper.go
```diff
@@ -653,8 +671,46 @@ func (k Keeper) HasPruningSequenceStart(ctx sdk.Context, portID, channelID strin
...
+		}
+
+		// Note packet receipts are only relevant for unordered channels.
+		if k.HasPacketReceipt(ctx, portID, channelID, start) {
+			k.deletePacketReceipt(ctx, portID, channelID, start)
```
AdityaSripal, line 705: Ditto

AdityaSripal (approved):
Approving as I only have a minor nit on optimistically deleting

Ok with any of the proposed names. Either the current, or Damian's suggestions. Let's come to final consensus and change once so we don't keep changing on every new opinion 😛

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,323 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+	err = path.EndpointA.ChanUpgradeOpen()
+	suite.Require().NoError(err)
+
+	err = path.EndpointA.UpdateClient()
+	suite.Require().NoError(err)
+}
+
+// sendMockPacket sends a packet from source to dest and acknowledges it on the source (completing the packet lifecycle).
+// Question(jim): find a nicer home for this?
+func (suite *KeeperTestSuite) sendMockPackets(source, dest *ibctesting.Endpoint, numPackets int) {
```
damiannolan, line 852: we should be able to replace endponts with passing the path. and then use `path.RelayPacket`

will look into it with @chatton

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,324 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+
+	err = path.EndpointA.ChanUpgradeOpen()
+	suite.Require().NoError(err)
+
+	err = path.EndpointA.UpdateClient()
+	suite.Require().NoError(err)
+}
+
+// sendMockPacket sends a packet from source to dest and acknowledges it on the source (completing the packet lifecycle).
+// Question(jim): find a nicer home for this?
```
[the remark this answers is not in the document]
damiannolan: just realised I should've replied here instead of a new comment :D https://github.com/cosmos/ibc-go/pull/5444/files#r1434023684

proto/ibc/core/channel/v1/tx.proto
colin-axner: `PruneAcknowledgements` was fine by me. I'd also propose `PrunePacketReceiveState`. My only gripe with `PruneStalePacketState` is that it doesn't indicate what type of packet state we are pruning

crodriguezvega: I will do then now the renaming to `PruneAcknowledgements` and MsgPruneAcknowledgements`.

modules/core/04-channel/keeper/keeper_test.go
damiannolan: https://github.com/cosmos/ibc-go/pull/5444/commits/998e7ba5096121ea3927a8b0ad2a4b87c7f58070

proto/ibc/core/channel/v1/tx.proto
crodriguezvega: Re-renaming done. ✅

modules/core/04-channel/keeper/keeper.go
crodriguezvega: Changing this now.

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,321 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+			suite.Require().ErrorIs(err, tc.expError)
+
+			// check on post state.
+			tc.post(pruned, left)
+		})
+	}
+}
+
+// UpgradeChannel performs a channel upgrade given a specific set of upgrade fields.
+// Question(jim): setup.coordinator.UpgradeChannel() wen?
```
damiannolan, line 825: 100 emoji

modules/core/04-channel/keeper/keeper_test.go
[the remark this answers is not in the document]
damiannolan: Can also look into moving this to a more generalized testing func in the testing pkg at some point! :)

damiannolan (approved):
LGTM!! 🚀 🎉 

Awesome work @DimitrisJim and amazing work on the testing, looks super thorough!
Thanks to everyone for their commits and reviews too! ❤️

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,321 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+
+				// channel upgraded, send additional packets and try and prune.
+				suite.sendMockPackets(path, 12)
+
+				// attempt to prune 5.
+				pruned, left, err := suite.chainA.App.GetIBCKeeper().ChannelKeeper.PruneAcknowledgements(
+					suite.chainA.GetContext(),
+					path.EndpointA.ChannelConfig.PortID,
+					path.EndpointA.ChannelID,
+					5,
```
colin-axner, line 757:
```suggestion
					17,
```
should we be attempting to prune post upgrade packet sends by setting it to a higher limit? To show that it safely returns after all possible have been pruned?

modules/core/04-channel/keeper/keeper.go
```diff
@@ -653,8 +665,42 @@ func (k Keeper) HasPruningSequenceStart(ctx sdk.Context, portID, channelID strin
...
+// The number of packet acks/receipts pruned is equal to limit. Pruning can only occur after a channel has been upgraded.
+//
+// Pruning sequence start keeps track of the packet ack/receipt that can be pruned next. When it reaches pruningSequenceEnd,
+// pruning is complete.
+func (k Keeper) PruneAcknowledgements(ctx sdk.Context, portID, channelID string, limit uint64) (uint64, uint64, error) {
+	if !k.HasPruningSequenceStart(ctx, portID, channelID) {
+		return 0, 0, errorsmod.Wrapf(types.ErrPruningSequenceStartNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
+	}
+
+	pruningSequenceStart := k.GetPruningSequenceStart(ctx, portID, channelID)
```
colin-axner, line 678: I find it odd that the get func here doesn't return a `found` bool. But I guess it is fine

modules/core/04-channel/keeper/keeper_test.go
```diff
@@ -541,3 +542,321 @@ func (suite *KeeperTestSuite) TestUnsetParams() {
...
+
+			receipts := suite.chainA.App.GetIBCKeeper().ChannelKeeper.GetAllPacketReceipts(suite.chainA.GetContext())
+			suite.Require().Len(receipts, int(expReceiptsLen))
+
+			start := suite.chainA.App.GetIBCKeeper().ChannelKeeper.GetPruningSequenceStart(suite.chainA.GetContext(), path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID)
+			suite.Require().Equal(start, expPruningSequenceStart)
+		}
+	)
+
+	testCases := []struct {
```
colin-axner, line 567: could we add a test case where a middle sequence doesn't have an ack stored (because a timeout occurred)?

colin-axner (approved):
Really top notch work @DimitrisJim! :crown: Thank you so much!!

modules/core/04-channel/keeper/keeper.go
damiannolan: I agree, I found I had the same reaction too! 

But apparently the argument is saving gas with a `Has` check? :D lol

damiannolan:
Opened https://github.com/cosmos/ibc-go/issues/5476 for testing improvements. Agree with @colin-axner's points!

modules/core/04-channel/keeper/keeper.go
colin-axner: Isn't this increasing gas? I suspect the majority of calls to this function to have the sequence set, which results in extra gas costs

damiannolan: I think it would increase gas in the successful case of it being set. And save gas in the scenario of it not being set

Issue #5437, opened 2023-12-18
Allow `host_connection_id` in version metadata to be empty on ChanOpenInit


srdtrk:
## Problem Definition

`host_connection_id` is not used by the controller. It is also very difficult for a controller contract to obtain the correct value for this parameter on ChanOpenInit (see https://github.com/cosmos/ibc-go/issues/3942).

## Proposal

We should allow this to be empty at `ChanOpenInit` and let the icahost fill in the value during version negotiation.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

GNaD13:
Hi @srdtrk, Can i take this issue?

crodriguezvega:
> Hi @srdtrk, Can i take this issue?

Yes, thank you @GNaD13!

PR #5533, opened 2024-01-07
Allow host_connection_id in version metadata to be empty on ChanOpenInit

GNaD13:
## Description

closes: #5437

```text
type: commit message
```

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
	- Improved interchain accounts' registration and governance proposal execution process.
- **Refactor**
	- Enhanced code readability by refactoring version string setup and handshake logic.
- **Tests**
	- Updated and added test cases for ensuring consistency and correctness in version handling and metadata validation.
- **Documentation**
	- Clarified comments on host connection identifiers in metadata functions.


modules/apps/27-interchain-accounts/host/keeper/handshake_test.go
```diff
@@ -47,16 +47,19 @@ func (suite *KeeperTestSuite) TestOnChanOpenTry() {
 		path     *ibctesting.Path
 		chanCap  *capabilitytypes.Capability
 		metadata icatypes.Metadata
+		version  string
 	)
 
 	testCases := []struct {
 		name     string
 		malleate func()
+		assert   func()
```
srdtrk, line 56: You shouldn't add an assert function since you can just check the `host_connection_id` after every successful handshake anyway.

modules/apps/27-interchain-accounts/types/metadata.go
```diff
@@ -169,6 +169,9 @@ func validateConnectionParams(metadata Metadata, controllerConnectionID, hostCon
 	}
 
 	if metadata.HostConnectionId != hostConnectionID {
+		if metadata.HostConnectionId == "" {
+			return nil
```
srdtrk, line 173: This is not how we should do it. It is possible that we want to add other checks after this in the future. So, the only `nil` returned should be at the end of the function, not inside this if statement

srdtrk (changes requested):
Thanks for the work @GNaD13. I noted some problems I had with the PR.

Also I'm wondering if we should change the `NewDefaultMetadata` function in controller's `OnChanInit` ([see here](https://github.com/cosmos/ibc-go/blob/677865d5a835bcf70baa33adde023fac32b92ef2/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L56)) to not require the `counterparty_connection_id`?

I could also take over the PR if you think these would be too much work for you @GNaD13 @crodriguezvega

crodriguezvega:
> Also I'm wondering if we should change the `NewDefaultMetadata` function in controller's `OnChanInit` ([see here](https://github.com/cosmos/ibc-go/blob/677865d5a835bcf70baa33adde023fac32b92ef2/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L56)) to not require the `counterparty_connection_id`?

That sounds fine to me, but that change would be API breaking. Do you think it's ok for users to wait until the v9 to release this? Otherwise we could add a new function `NewDefaultMetadataWithoutHostConnection` (or something like that) only with the controller connection ID.

> I could also take over the PR if you think these would be too much work for you @GNaD13 @crodriguezvega

That's also fine by me. Is it ok for you if we continue from here, @GNaD13? Thank you very much for opening the PR!

modules/apps/27-interchain-accounts/types/metadata.go
crodriguezvega: I think we need to be a bit carefully with the validation of `metadata.HostConnectionId` because it will be valid to be empty in `ChanOpenInit` but it should not be empty anymore in `ChanOpenAck`, right?

crodriguezvega:
> Yeah, that's also fine to me. It will be great if we could work together

Agree! Maybe then @srdtrk can comment the changes that would be needed and we can all together continue from there. :)

srdtrk:
@crodriguezvega I think we can break API in this PR to main. We can worry about the API if we backport (we might not backport). So I'd say, we should break `NewDefaultMetadata`. Sounds good, then I'll continue to give suggestions under my review.

modules/apps/27-interchain-accounts/types/metadata.go
srdtrk: I'd make the following suggestion @GNaD13 :
1. Revert your changes to validateConnectionParams.
2. In `ChanOpenTry`, before sending the metadata to validation, override `metadata.HostConnectionId` to be `connectionHops[0]` regardless of whether or not it is filled.
3. Send to validation after step 2.

Note that this is okay because the host is allowed to propose a new version string in `ChanOpenTry`. Does this sound good @crodriguezvega?

GNaD13: If so, I think we will also remove the validation in [ChanOpenInit](https://github.com/cosmos/ibc-go/blob/dc99ee43c1ade80094d56acd00ca16097b308d44/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L64)

GNaD13:
> @crodriguezvega I think we can break API in this PR to main. We can worry about the API if we backport (we might not backport). So I'd say, we should break `NewDefaultMetadata`. Sounds good, then I'll continue to give suggestions under my review.

@srdtrk I've updated the code. Thanks for your suggestion. Could you take a look pls?

srdtrk (commented):
Thanks for the improvements. I've added some followups. 

I'd also like you to update the end to end tests to use an empty `host_connection_id`. But I think we can open a separate PR for that if you don't want to.

modules/apps/27-interchain-accounts/host/keeper/handshake_test.go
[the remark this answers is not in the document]
GNaD13: I added a check for this

modules/apps/27-interchain-accounts/types/metadata.go
```diff
@@ -49,7 +49,8 @@ func NewDefaultMetadata(controllerConnectionID, hostConnectionID string) Metadat
 // NewDefaultMetadataString creates and returns a new JSON encoded version string containing the default ICS27 Metadata values
 // with the provided controller and host connection identifiers
 func NewDefaultMetadataString(controllerConnectionID, hostConnectionID string) string {
```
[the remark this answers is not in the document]
GNaD13: I removed it

GNaD13:
> Thanks for the improvements. I've added some followups.
> 
> I'd also like you to update the end to end tests to use an empty `host_connection_id`. But I think we can open a separate PR for that if you don't want to.

@srdtrk I've updated the code. Could you take a look pls?

e2e/tests/interchain_accounts/utils.go
```diff
@@ -0,0 +1,12 @@
+package interchainaccounts
+
+import (
+	icatypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/types"
+)
+
+func DefaultTestMetadataVersionString(controllerConnectionID, hostConnectionID string) string {
```
srdtrk, line 7: I'd like to pass an empty version string in the e2e test to ensure it works. So this function should not be needed and you can just use `NewDefaultMetadataString`

modules/apps/27-interchain-accounts/controller/migrations/v6/migrations_test.go
```diff
@@ -17,6 +17,12 @@ import (
 	ibcmock "github.com/cosmos/ibc-go/v8/testing/mock"
 )
 
+var (
+	metadata        = icatypes.NewMetadata(icatypes.Version, ibctesting.FirstConnectionID, ibctesting.FirstConnectionID, "", icatypes.EncodingProtobuf, icatypes.TxTypeSDKMultiMsg)
+	versionBytes, _ = icatypes.ModuleCdc.MarshalJSON(&metadata)
+	version         = string(versionBytes)
+)
+
```
srdtrk, line 25: Can you move this to inside `SetupTest` since it seems it is the only place it is used?

srdtrk (commented):
This looks very good. I added a few items. The only issue is that the callbacks module is not compiling since we broke the API. So you need to go into `modules/apps/callbacks` to fix it. After these changes, I'll give my approval.

GNaD13:
> Please don't worry about the e2e tests failing related to channel upgrades

yeah, I got it

modules/apps/27-interchain-accounts/controller/migrations/v6/migrations_test.go
GNaD13: I updated this code

e2e/tests/interchain_accounts/utils.go
srdtrk: I didn't mean you should change the function name. I meant that you could use the `NewDefaultMetadataString` function from icatypes instead of defining a new function here @GNaD13

modules/apps/27-interchain-accounts/controller/migrations/v6/migrations_test.go
srdtrk: This is resolved

srdtrk:
Please see https://github.com/cosmos/ibc-go/pull/5533#discussion_r1455253792

e2e/tests/interchain_accounts/utils.go
GNaD13: I got it

GNaD13: @srdtrk about the e2e test. If we use icatypes NewDefaultMetadataString with only controller port ID, it may get an error when ChanOpenInit check [here](https://github.com/decentrio/ibc-go/blob/ad3e34427557fc08f97caa01b727dc34a0d9317d/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L64). Should we use empty version so controller chain could create [metadata](https://github.com/decentrio/ibc-go/blob/ad3e34427557fc08f97caa01b727dc34a0d9317d/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L57) itself?

srdtrk: Oh, I see the issue now. Sorry about my oversight. We can go back to using your `DefaultTestMetadataVersionString`. I'll approve this PR after the e2e are passing

e2e/tests/upgrades/genesis_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 
+	interchainaccounts "github.com/cosmos/ibc-go/e2e/tests/interchain_accounts"
```
srdtrk, line 23: this doesn't cause a circular dependency right?

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -42,22 +42,23 @@ func (k Keeper) OnChanOpenInit(
 		err      error
 		metadata icatypes.Metadata
 	)
+
 	if strings.TrimSpace(version) == "" {
-		connection, err := k.channelKeeper.GetConnection(ctx, connectionHops[0])
+		_, err := k.channelKeeper.GetConnection(ctx, connectionHops[0])
 		if err != nil {
 			return "", err
 		}
```
srdtrk, line 50: If connection is not used any longer, this should just be removed

srdtrk (approved):
Lgtm. Just added 1 more improvement request.

e2e/tests/upgrades/genesis_test.go
GNaD13: yeah, I've checked it. It doesn't cause a circular dependency

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
GNaD13: I removed these code

crodriguezvega: I removed it, since the connection is retrieved anyway in `ChanOpenInit`, so if it cannot be found, it will already error there.

modules/apps/27-interchain-accounts/controller/keeper/handshake_test.go
```diff
@@ -32,6 +33,13 @@ func (suite *KeeperTestSuite) TestOnChanOpenInit() {
 			func() {},
 			nil,
 		},
+		{
+			"success: empty host connection ID",
+			func() {
+				path.EndpointB.ConnectionID = ""
```
crodriguezvega, line 39: What is this test case actually testing?

GNaD13: This test is for my old code. I removed it. Thank you sir

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -42,22 +42,18 @@ func (k Keeper) OnChanOpenInit(
 		err      error
 		metadata icatypes.Metadata
 	)
-	if strings.TrimSpace(version) == "" {
```
crodriguezvega, line 45: I am wondering now if we should just leave this as is. For our implementation of ICA controller it's easy to figure out the host connection ID, so we could just use it. Other implementations can leave it empty, if they cannot retrieve it, and with the change added to `OnChanOpenTry` in the host, then the host connection ID will be set. The change in `OnChanOpenTry` is actually what enables to leave the host connection ID empty in the proposed metadata of the controller. 

What do you think, @srdtrk?

modules/apps/27-interchain-accounts/types/metadata.go
```diff
@@ -33,11 +33,11 @@ func NewMetadata(version, controllerConnectionID, hostConnectionID, accAddress,
 }
 
 // NewDefaultMetadata creates and returns a new ICS27 Metadata instance containing the default ICS27 Metadata values
-// with the provided controller and host connection identifiers
-func NewDefaultMetadata(controllerConnectionID, hostConnectionID string) Metadata {
```
crodriguezvega, line 37: We could also just leave this function and `NewDefaultMetadataString` as they are, with two parameters, and if any implementation cannot provide a host connection ID, then they can just pass an empty string.

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
GNaD13: Yeah, I think we could do that. Actually, that is my first approach. The only thing is we need to allow the [validation](https://github.com/cosmos/ibc-go/blob/97a52e35f35dadfc08aa9e852196e151eeaa56af/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L59) on controller chain that allow empty host connectionID

crodriguezvega (approved):
Thank you, @GNaD13. The changes look good to me now.

srdtrk (approved):

Issue #5284, opened 2023-11-30
ICA host packet failure events not being emitted


dferendo:
## Summary of Bug

When a controller sends a packet to a host and the packet fails, the controller receives ["error handling packet: see events for details"](https://github.com/cosmos/ibc-go/tree/v8.0.0/modules/core/04-channel/types/acknowledgement.go#L16). When using the query `binaryd q interchain-accounts host packet-events` on the host chain, the event is not found. 

Digging deeper, in [msg_server.go](https://github.com/cosmos/ibc-go/blob/v8.0.0/modules/core/keeper/msg_server.go#L488), a cache context is created:  

```
	// Perform application logic callback
	//
	// Cache context so that we may discard state changes from callback if the acknowledgement is unsuccessful.
	cacheCtx, writeFn = ctx.CacheContext()
	ack := cbs.OnRecvPacket(cacheCtx, msg.Packet, relayer)
	if ack == nil || ack.Success() {
		// write application state changes for asynchronous and successful acknowledgements
		writeFn()
	}
```

The host [OnRecvPacket](https://github.com/cosmos/ibc-go/blob/v8.0.0/modules/apps/27-interchain-accounts/host/ibc_module.go#L113) is called and emits an event regardless successful or not:

```
	// Emit an event indicating a successful or failed acknowledgement.
	keeper.EmitAcknowledgementEvent(ctx, packet, ack, err)
````

Circling back to [msg_server.go](https://github.com/cosmos/ibc-go/blob/v8.0.0/modules/core/keeper/msg_server.go#L488), when the ack is not successful nothing is saved to the state thus the events submitted in OnRecvPacket are discarded. 

Issue #4173 can also be closed by fixing this bug and creating the e2e test. Possible solution is to add the events to the main context.

## Expected Behaviour

Events are emitted when ICA host packages fail

## Version

v8.0.0

## Steps to Reproduce

Emitting events in the [mocked ibc module](https://github.com/cosmos/ibc-go/blob/v8.0.0/testing/mock/ibc_module.go#L125) and checking if events are emitted in [msg_server_test](https://github.com/cosmos/ibc-go/blob/v8.0.0/modules/core/keeper/msg_server_test.go#L33) should reproduce this issue.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

migueldingli1997:
[...]

PR #5541, opened 2024-01-08
imp(core): allow huckleberry events with a prefix

srdtrk:
## Description

Event emissions on erroneous IBC application callbacks were removed due to the huckleberry [security advisory](https://github.com/cosmos/ibc-go/security/advisories/GHSA-3v7p-4x7p-4rx7). Since then, many users have been unable to debug their IBC applications effectively causing a considerable uptick in support requests across many chains.

This PR enables emission of events on erroneous IBC application callbacks by appending a prefix to all event attribute keys

closes: #5284

```text
imp(core): allow huckleberry events with a prefix
```

---

write a little note why.


modules/core/types/events.go
```diff
@@ -0,0 +1,29 @@
...
+	newEvents := make(sdk.Events, len(events))
+	for i, event := range events {
+		newAttributes := make([]sdk.Attribute, len(event.Attributes))
+		for j, attribute := range event.Attributes {
+			newAttributes[j] = sdk.NewAttribute(ErrorAttributeKeyPrefix+attribute.Key, attribute.Value)
+		}
+
+		// no need to append the error attribute prefix to the event type because
+		// the event type is not associated to a value that can be misinterpreted
+		newEvents[i] = sdk.NewEvent(event.Type, newAttributes...)
```
colin-axner, line 25: can we elaborate on this reasoning? I feel it is safer to prefix the type as well. Is there a direct benefit to not prefixing the type?

modules/core/types/events.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+const ErrorAttributeKeyPrefix = "ibccallbackerror-"
+
+// ConvertToErrorEvents converts all events to error events by appending the
+// error attribute prefix to each event's attribute key.
+func ConvertToErrorEvents(events sdk.Events) sdk.Events {
```
colin-axner, line 11: Can we make this more explicit to our use case? I don't really want to encourage reusing this short term solution. I would be in favour of keeping private and moving to the file it is used

modules/core/keeper/msg_server.go
```diff
@@ -494,6 +494,9 @@ func (k Keeper) RecvPacket(goCtx context.Context, msg *channeltypes.MsgRecvPacke
 	if ack == nil || ack.Success() {
 		// write application state changes for asynchronous and successful acknowledgements
 		writeFn()
+	} else {
+		// Modify events in cached context to reflect unsuccessful acknowledgement
+		ctx.EventManager().EmitEvents(coretypes.ConvertToErrorEvents(cacheCtx.EventManager().Events()))
```
colin-axner, line 499:
```suggestion
	} else if !ack.Success() {
		// Modify events in cached context to reflect unsuccessful acknowledgement
		ctx.EventManager().EmitEvents(coretypes.ConvertToErrorEvents(cacheCtx.EventManager().Events()))
```
I don't think the intention is to emit error events for async acks? Might be best to be explicit?

srdtrk: I think this suggestion is already enforced. It is not possible for `ack.Success()` to be true here anyway. Async acknowledgements are handled in the [if case](https://github.com/cosmos/ibc-go/blob/07fca7efc918dbc8e6cc04f13fd04c101bb87664/modules/core/keeper/msg_server.go#L494). Is there some misunderstanding?

modules/core/types/events.go
srdtrk: I thought it might be easier for users to index events by module that emitted the event. In that it wouldn't break their pre-existing clients too much.

The reasoning isn't solid but what I meant was that event types are only used for indexing attributes by module. Event types aren't the key to some value, they are only used to retrieve the attributes.

But I'm open to prefixing this too

modules/core/types/events.go
srdtrk: made private and moved to the file it is used in (`msg_server.go`)

modules/core/keeper/msg_server.go
charleenfei: I think async acks are handled in `ack == nil` but I don't see the harm in being explicit with `else if !ack.Success()`

modules/core/keeper/msg_server_test.go
```diff
@@ -419,6 +445,14 @@ func (suite *KeeperTestSuite) TestHandleAcknowledgePacket() {
 				// replay should not error as it is treated as a no-op
 				_, err := keeper.Keeper.Acknowledgement(*suite.chainA.App.GetIBCKeeper(), suite.chainA.GetContext(), msg)
 				suite.Require().NoError(err)
+
```
charleenfei, line 448: you could consider refactoring these test cases to use a function that asserts the expected test outcomes as in [this pr](https://github.com/cosmos/ibc-go/pull/5549/files) rather than checking against these `if` conditions, might be easier to read. but can also be done in a followup

srdtrk: Yeah, we could do it. I think this better be a followup because these "if conditions" are already used commonly in this file for `async`. So, such a refactor would change unrelated testing code too

modules/core/keeper/msg_server_test.go
```diff
@@ -419,6 +445,14 @@ func (suite *KeeperTestSuite) TestHandleAcknowledgePacket() {
...
+				if replay {
```
chatton, line 449: small nit, to be more consistent maybe we could make this a `tc.replay`, we seem to be mixing things like `tc.expPass` and standalone bools like `replay`.

modules/core/keeper/msg_server_test.go
```diff
@@ -436,6 +470,7 @@ func (suite *KeeperTestSuite) TestHandleTimeoutPacket() {
 		packet    channeltypes.Packet
 		packetKey []byte
 		path      *ibctesting.Path
+		noop      bool // indicate no-op
```
chatton, line 473: same suggestion about adding this to the test case rather than floating bool

srdtrk: done

modules/core/keeper/msg_server_test.go
srdtrk: done

charleenfei (approved):
Thanks for the changes!

modules/core/keeper/msg_server.go
colin-axner: ah yes, that's correct. I misread the first if statement

modules/core/keeper/msg_server_test.go
```diff
@@ -57,15 +58,15 @@ func (suite *KeeperTestSuite) TestHandleRecvPacket() {
 			suite.Require().NoError(err)
 
 			packet = channeltypes.NewPacket(ibctesting.MockPacketData, sequence, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, timeoutHeight, 0)
-		}, true, false},
+		}, true, false, false, false},
```
colin-axner, line 61: maybe we should open an issue to make these named args in a followup (4 booleans are hard to process what they indicate without a name)

modules/core/keeper/msg_server_test.go
```diff
@@ -186,14 +187,26 @@ func (suite *KeeperTestSuite) TestHandleRecvPacket() {
 				_, exists := suite.chainB.GetSimApp().ScopedIBCMockKeeper.GetCapability(suite.chainB.GetContext(), ibcmock.GetMockRecvCanaryCapabilityName(packet))
 				if tc.expRevert {
 					suite.Require().False(exists, "capability exists in store even after callback reverted")
+
+					// context events should contain error events
+					suite.Require().Contains(events, keeper.ConvertToErrorEvents(sdk.Events{ibcmock.NewMockRecvPacketEvent()})[0])
```
colin-axner, line 192: should we have not contains to ensure it doesn't have successful events as well?

modules/core/types/events.go
colin-axner: Hmm yea. My main concern is accidental misuse, that is using the event type to indicate an action occurred without caring about particular values. But I think you make a fair point that event types aren't the key to some value. I'm not too concerned about breaking pre-existing clients as pre-existing clients currently cannot access error events. If the primary use case here is simply for debugging, I'm not sure I see much benefit in not error prefixing the type defensively, but I don't feel strongly, just feel slightly uncomfortable allowing any surface area for misinterpretation of events

colin-axner (approved):
Everything looks very great to me! My only hesitation is not prefixing the event type, which I have left my thoughts on. I'm happy to go with whichever direction you think is best

modules/core/keeper/msg_server_test.go
srdtrk: added assetion

Issue #5523, opened 2024-01-05
When a channel reopens the ordering and metadata must not change


crodriguezvega:
Implement this pseudocode from the [spec](https://github.com/cosmos/ibc/blob/main/spec/app/ics-027-interchain-accounts/README.md#channel-lifecycle-management) in [`OnChanOpenInit` of the controller](https://github.com/cosmos/ibc-go/blob/fadf8f2b0ab184798d021d220d877e00c7634e26/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L24):

```typescript
// only open the channel if:
// - there is no active channel already set (with status OPEN)
// OR
// - there is already an active channel (with status CLOSED) AND
// the metadata matches exactly the existing metadata in the 
// version string of the active channel AND the ordering of the 
// new channel matches the ordering of the active channel.
activeChannelId, activeChannelFound = GetActiveChannelID(portId, connectionId)
if activeChannelFound {
  activeChannel = provableStore.get(channelPath(portId, activeChannelId))
  abortTransactionUnless(activeChannel.state === CLOSED)
  previousOrder = activeChannel.order
  abortTransactionUnless(previousOrder === order)
  previousMetadata = UnmarshalJSON(activeChannel.version)
  abortTransactionUnless(previousMetadata === metadata)
}
```

ThanhNhann:
Can I handle this issue?

crodriguezvega:
> Can I handle this issue? And I see the link you add with `OnChanOpenInit of the controller` maybe wrong

@ThanhNhann Thank you for the offer. The issues linked to #3996 are a bit time sensitive, since we want to include the whole feature in the release of v8.1 by the end of the month, so I think it will be better if we manage these internally. Thank you again!

srdtrk:
Why are we enforcing this? I'd like to allow ordering (and even encoding) to change.

crodriguezvega:
> Why do we want to enforce this? I'd like to allow ordering (and even encoding) to change.

Then you would be able to change ordering/encoding without the need to go through a channel upgrade. It sounds to me like a sneaky way of doing an upgrade because you remove the negotiation part with the counterparty. I know it is unlikely if the counterparty runs also ibc-go, but in a future world where the counterpart might run a different IBC implementation I guess it would be good to allow the counptarty also to have a saying in the change of the channel parameters, otherwise the controller is changing them unilaterally.

But I am fine to remove the restriction if @AdityaSripal agrees.

[...]

Issue #3996, opened 2023-07-03
Allow Unordered ICA Channels


srdtrk:
## Summary

Currently, ICA only allows ordered channels. This causes the channel to close if a timeout occurs, forcing the user to reopen it.

## Problem Definition

ICA only allows ordered channels because if ordering is not enforced, then the relayer could try malicious attacks by reordering/censoring some of the ica-controller's transactions. However, this requirement can be dropped because the ica-controller can choose not to send any new transactions until it receives an acknowledgement or a timeout packet.

## Proposal

```[tasklist]
- [ ] https://github.com/cosmos/ibc-go/issues/5522
- [ ] https://github.com/cosmos/ibc-go/issues/5523
- [ ] https://github.com/cosmos/ibc-go/issues/5524
- [ ] https://github.com/cosmos/ibc-go/issues/5580
- [ ] https://github.com/cosmos/ibc-go/issues/5629
- [ ] https://github.com/cosmos/ibc-go/issues/5602
- [ ] https://github.com/cosmos/ibc-go/issues/5605
- [ ] https://github.com/cosmos/ibc-go/issues/5525
- [ ] https://github.com/cosmos/ibc-go/issues/5534
- [ ] https://github.com/cosmos/ibc-go/issues/5526
```
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

migueldingli1997:
This would be very helpful for our use-case as well. We are using ICA round the clock with low timeout values for the transactions (because we want them to pass in a narrow time range, or not pass). If there is any event that causes this timeout then we're suddenly stuck with a closed channel that has to be re-opened.

womensrights:
Great to hear you are using ICA! Could you tell us a bit more about your precise use case if you are happy for it to be public? We had heard that ordered channels for ICA was a pain point for a few users but when we looked in more detail at the number of channel closures happening is actually really small. For example on Stride, the number of ICA timeouts has been 40/ 35,261 which is 0.11% of transactions since genesis and for Quicksilver it was 68 of 312,565 ICA txs since Jan. so approx 0.02% therefore the change would be a relatively small improvement it seems but would like to hear more about your own experience

hxrts:
> approx 0.02% therefore the change would be a relatively small improvement it seems

[...]

PR #5562, opened 2024-01-10
When a channel reopens the ordering and metadata must not change

chatton:
## Description

closes: #5523

I also did a small refactor of the test to check for specific errors. Hopefully it doesn't bloat the PR too much!

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/controller/keeper/handshake.go
+++ b/modules/apps/27-interchain-accounts/controller/keeper/handshake.go
@@ -72,8 +72,12 @@ func (k Keeper) OnChanOpenInit(
 			panic(fmt.Errorf("active channel mapping set for %s but channel does not exist in channel store", activeChannelID))
 		}
 
-		if channel.IsOpen() {
+		if channel.State == channeltypes.OPEN {
 			return "", errorsmod.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s is already OPEN", activeChannelID, portID)
+		}
+
+		if channel.Ordering != order {
+			return "", errorsmod.Wrapf(channeltypes.ErrInvalidChannelOrdering, "expected channel with ordering %s, got %s", order, channel.Ordering)
 		}
 
 		appVersion, found := k.GetAppVersion(ctx, portID, activeChannelID)
--- a/modules/apps/27-interchain-accounts/controller/keeper/handshake.go
+++ b/modules/apps/27-interchain-accounts/controller/keeper/handshake.go
@@ -77,7 +77,7 @@ func (k Keeper) OnChanOpenInit(
 		}
 
 		if channel.Ordering != order {
-			return "", errorsmod.Wrapf(channeltypes.ErrInvalidChannelOrdering, "expected channel with ordering %s, got %s", order, channel.Ordering)
+			return "", errorsmod.Wrapf(channeltypes.ErrInvalidChannelOrdering, "order cannot change when reopening a channel expected %s, got %s", channel.Ordering, order)
 		}
 
 		appVersion, found := k.GetAppVersion(ctx, portID, activeChannelID)
--- a/modules/apps/27-interchain-accounts/controller/keeper/handshake_test.go
+++ b/modules/apps/27-interchain-accounts/controller/keeper/handshake_test.go
@@ -25,12 +25,12 @@ func (suite *KeeperTestSuite) TestOnChanOpenInit() {
 	testCases := []struct {
 		name     string
 		malleate func()
-		expPass  bool
+		expError error
 	}{
 		{
 			"success",
 			func() {},
-			true,
+			nil,
 		},
 		{
 			"success: previous active channel closed",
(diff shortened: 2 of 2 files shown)


crodriguezvega:
Thanks, @chatton. Should we try to target all the PRs for https://github.com/cosmos/ibc-go/issues/3996 to a feature branch?

chatton:
whoops! Will do @crodriguezvega

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -72,10 +72,14 @@ func (k Keeper) OnChanOpenInit(
 			panic(fmt.Errorf("active channel mapping set for %s but channel does not exist in channel store", activeChannelID))
 		}
 
-		if channel.IsOpen() {
+		if channel.State == channeltypes.OPEN {
 			return "", errorsmod.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s is already OPEN", activeChannelID, portID)
 		}
 
+		if channel.Ordering != order {
+			return "", errorsmod.Wrapf(channeltypes.ErrInvalidChannelOrdering, "expected channel with ordering %s, got %s", order, channel.Ordering)
```
colin-axner, line 80:
```suggestion
			return "", errorsmod.Wrapf(channeltypes.ErrInvalidChannelOrdering, "order cannot change when reopening a channel expected %s, got %s", channel.Ordering, order)
```

colin-axner (approved):
Great cleanup! :) thanks for switching those tests to more explicit error checking

DimitrisJim (approved):
:rocket:

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
```diff
@@ -72,10 +72,14 @@ func (k Keeper) OnChanOpenInit(
 			panic(fmt.Errorf("active channel mapping set for %s but channel does not exist in channel store", activeChannelID))
 		}
 
-		if channel.IsOpen() {
+		if channel.State == channeltypes.OPEN {
```
crodriguezvega, line 75: Thinking now if we should change this condition to `channel.State != channeltypes.CLOSED`... I am thinking of the following scenario: an ICA channel is going through a channel upgrade and the channel state is either `FLUSHING` or `FLUSH_COMPLETE`, and a relayer submits a `MsgChannelOpenInit` with the same metadata version as the channel going through the upgrade. With the current check the opening handshake would go through, right? I guess we should prevent that?

If we need to change this, we can do it in a follow up PR.

crodriguezvega (commented):
Thank you, @chatton! Will you implement the restriction to make sure that the metadata shouldn't change in a follow up PR?

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
chatton: good call, maybe we can create a follow up issues for it, (this PR is not changing any logic in this condition)

chatton: This condition looks like it's already being handled [here](https://github.com/cosmos/ibc-go/blob/d58cf6f057b4df8986f7328622b6033995ca8a46/modules/apps/27-interchain-accounts/controller/keeper/handshake.go#L88-L90)

DimitrisJim: wouldn't it fail since channel open would try and claim and already claimed port cap?

chatton: `ChanOpenInit` on the controller doesn't seem to claim any capabilities, this happens in `OnChanOpenTry` on the host chain

AdityaSripal (commented):
I thought this was how we were going to very quickly enable changing from ORDERED to UNORDERED @srdtrk? This prevents that flow

colin-axner:
> I thought this was how we were going to very quickly enable changing from ORDERED to UNORDERED @srdtrk? This prevents that flow

We discussed this. At the moment there's no restriction preventing relayers from reopening channels on behalf of the ICA owner. If you allow the ordering to change, a malicious relayer could change the order type without the permission of the owner. We decided to move forward with the initial restriction and adopt a more flexible approach when there's better guarantees for changing ordering via reopening

modules/apps/27-interchain-accounts/controller/keeper/handshake.go
colin-axner: this seems like a valid concern to me. We should change to != CLOSED. Active channel gets set upon channel opening so we don't need to worry about edge cases with INIT and TRY

crodriguezvega: > wouldn't it fail since channel open init would try and claim and already claimed port cap?

The capability is claimed [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/ibc_middleware.go#L68), but the thing is that when reopening a closed ICA channel we are actually creating a new with the same metadata information as the previous one, so that's why claiming the capability will not fail.

> wouldn't it fail since channel open init would try and claim and already claimed port cap?

I will open an issue.

PR #5598, opened 2024-01-12
chore: recurring typo

Frierened:
Fixed one recurring typo "lastest". It also appears in the versioned_docs but I believe I shouldn't edit them.
Doesn't require testing or changelog entry.
Have a nice evening.


crodriguezvega (approved):
Thanks, @Frierened!

DimitrisJim (approved):
thanks!

Issue #5605, opened 2024-01-15
Allow channel version to change in icahost


srdtrk:
## Summary

This proposal aims to enhance the ICA channel reopening process by removing the consistency checks in channel ordering and version string from `icahost` only

## Problem Definition

Our current approach to transitioning from ordered to unordered ICA channels uses channel upgradability. We are not allowed to change the channel parameters while reopening an ICA channel. This is to prevent unauthorized relayers from modifying the ICA channel without ICA owner's consent.

## Proposal

We propose an approach where `icahost` is to accept changes in channel parameters upon reopening, while `icacontroller` maintains the right to reject these changes. This strategy offers two significant advantages:

1. **Flexibility for External Implementations**: External `icacontroller` implementations, particularly those prohibiting unauthorized reopening, can seamlessly transition their channels. They can do so by simply reopening with altered parameters.
2. **Future-proofing `icacontroller` Enhancements**: This change ensures that potential future enhancements to `icacontroller`, which might include provisions for reopening channels with new parameters, are not impeded by `icahost` restrictions.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

srdtrk:
closed by https://github.com/cosmos/ibc-go/pull/5613

PR #5613, opened 2024-01-16
imp(ica/host): removed previous version validation check

srdtrk:
## Description

closes: #5605 

```text
imp(ica/host): removed previous version validation check
```

---

write a little note why.


srdtrk:
Review this carefully and consider security implications before approving

modules/apps/27-interchain-accounts/host/keeper/handshake.go
```diff
@@ -53,15 +53,6 @@ func (k Keeper) OnChanOpenTry(
 		if channel.IsOpen() {
 			return "", errorsmod.Wrapf(icatypes.ErrActiveChannelAlreadySet, "existing active channel %s for portID %s is already OPEN", activeChannelID, portID)
 		}
-
-		appVersion, found := k.GetAppVersion(ctx, portID, activeChannelID)
```
colin-axner, line 57: just for my own reasoning. The previous metadata has:
version, controller/host connection id's, address, encoding, tx type

Only the host connection id and address cannot change? The host connection id check above in `ValidateHostMetadata`, but is the address? What happens if the controller reopens a channel modifying the interchain acc address to be different than what it is. It looks like it'll be overwritten later on. I feel like we should document these cases or make them more explicit

modules/apps/27-interchain-accounts/host/keeper/handshake_test.go
```diff
@@ -95,6 +95,27 @@ func (suite *KeeperTestSuite) TestOnChanOpenTry() {
...
+			func() {
+				// create a new channel and set it in state
+				ch := channeltypes.NewChannel(channeltypes.CLOSED, channeltypes.ORDERED, channeltypes.NewCounterparty(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID), []string{path.EndpointA.ConnectionID}, TestVersion)
+				suite.chainB.GetSimApp().GetIBCKeeper().ChannelKeeper.SetChannel(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, ch)
+
+				// set the active channelID in state
+				suite.chainB.GetSimApp().ICAHostKeeper.SetActiveChannelID(suite.chainB.GetContext(), path.EndpointB.ConnectionID, path.EndpointA.ChannelConfig.PortID, path.EndpointB.ChannelID)
+
+				// set the previous encoding to be proto3json. And then set the new encoding to be protobuf.
+				metadata.Encoding = icatypes.EncodingProto3JSON
```
damiannolan, line 109: > And then set the new encoding to be protobuf.

This is because `path.EndpointA.ChannelConfig.Version` is passed as `counterpartyVersion` on L312, correct? 
Maybe we could make that more explicit in the comment?

damiannolan (approved):
Spent a few mins to reason through flows with this check being removed! I think it should be all good and makes sense. Most of the main checks are still done in `ValidateHostMetadata`

LGTM! 🙏🏻

modules/apps/27-interchain-accounts/host/keeper/handshake.go
damiannolan: If I'm understanding correctly, the address is looked up with a key built from host connection id and counterparty port id, right? And counterparty port ID needs to proven by core before getting here and if host connection id cannot change then it should be all fine.

colin-axner: It's just a bit odd to me because the relayer can pass in:
metadata.Address = invalid-address
and our code won't error
it'll return 
metadata.Address = valid-address

just feels a bit odd if a middleware relies on the value before the callback

srdtrk: That's right. Moreover, the removed code (`IsPreviousMetadataEqual`) never actually checked if the addresses were the same... It only checked other fields

colin-axner: I like being a bit defensive/explicit in these cases:
```go
// loosen restrictions, but keep explicit checks
// the host connectionID and address must remain the same between reopening
appVersion, found := k.GetAppVersion(ctx, portID, activeChannelID)
if !found {
	panic(fmt.Errorf("active channel mapping set for %s, but channel does not exist in channel store", activeChannelID))
}
		
previousMetadata, err := MetadataFromVersion(previousVersion)
if err != nil {
     return err // or panic
 }
 
if previousMetadata.HostConnectionId != metadata.HostConnectionId {
   return err
}
 
if previousMetadata.Address != metadata.Address {
     return err
}   
 

```

but I don't feel strongly

damiannolan: I was going to suggest the same about at least explicitly checking the address. I didn't think of the middleware point you brought up, good catch. 

I'd be in favour of adding the extra defensive checks because why not! But also not really gonna block on it

colin-axner: ah okay, thanks for the context. Should we add a comment in case someone has the same question as me? 

maybe something like:
```go
// if a channel is being reopened, we allow the controller to propose new fields
// which are not exactly the same as the previous. The provided address will 
// be overwritten with the correct one before the metadata is returned
```

srdtrk: Added the comment

crodriguezvega: Should we check at least that version is `ics27-1`?

srdtrk: This is still being checked [here](https://github.com/cosmos/ibc-go/blob/198b525a0b870df2c3198e56151266bd334daf3e/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L42) @crodriguezvega

PR #5651, opened 2024-01-18
fix: add replay protection on upgraded channels 

colin-axner:
## Description

After performing a channel upgrade, we have reached a new checkpoint state where we should not process any old packets. Replay protection is currently given by packet receipts on unordered channels and next sequence recv on ordered channels. Because upgrading may result in invalid packet receipt state (either by pruning old packet receipts or upgrading from an ordered to unordered channel) we need to add an additional defensive check which ensures packets are only being received after the next checkpoint sequence, otherwise replay attacks (such as double spends) are possible

closes: #XXXX

```text
type: commit message
```

---

write a little note why.


modules/core/04-channel/keeper/keeper.go
```diff
@@ -623,17 +623,17 @@ func (k Keeper) HasInflightPackets(ctx sdk.Context, portID, channelID string) bo
 	return iterator.Valid()
 }
 
-// SetPruningSequenceEnd sets the channel's pruning sequence end to the store.
-func (k Keeper) SetPruningSequenceEnd(ctx sdk.Context, portID, channelID string, sequence uint64) {
+// SetRecvStartSequence sets the channel's recv start sequence to the store.
```
damiannolan, line 626: should we add an extra note, with something like:

> NOTE: RecvStartSequence is only set for channels which have undergone an upgrade. This sequences acts as a checkpoint for pruning historical channel data 

could also add a note saying its used as the upper bound when pruning

modules/core/04-channel/keeper/packet.go
```diff
@@ -197,9 +197,20 @@ func (k Keeper) RecvPacket(
 		return errorsmod.Wrap(err, "couldn't verify counterparty packet commitment")
 	}
 
+	// REPLAY PROTECTION: The recvStartSequence will prevent historical proofs from allowing replay
+	// attacks on packets processed in previous lifecycles of a channel. After a successful channel
+	// upgrade all packets under the recvStartSequence will have been processed and thus should be
+	// rejected.
+	recvStartSequence, _ := k.GetRecvStartSequence(ctx, packet.GetDestPort(), packet.GetDestChannel())
+	if packet.GetSequence() < recvStartSequence {
+		return errorsmod.Wrap(types.ErrPacketReceived, "packet already processed in previous channel upgrade")
```
chatton, line 206: Does `previous version` sound a bit better here?

```suggestion
		return errorsmod.Wrap(types.ErrPacketReceived, "packet already processed in a previous channel version")
```

modules/core/integration_test.go
damiannolan, line 1: what a juicy test 🚀 🥇

damiannolan (approved):
Looks great, lgtm! I just left a minor suggestion on godoc. 
Thanks @colin-axner!

modules/core/04-channel/keeper/keeper.go
charleenfei: or maybe something specific like "This sequence indicates the first packet sequence which should be processed after an upgrade"

modules/core/integration_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	clienttypes "github.com/cosmos/ibc-go/v8/modules/core/02-client/types"
+	channeltypes "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+	host "github.com/cosmos/ibc-go/v8/modules/core/24-host"
+	ibctesting "github.com/cosmos/ibc-go/v8/testing"
+	ibcmock "github.com/cosmos/ibc-go/v8/testing/mock"
+)
+
+// If packet receipts are pruned, it is possible to double spend via a
```
chatton, line 11:
```suggestion
// If packet receipts are pruned, it should not be possible to double spend via a
```

chatton (approved):
LGTM really nice job with this one, left a few minor wording suggestions but fix looks solid, nice job :1st_place_medal:

modules/core/04-channel/keeper/packet.go
crodriguezvega, line 206: Would it be possible to do `return types.ErrNoOpMsg` instead of returning an error? Just the same as we do in case the packet receipt for the sequence already exists, so that we prevent consuming fees. We could log the error message instead to give this information.

crodriguezvega (commented):
Nice work, @colin-axner!

charleenfei (approved):
super nice work @colin-axner on the find and the fix!

modules/core/integration_test.go
charleenfei: i understood colin's comment to mean that double spending works if the proof is submitted, that's why we added the check against the packet recv seq start in addition to the proof verification? but im not super opinionated, i think it makes sense this way as well.

modules/core/04-channel/keeper/packet.go
colin-axner: There's a possibility one relayer is attempting to flush while another relayer completes the upgrade, but I think this would be very rare and in such cases, I'd still prefer to reject the entire transaction (the relayer is probably flushing stale packets in the tx)

colin-axner: I wanted to stay away from "version" here since you can upgrade a channel without changing its version

modules/core/04-channel/keeper/keeper.go
colin-axner: I thought about this, but none of the other set functions for `nextSequenceRecv` `nextSequenceAck` indicate their purpose. I wasn't sure where we should be documenting their behaviour (presumably at least in the spec). I can add a comment in you'd like?

modules/core/integration_test.go
colin-axner: yea exactly, I can update the comment to be more specific

colin-axner: https://github.com/cosmos/ibc-go/pull/5651/commits/f23573484920ac0974af4780f49145b9423a3819 https://github.com/cosmos/ibc-go/pull/5651/commits/7d5f6e85fad217a1567828756a1738947698a7a1

AdityaSripal (approved):
Excellent work @colin-axner !!

modules/core/04-channel/keeper/keeper.go
colin-axner: https://github.com/cosmos/ibc-go/pull/5651/commits/b2a1f50b39029594c48c1189e4853e7b24d597f6

colin-axner:
I made `setRecvStartSequence` private to avoid any accidental pitfalls

modules/core/04-channel/keeper/packet.go
```diff
@@ -197,9 +197,20 @@ func (k Keeper) RecvPacket(
 		return errorsmod.Wrap(err, "couldn't verify counterparty packet commitment")
 	}
 
+	// REPLAY PROTECTION: The recvStartSequence will prevent historical proofs from allowing replay
+	// attacks on packets processed in previous lifecycles of a channel. After a successful channel
+	// upgrade all packets under the recvStartSequence will have been processed and thus should be
+	// rejected.
+	recvStartSequence, _ := k.GetRecvStartSequence(ctx, packet.GetDestPort(), packet.GetDestChannel())
```
DimitrisJim, line 204: shouldn't we only perform check if `found`? This defaults to `0` and would every packet would be marked as `recv`ed, no?

DimitrisJim (approved):
ignore brain-fart review, lgtm!

PR #5664, opened 2024-01-18
items from first security audit session

crodriguezvega:
## Description

I have created this branch where we can tackled the issues from the audit. @DimitrisJim @charleenfei please pick any items from the list and push to this branch, I thought doing it like this would be easier and faster than opening individual issues (and then we only need to backport one PR). I think the issues are small enough that should be easy to review all of them in one PR.

closes: #XXXX

```text
type: commit message
```

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/keeper/upgrade.go
+++ b/modules/core/04-channel/keeper/upgrade.go
@@ -626,9 +626,21 @@ func (k Keeper) ChanUpgradeCancel(ctx sdk.Context, portID, channelID string, err
 		return errorsmod.Wrap(commitmenttypes.ErrInvalidProof, "cannot submit an empty error receipt proof unless the sender is authorized to cancel upgrades AND channel is not in FLUSHCOMPLETE")
 	}
 
-	// the error receipt should also have a sequence greater than or equal to the current upgrade sequence.
-	if errorReceipt.Sequence < channel.UpgradeSequence {
-		return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be greater than or equal to current upgrade sequence (%d)", errorReceipt.Sequence, channel.UpgradeSequence)
+	// the error receipt should also have a sequence greater than or equal to the current upgrade sequence,
+	// except when the channel state if FLUSHCOMPLETE, in which case the sequences must match. This is required
+	// to guarantee that this channel end does not cancel the upgrade in case the counterparty has completed
+	// the upgrade and moved to OPEN, initiated a new upgrade (and thus incremented the upgrade sequence) and
+	// cancelled, all in the same block. Otherwise, the channel would open, but one end would have upgraded and
+	// the other not.
+	// DO NOT DELETE: this acts as protection against replay attacks.
+	if channel.State == types.FLUSHCOMPLETE {
+		if errorReceipt.Sequence != channel.UpgradeSequence {
+			return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be equal to current upgrade sequence (%d)", errorReceipt.Sequence, channel.UpgradeSequence)
+		}
+	} else {
+		if errorReceipt.Sequence < channel.UpgradeSequence {
+			return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be greater than or equal to current upgrade sequence (%d)", errorReceipt.Sequence, channel.UpgradeSequence)
+		}
 	}
 
 	// get underlying connection for proof verification
--- a/modules/core/04-channel/keeper/upgrade_test.go
+++ b/modules/core/04-channel/keeper/upgrade_test.go
@@ -1649,6 +1649,30 @@ func (suite *KeeperTestSuite) TestChanUpgradeCancel() {
 			expError: types.ErrInvalidUpgradeSequence,
 		},
 		{
+			name: "error receipt sequence not equal to channel upgrade sequence when channel in FLUSHCOMPLETE",
+			malleate: func() {
+				channel := path.EndpointA.GetChannel()
+				channel.State = types.FLUSHCOMPLETE
+				path.EndpointA.SetChannel(channel)
+
+				var ok bool
+				errorReceipt, ok = suite.chainB.GetSimApp().IBCKeeper.ChannelKeeper.GetUpgradeErrorReceipt(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
+				suite.Require().True(ok)
+
+				errorReceipt.Sequence = path.EndpointA.GetChannel().UpgradeSequence - 1
+
+				suite.chainB.GetSimApp().IBCKeeper.ChannelKeeper.SetUpgradeErrorReceipt(suite.chainB.GetContext(), path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, errorReceipt)
+
+				suite.coordinator.CommitBlock(suite.chainB)
+
+				suite.Require().NoError(path.EndpointA.UpdateClient())
+
+				upgradeErrorReceiptKey := host.ChannelUpgradeErrorKey(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID)
+				errorReceiptProof, proofHeight = suite.chainB.QueryProof(upgradeErrorReceiptKey)
+			},
+			expError: types.ErrInvalidUpgradeSequence,
+		},
+		{
 			name: "connection not found",
 			malleate: func() {
 				channel := path.EndpointA.GetChannel()
--- a/modules/core/keeper/msg_server.go
+++ b/modules/core/keeper/msg_server.go
@@ -954,7 +954,7 @@ func (k Keeper) ChannelUpgradeConfirm(goCtx context.Context, msg *channeltypes.M
 
 	// Move channel to OPEN state if both chains have finished flushing in-flight packets.
 	// Counterparty channel state has been verified in ChanUpgradeConfirm.
-	if msg.CounterpartyChannelState == channeltypes.FLUSHCOMPLETE && !k.ChannelKeeper.HasInflightPackets(ctx, msg.PortId, msg.ChannelId) {
+	if channel.State == channeltypes.FLUSHCOMPLETE && msg.CounterpartyChannelState == channeltypes.FLUSHCOMPLETE {
 		upgrade, found := k.ChannelKeeper.GetUpgrade(ctx, msg.PortId, msg.ChannelId)
 		if !found {
 			return nil, errorsmod.Wrapf(channeltypes.ErrUpgradeNotFound, "failed to retrieve channel upgrade: port ID (%s) channel ID (%s)", msg.PortId, msg.ChannelId)


DimitrisJim:
sgtm, would prefer PRs since its an easier workflow to not make mistakes in but lets give this a shot. Would be nice to sync up via slack when people plan to push so we don't end up with weird conflicts. also no force pushing!

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -185,13 +186,36 @@ func (suite *TypesTestSuite) TestIsUpgradeError() {
...
+		{
+			"true with wrapped upgrade err",
+			func() {
+				upgradeError := types.NewUpgradeError(1, types.ErrInvalidChannel)
+				err = errorsmod.Wrap(upgradeError, "wrapped upgrade error")
+			},
+			true,
+		},
+		{
+			"true with Errof upgrade error",
```
DimitrisJim, line 198:
```suggestion
			"true with Errof wrapped upgrade error",
```

crodriguezvega (commented):
Changes look great to me, pending the decision on the order of calling the callback. Great work @DimitrisJim and @charleenfei completing all the audit issues in less than 24 hours! 🦾

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -626,9 +627,21 @@ func (k Keeper) ChanUpgradeCancel(ctx sdk.Context, portID, channelID string, err
...
+	// the upgrade and moved to OPEN, initiated a new upgrade (and thus incremented the upgrade sequence) and
+	// cancelled, all in the same block. Otherwise, the channel would open, but one end would have upgraded and
+	// the other not.
+	// DO NOT DELETE: this acts as protection against replay attacks.
+	if channel.State == types.FLUSHCOMPLETE {
+		if errorReceipt.Sequence != channel.UpgradeSequence {
+			return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be equal to current upgrade sequence (%d)", errorReceipt.Sequence, channel.UpgradeSequence)
+		}
+	} else {
+		if errorReceipt.Sequence < channel.UpgradeSequence {
```
chatton, line 642: feels a bit odd to not have an `else if` here instead of an `else` followed by an `if`

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -1574,28 +1574,63 @@ func (suite *KeeperTestSuite) TestChanUpgradeCancel() {
...
+			},
 			expError: nil,
 		},
 		{
-			name: "success with flushing state",
+			name: "success with flush complete state",
 			malleate: func() {
 				channel := path.EndpointA.GetChannel()
-				channel.State = types.FLUSHING
+				channel.State = types.FLUSHCOMPLETE
```
chatton, line 1586: [nit] we can use `SetChannelState` instead of setting the whole channel here.

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -1574,28 +1574,63 @@ func (suite *KeeperTestSuite) TestChanUpgradeCancel() {
...
+				errorReceiptProof, proofHeight = suite.chainB.QueryProof(upgradeErrorReceiptKey)
 			},
 			expError: nil,
 		},
 		{
 			name: "upgrade cannot be cancelled in FLUSHCOMPLETE with invalid error receipt",
 			malleate: func() {
+				channel := path.EndpointA.GetChannel()
+				channel.State = types.FLUSHCOMPLETE
+				path.EndpointA.SetChannel(channel)
```
chatton, line 1612: same nit about `SetChannelState`

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -1615,22 +1650,21 @@ func (suite *KeeperTestSuite) TestChanUpgradeCancel() {
...
 			malleate: func() {
-				path.EndpointA.Chain.DeleteKey(host.ChannelKey(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID))
-			},
-			expError: types.ErrChannelNotFound,
-		},
-		{
-			name: "upgrade not found",
-			malleate: func() {
-				path.EndpointA.Chain.DeleteKey(host.ChannelUpgradeKey(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID))
+				channel := path.EndpointA.GetChannel()
```
chatton, line 1655: ditto

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -185,13 +186,36 @@ func (suite *TypesTestSuite) TestIsUpgradeError() {
...
+		{
+			"true with wrapped upgrade err",
+			func() {
+				upgradeError := types.NewUpgradeError(1, types.ErrInvalidChannel)
+				err = errorsmod.Wrap(upgradeError, "wrapped upgrade error")
+			},
+			true,
+		},
+		{
+			"true with Errof wrapped upgrade error",
```
chatton, line 198:
```suggestion
			"true with Errorf wrapped upgrade error",
```

modules/core/04-channel/types/upgrade_test.go
```diff
@@ -185,13 +186,36 @@ func (suite *TypesTestSuite) TestIsUpgradeError() {
 			func() {},
 			true,
 		},
+		{
```
chatton, line 189: might be nice to add a test case for a deeply nested error

```golang
		{
			"true with nested Errorf wrapped upgrade error",
			func() {
				err = fmt.Errorf("%w", fmt.Errorf("%w", fmt.Errorf("%w", types.NewUpgradeError(1, types.ErrInvalidChannel))))
			},
			true,
		},
```

modules/core/04-channel/types/upgrade.go
```diff
@@ -119,8 +119,17 @@ func (u *UpgradeError) GetErrorReceipt() ErrorReceipt {
...
+	for {
+		_, ok := err.(*UpgradeError)
+		if ok {
+			return true
+		}
+
+		if err = errors.Unwrap(err); err == nil {
+			return false
+		}
+	}
```
chatton, line 134: the presence of this function, means we can actually update the `Unwrap` fn of the `UpgradeError` type, I believe it can just be

```golang
// Unwrap returns the base error that caused the upgrade to fail.
func (u *UpgradeError) Unwrap() error {
	return errors.Unwrap(u.err)
}
```
And the corresponding test can be updated to 

```golang
// TestUpgradeErrorUnwrap tests that the underlying error is not modified when Unwrap is called.
func (suite *TypesTestSuite) TestUpgradeErrorUnwrap() {
	baseUnderlyingError := errorsmod.Wrap(types.ErrInvalidChannel, "base error")
	wrappedErr := errorsmod.Wrap(baseUnderlyingError, "wrapped error")
	upgradeError := types.NewUpgradeError(1, wrappedErr)

	originalUpgradeError := upgradeError.Error()
	unWrapped := errors.Unwrap(upgradeError)

	postUnwrapUpgradeError := upgradeError.Error()

	suite.Require().Equal(baseUnderlyingError, unWrapped, "unwrapped error was not equal to base underlying error")
	suite.Require().Equal(originalUpgradeError, postUnwrapUpgradeError, "original error was modified when unwrapped")
}
```

I think I like this better since it becomes more of a direct passthrough.

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -626,9 +627,21 @@ func (k Keeper) ChanUpgradeCancel(ctx sdk.Context, portID, channelID string, err
...
-	// the error receipt should also have a sequence greater than or equal to the current upgrade sequence.
-	if errorReceipt.Sequence < channel.UpgradeSequence {
-		return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be greater than or equal to current upgrade sequence (%d)", errorReceipt.Sequence, channel.UpgradeSequence)
+	// the error receipt should also have a sequence greater than or equal to the current upgrade sequence,
+	// except when the channel state if FLUSHCOMPLETE, in which case the sequences must match. This is required
+	// to guarantee that this channel end does not cancel the upgrade in case the counterparty has completed
+	// the upgrade and moved to OPEN, initiated a new upgrade (and thus incremented the upgrade sequence) and
+	// cancelled, all in the same block. Otherwise, the channel would open, but one end would have upgraded and
+	// the other not.
+	// DO NOT DELETE: this acts as protection against replay attacks.
```
colin-axner, line 636: Could we do something like this: https://github.com/cosmos/ibc-go/pull/5651/files#diff-927ca91645398beb24073f511453a335cb319ec6112ccdbe1506f389ad013a78R200

```go
// REPLAY PROTECTION: <why it provides replay protection>
```

chatton (commented):
Looks  great thank you! I left a couple of small suggestions

modules/core/04-channel/keeper/upgrade.go
DimitrisJim: how strong of an opinion on this? :laughing: I tend to like this sort of `if one_type_of_condition: ... else: if other_type_of_condition` over going with `else if` but can change if I'm the weird one.

modules/core/04-channel/types/upgrade.go
DimitrisJim: this would return `nil` on unwrapping a simple wrapped err no? `UpgradeError { err: types.ErrInvalidChannel }` would call `errors.Unwrap(types.ErrInvalidChannel)` which returns `nil` yea?

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -347,8 +347,10 @@ func (k Keeper) WriteUpgradeAckChannel(ctx sdk.Context, portID, channelID string
 	}
 
 	if !k.HasInflightPackets(ctx, portID, channelID) {
+		previousState := channel.State
 		channel.State = types.FLUSHCOMPLETE
 		k.SetChannel(ctx, portID, channelID, channel)
+		k.Logger(ctx).Info("channel state updated", "port-id", portID, "channel-id", channelID, "previous-state", previousState, "new-state", channel.State)
```
colin-axner, line 353: Hmm, I guess the reason we had to the logger at the bottom is that `startFlushing` might move the channel state from OPEN to FLUSHING, but, the previous channel state also could have been FLUSHING. A bit of a rock and a hard place

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -626,9 +627,20 @@ func (k Keeper) ChanUpgradeCancel(ctx sdk.Context, portID, channelID string, err
...
 
-	// the error receipt should also have a sequence greater than or equal to the current upgrade sequence.
-	if errorReceipt.Sequence < channel.UpgradeSequence {
-		return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be greater than or equal to current upgrade sequence (%d)", errorReceipt.Sequence, channel.UpgradeSequence)
+	// REPLAY PROTECTION: the error receipt should also have a sequence greater than or equal to the current upgrade sequence,
+	// except when the channel state is FLUSHCOMPLETE, in which case the sequences must match. This is required
+	// to guarantee that this channel end does not cancel the upgrade in case the counterparty has completed
+	// the upgrade and moved to OPEN, initiated a new upgrade (and thus incremented the upgrade sequence) and
+	// cancelled, all in the same block. Otherwise, the channel would open, but one end would have upgraded and
+	// the other not.
```
colin-axner, line 635:
```suggestion
	// REPLAY PROTECTION: The error receipt MUST have a sequence greater than or equal to the current upgrade sequence,
	// except when the channel state is FLUSHCOMPLETE, in which case the sequences MUST match. This is required
	// to guarantee that when a counterparty successfully completes an upgrade and moves to OPEN, this channel
	// cannot cancel its upgrade.  Without this strict sequence check, it would be possible for the counterparty
	// to complete its upgrade, move to OPEN, initiate a new upgrade (and thus increment the upgrade sequence) and
	// then cancel the new upgrade, all in the same block. This results in a valid error receipt being written at channel.UpgradeSequence + 1. 
	// The desired behaviour in this circumstance is for this channel to complete its current upgrade despite proof
	// of an error receipt at a greater upgrade sequence
```

modules/core/04-channel/keeper/upgrade.go
```diff
@@ -626,9 +627,20 @@ func (k Keeper) ChanUpgradeCancel(ctx sdk.Context, portID, channelID string, err
...
+	if channel.State == types.FLUSHCOMPLETE {
+		if errorReceipt.Sequence != channel.UpgradeSequence {
+			return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be equal to current upgrade sequence (%d)", errorReceipt.Sequence, channel.UpgradeSequence)
```
colin-axner, line 638:
```suggestion
			return errorsmod.Wrapf(types.ErrInvalidUpgradeSequence, "error receipt sequence (%d) must be equal to current upgrade sequence (%d) when the channel is in FLUSHCOMPLETE", errorReceipt.Sequence, channel.UpgradeSequence)
```

modules/core/04-channel/keeper/upgrade.go
colin-axner: we could remove the else entirely? 

```go
if channel.State == FLUSHCOMPLETE && errorReceipt.Sequence != channel.UpgradeSequence {
    return err
}

if errorReceipt < channel.UpgradeSequence {
    return err
}
```

DimitrisJim: flattened if down as a good compromise.

colin-axner (approved):
Thanks everyone pushing commits! Big thanks to @DimitrisJim for following up on review comments!

modules/core/04-channel/types/upgrade.go
damiannolan: Yeah, `nil` is returned if `err` is just a single unwrapped error in that case

damiannolan (approved):
Thank you all! 🚀

AdityaSripal (approved):
Nice work!

PR #5668, opened 2024-01-19
fix: allow `RecvPacket` and `WriteAcknowledgement` when channel is `FLUSHING` or `FLUSHCOMPLETE`

crodriguezvega:
## Description

While doing testing with hermes Luca an I found that hermes couldn't relay packets when the channel end was `FLUSHING` because `RecvPacket` RPC handler failed on `WriteAcknowledgement`.

closes: #XXXX

### Commit Message / Changelog Entry

```text
type: commit message
```

see the [guidelines](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) for commit messages. (view raw markdown for examples)

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [x] Targeted PR against the correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#pull-request-targeting)).
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/build/building-modules/11-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [commit message](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) to be used for the changelog entry in the PR description for review.
- [ ] Re-reviewed `Files changed` in the Github PR explorer.
- [ ] Review `Codecov Report` in the comment section below once CI passes.

PR #5671, opened 2024-01-21
imp: check order in validate basic of `MsgRegisterInterchainAccount`

crodriguezvega:
## Description

Also found while testing with hermes: a valid `MsgRegisterInterchainAccount` messages can be constructed and submitted with `NONE` order. The channel will not be created because an [error is returned in `ValidateBasic` of `MsgChannelOpenInit`](https://github.com/cosmos/ibc-go/blob/04531d83bf51d949f4bf2d61e88e419d1620ea32/modules/core/04-channel/types/channel.go#L71), but I thought it would be nicer to do the same check in the `ValidateBasic` of `MsgRegisterInterchainAccount`.

However, I was also thinking that maybe we actually shouldn't do this check at all in `ValidateBasic` and just set the order to `ORDERED` if the [value passed in `registerInterchainAccount`](https://github.com/cosmos/ibc-go/blob/04531d83bf51d949f4bf2d61e88e419d1620ea32/modules/apps/27-interchain-accounts/controller/keeper/account.go#L54) is `NONE`. That way I think we preserve backwards compatibility better, because upstream users don't need to change the existing code they may have creating a `MsgRegisterInterchainAccount`, otherwise, they will have to change their code and set a valid ordering. 

We could leave the `ValidateBasic` check of this PR and merge it to main, but not backport; and then in `release/v8.1.x` we could implement the other solution. 

Looking forward to reading people's thoughts.

closes: #XXXX

```text
type: commit message
```

---

write a little note why.


modules/core/04-channel/types/channel.go
```diff
@@ -68,7 +70,7 @@ func (ch Channel) ValidateBasic() error {
 	if ch.State == UNINITIALIZED {
 		return ErrInvalidChannelState
 	}
-	if !(ch.Ordering == ORDERED || ch.Ordering == UNORDERED) {
+	if !slices.Contains([]Order{ORDERED, UNORDERED}, ch.Ordering) {
```
crodriguezvega, line 73: I took the liberty to make this small improvement.

modules/apps/27-interchain-accounts/controller/types/msgs_test.go
```diff
@@ -72,6 +72,13 @@ func TestMsgRegisterInterchainAccountValidateBasic(t *testing.T) {
 			},
 			false,
 		},
+		{
+			"order is not valid",
+			func() {
+				msg.Order = channeltypes.NONE
```
crodriguezvega, line 78: Small nit I just realised: in the protos we called this field `order` for `MsgRegisterInterchainAccount, while in `MsgChannelOpenInit` is called `ordering`. Should we rename `order` to `ordering` for consistency?

DimitrisJim: yea, think we should!

modules/apps/27-interchain-accounts/controller/types/msgs.go
```diff
@@ -41,6 +42,10 @@ func (msg MsgRegisterInterchainAccount) ValidateBasic() error {
 		return errorsmod.Wrap(err, "invalid connection ID")
 	}
 
+	if !slices.Contains([]channeltypes.Order{channeltypes.ORDERED, channeltypes.UNORDERED}, msg.Order) {
```
DimitrisJim, line 45: I asked about this in security audit and though it does indeed get handled later on I do think having explicit ordering check here also makes sense. Only worry is something like https://github.com/cosmos/ibc-go/pull/5668 occurring again, maybe we can add these checks in spec too?

modules/apps/27-interchain-accounts/controller/types/msgs_test.go
crodriguezvega: Yeah, go for it!

modules/apps/27-interchain-accounts/controller/types/msgs.go
crodriguezvega: Sounds good to include it in the spec as well. I have [this PR](https://github.com/cosmos/ibc/pull/1047) with ICS 27 updates, so I can include it there.

modules/apps/27-interchain-accounts/controller/types/msgs_test.go
DimitrisJim: ah, would be nice to backport that change and just saw about BP this PR. can open as separate instead to have it backport cleanly

Issue #5771, opened 2024-01-30
Remove legacy params subspace args from ibc keepers


damiannolan:
## Summary

We recently migrated to self-managed params within our modules. I think we should be able to remove the `legacySubspace` args from the `NewKeeper` funcs of each module/submodule within the codebase.

This needs to be removed before ibc-go v11

---

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

neitdung:
Hi @damiannolan, can I take this one?

damiannolan:
@neitdung I've assigned you to the issue! Thank you very much <3

PR #5779, opened 2024-01-31
 refactor: remove legacySubspace args from the NewKeeper

neitdung:
## Description

- Remove  `legacySubspace`  args from `NewKeeper` funcs of each module/submodule
- Change `m.keeper.legacySubspace.GetParamSet(ctx, &params)` to `m.keeper.GetParams(ctx)`

closes: #5771  

```text
refactor: remove legacy params subspace args from ibc keepers
```

---

write a little note why.


modules/apps/27-interchain-accounts/controller/keeper/migrations.go
```diff
@@ -55,8 +55,7 @@ func (m Migrator) AssertChannelCapabilityMigrations(ctx sdk.Context) error {
 // MigrateParams migrates the controller submodule's parameters from the x/params to self store.
 func (m Migrator) MigrateParams(ctx sdk.Context) error {
 	if m.keeper != nil {
-		var params controllertypes.Params
-		m.keeper.legacySubspace.GetParamSet(ctx, &params)
+		params := m.keeper.GetParams(ctx)
```
colin-axner, line 58: I believe this migration can no longer be successfully performed, so I think we should panic if a chain tries to execute it. (I would remove all the code within the `if m.keeper != nil` block and replace it with a panic, indicating that a in-place migration must occur to the `v8.x` line before upgrading to `v9.x`

neitdung: @colin-axner Thanks for pointing that for me but I think we should return `ErrInvalidVersion` instead panic because this will not change function return type then we don't need to remove migration test and after this function, it still panic if `MigrateParams` be executed:

colin-axner:
I'm going to mark this pr as draft. Thanks @neitdung for the contribution! Internally there is some hesitation to remove the ability to migrate to v9 directly from a pre-v8 version unless necessary. We are going to double check with the SDK on if this will be required regardless and then we will get back to you on what our desired approach is

crodriguezvega:
@neitdung We discussed internally and decided that users will have to upgrade through either v8 or v9. The SDK team will also force users to upgrade through v0.47 or v0.50, so that our upgrade path would be in lockstep with theirs. For that reason, we cannot tackle this issue until v9 is released. I will close the PR for now and I have added the issue already to the v10 milestone, so that we don't lose sight of it. Thank you very much for opening this PR; hopefully you can pick this up later in the year. :)

Issue #5310, opened 2023-12-05
Querier Approach for Conditional Clients


srdtrk:
## Summary

This proposal introduces a querier-based mechanism in `ibc-go` to enable conditional clients, based on the requirements and workflows outlined in [#5112](https://github.com/cosmos/ibc-go/issues/5112). It specifically enables light clients to query other light clients through gRPC, offering a non-intrusive solution. A key feature of this proposal is the introduction of a non-state-modifying proof verification query within the `02-client` module, which streamlines the process of implementing conditional clients, particularly within the `08-wasm` module. The document details necessary modifications to both the `02-client` and `08-wasm` modules and includes a [proof of concept branch](https://github.com/cosmos/ibc-go/tree/serdar/poc-querier-conditional-clients). Additionally, the proposal contrasts this approach with an alternative method based on a gRPC message server, to provide a comprehensive view of potential gRPC based solutions.

## Problem Definition

Issue [#5112](https://github.com/cosmos/ibc-go/issues/5112) introduced the concept of conditional execution in light clients, dependent on the state of another light client. Traditional light client modules in our framework lack the inherent capability to execute code conditionally based on the state of other modules. While we have extended the functionality in the `08-wasm` module (see [#5192](https://github.com/cosmos/ibc-go/issues/5192)) by allowing chains to provide a custom querier, a more comprehensive solution is needed to facilitate conditional execution between light clients.

At present, the interaction between light client modules and the core `02-client` module is unidirectional: the `02-client` can only send information to light clients without receiving any queries in return. Our proposal aims to enable light clients to query other light clients (or even modules if enabled) via the [`GRPCQueryRouter`](https://github.com/cosmos/cosmos-sdk/blob/v0.50.1/baseapp/grpcserver.go#L23) in BaseApp. 

The `02-client` already includes a [query gRPC service](https://github.com/cosmos/ibc-go/blob/v8.0.0/proto/ibc/core/client/v1/query.proto#L13-L63) in its keeper, which enables querying light client states. A gap in the current model is the lack of queries for proof verification. To address this, we propose introducing non-state-modifying proof verification queries within the `02-client` module. This enhancement will reduce the engineering complexity and dependencies involved in building conditional clients.
Here is an example workflow to illustrate the proposed solution based on [#5112](https://github.com/cosmos/ibc-go/issues/5112):

1. Packet is sent on an optimistic rollup.
2. State header for this execution is submitted to Celestia.
[...]

PR #5821, opened 2024-02-07
feat: add rpc `VerifyMembershipProof` - querier approach for conditional clients

damiannolan:
## Description

Adds the new `Query` service rpc `VerifyMembershipProof` outlined in #5310. Big thanks to @srdtrk for the initial deep dive on this issue.

closes: #5310 

---

write a little note why.


damiannolan:
Would like to add a couple of more tests:
- 08-wasm stargate querier 
- Gas is correctly consumed from context

Please raise any more in the comments if applicable!

proto/ibc/core/client/v1/query.proto
```diff
@@ -205,3 +216,27 @@ message QueryUpgradedConsensusStateResponse {
...
+  // optional time delay
+  uint64 time_delay = 6;
+  // optional block delay
+  uint64 block_delay = 7;
+}
+
+// QueryVerifyMembershipProofResponse is the response type for the Query/VerifyMembershipProof RPC method
+message QueryVerifyMembershipProofResponse {
+  // result indicating success or failure of proof verification.
+  bool result = 1;
```
damiannolan, line 241: This could also be `success`, dropping the comment here if anyone feels one is better than the other. I feel indifferent

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -328,3 +328,54 @@ func (k Keeper) UpgradedConsensusState(c context.Context, req *types.QueryUpgrad
...
+	clientState, found := k.GetClientState(cachedCtx, req.ClientId)
+	if !found {
+		return nil, status.Error(codes.NotFound, errorsmod.Wrap(types.ErrClientNotFound, req.ClientId).Error())
+	}
+
+	if err := clientState.VerifyMembership(ctx, k.ClientStore(cachedCtx, req.ClientId), k.cdc, req.ProofHeight, req.TimeDelay, req.BlockDelay, req.Proof, req.MerklePath, req.Value); err != nil {
+		return &types.QueryVerifyMembershipProofResponse{
+			Result: false,
+		}, nil
+	}
```
damiannolan, line 376: Should we log the error or not? Seems like possibly worth doing if we are essentially swallowing it with the return value.

proto/ibc/core/client/v1/query.proto
```diff
@@ -60,6 +62,15 @@ service Query {
   rpc UpgradedConsensusState(QueryUpgradedConsensusStateRequest) returns (QueryUpgradedConsensusStateResponse) {
     option (google.api.http).get = "/ibc/core/client/v1/upgraded_consensus_states";
   }
+
+  // VerifyMembershipProof queries an IBC light client for proof verification of a value at a given key path.
+  rpc VerifyMembershipProof(QueryVerifyMembershipProofRequest) returns (QueryVerifyMembershipProofResponse) {
```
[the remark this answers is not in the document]
damiannolan: If that's the consensus then sure thing! I just went off what was written in the issue write up :)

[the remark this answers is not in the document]
colin-axner: I like `VerifyMembership`

proto/ibc/core/client/v1/query.proto
colin-axner: I like `success`, not sure what a bool for `result` would mean

colin-axner:
```
if resp.Success {

}
```

vs 
```
if resp.Result {

}
```

damiannolan: Aye, when its written like that I completely agree 👍🏻

modules/core/02-client/keeper/grpc_query.go
damiannolan: So you'd opt for returning an error? Only thing is you'll never have a response where `success: false` in that case, but it's probably fine 🤷🏻‍♂️ I could take either personally.

proto/ibc/core/client/v1/query.proto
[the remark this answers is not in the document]
damiannolan: sorted!

modules/light-clients/08-wasm/types/querier_test.go
```diff
@@ -157,6 +161,81 @@ func (suite *TypesTestSuite) TestStargateQuery() {
 				})
 			},
 		},
+		{
+			"success: verify membership query",
```
damiannolan, line 165: This test sets a key-value in the ibc store, using `"mock-key": "mock-value"`.  

It registers the query callback for `VerifyMembership` on the `Status` call to the wasm client using the mockVM. 
It also needs to register a callback for `VerifyMembership` since this is all mocked, we can still unmarshal and verify it tho within the callback registered for `VerifyMembership`.

The only alternative I see to this is setting up an additional ibctesting path with a regular tendermint client, and querying some proof to be verified by that using its client ID as the target. Given that we use a custom `wasmtesting.Endpoint` for these tests, I decided not to mix and match endpoint functionality between testing pkgs. I think the tests serves it purpose, it routes, decodes and verifies a proof.

crodriguezvega: Maybe for clarification you could add a comment of why the `VerifyMembership` callback is needed in the test and explain how the flow goes?

crodriguezvega (approved):
Nice and clean. Thank you for the PR, @damiannolan; and thank you @srdtrk for the ground work.

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -328,3 +328,54 @@ func (k Keeper) UpgradedConsensusState(c context.Context, req *types.QueryUpgrad
...
+	defer func() {
+		ctx.GasMeter().ConsumeGas(cachedCtx.GasMeter().GasConsumed(), "verify membership query")
+	}()
+
+	clientState, found := k.GetClientState(cachedCtx, req.ClientId)
+	if !found {
+		return nil, status.Error(codes.NotFound, errorsmod.Wrap(types.ErrClientNotFound, req.ClientId).Error())
+	}
+
+	if err := clientState.VerifyMembership(ctx, k.ClientStore(cachedCtx, req.ClientId), k.cdc, req.ProofHeight, req.TimeDelay, req.BlockDelay, req.Proof, req.MerklePath, req.Value); err != nil {
```
colin-axner, line 372: Is this supposed to be acting on the `ctx`? 
```suggestion
	if err := clientState.VerifyMembership(cachedCtx, k.ClientStore(cachedCtx, req.ClientId), k.cdc, req.ProofHeight, req.TimeDelay, req.BlockDelay, req.Proof, req.MerklePath, req.Value); err != nil {
```

? (should this have a test if so?)

damiannolan: 100% great catch, will add the state change to the test I mentioned above. It should catch this :)

damiannolan: actually maybe it would not catch it, since `cacheCtx` is used for `k.ClientStore()`

will add the test regardless, we should probably pass the `cacheCtx` instead of `ctx` as well 👍🏻

modules/core/02-client/keeper/grpc_query_test.go
```diff
@@ -770,3 +774,143 @@ func (suite *KeeperTestSuite) TestQueryClientParams() {
...
+			suite.SetupTest() // reset
+
+			path = ibctesting.NewPath(suite.chainA, suite.chainB)
+			path.Setup()
+
+			tc.malleate()
+
+			ctx := suite.chainA.GetContext()
+			initialGas := ctx.GasMeter().GasConsumed()
+			res, err := suite.chainA.QueryServer.VerifyMembership(ctx, req)
```
colin-axner, line 899: would be nice if we could have some state change to verify in tests

modules/core/02-client/keeper/grpc_query_test.go
```diff
@@ -770,3 +774,143 @@ func (suite *KeeperTestSuite) TestQueryClientParams() {
...
+				bz, err := suite.chainB.Codec.Marshal(&channel)
+				suite.Require().NoError(err)
+
+				channelProof, proofHeight := path.EndpointB.QueryProof(host.ChannelKey(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID))
+
+				merklePath := commitmenttypes.NewMerklePath(host.ChannelPath(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID))
+				merklePath, err = commitmenttypes.ApplyPrefix(suite.chainB.GetPrefix(), merklePath)
+				suite.Require().NoError(err)
+
+				req = &types.QueryVerifyMembershipRequest{
```
colin-axner, line 802: should this be the default setup? I guess it is okay since we assert the error

modules/core/02-client/keeper/grpc_query.go
damiannolan: Seems like my hypothesis was correct here. Passing `cachedCtx` anyways to avoid any nonsense or trickery

modules/light-clients/08-wasm/types/querier_test.go
damiannolan: I added a comment :)

modules/light-clients/08-wasm/types/querier.go
```diff
@@ -128,9 +128,13 @@ func NewDefaultQueryPlugins() *QueryPlugins {
 // AcceptListStargateQuerier allows all queries that are in the provided accept list.
 // This function returns protobuf encoded responses in bytes.
 func AcceptListStargateQuerier(acceptedQueries []string) func(sdk.Context, *wasmvmtypes.StargateQuery) ([]byte, error) {
+	defaultAcceptList := []string{
```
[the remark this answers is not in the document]
damiannolan: done!

modules/core/02-client/keeper/grpc_query_test.go
damiannolan: done in https://github.com/cosmos/ibc-go/pull/5821/commits/da3b02e8a551704dfd77fd516fe314a4efc615f4

modules/core/02-client/keeper/grpc_query_test.go
damiannolan: you mean moved to the test body rather than constructed in malleate?

I don't feel too strongly about it! Let me know if you'd like it changed before merge 👍🏻

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -328,3 +328,55 @@ func (k Keeper) UpgradedConsensusState(c context.Context, req *types.QueryUpgrad
 		UpgradedConsensusState: protoAny,
 	}, nil
 }
+
+// VerifyMembership implements the Query/VerifyMembership gRPC method
+func (k Keeper) VerifyMembership(c context.Context, req *types.QueryVerifyMembershipRequest) (*types.QueryVerifyMembershipResponse, error) {
```
crodriguezvega, line 333: Should we document this addition and the fact that there is going to be a set of default allowed queries made available to the Querier? Maybe in the 08-wasm docs?

crodriguezvega:
I added the `backport-to-v8.2.x` label to the PR, but I guess we also need to create a new release branch for 08-wasm that will depend on ibc-go v8.2 and backport to that branch the 08-wasm changes of the PR? Like `release/v0.2.x+ibc-go-v8.2.x-wasmvm-v2.0.x` (and then we can also upgrade to the latest wasmvm version)?

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -328,3 +328,55 @@ func (k Keeper) UpgradedConsensusState(c context.Context, req *types.QueryUpgrad
...
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "empty request")
+	}
+
+	if err := host.ClientIdentifierValidator(req.ClientId); err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
```
colin-axner, line 340: I think we should error if the client type is solo machine?

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -328,3 +328,55 @@ func (k Keeper) UpgradedConsensusState(c context.Context, req *types.QueryUpgrad
 		UpgradedConsensusState: protoAny,
 	}, nil
 }
+
+// VerifyMembership implements the Query/VerifyMembership gRPC method
```
colin-axner, line 332: should we add more godoc? (no state changes occur, we charge gas if possible, meant for state machine usage, delegates to light cilents)

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -328,3 +328,55 @@ func (k Keeper) UpgradedConsensusState(c context.Context, req *types.QueryUpgrad
...
+	}
+
+	if len(req.Proof) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "empty proof")
+	}
+
+	if req.ProofHeight.IsZero() {
+		return nil, status.Error(codes.InvalidArgument, "proof height must be non-zero")
+	}
+
```
colin-axner, line 349: I think proof height is allowed to be empty? (for localhost?)

modules/core/02-client/keeper/grpc_query.go
colin-axner: localhost as well?

modules/core/02-client/keeper/grpc_query_test.go
colin-axner: fine as is

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -328,3 +328,55 @@ func (k Keeper) UpgradedConsensusState(c context.Context, req *types.QueryUpgrad
...
+	defer func() {
+		ctx.GasMeter().ConsumeGas(cachedCtx.GasMeter().GasConsumed(), "verify membership query")
+	}()
+
+	clientState, found := k.GetClientState(cachedCtx, req.ClientId)
+	if !found {
+		return nil, status.Error(codes.NotFound, errorsmod.Wrap(types.ErrClientNotFound, req.ClientId).Error())
+	}
+
+	if err := clientState.VerifyMembership(cachedCtx, k.ClientStore(cachedCtx, req.ClientId), k.cdc, req.ProofHeight, req.TimeDelay, req.BlockDelay, req.Proof, req.MerklePath, req.Value); err != nil {
```
colin-axner, line 372: if we ever add a mock client, I think this would be a good test case we could use it on. I think it is nice to be testing the state change in 02-client rather than 08-wasm, but it is fine for now

colin-axner (approved):
Excellent work!!! Thanks for all the followup changes. My only requested addition is to disable light clients on the verify membership query which we don't want to support initially to limit any potential side effects we aren't trying to account for. I'm okay doing this in a followup though since the diffs should be quite minimal and mainly an addition

modules/core/02-client/keeper/grpc_query.go
chatton: do we also want to error if the client is not active, i.e.

`k.GetClientStatus(cachedCtx, clientState, req.ClientId) != exported.Active`

chatton: > I think we should error if the client type is solo machine?

This would involve adding solomachine types as a concrete dependency in 02-client which I think is currently not the case, (it is already with localhost though).

modules/light-clients/08-wasm/types/querier.go
```diff
@@ -130,7 +135,7 @@ func NewDefaultQueryPlugins() *QueryPlugins {
 func AcceptListStargateQuerier(acceptedQueries []string) func(sdk.Context, *wasmvmtypes.StargateQuery) ([]byte, error) {
 	return func(ctx sdk.Context, request *wasmvmtypes.StargateQuery) ([]byte, error) {
 		// A default list of accepted queries can be added here.
```
chatton, line 137: think this comment can be adjusted since we are adding this default list now

chatton (approved):
Looks good! Left some small suggestions but nothing major :+1:

modules/core/02-client/keeper/grpc_query.go
damiannolan: https://github.com/cosmos/ibc-go/issues/5846

modules/core/02-client/keeper/grpc_query.go
damiannolan: > I think we should error if the client type is solo machine?

Why would we error if it is solomachine client type? State changes are discarded so no sequence incrementing will happen. Shouldn't it be fine to let it go through, just sig verify on whatever bytes? Is it that solomachine is only signing one particular set of bytes for a one sequence? i.e. you will be able to verify for example, a channel end signed at sequence x? so in theory the `VerifyMembership` query is only available for proof one value at a particular point in time. Does that make sense?

> localhost as well?

Yeah, if you query a chain for a proof and have access to an rpc for a header to get the app hash then you should be able to verify off chain and it would be spammy(?) I guess to allow a query to go through like that. Fee like there's no reason to allow it.

> This would involve adding solomachine types as a concrete dependency in 02-client which I think is currently not the case, (it is already with localhost though)

Solomachine client type is in `exported` afaik.

modules/core/02-client/keeper/grpc_query.go
damiannolan: if we disable localhost then we should be sweet? :)

modules/core/02-client/keeper/grpc_query.go
damiannolan: Created this issue: https://github.com/cosmos/ibc-go/issues/5848

modules/core/02-client/keeper/grpc_query.go
damiannolan: done in https://github.com/cosmos/ibc-go/pull/5821/commits/00f59e866d2d8df3e2aca3c31ce9c68340bd4763

modules/core/02-client/keeper/grpc_query.go
colin-axner: solomachine client type is a string, it can be duplicated

> Is it that solomachine is only signing one particular set of bytes for a one sequence?

yes it is.

PR #5827, opened 2024-02-09
02-client routing: implement `VerifyUpgradeAndUpdateState` in light client modules

crodriguezvega:
## Description

closes: #XXXX

```text
type: commit message
```

---

write a little note why.


modules/core/keeper/msg_server.go
```diff
@@ -69,17 +69,13 @@ func (k Keeper) UpdateClient(goCtx context.Context, msg *clienttypes.MsgUpdateCl
...
-	if err != nil {
-		return nil, err
-	}
-
-	if err = k.ClientKeeper.UpgradeClient(ctx, msg.ClientId, upgradedClient, upgradedConsState,
-		msg.ProofUpgradeClient, msg.ProofUpgradeConsensusState); err != nil {
+	if err := k.ClientKeeper.UpgradeClient(
+		ctx, msg.ClientId,
+		msg.ClientState.Value,
+		msg.ConsensusState.Value,
```
crodriguezvega, line 75: @colin-axner @damiannolan I think just passing doing `.Value` here and passing the bytes of the proton-encoded client and consensus states might not be enough. See test failing [here](https://github.com/cosmos/ibc-go/actions/runs/7849754768/job/21423773040?pr=5827), it looks like if we just pass the bytes we miss the type URL information that we need to unmarshal the bytes.

modules/light-clients/07-tendermint/light_client_module.go
```diff
@@ -295,30 +303,45 @@ func (lcm LightClientModule) VerifyUpgradeAndUpdateState(
...
+
+	if err := cdc.UnmarshalInterface(newClient, &newClientState); err != nil {
 		return err
 	}
-
+	newTmClientState, ok := newClientState.(*ClientState)
+	if !ok {
+		return errorsmod.Wrapf(clienttypes.ErrInvalidClient, "expected client state type %T, got %T", (*ClientState)(nil), newClientState)
+	}
 	if err := newClientState.Validate(); err != nil {
```
crodriguezvega, line 319: @colin-axner @damiannolan This validation fails if the trust level is not within [0, 1/3] (see test failing [here](https://github.com/cosmos/ibc-go/actions/runs/7849754768/job/21423772705?pr=5827)). However, the upgraded client state that the counterparty chain commits should have all the custom fields zeroed out [since that's what we verify in `VerifyUpgradeAndUpdateState` function in the `ClientState`](https://github.com/cosmos/ibc-go/blob/d957ef80d72a260af3f6e66bbf5d7bea85473e2b/modules/light-clients/07-tendermint/upgrade.go#L73). So I think we either cannot do this validation here or we need to validate only the fields that shouldn't be zeroed out.

modules/core/keeper/msg_server.go
damiannolan: See https://github.com/cosmos/ibc-go/pull/5827/commits/3e311639ef002c2272ed02e11cbd5aee1bc543d3

We should unmarshal to concrete types and not interfaces within the light client module.

modules/light-clients/07-tendermint/light_client_module.go
damiannolan: I saw this failing test when working on the one above. I didn't dig any deeper on this one yet. But from what you're saying here it sounds correct. I'm not sure exactly what pieces of code we'll need to change yet, can follow up o this one next week! :)

modules/core/02-client/keeper/client.go
```diff
@@ -122,46 +121,41 @@ func (k Keeper) UpdateClient(ctx sdk.Context, clientID string, clientMsg exporte
 
 // UpgradeClient upgrades the client to a new client state if this new client was committed to
 // by the old client at the specified upgrade height
-func (k Keeper) UpgradeClient(ctx sdk.Context, clientID string, upgradedClient exported.ClientState, upgradedConsState exported.ConsensusState,
-	upgradeClientProof, upgradeConsensusStateProof []byte,
+func (k Keeper) UpgradeClient(
+	ctx sdk.Context,
+	clientID string,
+	upgradedClient, upgradedConsState, upgradeClientProof, upgradeConsensusStateProof []byte,
```
damiannolan, line 127: Isn't this an API breaking change? shouldn't the API remain the same here in the 02-client keeper and continue to pass the `exported` interfaces.

We can then pass `.Value` bytes when calling into the light client module.

modules/core/keeper/msg_server.go
damiannolan, line 75: essentially push this down a level into 02-client

modules/core/keeper/msg_server_test.go
```diff
@@ -888,7 +888,7 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
 		tc.setup()
 
 		ctx := suite.chainA.GetContext()
-		_, err = keeper.Keeper.UpgradeClient(*suite.chainA.App.GetIBCKeeper(), ctx, msg)
+		_, err = suite.chainA.GetSimApp().GetIBCKeeper().UpgradeClient(ctx, msg)
```
damiannolan, line 891: decided to change this while I was here, makes no difference but kinda looks weird how it was 🧹 

I think these tests could be improved or cleaned up some more. But maybe outside the scope of this PR

modules/core/02-client/keeper/client.go
crodriguezvega: It is API breaking, but I think that this refactoring will be API breaking anyway and will be released in a major release. I basically did here [a similar change as was done in `CreateClient`](https://github.com/cosmos/ibc-go/blob/f4ee20bf8c9139c7806c6410785ae37073bc7702/modules/core/02-client/keeper/client.go#L20).

modules/core/keeper/msg_server.go
crodriguezvega: Thanks, Damian. I changed this to unmarshal to interface because [the tests are generally using `client types.MarshalClientState`](https://github.com/cosmos/ibc-go/blob/055a3e39c3da309e013ec4763f78d33b74be59c9/modules/core/02-client/keeper/client_test.go#L481), which [calls `cdc.MarshalInterface`](https://github.com/cosmos/ibc-go/blob/055a3e39c3da309e013ec4763f78d33b74be59c9/modules/core/02-client/types/encoding.go#L35), so when trying to unmarshal to the concrete type I would get [errors like the ones you see now in the PR](https://github.com/cosmos/ibc-go/actions/runs/7952186466/job/21706474023?pr=5827).

So I guess my question would be: do we expect the client state (and consensus state) received in the messages be marshalled as an interface or a concrete type? I am not sure if this question really makes sense, though. Maybe relayers will just send the bytes for the marshalled concrete type (it seems like [that's what hermes does](https://github.com/informalsystems/hermes/blob/f986dc82cff5aafef11befee44b1d2eb8d179ad9/crates/relayer/src/client_state.rs#L149-L152), although the [go relayer seems to marshal the interface type](https://github.com/cosmos/relayer/blob/11b7de318b5527b0f097475df1865fe472f76ff1/relayer/chains/cosmos/tx.go#L708-L716)?), so maybe we just need to change our tests to also marshal to the concrete type? Sorry, my protobuf foo is pretty weak, so I might just not making any sense, hehe. But having a bit of clarity on this would be helpful for me to unblock this work.

modules/light-clients/07-tendermint/light_client_module.go
damiannolan: I removed the basic validation of client state and consensus state at the light client module level. ClientState is reconstructed within the method `VerifyUpgradeAndUpdateState` and validated there. 

There is still some test failures wrt encoding from what I can see, but I think they should be manageable. Is the plan to restructure tests to invoke the client module in this PR or to leave it for a follow up? 

Happy to help out where needed :)

modules/light-clients/08-wasm/light_client_module.go
```diff
@@ -276,14 +277,69 @@ func (lcm LightClientModule) RecoverClient(ctx sdk.Context, clientID, substitute
...
-// // DEPRECATED: will be removed as performs internal functionality
-func (LightClientModule) VerifyUpgradeAndUpdateState(ctx sdk.Context, clientID string, newClient []byte, newConsState []byte, upgradeClientProof, upgradeConsensusStateProof []byte) error {
-	return nil
+// Upgrade functions
+// NOTE: proof heights are not included as upgrade to a new revision is expected to pass only on the last
+// height committed by the current revision. Clients are responsible for ensuring that the planned last
+// height of the current revision is somehow encoded in the proof verification process.
+// This is to ensure that no premature upgrades occur, since upgrade plans committed to by the counterparty
+// may be cancelled or modified before the last planned height.
+// If the upgrade is verified, the upgraded client and consensus states must be set in the client store.
```
colin-axner, line 286: godoc should probably be rewritten to be wasm specific, not a copy/paste of the API godoc

modules/light-clients/08-wasm/light_client_module.go
```diff
@@ -276,14 +277,69 @@ func (lcm LightClientModule) RecoverClient(ctx sdk.Context, clientID, substitute
...
+		return errorsmod.Wrapf(clienttypes.ErrInvalidClientType, "expected: %s, got: %s", types.Wasm, clientType)
+	}
+
+	var (
+		cdc               = lcm.keeper.Codec()
+		newClientState    exported.ClientState
+		newConsensusState exported.ConsensusState
+	)
+
+	if err := cdc.UnmarshalInterface(newClient, &newClientState); err != nil {
```
colin-axner, line 310: I wonder if we could unmarshal directly into the concrete type and drop the cast assertions that follow?

modules/light-clients/07-tendermint/light_client_module_test.go
```diff
@@ -137,3 +144,173 @@ func (suite *TendermintTestSuite) TestRecoverClient() {
...
+			ibcerrors.ErrInvalidHeight,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		suite.Run(tc.name, func() {
+			suite.SetupTest() // reset
+			cdc := suite.chainA.App.AppCodec()
+			ctx := suite.chainA.GetContext()
```
colin-axner, line 265: super nit: one thing I dislike about assigning the context to a variable rather than dynamically requesting it is that this variable has the incorrect block height set in the context. The `tc.malleate()` call will commit a block on chainA on the "success" case, resulting in the context having an out of date block header when calling `VerifyUpgradeAndUpdateState`. I guess it is fine in this test, but can lead to confusing errors if it needs to be accurate

crodriguezvega: Got it. I changed it now.

modules/light-clients/08-wasm/light_client_module.go
crodriguezvega: Fixed.

modules/core/02-client/keeper/client_test.go
```diff
@@ -351,64 +351,35 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
...
 
-				upgradedClientProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedClientKey(int64(lastHeight.GetRevisionHeight())), cs.GetLatestHeight().GetRevisionHeight())
-				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(lastHeight.GetRevisionHeight())), cs.GetLatestHeight().GetRevisionHeight())
+				upgradedClientProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedClientKey(int64(upgradeHeight.GetRevisionHeight())), cs.GetLatestHeight().GetRevisionHeight())
+				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(upgradeHeight.GetRevisionHeight())), cs.GetLatestHeight().GetRevisionHeight())
 			},
 			expPass: true,
 		},
-		{
-			name: "client state not found",
```
crodriguezvega, line 360: I deleted this test because this is tested in the light client module.

modules/core/02-client/keeper/client_test.go
```diff
@@ -440,71 +412,44 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
 				cs, found := suite.chainA.App.GetIBCKeeper().ClientKeeper.GetClientState(suite.chainA.GetContext(), path.EndpointA.ClientID)
 				suite.Require().True(found)
 
-				upgradedClientProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedClientKey(int64(lastHeight.GetRevisionHeight())), cs.GetLatestHeight().GetRevisionHeight())
-				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(lastHeight.GetRevisionHeight())), cs.GetLatestHeight().GetRevisionHeight())
-			},
-			expPass: false,
-		},
-		{
-			name: "unsuccessful upgrade: upgraded height is not greater than current height",
```
crodriguezvega, line 449: I moved this test to the light client module, but can move it back here in the PR where `LatestHeight` is added to the light client module and then we can move the check back to 02-client.

modules/light-clients/08-wasm/light_client_module.go
crodriguezvega: As discussed on Monday, we need to unmarshal into the interface.

crodriguezvega:
I think this is again ready for review after the discussion on Monday. There are a couple of tests that fail because we need to add the right value for the latest height in the create and upgrade client events, but that should be fixed in https://github.com/cosmos/ibc-go/pull/5866.

modules/core/keeper/msg_server.go
```diff
@@ -69,17 +69,23 @@ func (k Keeper) UpdateClient(goCtx context.Context, msg *clienttypes.MsgUpdateCl
 func (k Keeper) UpgradeClient(goCtx context.Context, msg *clienttypes.MsgUpgradeClient) (*clienttypes.MsgUpgradeClientResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	upgradedClient, err := clienttypes.UnpackClientState(msg.ClientState)
+	upgradedClientState, err := k.cdc.Marshal(msg.ClientState)
```
colin-axner, line 72:
```suggestion
        // Backwards Compatibility: the light client module is expecting
        // to unmarshal an interface so we marshal the client state Any
	upgradedClientState, err := k.cdc.Marshal(msg.ClientState)
```

modules/core/keeper/msg_server.go
```diff
@@ -69,17 +69,23 @@ func (k Keeper) UpdateClient(goCtx context.Context, msg *clienttypes.MsgUpdateCl
...
 	if err != nil {
 		return nil, err
 	}
-	upgradedConsState, err := clienttypes.UnpackConsensusState(msg.ConsensusState)
+
+	upgradedConsensusState, err := k.cdc.Marshal(msg.ConsensusState)
```
colin-axner, line 77:
```suggestion
        // Backwards Compatibility: the light client module is expecting
        // to unmarshal an interface so we marshal the consensus state Any
	upgradedConsensusState, err := k.cdc.Marshal(msg.ConsensusState)
```

some sort of in line code doc like this can be useful

modules/light-clients/07-tendermint/light_client_module.go
```diff
@@ -256,30 +266,39 @@ func (lcm LightClientModule) VerifyUpgradeAndUpdateState(
...
 	if !found {
 		return errorsmod.Wrap(clienttypes.ErrClientNotFound, clientID)
 	}
 
-	return clientState.VerifyUpgradeAndUpdateState(ctx, cdc, clientStore, &newClientState, &newConsensusState, upgradeClientProof, upgradeConsensusStateProof)
+	// last height of current counterparty chain must be client's latest height
+	lastHeight := clientState.GetLatestHeight()
+	if !newClientState.GetLatestHeight().GT(lastHeight) {
+		return errorsmod.Wrapf(ibcerrors.ErrInvalidHeight, "upgraded client height %s must be at greater than current client height %s", newClientState.GetLatestHeight(), lastHeight)
+	}
```
colin-axner, line 301: Add back to 02-client level once latest height pr is merged?

colin-axner (approved):
Thanks @crodriguezvega!!

damiannolan:
Resolving conflicts.. 🏃🏻

modules/core/02-client/keeper/client.go
```diff
@@ -122,46 +121,41 @@ func (k Keeper) UpdateClient(ctx sdk.Context, clientID string, clientMsg exporte
...
+		return errorsmod.Wrapf(types.ErrClientNotFound, "clientID (%s)", clientID)
+	}
 
-	if !upgradedClient.GetLatestHeight().GT(lastHeight) {
-		return errorsmod.Wrapf(ibcerrors.ErrInvalidHeight, "upgraded client height %s must be at greater than current client height %s",
-			upgradedClient.GetLatestHeight(), lastHeight)
+	lightClientModule, found := k.router.GetRoute(clientID)
+	if !found {
+		return errorsmod.Wrap(types.ErrRouteNotFound, clientID)
 	}
```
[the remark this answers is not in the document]
damiannolan: discussed async, this check must be moved to the light client module, and this functionality should likely only exist in the tendermint client module.

At first glance, we do not have the `clientID` of the `upgradedClientState`. However, it should be the same `clientID` as the `clientState` (target). However, we need the latest height from the `upgradedClientState` concrete implementation here - this is the client which the counterparty stores in it upgrade path and we prove against within the tm client module.

modules/core/02-client/keeper/client_test.go
```diff
@@ -448,73 +448,73 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
...
-				suite.Require().True(ok)
-
-				upgradedClientProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedClientKey(int64(lastHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
-				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(lastHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
+				upgradedClientProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedClientKey(int64(upgradeHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
+				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(upgradeHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
 			},
 			expPass: false,
 		},
+		// TODO: Needs to move to light client module?
```
damiannolan, line 456: this test needs to move given that we have removed the height check

modules/core/02-client/keeper/client_test.go
damiannolan: https://github.com/cosmos/ibc-go/pull/5827#discussion_r1520067044

I guess it can just be removed entirely now?

modules/core/keeper/msg_server.go
```diff
@@ -70,17 +70,26 @@ func (k Keeper) UpdateClient(goCtx context.Context, msg *clienttypes.MsgUpdateCl
 func (k Keeper) UpgradeClient(goCtx context.Context, msg *clienttypes.MsgUpgradeClient) (*clienttypes.MsgUpgradeClientResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
-	upgradedClient, err := clienttypes.UnpackClientState(msg.ClientState)
+	// Backwards Compatibility: the light client module is expecting
+	// to unmarshal an interface so we marshal the client state Any
+	upgradedClientState, err := k.cdc.Marshal(msg.ClientState)
```
AdityaSripal, line 75: It's a bit unfortunate that we're marshalling here and then unmarshalling later in the callstack. Seems wasteful...

AdityaSripal (commented):
Looks good! I suppose we have to keep the interfaces for users the same way, but would be nice to eventually just pass everything through and unmarshal at the light client module level.

modules/core/02-client/keeper/client_test.go
crodriguezvega: Sounds good. At the moment this check is performed in 07-tendermint and 08-wasm light client modules. Based on [your comment](https://github.com/cosmos/ibc-go/pull/5827/files#r1520066359), is that ok? Any reason why the check should not be done in 08-wasm?

modules/core/keeper/msg_server.go
damiannolan: I don't think this is necessary. I think this is symptom of how the tests were/are written and they just need to be refactored. After looking closer at the code and how it works, I think it should be possible to pass the `Any.Value` bytes along here and into the light client module and have it work.

Opened a draft which I'll pick up in the morning, but see here for happy path test working: https://github.com/cosmos/ibc-go/pull/5967/files#diff-cef346df04f3fb2455741328504cc2cbb06c7133cce6ede300347139fbdda74eR490

tldr; test code needs to define `*codectypes.Any` and pass their value bytes as args. We can then make the changes in the light client module to unmarshal to concrete types. When we reconstruct the clientState which is proven it has always used `MarshalInterface` given the concrete type and should continue to work like that. I think the tests just need more attention.

modules/core/02-client/keeper/client_test.go
charleenfei: we can have these tests in the lcm modules for 07-tendermint and 08-wasm already, i think! is that fine @crodriguezvega?

crodriguezvega: Yes, correct! The check and the tests are currently in the light client module. Happy to leave it there if that's the right place. :)

modules/core/02-client/keeper/client_test.go
```diff
@@ -353,21 +354,21 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
...
-				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(lastHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
+				upgradedClientProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedClientKey(int64(upgradeHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
+				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(upgradeHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
 			},
 			expPass: true,
 		},
 		{
 			name: "client state not found",
 			setup: func() {
 				// last Height is at next block
```
damiannolan, line 365:
```suggestion
				// upgrade height is at next block
```

modules/core/02-client/keeper/client_test.go
```diff
@@ -394,16 +395,15 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
 				// client is frozen
 
 				// last Height is at next block
```
damiannolan, line 397:
```suggestion
				// upgrade height is at next block
```

modules/core/02-client/keeper/client_test.go
```diff
@@ -425,49 +425,22 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
 			expPass: false,
 		},
 		{
-			name: "tendermint client VerifyUpgrade fails",
+			name: "light client module VerifyUpgradeAndUpdateState fails",
 			setup: func() {
 				// last Height is at next block
```
damiannolan, line 430:
```suggestion
				// upgrade height is at next block
```

modules/light-clients/08-wasm/light_client_module.go
```diff
@@ -257,6 +258,53 @@ func (lcm LightClientModule) RecoverClient(ctx sdk.Context, clientID, substitute
...
+	if err := cdc.UnmarshalInterface(newConsState, &newConsensusState); err != nil {
+		return err
+	}
+	newWasmConsensusState, ok := newConsensusState.(*types.ConsensusState)
+	if !ok {
+		return errorsmod.Wrapf(clienttypes.ErrInvalidConsensus, "expected consensus state type %T, got %T", (*types.ConsensusState)(nil), newConsensusState)
+	}
+	if err := newWasmConsensusState.ValidateBasic(); err != nil {
+		return err
+	}
```
damiannolan, line 295: This code needs to be aligned with 07-tendermint impl, shame that this is not caught with tests.. hmmm

modules/light-clients/08-wasm/light_client_module_test.go
damiannolan, line 1: And tests

damiannolan (changes requested):
I missed 08-wasm changes needed to aligned with 07-tendermint from my encoding PR. 
I'll handle them today and then LGTM! 🙏🏻

modules/core/02-client/keeper/client_test.go
```diff
@@ -381,8 +381,8 @@ func (suite *KeeperTestSuite) TestUpgradeClient() {
 				tmCs, ok := cs.(*ibctm.ClientState)
 				suite.Require().True(ok)
 
-				upgradedClientProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedClientKey(int64(lastHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
-				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(lastHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
+				upgradedClientProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedClientKey(int64(upgradeHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
+				upgradedConsensusStateProof, _ = suite.chainB.QueryUpgradeProof(upgradetypes.UpgradedConsStateKey(int64(upgradeHeight.GetRevisionHeight())), tmCs.LatestHeight.GetRevisionHeight())
 
 				path.EndpointA.ClientID = "wrongclientid"
```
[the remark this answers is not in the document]
damiannolan: can follow up this post merge, maybe? I think all these tests should be using `expError` and not `expPass` (bool)

damiannolan (approved):
LGTM!

Issue #3594, opened 2023-05-16
Consider relying solely on local clock time for transfer CLI relative timeouts


damiannolan:
## Summary

Relative timeouts currently lookup client state and consensus state in order to retrieve the latest timestamp of the counterparty chain's consensus state when constructing a packet with relative timeouts. 

It may be worth relying solely on local clock time to calculate a relative timeout.

Pros:
- Reduced the amount of the stateful queries on nodes when using a CLI client

Cons:
- Local clock time must be accurate 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Make transfer and ICA CLIs consistent regarding handling of timeouts.

PR #5908, opened 2024-02-26
refactor: use local clock time only for transfer CLI relative timeouts

damiannolan:
## Description

- Refactors transfer CLI to use local clock time only for relative timeouts.
- Removes the stateful query on the node to client and consensus states.
- Disables relative timeout by block height.

closes: #3594 

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/controller/client/cli/tx.go
+++ b/modules/apps/27-interchain-accounts/controller/client/cli/tx.go
@@ -81,8 +81,10 @@ func newSendTxCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "send-tx [connection-id] [path/to/packet_msg.json]",
 		Short: "Send an interchain account tx on the provided connection.",
-		Long: strings.TrimSpace(`Submits pre-built packet data containing messages to be executed on the host chain 
-and attempts to send the packet. Packet data is provided as json, file or string. A relative timeout timestamp can be provided with flag {packet-timeout-timestamp}.
+		Long: strings.TrimSpace(`Submits pre-built packet data containing messages to be executed on the host chain and attempts to send the packet. 
+Packet data is provided as json, file or string. A timeout timestamp can be provided using the flag {packet-timeout-timestamp}. 
+By default timeout timestamps are calculated relatively, adding {packet-timeout-timestamp} to the user's local system clock time. 
+Absolute timeout timestamp values can be used by setting the {absolute-timeouts} flag to true.
 If no timeout value is set then a default relative timeout value of 10 minutes is used.`),
 		Args: cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/modules/apps/27-interchain-accounts/controller/client/cli/tx.go
+++ b/modules/apps/27-interchain-accounts/controller/client/cli/tx.go
@@ -115,17 +115,17 @@ If no timeout value is set then a default relative timeout value of 10 minutes i
 			}
 
 			// NOTE: relative timeouts using block height are not supported.
-			if !absoluteTimeouts {
-				// use local clock time as reference time for calculating timeout timestamp.
-				if timeoutTimestamp != 0 {
-					now := time.Now().UnixNano()
-					if now > 0 {
-						timeoutTimestamp = uint64(now) + timeoutTimestamp
-					} else {
-						return errors.New("local clock time is not greater than Jan 1st, 1970 12:00 AM")
-					}
-				}
+			if !absoluteTimeouts && timeoutTimestamp == 0 {
+				return errors.New("relative timeouts must provide a non zero value timestamp")
 			}
+
+			// use local clock time as reference time for calculating timeout timestamp.
+			now := time.Now().UnixNano()
+			if now <= 0 {
+				return errors.New("local clock time is not greater than Jan 1st, 1970 12:00 AM")
+			}
+
+			timeoutTimestamp = uint64(now) + timeoutTimestamp
 
 			msg := types.NewMsgSendTx(owner, connectionID, timeoutTimestamp, icaMsgData)
 
--- a/modules/apps/27-interchain-accounts/controller/client/cli/tx.go
+++ b/modules/apps/27-interchain-accounts/controller/client/cli/tx.go
@@ -29,6 +29,12 @@ const (
 	flagAbsoluteTimeouts       = "absolute-timeouts"
 )
 
+// defaultRelativePacketTimeoutTimestamp is the default packet timeout timestamp (in nanoseconds)
+// relative to the current block timestamp of the counterparty chain provided by the client
+// state. The timeout is disabled when set to 0. The default is currently set to a 10 minute
+// timeout.
+var defaultRelativePacketTimeoutTimestamp = uint64((time.Duration(10) * time.Minute).Nanoseconds())
+
 func newRegisterInterchainAccountCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "register [connection-id]",
@@ -133,7 +139,7 @@ If no timeout value is set then a default relative timeout value of 10 minutes i
 		},
 	}
 
-	cmd.Flags().Uint64(flagPacketTimeoutTimestamp, icatypes.DefaultRelativePacketTimeoutTimestamp, "Packet timeout timestamp in nanoseconds from now. Default is 10 minutes.")
+	cmd.Flags().Uint64(flagPacketTimeoutTimestamp, defaultRelativePacketTimeoutTimestamp, "Packet timeout timestamp in nanoseconds from now. Default is 10 minutes.")
 	cmd.Flags().Bool(flagAbsoluteTimeouts, false, "Timeout flags are used as absolute timeouts.")
 	flags.AddTxFlagsToCmd(cmd)
 
--- a/modules/apps/27-interchain-accounts/controller/client/cli/tx.go
+++ b/modules/apps/27-interchain-accounts/controller/client/cli/tx.go
@@ -1,9 +1,11 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


modules/apps/transfer/types/packet.go
```diff
@@ -22,6 +22,7 @@ var (
 	// DefaultRelativePacketTimeoutHeight is the default packet timeout height (in blocks) relative
 	// to the current block height of the counterparty chain provided by the client state. The
 	// timeout is disabled when set to 0.
+	// Deprecated: This value will be removed in the next major release of IBC-go.
```
damiannolan, line 25: I can remove this now default value now if everyone is on board.

damiannolan (commented):
I'm not sure if I would consider this an API breaking change but it definitely has implications on end users. If its going to target v9 anyway I suppose it doesn't matter.

Let me know and I can remove the value which I've currently marked as deprecated.

damiannolan:
Published as r4r for e2es, and will manually test tomorrow morning 👍🏻

modules/apps/transfer/client/cli/tx.go
```diff
@@ -34,11 +31,9 @@ func NewTransferTxCmd() *cobra.Command {
 		Use:   "transfer [src-port] [src-channel] [receiver] [amount]",
 		Short: "Transfer a fungible token through IBC",
 		Long: strings.TrimSpace(`Transfer a fungible token through IBC. Timeouts can be specified
-as absolute or relative using the "absolute-timeouts" flag. Timeout height can be set by passing in the height string
-in the form {revision}-{height} using the "packet-timeout-height" flag. Relative timeout height is added to the block
-height queried from the latest consensus state corresponding to the counterparty channel. Relative timeout timestamp 
-is added to the greater value of the local clock time and the block timestamp queried from the latest consensus state 
-corresponding to the counterparty channel. Any timeout set to 0 is disabled.`),
+as absolute using the "absolute-timeouts" flag. Timeout height can be set by passing in the height string
+in the form {revision}-{height} using the "packet-timeout-height" flag. Note, relative timeout height is not supported. Relative timeout timestamp
```
crodriguezvega: Just a nit: it looks like in the other CLIs we enclose the flags between `{}`. Not sure if this affects the formatting of the flag on the command line...

```suggestion
in the form {revision}-{height} using the {packet-timeout-height} flag. Note, relative timeout height is not supported. Relative timeout timestamp 
```

modules/apps/27-interchain-accounts/controller/client/cli/tx.go
```diff
@@ -101,18 +104,37 @@ appropriate relative timeoutTimestamp must be provided with flag {relative-packe
...
-			msg := types.NewMsgSendTx(owner, connectionID, relativeTimeoutTimestamp, icaMsgData)
+			absoluteTimeouts, err := cmd.Flags().GetBool(flagAbsoluteTimeouts)
+			if err != nil {
+				return err
+			}
+
+			// NOTE: relative timeouts using block height are not supported.
+			if !absoluteTimeouts {
+				// use local clock time as reference time for calculating timeout timestamp.
+				if timeoutTimestamp != 0 {
```
crodriguezvega, line 120: Would it be possible to combine these two conditions in a single `if` statement? Like `if !absoluteTimeouts && timeoutTimestamp != 0`.

crodriguezvega: Also, is it possible to have `absoluteTimeouts = true` and `timeoutTimestamp = 0`? In that case we would broadcast a message with a 0 timeout? Would it be better just to return an error early instead?

damiannolan: I can see if I can simply the logic here. 

> absoluteTimeouts = true and timeoutTimestamp = 0

This should be possible, but if so then timeHeight needs to be set. Either way it would fail basic validation if neither is set. Maybe the logic should be like:
 `if !absoluteTimeouts && timeoutHeight.IsZero() => err`

modules/apps/transfer/client/cli/tx.go
```diff
@@ -85,62 +81,18 @@ corresponding to the counterparty channel. Any timeout set to 0 is disabled.`),
...
-						return err
-					}
-				}
-
 				if !timeoutHeight.IsZero() {
-					absoluteHeight := clientHeight
-					absoluteHeight.RevisionNumber += timeoutHeight.RevisionNumber
-					absoluteHeight.RevisionHeight += timeoutHeight.RevisionHeight
-					timeoutHeight = absoluteHeight
+					return errors.New("relative timeouts using block height is not supported")
```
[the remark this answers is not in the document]
colin-axner: I think the absolute timeout height is parsed from that flag

modules/apps/transfer/client/cli/tx.go
```diff
@@ -85,62 +80,18 @@ corresponding to the counterparty channel. Any timeout set to 0 is disabled.`),
...
 				}
 
-				// use local clock time as reference time if it is later than the
-				// consensus state timestamp of the counterparty chain, otherwise
-				// still use consensus state timestamp as reference.
-				// for localhost clients local clock time is always used.
+				// use local clock time as reference time for calculating timeout timestamp.
 				if timeoutTimestamp != 0 {
```
colin-axner, line 91: I think the code should actually be:
```go
if !absolute {
    if !timeoutHeight.IsZero() {
        return err
    }
    
    if timeoutTimestamp == 0 {
        return err
    }
    
    now := time.Now().UnixNano()
    if now <= 0 {
        return errors.New("local clock time is not greater than Jan 1st, 1970 12:00 AM")
    }
    
    timeoutTimestamp = uint64(now) + timeoutTimestamp
}
```

damiannolan: Happy to go with this, but if timeouts are not absolute - someone would also have to explicitly provide a zero value timestamp to trigger the err here. Because will be default 10 mins if not overridden. Still tho, returning the err in this case makes perfect sense. Will implement as suggested, ty!

damiannolan: Plus one level of if indentation is removed here 😂

modules/apps/transfer/client/cli/tx.go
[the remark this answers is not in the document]
damiannolan: It is indeed!

modules/apps/27-interchain-accounts/controller/client/cli/tx.go
```diff
@@ -22,10 +24,17 @@ const (
 	// The controller chain channel version
 	flagVersion = "version"
 	// The channel ordering
-	flagOrdering              = "ordering"
-	flagRelativePacketTimeout = "relative-packet-timeout"
+	flagOrdering               = "ordering"
+	flagPacketTimeoutTimestamp = "packet-timeout-timestamp"
+	flagAbsoluteTimeouts       = "absolute-timeouts"
```
colin-axner, line 29: what is this used for? Unless I am missing something the code never allows an absolute timeout to be used?

colin-axner: we should either remove the absolute flag or allow absolute timestamps to be specified

crodriguezvega: Isn't it the case that absolute timeout timestamp is used when `flagAbsoluteTimeouts` is true and `flagPacketTimeoutTimestamp` is also used?

crodriguezvega: Isn't it the case that absolute timeout timestamp is used when `flagAbsoluteTimeouts` is true and `flagPacketTimeoutTimestamp` is also used?

modules/apps/27-interchain-accounts/controller/client/cli/tx.go
```diff
@@ -73,8 +82,8 @@ func newSendTxCmd() *cobra.Command {
 		Use:   "send-tx [connection-id] [path/to/packet_msg.json]",
 		Short: "Send an interchain account tx on the provided connection.",
 		Long: strings.TrimSpace(`Submits pre-built packet data containing messages to be executed on the host chain 
-and attempts to send the packet. Packet data is provided as json, file or string. An 
-appropriate relative timeoutTimestamp must be provided with flag {relative-packet-timeout}`),
+and attempts to send the packet. Packet data is provided as json, file or string. A relative timeout timestamp can be provided with flag {packet-timeout-timestamp}.
```
crodriguezvega, line 85: If it's correct that absolute timeout timestamps can be used when `flagAbsoluteTimeouts` is true and `flagPacketTimeoutTimestamp` is also set, should we update this doc string to clarify that? At the moment, the way it reads I think a reader would think `flagPacketTimeoutTimestamp` can only be used for relative timeouts...

damiannolan: Updated in https://github.com/cosmos/ibc-go/pull/5908/commits/7f673fb86238c22d85d464de0195f6e94f970765

feel free to push another commit if you want to change it.

modules/apps/27-interchain-accounts/controller/client/cli/tx.go
colin-axner: yes, (this logic was missing when I commented). In addition, the previous logic did not allow absolute timeout submission which is why I wasn't sure which direction was being chosen

colin-axner (approved):
Thanks @damiannolan !

chatton (approved):
LGTM thanks @damiannolan

modules/apps/27-interchain-accounts/controller/client/cli/tx.go
```diff
@@ -101,18 +112,39 @@ appropriate relative timeoutTimestamp must be provided with flag {relative-packe
...
+				// use local clock time as reference time for calculating timeout timestamp.
+				now := time.Now().UnixNano()
+				if now <= 0 {
+					return errors.New("local clock time is not greater than Jan 1st, 1970 12:00 AM")
+				}
+
+				timeoutTimestamp = uint64(now) + timeoutTimestamp
+			}
+
+			msg := types.NewMsgSendTx(owner, connectionID, timeoutTimestamp, icaMsgData)
```
yihuang, line 140: Isn't `NewMsgSendTx` expects a relative timeout here, but we passed an absolute one?

damiannolan: Ah yea, I think you are correct @yihuang. Good catch. This cli should be revert to use relative timestamps only and included in a patch release.

Did this cause some problem for you?

Issue #5283, opened 2023-11-30
Bump `wasmvm` to v2.0


crodriguezvega:
Issue to keep track of the upgrade. Release is not available yet.

Migration docs are available [here](https://github.com/CosmWasm/wasmvm/blob/main/docs/MIGRATING.md).

PR #5909, opened 2024-02-26
deps: bump wasmvm v2.0.0

crodriguezvega:
## Description

~~I will mark the PR as r4r when we can use the final v2.0.0 release (using v2.0.0-rc.2 at the moment).~~

closes: #5283 

```text
type: commit message
```

---

write a little note why.

## Summary by CodeRabbit

- **New Features**
	- Enhanced support for multiple capabilities in light clients, moving from a singular to multiple capabilities support.
	- Introduced a new gas limit parameter and additional gas-related values for improved gas management in smart contracts.
	- Randomized data directory paths for testing, ensuring unique environments for each test run.

- **Improvements**
	- Updated `wasmvm` and `wasmvmtypes` dependencies to version 2 across various modules for better performance and compatibility.
	- Refined gas cost constants and introduced a new `WasmGasRegisterConfig` struct for more flexible gas cost configuration.
	- Improved type safety and error handling by updating function return types to use `ContractResult` and `QueryResult`.

- **Documentation**
	- Updated the CHANGELOG to reflect the dependency update to `wasmvm` version 2.0.0.

- **Bug Fixes**
	- Addressed various minor bugs related to import paths and function signatures to align with the new `wasmvm` version 2 updates.


modules/light-clients/08-wasm/types/vm.go
```diff
@@ -26,16 +26,16 @@ var (
 	// wasmvmAPI is a wasmvm.GoAPI implementation that is passed to the wasmvm, it
 	// doesn't implement any functionality, directly returning an error.
 	wasmvmAPI = wasmvm.GoAPI{
-		HumanAddress:     humanAddress,
-		CanonicalAddress: canonicalAddress,
+		HumanizeAddress:     humanizeAddress,
+		CanonicalizeAddress: canonicalizeAddress,
```
webmaster128, line 30: Here you need a validate implementation too now. Can be the same style as the other two but needs to be set.

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -114,7 +114,10 @@ func wasmInstantiate(ctx sdk.Context, cdc codec.BinaryCodec, clientStore storety
 	checksum := cs.Checksum
 	resp, err := instantiateContract(ctx, clientStore, checksum, encodedData)
```
webmaster128, line 115: This variable name "resp" is misleading now since the variable contains a result (i.e. success or error struct) and response only referred to the success case

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -114,7 +114,10 @@ func wasmInstantiate(ctx sdk.Context, cdc codec.BinaryCodec, clientStore storety
...
 	if err != nil {
-		return errorsmod.Wrap(ErrWasmContractCallFailed, err.Error())
+		return errorsmod.Wrap(ErrVMError, err.Error())
+	}
+	if resp.Err != "" {
+		return errorsmod.Wrap(ErrWasmContractCallFailed, resp.Err)
 	}
 
 	if err = checkResponse(resp); err != nil {
```
webmaster128, line 123: To restore the old behaviour, I'd pass `result.Ok` as the argument here and only handle the response internally

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -287,25 +295,25 @@ func getEnv(ctx sdk.Context, contractAddr string) wasmvmtypes.Env {
...
 
-func canonicalAddress(human string) ([]byte, uint64, error) {
+func canonicalizeAddress(human string) ([]byte, uint64, error) {
 	return nil, 0, errors.New("canonicalAddress not implemented")
 }
 
 // checkResponse returns an error if the response from a sudo, instantiate or migrate call
 // to the Wasm VM contains messages, events or attributes.
-func checkResponse(response *wasmvmtypes.Response) error {
+func checkResponse(response *wasmvmtypes.ContractResult) error {
```
webmaster128, line 308:
```suggestion
func checkResponse(response *wasmvmtypes.Response) error {
```

I'd keep this type can adjust the callers

webmaster128 (commented):
Good stuff. Thanks for the early integration. Some first notes

modules/light-clients/08-wasm/types/vm.go
crodriguezvega: Good point. I can change the name.

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -112,12 +113,15 @@ func wasmInstantiate(ctx sdk.Context, cdc codec.BinaryCodec, clientStore storety
...
 
 	checksum := cs.Checksum
-	resp, err := instantiateContract(ctx, clientStore, checksum, encodedData)
+	res, err := instantiateContract(ctx, clientStore, checksum, encodedData)
 	if err != nil {
-		return errorsmod.Wrap(ErrWasmContractCallFailed, err.Error())
+		return errorsmod.Wrap(ErrVMError, err.Error())
+	}
+	if res.Err != "" {
+		return errorsmod.Wrap(ErrWasmContractCallFailed, res.Err)
```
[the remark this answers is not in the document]
webmaster128: No, there are now two layers of error: errors caused by the VM (`err`) and successful VM executions that contain an error caused by the contract (`res.Err`). Before 2.0 both error cases were compressed into one in wasmvm. But this is losing information on the caller side, which we need in wasmd. The reason is that we want to utilize the knowledge that `res.Err` is guaranteed to deterministic. See also:
- The `DeterministicError` and its usage in https://github.com/CosmWasm/wasmd/pull/1804/files
- This document discussing the whole topic a bit more: https://github.com/CosmWasm/cosmwasm/blob/main/docs/ERROR_HANDLING.md

modules/light-clients/08-wasm/testing/simapp/app.go
```diff
@@ -462,9 +464,14 @@ func NewSimApp(
 	// Function DefaultWasmConfig can also be used to use default values.
 	//
 	// In the code below we use the second method because we are not using x/wasm in this app.go.
+
+	// NOTE: a random string is appended to the data directory to ensure that every test
```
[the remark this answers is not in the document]
crodriguezvega: Yes, I agree it's a bit hacky and I agree we can open an issue and fix it later. 👍

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -151,16 +155,19 @@ func wasmSudo[T ContractResult](ctx sdk.Context, cdc codec.BinaryCodec, clientSt
 	}
 
 	checksum := cs.Checksum
-	resp, err := callContract(ctx, clientStore, checksum, encodedData)
+	res, err := callContract(ctx, clientStore, checksum, encodedData)
 	if err != nil {
-		return result, errorsmod.Wrap(ErrWasmContractCallFailed, err.Error())
+		return result, errorsmod.Wrap(ErrVMError, err.Error())
+	}
+	if res.Err != "" {
```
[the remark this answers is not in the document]
crodriguezvega: I don't feel strongly about either, so I am fine with whatever people prefer. Maybe I am slightly inclined to leave it like this, so that `checkResponse` is limited to checking that the contract doesn't return data that we expect, but I am fine with both.

modules/light-clients/08-wasm/types/gas_register.go
```diff
@@ -1,7 +1,7 @@
 package types
```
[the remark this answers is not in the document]
crodriguezvega: Yes, I took it from the v2-rc tag, and I just updated it with the v2 tag (there were a couple of more changes).

docs/docs/03-light-clients/04-wasm/03-integration.md
```diff
@@ -214,10 +214,10 @@ app.WasmClientKeeper = ibcwasmkeeper.NewKeeperWithVM(
...
-- `MemoryCacheSize` sets [the size in MiB of an in-memory cache for e.g. module caching](https://github.com/CosmWasm/wasmvm/blob/1638725b25d799f078d053391945399cb35664b1/lib.go#L29C16-L29C104). It is not consensus-critical and should be defined on a per-node basis, often in the range 100 to 1000 MB. [`wasmd` reads this value of](https://github.com/CosmWasm/wasmd/blob/36416def20effe47fb77f29f5ba35a003970fdba/app/app.go#L579). Default value is 256.
-- `ContractDebugMode` is a [flag to enable/disable printing debug logs from the contract to STDOUT](https://github.com/CosmWasm/wasmvm/blob/1638725b25d799f078d053391945399cb35664b1/lib.go#L28). This should be false in production environments. Default value is false.
+- `DataDir` is the [directory for Wasm blobs and various caches](https://github.com/CosmWasm/wasmvm/blob/v2.0.0/lib.go#L25). In `wasmd` this is set to the [`wasm` folder under the home directory](https://github.com/CosmWasm/wasmd/blob/36416def20effe47fb77f29f5ba35a003970fdba/app/app.go#L578).
```
damiannolan, line 217: Should it be mentioned that this should be different to that used by x/wasm?

modules/light-clients/08-wasm/testing/simapp/app.go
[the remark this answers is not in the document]
damiannolan: Go testing lib offers an API that might be useful here. Happy to see if it can be used in the follow up issue. See https://pkg.go.dev/testing#T.TempDir

modules/light-clients/08-wasm/types/errors.go
```diff
@@ -19,4 +19,5 @@ var (
 	ErrWasmContractCallFailed          = errorsmod.Register(ModuleName, 14, "wasm contract call failed")
 	ErrWasmInvalidResponseData         = errorsmod.Register(ModuleName, 15, "wasm contract returned invalid response data")
 	ErrWasmInvalidContractModification = errorsmod.Register(ModuleName, 16, "wasm contract made invalid state modifications")
+	ErrVMError                         = errorsmod.Register(ModuleName, 17, "wasm VM error")
```
damiannolan, line 22: would the intension to be to use in this in favour of `ErrWasmContractCallFailed` in all places? 

Is there a difference between the two errors and when one should be used in favour of the other?

modules/light-clients/08-wasm/types/vm.go
```diff
@@ -26,16 +26,17 @@ var (
 	// wasmvmAPI is a wasmvm.GoAPI implementation that is passed to the wasmvm, it
 	// doesn't implement any functionality, directly returning an error.
 	wasmvmAPI = wasmvm.GoAPI{
-		HumanAddress:     humanAddress,
-		CanonicalAddress: canonicalAddress,
+		HumanizeAddress:     humanizeAddress,
+		CanonicalizeAddress: canonicalizeAddress,
+		ValidateAddress:     validateAddress,
```
damiannolan, line 31: Is this called by something outside our reach? I see it returns an error, same with the other two above. I guess this is intended not to be called by contracts?

damiannolan (approved):
Awesome work @crodriguezvega! Thank you 🙏🏻 

And thanks to @webmaster128 for quick review as well! LGTM

modules/light-clients/08-wasm/types/vm.go
webmaster128: To avoid null pointer crashes further down the road, latest versions of wasmvm requires all the fields to be set to a non-nil value

modules/light-clients/08-wasm/testing/simapp/app.go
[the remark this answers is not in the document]
damiannolan: > This is required because wasm VM forbids 2 or more different VM instances running in the same data directory.

Is this because go test is running tests in parallel and the wasvm module doesn't like it that we spin up multiple simapps?

I realised my last suggestion needs a `testing.T/B/F`. Can try to see if https://pkg.go.dev/os#MkdirTemp works instead albeit, its similar to what you're doing manually here.

modules/light-clients/08-wasm/types/errors.go
crodriguezvega: Not in all places. The way I've done it is like this:

```
res, err := <contract call>
if err != nil {
  return result, errorsmod.Wrap(ErrVMError, err.Error())
}
if res.Err != "" {
  return result, errorsmod.Wrap(ErrWasmContractCallFailed, res.Err)
}
```

damiannolan: Aha! Perfect, that makes sense!

modules/light-clients/08-wasm/testing/simapp/app.go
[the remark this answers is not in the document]
crodriguezvega: > Is this because go test is running tests in parallel and the wasvm module doesn't like it that we spin up multiple simapps?

Yes, correct. When multiple tests are running in parallel and they all use the same directory for the wasm data, then wasmvm doesn't like that.

Issue #3560, opened 2023-05-04
Support SDK dependency injection for v2 apps


alpe:
## Summary
SDK 47 ships with the new dependency-injection framework used in [app_v2](https://github.com/cosmos/cosmos-sdk/blob/v0.47.2/simapp/app_v2.go#L204).

As wasmd has a dependency to the ibc (core and other) keepers, we can not update our module to be used with the dependency-injection framework. 
There is no workaround AFAIK

## Proposal
Please make the modules compatible to work with app v1 + v2.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Thank you, @alpe, for opening this issue. How pressing is your need to have ibc-go compatible with app v2? Having also app v1 is blocking any urges development on your side?

alpe:
The v1 app is working with our sdk v47 integration. But as v2 had become the "preferred" setup (in build tags) we were eager to provide a fully compatible module with our official release.
This is not possible due to our dependency on ibc-go in the wasmd module. We set the new goal for v2 support in the v0.41 release.
Therefore it is not super urgent but blocking the integration work on our side.

crodriguezvega:
Thanks for the information, @alpe. What is your timeline for v0.41? 

And If I understand correctly, then you would like to have this available in the v7 release line, right?

alpe:
v7.x would be nice to have a small change set only. 
We don't have a strict release schedule for v0.41 but target middle of April. As the work on our side had started already, it would be great to not have this pending for too long.

In https://github.com/CosmWasm/wasmd/pull/1287 we have a mixed v2 with v1 legacy module setup in /app. That could be some template for your test setup before adding full v2 support to the module.

crodriguezvega:
```[tasklist]
### Tasks
- [ ] Add a draft title or issue reference here
```

crodriguezvega:
- [ ] Add buf pulsar generation (in progress, branch [here](https://github.com/cosmos/ibc-go/tree/carlos/add-buf-gen-pulsar) -[previous PR](https://github.com/cosmos/ibc-go/pull/4039)).
- [ ] Add depinject support to code module (in progress, branch [here](https://github.com/cosmos/ibc-go/tree/carlos/3560-depinject-support-code) - [previous PR](https://github.com/cosmos/ibc-go/pull/4040))
[...]

PR #5968, opened 2024-03-12
feat(depinject): add buf gen pulsar

crodriguezvega:
## Description

- I welcome ideas on how to call the new module instead of `api`.

ref: #3560

---

write a little note why.


crodriguezvega:
We could also just use this `api` module also for what we want to do [here](https://github.com/cosmos/ibc-go/issues/2471#issuecomment-1941975081).

damiannolan:
Is the feat/depinject target branch clean? And up to date with main?

crodriguezvega:
> Is the feat/depinject target branch clean? And up to date with main?

Yes and yes!

scripts/protocgen-pulsar.sh
damiannolan, line 1: Should we just combine this step into the `protocgen.sh` script? Seems like maybe more overhead to have separate scripts for this, or is there any benefit? 

Looks to me that the sdk team has it all in the one script nowadays

damiannolan:
> 2024/04/16 15:30:37 Message ibc.applications.interchain_accounts.v1.InterchainAccountPacketData contains the reserved field name ibc.applications.interchain_accounts.v1.InterchainAccountPacketData.type which conflicts with protoreflect.Message interface implementation.
This field will be suffixed with an underscore '_'.
If you can change the message field name, please do so.
In a future iteration of pulsar we may make a breaking change to this practice in order to be compliant with field naming of the original golang protobuf implementation.

This msg is displayed in my shell when I run `make proto-all`. AFAIK the sdk team will be removing the need for pulsar(and protov2 spec) with the next release. They want to make it work with gogoproto which means the codegen will change and I think this problem/error msg will go away. The depinject apis will remain the same tho.

scripts/protocgen-pulsar.sh
damiannolan: https://github.com/cosmos/ibc-go/pull/5968/commits/b65ad37631c3b5ed002e60b3e4c20739d8893c04

proto/buf.gen.pulsar.yaml
damiannolan, line 1: As I mentioned in https://github.com/cosmos/ibc-go/pull/5968#issuecomment-2059384286, we should no longer need pulsar with the next release. My thinking is that we can go ahead with this for now and add the pulsar codegen but when it works with gogoproto we just remove what's not needed and swap it out - should produce diffs to the codegen (i.e. removing the protov2 msg APIs and back to v1 with gogoproto - from what I understand this relates the reflection apis which are used in protobuf and sdk)

damiannolan (approved):

damiannolan:
> Actually, scratch that. Can't get this to build locally still. Is cd-ing into `api` and doing `go build ./...` working for others? I would assume it should?

Same experience for me.
 
> `go build` complains about others too. The generation for some reason might have mucked things up?

Very possible, I can try to look into this a bit more.

crodriguezvega:
> go build complains about others too. The generation for some reason might have mucked things up?

Oh, I hadn't even tried that... Does building work in the api folder of the SDK?

api/ibc/applications/transfer/v1/tx.pulsar.go
```diff
@@ -0,0 +1,2592 @@
...
+	state         protoimpl.MessageState
+	sizeCache     protoimpl.SizeCache
+	unknownFields protoimpl.UnknownFields
+
+	// the port on which the packet will be sent
+	SourcePort string `protobuf:"bytes,1,opt,name=source_port,json=sourcePort,proto3" json:"source_port,omitempty"`
+	// the channel by which the packet will be sent
+	SourceChannel string `protobuf:"bytes,2,opt,name=source_channel,json=sourceChannel,proto3" json:"source_channel,omitempty"`
+	// the tokens to be transferred
+	Token *types.Coin `protobuf:"bytes,3,opt,name=token,proto3" json:"token,omitempty"`
```
damiannolan, line 2187: What instructs the code generator to import `Coin` here as:

```
types "github.com/cosmos/cosmos-sdk/types"
```

I see in that in x/gov for example the `InitialDeposit` coin is imported as `v1beta1.Coin` (from the `api` module):

```
v1beta1 "cosmossdk.io/api/cosmos/base/v1beta1"
```

The same is true for other dependent types in ibc protos such as:
- `BaseAccount` used in ica
- `codectypes.Any` used for multiple different fields such as pub keys, client/consensus state in `MsgCreateClient`, etc
- `CommitmentProof` imported from ics23 protos

Due to each of these types not implementing the protov2 api `go build` fails on the api package:
```
value.ProtoReflect undefined (type *"github.com/cosmos/cosmos-sdk/x/auth/types".BaseAccount has no field or method ProtoReflect)
```

Do you guys have any ideas how to get around this? @tac0turtle @julienrbrt 

If we are working towards depinject/appmodule without protov2 then I assume this won't be a problem? Are we better waiting until then to address this, and in the meantime just put try to put the pieces in place for module inputs/outputs etc?

julienrbrt: You need to add an overwrite in your buf.gen.pulsar.yaml.

julienrbrt: Example: https://github.com/cosmos/cosmos-sdk/blob/main/x/bank/proto/buf.gen.pulsar.yaml

damiannolan: Thanks, will give it a shot!

damiannolan: I suspect this should work for sdk types but I'm concerned that ics23 will still not have protov2 methods

damiannolan: The `Anys` in cosmos-sdk seem to picking up `anybp` (google import) in the api module, instead of `codectypes.Any`.

I see this diff on main, but on sdk v0.50 release tags codectypes is still used. 

```diff
-    opt: plugins=grpc,Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types
+    opt: plugins=grpc,Mgoogle/protobuf/any.proto=github.com/cosmos/gogoproto/types/any
```

Is there some kind of magic needed to use google anys in the api pkg too? I followed your suggestion for the override and it resolved sdk type imports like coin and base account. Still have `go build` failing in `api` here due to Anys and ics23

julienrbrt: Yeah, that's on main, you should still use the sdk one for gogoproto. However, this shouldn't be needed for pulsar.
Concerning ics23, yep, you may need to generate it here too.

Why are you generating things with `go-grpc` btw? Only `go-pulsar` is sufficient.

api/go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/cosmos/ibc-go/api
+
+go 1.22.0
+
+// replace github.com/cosmos/ics23/go/api => ../../ics23/go/api
```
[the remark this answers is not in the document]
damiannolan: beautiful! I've done a replace to your branch with the go pseudoversion!

.github/workflows/api_module.yml
damiannolan, line 1: Thanks for adding this @DimitrisJim, good call. It looks to be passing now! 🙌🏻

api/go.mod
```diff
@@ -0,0 +1,28 @@
+module github.com/cosmos/ibc-go/api
+
+go 1.22.0
+
+// NOTE: This replaces points the ics23 dependency to a protov2 compatible encoding format which also uses pulsar.
```
damiannolan, line 5:
```suggestion
// NOTE: This replace points the ics23 code deps used by 23-commitment to a branch using pulsar codegen (i.e. protov2 compatible encoding interfaces)
```

proto/buf.gen.pulsar.yaml
```diff
@@ -0,0 +1,15 @@
...
+    except:
+      - buf.build/googleapis/googleapis
+      - buf.build/cosmos/gogo-proto
+      - buf.build/cosmos/cosmos-proto
+    override:
+      buf.build/cosmos/cosmos-sdk: cosmossdk.io/api
+plugins:
+  - name: go-pulsar
+    out: ../api
+    opt: Mcosmos/app/v1alpha1/module.proto=cosmossdk.io/api/cosmos/app/v1alpha1,Mcosmos/ics23/v1/proofs.proto=github.com/cosmos/ics23/go/api
```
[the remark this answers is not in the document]
damiannolan: yeah, I think this is how you do overriding. 

maybe @julienrbrt could give us a more detailed description so we have a better understanding.

damiannolan (approved):
LGTM!

api/go.mod
```diff
@@ -0,0 +1,28 @@
+module github.com/cosmos/ibc-go/api
+
+go 1.22.0
```
damiannolan, line 3: should we be using go 1.21?

damiannolan: I don't think this really matters since its hitting a feat branch for now. I imagine by release time all should be updated to go 1.22

crodriguezvega: Yeah, v9 will be in 1.22.

crodriguezvega:
Great work taking this to the finish line, @damiannolan and @DimitrisJim. 👏

Issue #4140, opened 2023-07-20
Missing checks in gRPC queries


ancazamfir:
## Summary of Bug
Some gRPC queries are missing validation.

## Expected Behaviour
Packet level query with nonexistent channel should return error instead of empty output.

## Version
v7.2.0, also checked the code in main

## Steps to Reproduce
Here is an example
```
$ grpcurl --plaintext -d '{"port_id":"transfer", "channel_id":"channel-43"}' stride-grpc.polkachu.com:12290 ibc.core.channel.v1.Query.PacketCommitments
{
  "pagination": {

  },
  "height": {
    "revisionNumber": "1",
    "revisionHeight": "4626258"
  }
}
$ grpcurl --plaintext -d '{"port_id":"icacontroller-osmosis-1.FEE", "channel_id":"channel-43"}' stride-grpc.polkachu.com:12290 ibc.core.channel.v1.Query.PacketCommitments
{
  "pagination": {

  },
  "height": {
    "revisionNumber": "1",
    "revisionHeight": "4626258"
  }
}
```
This is the nonexistent channel:
```
$ grpcurl --plaintext -d '{"port_id":"transfer", "channel_id":"channel-43"}' stride-grpc.polkachu.com:12290 ibc.core.channel.v1.Query.Channel
ERROR:
  Code: NotFound
  Message: port-id: transfer, channel-id channel-43: channel not found
```

The existing one:
```
$ grpcurl --plaintext -d '{"port_id":"icacontroller-osmosis-1.FEE", "channel_id":"channel-43"}' stride-grpc.polkachu.com:12290 ibc.core.channel.v1.Query.Channel
{
  "channel": {
    "state": "STATE_OPEN",
...
```
There are other queries that are missing the channel check (PacketReceipt, PacketCommitment, PacketAcknowledgement...), didn't check them all.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Thanks for raising this issue, @ancazamfir. I tried to be as exhaustive as possible, and I see that the issue is present in the following gRPC handlers.

In 29-fee:
- [`IncentivizedPacketsForChannel`l](https://github.com/cosmos/ibc-go/blob/31b3c6000ca05be316d5bb49d34e073ed07a82e8/modules/apps/29-fee/keeper/grpc_query.go#L71)
[...]

PR #6059, opened 2024-03-26
chore: Add gRPC validation for EscrowAddress

bznein:
## Description

Added gRPC validation for EscrowAddress in apps/transfer/keeper

ref: #4140 

write a little note why.

## Summary by CodeRabbit

- **New Features**
	- Improved validation logic for gRPC request parameters in the transfer service.
- **Bug Fixes**
	- Enhanced error handling for non-existent channels in transfer queries.
- **Tests**
	- Added new test cases for transfer service to cover scenarios with invalid channel identifiers.

### DIFF
--- a/modules/apps/transfer/keeper/grpc_query.go
+++ b/modules/apps/transfer/keeper/grpc_query.go
@@ -126,8 +126,7 @@ func (k Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRe
 	}
 
 	ctx := sdk.UnwrapSDKContext(c)
-	found := k.channelKeeper.HasChannel(ctx, req.PortId, req.ChannelId)
-	if !found {
+	if !k.channelKeeper.HasChannel(ctx, req.PortId, req.ChannelId) {
 		return nil, status.Error(
 			codes.NotFound,
 			errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
--- a/modules/apps/transfer/keeper/grpc_query.go
+++ b/modules/apps/transfer/keeper/grpc_query.go
@@ -16,7 +16,7 @@ import (
 
 	"github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
 	errors "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
-	host "github.com/cosmos/ibc-go/v7/modules/core/24-host"
+	host "github.com/cosmos/ibc-go/v8/modules/core/24-host"
 )
 
 var _ types.QueryServer = (*Keeper)(nil)
--- a/modules/apps/transfer/keeper/grpc_query.go
+++ b/modules/apps/transfer/keeper/grpc_query.go
@@ -15,7 +15,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/types/query"
 
 	"github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
-	errors "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+	channeltypes "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
 	host "github.com/cosmos/ibc-go/v8/modules/core/24-host"
 )
 
@@ -130,7 +130,7 @@ func (k Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRe
 	if !found {
 		return nil, status.Error(
 			codes.NotFound,
-			errorsmod.Wrapf(errors.ErrChannelNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
+			errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
 		)
 	}
 
--- a/modules/apps/transfer/keeper/grpc_query.go
+++ b/modules/apps/transfer/keeper/grpc_query.go
@@ -129,7 +129,7 @@ func (k Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRe
 	if !k.channelKeeper.HasChannel(ctx, req.PortId, req.ChannelId) {
 		return nil, status.Error(
 			codes.NotFound,
-			errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
+			errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "port ID (%s) channel ID (%s)", req.PortId, req.ChannelId).Error(),
 		)
 	}
 
--- a/modules/apps/transfer/keeper/grpc_query.go
+++ b/modules/apps/transfer/keeper/grpc_query.go
@@ -126,7 +126,7 @@ func (k Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRe
 	}
 
 	ctx := sdk.UnwrapSDKContext(c)
-	_, found := k.channelKeeper.GetChannel(ctx, req.PortId, req.ChannelId)
+	found := k.channelKeeper.HasChannel(ctx, req.PortId, req.ChannelId)
 	if !found {
 		return nil, status.Error(
 			codes.NotFound,
--- a/modules/apps/transfer/keeper/grpc_query.go
+++ b/modules/apps/transfer/keeper/grpc_query.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
 	errors "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+	host "github.com/cosmos/ibc-go/v7/modules/core/24-host"
 )
 
 var _ types.QueryServer = (*Keeper)(nil)
@@ -113,7 +114,7 @@ func (k Keeper) DenomHash(c context.Context, req *types.QueryDenomHashRequest) (
 }
 
 // EscrowAddress implements the EscrowAddress gRPC method
-func (Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRequest) (*types.QueryEscrowAddressResponse, error) {
+func (k Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRequest) (*types.QueryEscrowAddressResponse, error) {
 	if req == nil {
 		return nil, status.Error(codes.InvalidArgument, "empty request")
 	}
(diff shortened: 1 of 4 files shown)


modules/apps/transfer/keeper/grpc_query_test.go
```diff
@@ -245,12 +245,24 @@ func (suite *KeeperTestSuite) TestEscrowAddress() {
...
+			},
+			false,
+		},
 	}
 
 	for _, tc := range testCases {
 		tc := tc
 		suite.Run(fmt.Sprintf("Case %s", tc.msg), func() {
 			suite.SetupTest() // reset
+			path := ibctesting.NewTransferPath(suite.chainA, suite.chainB)
```
bznein, line 264: I added this part to enable creation of the channel, otherwise the test for the "happy path" would fail.

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -142,3 +157,15 @@ func (k Keeper) TotalEscrowForDenom(c context.Context, req *types.QueryTotalEscr
 		Amount: amount,
 	}, nil
 }
+
+func validategRPCRequest(portID, channelID string) error {
```
bznein, line 161: This function is copy-pasted from here: https://github.com/cosmos/ibc-go/blob/1323a733d5b0d83478f1362ebc64cb1e42fa8caa/modules/core/04-channel/keeper/grpc_query.go#L641-L651

we could technically export that one instead but since it's internal logic I'm not sure that would be better. Open to change it in that case.

chatton:
E2E failure looks unrelated, we can re-run it :+1:

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/cosmos/cosmos-sdk/types/query"
 
 	"github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	errors "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
```
chatton, line 18: we typically use `channeltypes` as the alias here, the errors are just in the package along with the other types.

modules/apps/transfer/types/expected_keepers.go
```diff
@@ -39,6 +39,8 @@ type ChannelKeeper interface {
 	GetChannel(ctx sdk.Context, srcPort, srcChan string) (channel channeltypes.Channel, found bool)
 	GetNextSequenceSend(ctx sdk.Context, portID, channelID string) (uint64, bool)
 	GetAllChannelsWithPortPrefix(ctx sdk.Context, portPrefix string) []channeltypes.IdentifiedChannel
+
```
chatton, line 42: [nit] can we axe this whitespace :sweat_smile:

chatton (commented):
Looks great! I think we can just change the alias to be more consistent then we should be good to go! Great work :muscle:

modules/apps/transfer/types/expected_keepers.go
bznein: Done

modules/apps/transfer/keeper/grpc_query.go
bznein: Done!

chatton (approved):
LGTM thank you @bznein !

modules/apps/transfer/keeper/grpc_query.go
bznein: Yeah it could be a nice clean-up after I complete the issue :)

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -112,13 +114,26 @@ func (k Keeper) DenomHash(c context.Context, req *types.QueryDenomHashRequest) (
...
-func (Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRequest) (*types.QueryEscrowAddressResponse, error) {
+func (k Keeper) EscrowAddress(c context.Context, req *types.QueryEscrowAddressRequest) (*types.QueryEscrowAddressResponse, error) {
 	if req == nil {
 		return nil, status.Error(codes.InvalidArgument, "empty request")
 	}
 
 	addr := types.GetEscrowAddress(req.PortId, req.ChannelId)
 
+	if err := validategRPCRequest(req.PortId, req.ChannelId); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
bznein: Good point, added!

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -112,13 +114,26 @@ func (k Keeper) DenomHash(c context.Context, req *types.QueryDenomHashRequest) (
...
+	}
+
+	ctx := sdk.UnwrapSDKContext(c)
+	found := k.channelKeeper.HasChannel(ctx, req.PortId, req.ChannelId)
+	if !found {
```
damiannolan, line 130: This could be condensed but I don't really feel strongly about it...

```suggestion
	if !k.channelKeeper.HasChannel(ctx, req.PortId, req.ChannelId) {
```

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -112,13 +114,26 @@ func (k Keeper) DenomHash(c context.Context, req *types.QueryDenomHashRequest) (
...
+		return nil, status.Error(
+			codes.NotFound,
+			errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "port-id: %s, channel-id %s", req.PortId, req.ChannelId).Error(),
```
damiannolan, line 133: mega formatting nit: see errors msgs in handshake.go 04-channel for example

```suggestion
			errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "port ID (%s) channel ID (%s)", req.PortId, req.ChannelId).Error(),
```

damiannolan: to be fair there's probably inconsistency in places, but I think the suggested might be mostly standard 😅

damiannolan (approved):
Nice one! Thanks for picking this up @bznein 

All I left was nits, lgtm!

modules/apps/transfer/keeper/grpc_query_test.go
```diff
@@ -245,12 +245,34 @@ func (suite *KeeperTestSuite) TestEscrowAddress() {
...
+			"failure - channel not found due to invalid channelID",
+			func() {
+				req = &types.QueryEscrowAddressRequest{
+					PortId:    ibctesting.TransferPort,
+					ChannelId: ibctesting.InvalidID,
+				}
+			},
+			false,
+		},
 	}
```
crodriguezvega, line 268: Should we add a test where both port ID and channel ID are valid, but no channel is found for those parameters? Because the tests will fail on `validategRPCRequest` because either port ID or channel ID is not valid, but we don't have a test case that fails if `HasChannel` returns false, right?

crodriguezvega (commented):
Thanks, @bznein. I left a comment because maybe there's one more test case we can add?

modules/apps/transfer/keeper/grpc_query_test.go
```diff
@@ -245,12 +245,34 @@ func (suite *KeeperTestSuite) TestEscrowAddress() {
 			},
 			true,
 		},
+		{
+			"failure - channel not found due to invalid portID",
```
crodriguezvega, line 249: I think I would put it like this, because the code doesn't actually try to find the channel, it fails before that, right?

```suggestion
			"failure - invalid portID",
```

modules/apps/transfer/keeper/grpc_query_test.go
bznein: I might have misunderstood it :) invalidID is equal to IDIsInvalid (https://github.com/cosmos/ibc-go/blob/2555a7c504a904064d659e4c1a3a74000887f73d/testing/values.go#L37) but when we check for its validity in validategRPCRequest (https://github.com/cosmos/ibc-go/blob/main/modules/core/24-host/validate.go#L39-L60) I don't think this value will cause that function to return false? The ID is not blank, doesn't contain slashes, it's within the range for length, and satisfies the regex: https://github.com/cosmos/ibc-go/blob/main/modules/core/24-host/validate.go#L28

I just assumed that "invalidID" meant something like "an ID that would neer match an existing port/channel ID" but maybe I was wrong :)

bznein: To verify this I commented out the check for `HasChannel` and ran the test again, and got this failure:

```
--- FAIL: TestKeeperTestSuite (55.03s)
    --- FAIL: TestKeeperTestSuite/TestEscrowAddress (1.59s)
        --- FAIL: TestKeeperTestSuite/TestEscrowAddress/Case_failure_-_channel_not_found_due_to_invalid_portID (0.40s)
            grpc_query_test.go:287: 
                        Error Trace:    /workspaces/ibc-go/modules/apps/transfer/keeper/grpc_query_test.go:287
                                                                /go/pkg/mod/github.com/stretchr/testify@v1.9.0/suite/suite.go:115
                        Error:          An error is expected but got nil.
                        Test:           TestKeeperTestSuite/TestEscrowAddress/Case_failure_-_channel_not_found_due_to_invalid_portID
        --- FAIL: TestKeeperTestSuite/TestEscrowAddress/Case_failure_-_channel_not_found_due_to_invalid_channelID (0.43s)
            grpc_query_test.go:287: 
                        Error Trace:    /workspaces/ibc-go/modules/apps/transfer/keeper/grpc_query_test.go:287
                                                                /go/pkg/mod/github.com/stretchr/testify@v1.9.0/suite/suite.go:115
                        Error:          An error is expected but got nil.
                        Test:           TestKeeperTestSuite/TestEscrowAddress/Case_failure_-_channel_not_found_due_to_invalid_channelID
```

which leads me to think that we are testing for the correct path here?

bznein: (that being said, this means that we can -and should- probably add a test for `validateGRPCRequest`, though maybe this can be delayed until we maybe consolidate it into a single location, as discussed here https://github.com/cosmos/ibc-go/pull/6059#discussion_r1539199138)

modules/apps/transfer/keeper/grpc_query_test.go
bznein: I think this is what we cover in the discussion here https://github.com/cosmos/ibc-go/pull/6059#discussion_r1539724255 :)

modules/apps/transfer/keeper/grpc_query_test.go
crodriguezvega: Oh, yes, my bad. I misinterpreted the function :)

Yes, sounds good to wait to see where that function lands and test it once there.

crodriguezvega: Although to come back to [Jim's point](https://github.com/cosmos/ibc-go/pull/6059#discussion_r1539215613), maybe it is worth to add a simple test case for both port ID empty and channel ID empty.

bznein: Sounds good! I removed one of the two "channel not found" test because they were actually both just testing the same line (`hasChannel() == false`) and added one for an empty portID and one for an empty channelID

crodriguezvega: Great, thank you!

crodriguezvega (approved):
Great work, @bznein. Looking forward to the follow-up PRs to add the checks in the other gRPCs. :)

PR #6116, opened 2024-04-08
Add test for converting packet data from v1 to v3

chatton:
## Description

WIP

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/apps/transfer/internal/convert/convert.go
+++ b/modules/apps/transfer/internal/convert/convert.go
@@ -1,0 +1,82 @@
+package convert
+
+import (
+	v1types "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	v3types "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types/v3"
+	channeltypes "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+	"strconv"
+	"strings"
+)
+
+// PacketDataV1ToV3 converts a v1 packet data to a v2 packet data.
+func PacketDataV1ToV3(packetData v1types.FungibleTokenPacketData) v3types.FungibleTokenPacketData {
+	amount, err := strconv.ParseUint(packetData.Amount, 10, 64)
+	if err != nil {
+		panic(err)
+	}
+
+	v2Denom, trace := extractDenomAndTraceFromV1Denom(packetData.Denom)
+
+	// TODO: we should fail here, but some tests fail with this panic. We can re-visit.
+	// if v2Denom == "" {
+	// 	panic("base denom cannot be empty")
+	// }
+
+	return v3types.FungibleTokenPacketData{
+		Tokens: []*v3types.Token{
+			{
+				Denom:  v2Denom,
+				Amount: amount,
+				Trace:  trace,
+			},
+		},
+		Sender:   packetData.Sender,
+		Receiver: packetData.Receiver,
+		Memo:     packetData.Memo,
+	}
+}
+
+func extractDenomAndTraceFromV1Denom(v1Denom string) (string, []string) {
+	v1DenomTrace := v1types.ParseDenomTrace(v1Denom)
+
+	splitPath := strings.Split(v1Denom, "/")
+	pathSlice := extractPathAndBaseFromFullDenomSlice(splitPath)
+
+	if len(pathSlice) == 0 {
+		return v1DenomTrace.BaseDenom, []string(nil)
+	}
+
+	if len(pathSlice)%2 != 0 {
+		panic("pathSlice length is not even")
+	}
+
+	var trace []string
+	for i := 0; i < len(pathSlice); i += 2 {
+		trace = append(trace, strings.Join(pathSlice[i:i+2], "/"))
+	}
+
+	return v1DenomTrace.BaseDenom, trace
+}
+
+func extractPathAndBaseFromFullDenomSlice(fullDenomItems []string) []string {
+	var pathSlice []string
+
+	length := len(fullDenomItems)
+	for i := 0; i < length; i += 2 {
+		// The IBC specification does not guarantee the expected format of the
+		// destination port or destination channel identifier. A short term solution
+		// to determine base denomination is to expect the channel identifier to be the
+		// one ibc-go specifies. A longer term solution is to separate the path and base
+		// denomination in the ICS20 packet. If an intermediate hop prefixes the full denom
+		// with a channel identifier format different from our own, the base denomination
+		// will be incorrectly parsed, but the token will continue to be treated correctly
+		// as an IBC denomination. The hash used to store the token internally on our chain
+		// will be the same value as the base denomination being correctly parsed.
+		if i < length-1 && length > 2 && channeltypes.IsValidChannelID(fullDenomItems[i+1]) {
+			pathSlice = append(pathSlice, fullDenomItems[i], fullDenomItems[i+1])
+		} else {
+			break
+		}
+	}
+	return pathSlice
+}
--- a/modules/apps/transfer/internal/convert/convert_test.go
+++ b/modules/apps/transfer/internal/convert/convert_test.go
@@ -1,0 +1,102 @@
+package convert
+
+import (
+	v1types "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	v3types "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types/v3"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestConvertPacketV1ToPacketV2(t *testing.T) {
+	const (
+		sender   = "sender"
+		receiver = "receiver"
+	)
+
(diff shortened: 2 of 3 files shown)


modules/apps/transfer/internal/convert/convert.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"strings"
+
+	"github.com/cosmos/ibc-go/v8/modules/apps/transfer/internal/denom"
+	v1types "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	v3types "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types/v3"
+)
+
+// PacketDataV1ToV3 converts a v1 (ICS20-V1) packet data to a v3 (ICS20-V2) packet data.
+func PacketDataV1ToV3(packetData v1types.FungibleTokenPacketData) v3types.FungibleTokenPacketData {
```
chatton, line 12: I don't really like how we're  going from V1 -> V3 in terms of proto, but going from ICS20-V1 to V2, open to suggestions here.

modules/apps/transfer/internal/denom/denom.go
```diff
@@ -0,0 +1,41 @@
...
+	"strings"
+
+	"strings"
+
+	channeltypes "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+)
+
+// ExtractPathAndBaseFromFullDenom returns the trace path and the base denom from
+// the elements that constitute the complete denom.
+func ExtractPathAndBaseFromFullDenom(fullDenomItems []string) ([]string, string) {
```
chatton, line 13: note: this was moved from trace.go, and made internal. The one difference is a pathSlice is returned instead of a path. This makes the fn work for both use cases.

chatton:
This PR should be ready to go, today is my last day for a few weeks, so if any changes need to be made, feel free to take over this PR and get it over the line :+1:

modules/apps/transfer/internal/convert/convert.go
charleenfei: i'm ok with the `ICS20V1ToV2` framing as opposed to V1 -> V3 for PacketData!

charleenfei: we can do this in a followup pr

charleenfei (approved):

modules/apps/transfer/internal/convert/convert.go
srdtrk: Here are some potential ways to improve naming:
- Move v2 to v1 and v3 to v2.
- The name of the struct doesn't really mean much. So we could Name it `FungibleTokenPacketDataV2` and put it in v2.
- Move directly to `ics20-3`, SDK `v0.52` style.
- Keep as is

modules/apps/transfer/internal/convert/convert.go
```diff
@@ -0,0 +1,57 @@
...
+		Memo:     packetData.Memo,
+	}
+}
+
+// extractDenomAndTraceFromV1Denom extracts the base denom and remaining trace from a v1 IBC denom.
+func extractDenomAndTraceFromV1Denom(v1Denom string) (string, []string) {
+	v1DenomTrace := v1types.ParseDenomTrace(v1Denom)
+
+	splitPath := strings.Split(v1Denom, "/")
+	pathSlice, _ := denom.ExtractPathAndBaseFromFullDenom(splitPath)
```
srdtrk, line 37: This logic is already done in `ParseDenomTrace`. Instead of repeating the logic, we should probably just trim suffix `splitPath` using `v1DenomTrace.BaseDenom`

modules/apps/transfer/internal/convert/convert_test.go
```diff
@@ -0,0 +1,112 @@
...
+	const (
+		sender   = "sender"
+		receiver = "receiver"
+	)
+
+	testCases := []struct {
+		name        string
+		v1Data      v1types.FungibleTokenPacketData
+		v3Data      v3types.FungibleTokenPacketData
+		shouldPanic bool
```
srdtrk, line 22:
```suggestion
		expPanic bool
```
This can even be made a string/any to see if the panic message is the expected message

srdtrk (commented):
Looks good. Made some improvement suggestions

modules/apps/transfer/internal/convert/convert_test.go
damiannolan: couldn't we do `expPanic error` and do `RequirePanicsWithError` since there is no error return case?

modules/apps/transfer/internal/convert/convert.go
damiannolan: I suggested moving the proto definition to proto pkg v3 to see what it would look like seeing as it kind of makes sense to follow some proto rules. But maybe adding `FunginbleTokenPacketDataV2` or some other option is less noisy given that we are already on proto pkg v2.

When proto pkg v2 was added for packet data with amount as string we removed the v1 proto definition ([ref](https://github.com/cosmos/ibc-go/pull/350/files#diff-b92045f77047ee77e807794e59ed0ee6486f96087f31990086202a053df40c9e)), but this scenario is different given we want to maintain the existing packet data in parallel. 

There is a few different options here as @srdtrk has pointed out. I think we should avoid `ics20-3` at all costs - this would add too much confusion 😅 
I think another option is probably making the new packet data the proto pkg v2 `FungibleTokenPacketData` and renaming the existing one to have the V1 suffix or "Legacy" prefix or something. That might be a cleaner option completely, rather than introducing proto pkg v3 - I think this should be possible but will need to look into it a bit more, any thoughts on this?

modules/apps/transfer/internal/convert/convert.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"strings"
+
+	"github.com/cosmos/ibc-go/v8/modules/apps/transfer/internal/denom"
+	v1types "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	v3types "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types/v3"
+)
+
+// PacketDataV1ToV3 converts a v1 (ICS20-V1) packet data to a v3 (ICS20-V2) packet data.
```
damiannolan, line 11: so where is this being used at the moment? or it isn't?

charleenfei: it will eventually be used in the ibc module packet lifecycle methods (`OnRecvPacket`, `OnAckPacket` `OnTimeoutPacket`) as those methods take in a `channeltypes.Packet` and we need to ensure that the Packet's data is read correctly.

modules/apps/transfer/internal/convert/convert_test.go
charleenfei: i can update this!

modules/apps/transfer/internal/convert/convert.go
charleenfei: ```I think another option is probably making the new packet data the proto pkg v2 FungibleTokenPacketData and renaming the existing one to have the V1 suffix or "Legacy" prefix or something. ```

this makes the most sense to me, as I think having the proto pkg move from v1 -> v2 in parallel with ics20 moving from v1 -> v2 seems to make the most sense. If there is general agreement I can make this update in this  pr!

modules/apps/transfer/internal/convert/convert.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		Sender:   packetData.Sender,
+		Receiver: packetData.Receiver,
+		Memo:     packetData.Memo,
+	}
+}
+
+// extractDenomAndTraceFromV1Denom extracts the base denom and remaining trace from a v1 IBC denom.
+func extractDenomAndTraceFromV1Denom(v1Denom string) (string, []string) {
+	v1DenomTrace := v1types.ParseDenomTrace(v1Denom)
```
AdityaSripal, line 34: Seems to me like we can just use this function. And then split the trace in here further and return the base denom as-is

modules/apps/transfer/internal/convert/convert_test.go
```diff
@@ -0,0 +1,113 @@
...
+					{
+						Denom:  "atom/pool",
+						Amount: "1000",
+						Trace:  []string{"transfer/channel-0", "transfer/channel-1"},
+					},
+				}, sender, receiver, ""),
+			nil,
+		},
+		{
+			"success: longer trace with non transfer port",
```
AdityaSripal, line 82: Can we do non-transfer port with a denom with slash

AdityaSripal (approved):
The first function seems unnecessarily complicated at first glance. But everything else looks good

modules/apps/transfer/internal/convert/convert.go
charleenfei: https://github.com/cosmos/ibc-go/issues/6170 added this issue to track this topic so we can continue on in parallel!

Issue #6133, opened 2024-04-10
Remove Router from IBCKeeper and use PortKeeper


damiannolan:
## Summary

We currently maintain a reference to the IBC application router on the `IBCKeeper` as well as the `PortKeeper` submodule keeper.
This is often referred to as the "port router" as this is how we look up application callback stacks in order to route application handshake msgs and packet data to the appropriate handlers.

With moving towards a more consistent usage of standardising submodule keepers as reference types (see https://github.com/cosmos/ibc-go/pull/6081) we should now be able to remove the `Router` reference from the `IBCKeeper` and instead rely directly on the one in `PortKeeper`.

With this change we can also probably make `Router` unexported within `PortKeeper` and add a `Route()` method on the `PortKeeper` instead, and all application callback stack access can flow through there.

- [ ] Remove `Router` from `IBCKeeper`
- [ ] Add `GetRoute() types.IBCModule` to `PortKeeper`
- [ ] Replace usage of `Router` in `core/keeper/msg_server.go` with `PortKeeper.Route()`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

bznein:
@damiannolan I wouldn't mind taking this one :)

chatton:
go for it @bznein this one depends on #6081 to me merged first

PR #6138, opened 2024-04-10
chore(api!): replace Router with PortKeeper in IBCKeeper.

bznein:
## Description

Removes `Router` from `IBCKeeper` and replace calls to it with calls to the new function `PortKeeper.GetRoute()`

Note: I didn't add tests for this new method for the reason that `GetRoute` on the underlying router is already tested, and I wasn't able to get a proper set up for the test. I'm more than open to suggestions! 

closes: #6133

---

write a little note why.

- [] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).

## Summary by CodeRabbit

- **Refactor**
    - Enhanced channel operations by replacing `Router` with `PortKeeper` across modules for improved reliability.
    - Introduced `GetRoute()` function for better route handling in channel operations.
- **Bug Fixes**
    - Improved application callbacks for channel operations like opening, closing, and timeouts.

### DIFF
--- a/modules/core/05-port/keeper/keeper.go
+++ b/modules/core/05-port/keeper/keeper.go
@@ -82,7 +82,8 @@ func (k *Keeper) LookupModuleByPort(ctx sdk.Context, portID string) (string, *ca
 	return types.GetModuleOwner(modules), capability, nil
 }
 
-// GetRoute() returns a IBCModule for a given module.
+// GetRoute returns a IBCModule for a given module, and a boolean indicating
+// whether or not the route is present.
 func (k Keeper) GetRoute(module string) (types.IBCModule, bool) {
 	return k.Router.GetRoute(module)
 }
--- a/modules/core/05-port/keeper/keeper.go
+++ b/modules/core/05-port/keeper/keeper.go
@@ -82,8 +82,8 @@ func (k *Keeper) LookupModuleByPort(ctx sdk.Context, portID string) (string, *ca
 	return types.GetModuleOwner(modules), capability, nil
 }
 
-// GetRoute returns a IBCModule for a given module, and a boolean indicating
+// Route returns a IBCModule for a given module, and a boolean indicating
 // whether or not the route is present.
-func (k *Keeper) GetRoute(module string) (types.IBCModule, bool) {
-	return k.Router.GetRoute(module)
+func (k *Keeper) Route(clientID string) (types.IBCModule, bool) {
+	return k.Router.GetRoute(clientID)
 }
--- a/modules/core/05-port/keeper/keeper.go
+++ b/modules/core/05-port/keeper/keeper.go
@@ -84,6 +84,6 @@ func (k *Keeper) LookupModuleByPort(ctx sdk.Context, portID string) (string, *ca
 
 // GetRoute returns a IBCModule for a given module, and a boolean indicating
 // whether or not the route is present.
-func (k Keeper) GetRoute(module string) (types.IBCModule, bool) {
+func (k *Keeper) GetRoute(module string) (types.IBCModule, bool) {
 	return k.Router.GetRoute(module)
 }
--- a/modules/core/05-port/keeper/keeper.go
+++ b/modules/core/05-port/keeper/keeper.go
@@ -81,3 +81,8 @@ func (k Keeper) LookupModuleByPort(ctx sdk.Context, portID string) (string, *cap
 
 	return types.GetModuleOwner(modules), capability, nil
 }
+
+// GetRoute() returns a IBCModule for a given module.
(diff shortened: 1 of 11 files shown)


modules/core/05-port/keeper/keeper.go
```diff
@@ -81,3 +81,8 @@ func (k *Keeper) LookupModuleByPort(ctx sdk.Context, portID string) (string, *ca
 
 	return types.GetModuleOwner(modules), capability, nil
 }
+
+// GetRoute() returns a IBCModule for a given module.
+func (k Keeper) GetRoute(module string) (types.IBCModule, bool) {
```
chatton, line 86: I think we should make this a method on `*Keeper` instead of the struct since it's referencing a mutable type `Router`, and also to stay consistent with the other methods.

modules/core/05-port/keeper/keeper.go
```diff
@@ -81,3 +81,8 @@ func (k *Keeper) LookupModuleByPort(ctx sdk.Context, portID string) (string, *ca
 
 	return types.GetModuleOwner(modules), capability, nil
 }
+
+// GetRoute() returns a IBCModule for a given module.
```
chatton, line 85:
```suggestion
// GetRoute returns a IBCModule for a given module, and a boolean indicating
// whether or not the route is present.
```

modules/core/05-port/keeper/keeper.go
```diff
@@ -81,3 +81,8 @@ func (k Keeper) LookupModuleByPort(ctx sdk.Context, portID string) (string, *cap
...
+func (k Keeper) GetRoute(module string) (types.IBCModule, bool) {
+	return k.Router.GetRoute(module)
+}
```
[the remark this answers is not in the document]
chatton: I don't know if this is actually a problem, we could error in the case of the router not being sealed, but at the moment the only usage of the `Sealed()` fn is to check whether or not we can seal a router a second time.

chatton (approved):
LGTM, left a few small suggestions but nothing major :+1:

modules/core/05-port/keeper/keeper.go
bznein: Done!

modules/core/05-port/keeper/keeper.go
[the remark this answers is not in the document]
bznein: @chatton yeah I'm conflicted :) We could definitely check for `k.Router == nil || !k.Router.Sealed()` here and either panic or return `nil, false` in that case. WDYT?

crodriguezvega (approved):
You're on fire, @bznein! 🔥🧯

modules/core/keeper/keeper.go
```diff
@@ -33,7 +33,6 @@ type Keeper struct {
 	ConnectionKeeper *connectionkeeper.Keeper
 	ChannelKeeper    *channelkeeper.Keeper
 	PortKeeper       *portkeeper.Keeper
-	Router           *porttypes.Router
```
crodriguezvega, line 36: Removing this is API breaking, so we should document it in the migration docs of v9. I will also change the PR name to reflect this.

modules/core/05-port/keeper/keeper.go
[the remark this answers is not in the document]
chatton: yeah I wouldn't be against panicking if the router is not sealed, as it shouldn't be used when it is partially configured.

Normally wouldn't like to panic,  but the alternative of `false, nil` just means a silent failure IMO.

modules/core/05-port/keeper/keeper.go
```diff
@@ -81,3 +81,9 @@ func (k *Keeper) LookupModuleByPort(ctx sdk.Context, portID string) (string, *ca
 
 	return types.GetModuleOwner(modules), capability, nil
 }
+
+// GetRoute returns a IBCModule for a given module, and a boolean indicating
+// whether or not the route is present.
+func (k *Keeper) GetRoute(module string) (types.IBCModule, bool) {
```
damiannolan, line 87: we have another router in 02-client for light client modules and the func signature we use there is `Route(clientID string) (exported.LightClientModule, bool)`

it would be nice to have some symmetry between the two imo!

damiannolan (commented):
Code changes look great! Thanks for picking this up, just left one comment about aligning APIs as we now have two routers in ibc core - one for ibc apps and one for ibc client modules

modules/core/05-port/keeper/keeper.go
bznein: Sure that makes sense! Done

PR #6167, opened 2024-04-16
fix: avoid panic when migrate param for newly added host

mmsqe:
## Description

- **Problem**
    - when migrating ica from version 2 to 3 using ibc-go [v8.2.0](https://github.com/crypto-org-chain/cronos/blob/90fc8be8d3cd5f3aa0a33f41d1fab6b6fda5d5ed/go.mod#L27), ica host legacy params moving to store caused crash during migration.

- **Solution**
	- setting legacySubspace as nil to set params to default values during parameter migration.

---

write a little note why.

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced the migration process for parameters in the interchain accounts module to prevent potential crashes during parameter migration in specific scenarios.
- **New Features**
	- Added a method to check for the existence of parameter sets, improving robustness in parameter handling.
- **Tests**
	- Introduced new test cases to validate the successful migration of parameters even when no legacy parameters are provided.


damiannolan:
Hi @mmsqe, did you encounter a panic in the scenario you've suggested with the PR title? 

We normally encourage a short description when opening PRs that do not have an existing issue. 
It would be useful to know for example:
- The scope of this PR
- What the issue is related to and how you encountered the problem?
- What versions you encountered the issue, so that we can backport the fix to appropriate release branches

mmsqe:
> Hi @mmsqe, did you encounter a panic in the scenario you've suggested with the PR title?
> 
> We normally encourage a short description when opening PRs that do not have an existing issue. It would be useful to know for example:
> 
> * The scope of this PR
> * What the issue is related to and how you encountered the problem?
> * What versions you encountered the issue, so that we can backport the fix to appropriate release branches
> 
> edit: The scenario here is that a chain includes ica controller but does not include host functionality, in an upgrade the ica host is enabled and subkeeper is added. It is the same release which includes the params migration. This means an entry in the `versionMap` for running migrations exists (due to the pre-existing ica controller), however no params are set for the host? Am I understanding this correctly, more or less?

Thanks @damiannolan, you're right, just updated the description, see if it's ok.

modules/apps/27-interchain-accounts/host/keeper/migrations.go
```diff
@@ -22,10 +22,13 @@ func NewMigrator(k *Keeper) Migrator {
 func (m Migrator) MigrateParams(ctx sdk.Context) error {
 	if m.keeper != nil {
 		var params types.Params
-		m.keeper.legacySubspace.GetParamSet(ctx, &params)
-
-		if err := params.Validate(); err != nil {
-			return err
+		if m.keeper.legacySubspace == nil {
```
damiannolan, line 25: Just to clarify when icahost was added in your upgrade it was created with a `nil` legacy subspace? @mmsqe

damiannolan: Okay, thank you! In that case this should be fine.

I'm curious about a case in which an actual params subspace is created and passed but which does not have any params stored in the param set from a previous version. As genesis would already have been run for the ica module but not setup default params I would imagine it to be empty. I assume in that case `params.Validate()` should error below, and this would technically misconfiguration by the chain's app wiring.

damiannolan: Aha, yea I believe you're right @DimitrisJim. With `RegisterParamSet(&Params{})`, right? 

That's probably fine behaviour imo - given this would probably be the result of misconfig anyways - a param's change would need to be done for allowing msgs in `[]string` then. This is a pretty edge case scenario either way. Most chains deploy host first anyways, and controller later

damiannolan: Wondering if this migration code would be safer and cover edge cases such as described above (this would at least init params with all msgs allowed instead of the empty string slice, its a bit more explicit than relying on a side effect of `RegisterParamSet`):

```go
if m.keeper != nil {
        params := types.DefaultParams()
	if m.keeper.legacySubspace != nil {
		m.keeper.legacySubspace.GetParamSetIfExists(ctx, &params)
	}

	if err := params.Validate(); err != nil {
		return err
	}

	m.keeper.SetParams(ctx, params)
	m.keeper.Logger(ctx).Info("successfully migrated ica/host submodule to self-manage params")
}
```

Any thoughts? Happy to stick with the current changes tho if agreed.

damiannolan: > I'd assume we'd want to panic if we try to get something and it doesn't exist.

I'm not sure, and I don't feel particularly strongly about this - if it doesn't exist we've at least assigned safe defaults via `types.DefaultParams` rather than panicking. Do you have any feedback here @mmsqe, as someone maintaining a chain?

> One sidenote here, should we be doing same sanity checks for controller?

@mmsqe yes I think it would be nice to align the controller migration code as well, thank you! 🙏

modules/apps/27-interchain-accounts/types/expected_keepers.go
```diff
@@ -36,5 +36,5 @@ type PortKeeper interface {
 
 // ParamSubspace defines the expected Subspace interface for module parameters.
 type ParamSubspace interface {
-	GetParamSet(ctx sdk.Context, ps paramtypes.ParamSet)
```
damiannolan, line 39: Sorry to be a pain but can we keep `GetParamSet` since it would technically be an API breaking change? Happy to just add the `IfExists` method as addition, but I think we should keep this in order to preserve our release guarantees when backporting to v8.3

damiannolan (approved):
Thank you so much! LGTM

Issue #5733, opened 2024-01-25
Bundle the transition to FLUSHCOMPLETE into a single function


colin-axner:
## Summary

When transitioning to FLUSHCOMPLETE, two invariants MUST be true:
* the counterparty timeout has not elapsed
* there are no in-flight packets for our end to process

We have verified as the main branch stands today, the implementation is correct. To future proof maintenance, we can make this invariant easier to maintain by bundling these checks and state transitions into a single function. This may require some reworkings of the Ack and Confirm handlers as we currently check the timeout in the handler and perform the state transition in the write function

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
@hoangdv2429 Sorry for the late reply: yes, you can work on this one. Thank you!

PR #6327, opened 2024-05-17
refactor: bundle the transition to FLUSHCOMPLETE into a single function

TropicalDog17:
## Description

closes: #5733 

---

write a little note why.

## Summary by CodeRabbit

- **Refactor**
  - Improved the handling of packet flushing and channel state during upgrades for better efficiency and reliability.


modules/core/04-channel/keeper/packet.go
```diff
@@ -489,27 +489,35 @@ func (k *Keeper) AcknowledgePacket(
...
-				emitChannelFlushCompleteEvent(ctx, packet.GetSourcePort(), packet.GetSourceChannel(), channel)
-			}
+		if err := k.handleFlushState(ctx, &packet, &channel); err != nil {
+			return err
 		}
 	}
+	return nil
+}
+
+func (k *Keeper) handleFlushState(ctx sdk.Context, packet *types.Packet, channel *types.Channel) error {
```
crodriguezvega, line 499: I think it's better to return a boolean in the call site then you can do:

```go
if aborted := k.handleFlushState(ctx, &packet, &channel); aborted == true {
  return nil
}
```

```suggestion
func (k *Keeper) handleFlushState(ctx sdk.Context, packet *types.Packet, channel *types.Channel) bool {
```

modules/core/04-channel/keeper/packet.go
```diff
@@ -489,27 +489,35 @@ func (k *Keeper) AcknowledgePacket(
...
+	counterpartyUpgrade, found := k.GetCounterpartyUpgrade(ctx, packet.GetSourcePort(), packet.GetSourceChannel())
+	if !found {
+		return nil
+	}
+
+	timeout := counterpartyUpgrade.Timeout
+	selfHeight, selfTimestamp := clienttypes.GetSelfHeight(ctx), uint64(ctx.BlockTime().UnixNano())
+
+	if timeout.Elapsed(selfHeight, selfTimestamp) {
+		// packet flushing timeout has expired, abort the upgrade and return nil,
```
crodriguezvega, line 509:
```suggestion
		// packet flushing timeout has expired, abort the upgrade,
```

crodriguezvega (commented):
Thanks for opening the PR, @TropicalDog17.

I would be nice if we could also replace [this code](https://github.com/cosmos/ibc-go/blob/5069ab673b9cedc88916debaada7fdc2f293e513/modules/core/04-channel/keeper/timeout.go#L155-L172) with a call to `handleFlushState`. I just noticed actually that there might be a small bug in that code, since we are not returning if the upgrade is aborted, [as the comment indicates that we should](https://github.com/cosmos/ibc-go/blob/5069ab673b9cedc88916debaada7fdc2f293e513/modules/core/04-channel/keeper/timeout.go#L163).

However, in order to replace that code you need to make sure that we exit the function only if the upgrade is aborted, since there is [more logic to execute in case the upgrade is not aborted](https://github.com/cosmos/ibc-go/blob/5069ab673b9cedc88916debaada7fdc2f293e513/modules/core/04-channel/keeper/timeout.go#L175-L205).

TropicalDog17:
@crodriguezvega thanks for your review, I've refactored accordingly.

modules/core/04-channel/keeper/packet.go
```diff
@@ -489,27 +489,46 @@ func (k *Keeper) AcknowledgePacket(
...
 	}
 
 	return nil
 }
+
+// handleFlushState is called when a packet is acknowledged or timed out and the channel is in
+// FLUSHING state. It checks if the upgrade has timed out and if so, aborts the upgrade. If all
+// packets have completed their lifecycle, it sets the channel state to FLUSHCOMPLETE and
+// emits a channel_flush_complete event. Returns true if the upgrade was aborted, false otherwise.
+func (k *Keeper) handleFlushState(ctx sdk.Context, packet *types.Packet, channel *types.Channel) bool {
```
crodriguezvega, line 510: Instead of a pointer we could pass a copy of the whole struct, just to prevent that we could by mistake modify the packet within this function. I don't feel strongly about it, though, so I will leave others to also chime in.

```suggestion
func (k *Keeper) handleFlushState(ctx sdk.Context, packet types.Packet, channel *types.Channel) bool {
```

crodriguezvega (approved):
Great work, @TropicalDog17. Thank you!

modules/core/04-channel/keeper/packet.go
TropicalDog17: Yeah seems reasonable to pass a struct, I will update.

crodriguezvega (approved):

PR #6330, opened 2024-05-20
Use type with V2 suffix for package data

### LINKED ISSUE #6170: update proto definitions for ics20-v2 feature
>I suggested moving the proto definition to proto pkg v3 to see what it would look like seeing as it kind of makes sense to follow some proto rules. But maybe adding `FunginbleTokenPacketDataV2` or some other option is less noisy given that we are already on proto pkg v2. 
> When proto pkg v2 was added for packet data with amount as string we removed the v1 proto definition ([ref](https://github.com/cosmos/ibc-go/pull/350/files#diff-b92045f77047ee77e807794e59ed0ee6486f96087f31990086202a053df40c9e)), but this scenario is different given we want to maintain the existing packet data in parallel. There is a few different options here as @srdtrk has pointed out. I think we should avoid `ics20-3` at all costs - this would add too much confusion 😅 
> I think another option is probably making the new packet data the proto pkg v2 `FungibleTokenPacketData` and renaming the existing one to have the V1 suffix or "Legacy" prefix or something. That might be a cleaner option completely, rather than introducing proto pkg v3 - I think this should be possible but will need to look into it a bit more, any thoughts on this?

Originally posted by @damiannolan in https://github.com/cosmos/ibc-go/pull/6116#discussion_r1565709461_

chatton:
## Description

This PR moves test files and types out of the v3 package, and into the v2 package.

The types has been renamed to `FungibleTokenPacketDataV2` as discussed.

closes #6170


colin-axner (approved):
Quick one, nice!! Everything looks great! :sunflower:

Issue #5793, opened 2024-02-01
ICS20 V2 Support for Multi Denom 


chatton:
## Requirements document

Shout out to @charleenfei for all the work done so far in [this PoC branch](https://github.com/cosmos/ibc-go/pull/5763) that we've been working on.

In the linked PoC, we have the spec changes implemented and unit and e2e tests passing (additional ones still need to be added)

In the PoC, we have some full unit tests and E2E tests passing and can be used as a reference, [unit](https://github.com/cosmos/ibc-go/blob/366bb8376c22d7e27be5af2a1587599793ae62eb/modules/apps/transfer/ibc_module_test.go#L501-L587) and [e2e](https://github.com/cosmos/ibc-go/blob/8d5e927df2e4b32fdf0a56c786da50398594142a/e2e/tests/transfer/base_test.go#L155-L253).

This issue will outline the proposed order of PR creation for the changes to be most easily reviewed.

## IBC spec

[1020](https://github.com/cosmos/ibc/pull/1020)

## Implementation issues

- [x] #5794
- [x] #5799
- [x] #5795
- [x] #6114
- [x] #5796
- [x] #5797
- [x] #5798
- [x] #5801
- [x] #5803 
- [x] #5804
- [x] #5805
- [x] #5802
- [x] #6109

## Checklist

- [ ] Internal audit(s)
- [ ] External audit(s)
- [ ] Documentation
- [ ] Swagger
- [ ] Integration with ledger
- [ ] Integration with relayers:
  - [ ] Hermes
  - [ ] Rly

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

srdtrk:
When we are doing such a rewrite, is it also possible to migrate storage to `"cosmossdk.io/collections"`?

damiannolan:
We should also test ledger support since `MsgTransfer` is being modified. I will add to the checklist on the issue

Issue #6221, opened 2024-04-24
Migrate internal storage of trace string to array of trace info


colin-axner:
## Summary

Migrate internal trace information of ics20 token to store individual trace hop pairs, []string, to match ics20-v2 updates

## Problem Definition

In maintaining backwards compatibility, there's an unnecessary amount of parsing, joining occurring which would be largely simplified by performing this migration. 

Upon SendPacket:
1. ibc token lookup occurs from hash to trace info
2. [full joined path is returned](https://github.com/cosmos/ibc-go/blob/cian/charly/transfer-multi-denom-poc/modules/apps/transfer/keeper/relay.go#L98)
3. the full path is then [re-parsed](https://github.com/cosmos/ibc-go/blob/cian/charly/transfer-multi-denom-poc/modules/apps/transfer/keeper/relay.go#L136) (despite its denom, path being cached)
4. in this re-parsing, we parse out the [denom trace](https://github.com/cosmos/ibc-go/blob/cian/charly/transfer-multi-denom-poc/modules/apps/transfer/v2/transferv2.go#L40C2-L40C48) (despite it being available in step 1)
5. and then we reparse the path again (second time in this stack trace) to [format](https://github.com/cosmos/ibc-go/blob/cian/charly/transfer-multi-denom-poc/modules/apps/transfer/v2/transferv2.go#L55) the []trace pair which is set in the packet, but not internal state

In the expected flow:
1. ibc token lookup occurs from hash, returns ibc token info (denom, []trace)
2. any usage which needs full denom path (v1 compat) can do a basic strings.join on []trace

## Use cases

Mostly simplifies maintenance and optimizes the code. In the long run, it would remove the possibility for a [minor issue](https://github.com/cosmos/ibc-go/blob/4efbc8a1c0bf756fad5c7155756dae60b6d49321/modules/apps/transfer/types/trace.go#L95) (once support for ics20-v1 is dropped)

## Proposal

Perform an in-place migration which migrates the stored denom trace info from being a single string to an array of trace info. This makes internal storage compatible with the v2 approach rather than trying to reconstruct the v2 info based on v1 storage. Not essential, but likely worth the effort.

In practice we could remove/replace the `DenomTrace` type, ideally with a better named type

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #6352, opened 2024-05-22
[FEAT] ICS20-V2

chatton:
## Description

ICS20-V2 Feature branch

closes: #5793

---

write a little note why.

## Summary by CodeRabbit

- **Documentation**
  - Updated migration guide to reflect new parameters for `UnmarshalPacketData`.
  
- **New Features**
  - Introduced `DefaultTransferCoins` function for easier coin handling.
  - Added support for multi-token transfers in `MsgTransfer`.

- **Bug Fixes**
  - Corrected error handling and validation in `MsgTransfer` tests.
  
- **Refactor**
  - Updated `MsgTransfer` initialization to use `sdk.NewCoins` for better consistency.
  - Replaced `transfertypes.Version` with `transfertypes.V2` for channel versioning.
  
- **Tests**
  - Enhanced test cases to handle new `MsgTransfer` parameters and multi-token scenarios.
  
- **Chores**
  - Updated event attribute names for token amounts and refunds.
  - Modified `BankKeeper` interface to support variadic `sdk.Coin` arguments.


proto/ibc/applications/transfer/v1/tx.proto
```diff
@@ -49,6 +49,8 @@ message MsgTransfer {
   uint64 timeout_timestamp = 7;
   // optional memo
   string memo = 8;
+  // tokens to be transferred
+  repeated cosmos.base.v1beta1.Coin tokens = 9 [(gogoproto.nullable) = false, (amino.dont_omitempty) = true];
```
colin-axner, line 53: should we deprecate the `token` field?

proto/ibc/applications/transfer/v2/packet.proto
```diff
@@ -19,3 +21,27 @@ message FungibleTokenPacketData {
...
+}
+
+// Token defines a struct which represents a token to be transferred.
+message Token {
+  // the base token denomination to be transferred
+  string denom = 1;
+  // the token amount to be transferred
+  string amount = 2;
+  // the trace of the token
+  repeated string trace = 3;
```
colin-axner, line 46: maybe we should add godoc on what this means? ie is each string a port/channel pair or isolated port and isolated channel?

modules/apps/transfer/types/keys.go
```diff
@@ -38,11 +34,28 @@ const (
 	ParamsKey = "params"
 )
 
+const (
+	// V1 defines first version of the IBC transfer module
+	V1 = "ics20-1"
+
+	// V2 defines the current version the IBC transfer
+	// module supports
```
colin-axner, line 42: should we mention that V2 supports multidenom transfers and uses PacketDataV2?

modules/apps/transfer/ibc_module.go
```diff
@@ -352,11 +380,11 @@ func (IBCModule) OnChanUpgradeOpen(ctx sdk.Context, portID, channelID string, pr
 // UnmarshalPacketData attempts to unmarshal the provided packet data bytes
 // into a FungibleTokenPacketData. This function implements the optional
 // PacketDataUnmarshaler interface required for ADR 008 support.
-func (IBCModule) UnmarshalPacketData(bz []byte) (interface{}, error) {
-	var packetData types.FungibleTokenPacketData
-	if err := json.Unmarshal(bz, &packetData); err != nil {
+func (im IBCModule) UnmarshalPacketData(bz []byte) (interface{}, error) {
```
colin-axner, line 383: I would have assumed this returns the packet data based on the version, not that it converts to V2. I think it's fine if it converts to V2, but could be odd behaviour. It's technically okay since the interface only exists to extract from the memo in a standard way, which will be supported by V2 packet data in the exact same way as V1

e2e/testsuite/testsuite.go
```diff
@@ -154,6 +154,10 @@ func (s *E2ETestSuite) ConfigureRelayer(ctx context.Context, chainA, chainB ibc.
 	pathName := s.generatePathName()
 
 	channelOptions := ibc.DefaultChannelOpts()
+	// TODO: better way to do this.
+	// For now, set the version to the latest transfer module version
```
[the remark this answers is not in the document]
chatton: think this was just because `ibc.DefaultChannelOpts()` defaults to `ics20-1` so we are overriding that with `ics20-2`, this actually isn't too bad, we could also leave it as `empty` string but I think I'd prefer the explicit override here.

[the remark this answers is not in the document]
colin-axner: I see. This is because `ibc.DefaultChannelOpts()` is used by the grandpa wasm test. We could optionally change that code to use [`s.TransferChannelOpts()`](https://github.com/cosmos/ibc-go/blob/8eae033732183be839b3d368a6b97720768e7b3b/e2e/testsuite/testsuite.go#L454)?

modules/apps/transfer/types/token.go
```diff
@@ -0,0 +1,61 @@
...
+	errorsmod "cosmossdk.io/errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	denominternal "github.com/cosmos/ibc-go/v8/modules/apps/transfer/internal/denom"
+)
+
+// Validate validates a token denomination and trace identifiers.
+func (t Token) Validate() error {
+	if err := sdk.ValidateDenom(t.Denom); err != nil {
```
colin-axner, line 15: I don't think we want to do SDK validation on the denom. The reasoning for using `ibc/{hash}` was that other chains may use denominations which don't meet our chains validation (the validation function in SDK can be changed). I think mostly just check that the denom is not empty?

modules/apps/transfer/types/token.go
```diff
@@ -0,0 +1,61 @@
...
+		return errorsmod.Wrap(ErrInvalidDenomForTransfer, err.Error())
+	}
+
+	if len(t.Trace) == 0 {
+		return nil
+	}
+
+	trace := strings.Join(t.Trace, "/")
+	identifiers := strings.Split(trace, "/")
```
colin-axner, line 24: am I missing something? This looks redundant? I guess it is taking the "portid/channelid" pair and turning it into "portid", "channelid", might be nice to change the validate trace identifiers to not need to do this, or to add a comment explaining this logic since it looks odd

modules/apps/transfer/types/token.go
```diff
@@ -0,0 +1,61 @@
...
+// GetFullDenomPath returns the full denomination according to the ICS20 specification:
+// tracePath + "/" + baseDenom
+// If there exists no trace then the base denomination is returned.
+func (t Token) GetFullDenomPath() string {
+	trace := strings.Join(t.Trace, "/")
+	if len(trace) == 0 {
+		return t.Denom
+	}
+
+	return strings.Join(append(t.Trace, t.Denom), "/")
```
colin-axner, line 38: took me a second to parse this logic. Is string.Join an efficiency choice over: `return fmt.Sprintf("%s/%s", trace, t.Denom)`

fine as is, just curious

modules/apps/transfer/types/token.go
```diff
@@ -0,0 +1,61 @@
...
+
+	errorsmod "cosmossdk.io/errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	denominternal "github.com/cosmos/ibc-go/v8/modules/apps/transfer/internal/denom"
+)
+
+// Validate validates a token denomination and trace identifiers.
+func (t Token) Validate() error {
```
colin-axner, line 14: should we be validating the amount in this function?

modules/apps/transfer/types/packet.go
```diff
@@ -96,3 +96,98 @@ func (ftpd FungibleTokenPacketData) GetCustomPacketData(key string) interface{}
...
+
+	for _, token := range ftpd.Tokens {
+		amount, ok := sdkmath.NewIntFromString(token.Amount)
+		if !ok {
+			return errorsmod.Wrapf(ErrInvalidAmount, "unable to parse transfer amount (%s) into math.Int", token.Amount)
+		}
+
+		if !amount.IsPositive() {
+			return errorsmod.Wrapf(ErrInvalidAmount, "amount must be strictly positive: got %d", amount)
+		}
```
colin-axner, line 138: I feel this logic should be moved into `token.Validate()`

modules/apps/transfer/types/msgs.go
```diff
@@ -92,5 +99,48 @@ func (msg MsgTransfer) ValidateBasic() error {
 	if len(msg.Memo) > MaximumMemoLength {
 		return errorsmod.Wrapf(ErrInvalidMemo, "memo must not exceed %d bytes", MaximumMemoLength)
 	}
-	return ValidateIBCDenom(msg.Token.Denom)
+
+	for _, token := range msg.GetTokens() {
+		if !isValidToken(token) {
+			return errorsmod.Wrap(ibcerrors.ErrInvalidCoins, token.String())
```
colin-axner, line 105: I find this confusing. We refer to `msg.Tokens` as a token, but it is of type coin. This is because the existing field is called `Token` and not coin. What do people think about:
* rename `msg.GetTokens()` to `msg.GetCoins` (lets only refer to it as a token if it is the token type)
* remove `isValidToken` and use `coin.Validate()` (I believe it does all the same checks)

current terminology:
* coin -> sdk.Coin
* token -> transfertypes.Token
* path -> "portid/channelid/portid/channelid/denom"
* trace -> {"portid/channelid", "portid/channelid"} 
* ibc denom -> ibc/{hash} (intended for internal chain usage, not used in packet data)

The one exception would be that `MsgTransfer.Tokens` would be []sdk.Coin

It'd be nice to be consistent with naming conventions. I've already noticed myself getting tripped up a few times because I read "token" but the actual type was a `sdk.Coin`

In addition, I think we could make `ValidateIBCDenom` private

modules/apps/transfer/types/msgs_test.go
```diff
@@ -54,22 +54,27 @@ func TestMsgTransferValidation(t *testing.T) {
 		msg     *types.MsgTransfer
 		expPass bool
```
colin-axner, line 55: we should change these to `expError`. Since transfer is really important and we are changing things, it'd be nice to feel extra good about our test cases and make sure they are actually hitting the right checks

proto/ibc/applications/transfer/v1/tx.proto
damiannolan: I think it could be removed on the `token` potentially. This would need to be tested with ledger signing for compatability. If we deprecate and leave the tag, an empty `Coin` would always be displayed on ledger screen before signing (don't omit empty) but if we remove the tag and the `Coin` is empty because the new repeated field `tokens` is being used then I would imagine the empty coin is omitted and not displayed on the ledger screen (which I think is better ux).

damiannolan: Aside from above. The repeated field needs: `(amino.encoding) = "legacy_coins",`

proto/ibc/applications/transfer/v2/packet.proto
damiannolan: is there reason why we're not using something like `repeated Trace trace` where:

```proto
message Trace {
    string port_id = 1;
    string channel_id = 2;
}
```

I'm not sure if this was explored, just thought I'd mention here!

proto/ibc/applications/transfer/v1/tx.proto
damiannolan: sdk also uses `(gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"` for fields like this, maybe we should also be doing that. 

ref: https://github.com/cosmos/cosmos-sdk/blob/e8f28bf5db18b8d6b7e0d94b542ce4cf48fed9d6/proto/cosmos/bank/v1beta1/genesis.proto#L23-L26

proto/ibc/applications/transfer/v2/packet.proto
colin-axner: I don't think it was explored. I had the same curiosity as it removes string parsing based on "/" (for v2 packets forward)

damiannolan: what do you think about creating an issue for this? happy to defer til later but could be a nice improvement if we can remove string parsing

e2e/testsuite/testsuite.go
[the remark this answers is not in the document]
colin-axner: ah function signature mismatch. `DefaultChannelOpts()` returns a type whereas TransferChannelOptions returns a func

modules/apps/transfer/types/token.go
chatton: there was nothing fancy going on here, I agree with @DimitrisJim 's suggestion to simplify with the len check then keep join as is.

modules/apps/transfer/types/token.go
chatton: pretty sure this is redundant ad we can just pass `t.Trace` directly as is :thinking:

modules/apps/transfer/types/msgs.go
```diff
@@ -92,5 +99,48 @@ func (msg MsgTransfer) ValidateBasic() error {
...
+		return false
+	}
+
+	if coin.Amount.IsZero() {
+		return false
+	}
+
+	if coin.Amount.IsNegative() {
+		return false
+	}
```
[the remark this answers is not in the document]
chatton: Trying to remember now why this was needed, let's try and revert to this suggestion and see if we run into any issues.

e2e/testsuite/testsuite.go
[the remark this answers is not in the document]
colin-axner: https://github.com/cosmos/ibc-go/pull/6360

modules/apps/transfer/types/keys.go
colin-axner: https://github.com/cosmos/ibc-go/pull/6360

modules/apps/transfer/types/token.go
```diff
@@ -0,0 +1,61 @@
...
+
+	return strings.Join(append(t.Trace, t.Denom), "/")
+}
+
+// Tokens is a set of Token
+type Tokens []Token
+
+// String prints out the tokens array as a string.
+// If the array is empty, an empty string is returned
+func (tokens Tokens) String() string {
```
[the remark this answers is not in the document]
colin-axner: https://github.com/cosmos/ibc-go/pull/6360

modules/apps/transfer/types/token.go
chatton: the actual `t.Denom` field will never be an ibc denom in the format of `ibc/hash`, the sdk validation should be fine, in our code the we end up using the token denom to get the hash like this

```golang
fullDenomPath := token.GetFullDenomPath()
trace := types.ParseDenomTrace(fullDenomPath)
coin := sdk.NewCoin(trace.IBCDenom(), 100)
```

modules/apps/transfer/types/msgs.go
```diff
@@ -72,11 +73,17 @@ func (msg MsgTransfer) ValidateBasic() error {
...
 	}
-	if !msg.Token.IsValid() {
-		return errorsmod.Wrap(ibcerrors.ErrInvalidCoins, msg.Token.String())
+
+	if len(msg.Tokens) == 0 && !isValidToken(msg.Token) {
+		return errorsmod.Wrap(ErrInvalidAmount, "either token or token array must be filled")
+	}
+
+	if len(msg.Tokens) != 0 && isValidToken(msg.Token) {
+		return errorsmod.Wrap(ErrInvalidAmount, "cannot fill both token and token array")
```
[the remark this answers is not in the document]
chatton: yep I think I like that better!

proto/ibc/applications/transfer/v2/packet.proto
colin-axner: I think I'm in favour after digging through the code some more. I think only going part way to typed information about the ibc token still causes confusion.

v1 approach:
ibc token: "portid/channelid/portid/channelid/denom", amount

v2 approach:
ibc token: denom, amount, "portid/channelid", "portid, channelid"

For backwards compatibility we still need to parse from v1 and reformat as v1, which means that we have logic to go from:
[]string{"portid", "channelid", "portid", "channelid"}
to:
[]string{"portid/channelid", "portid/channelid"}
and back

I find this confusing, as it is hard to know if `[]string` is single items or multiple. I highlighted in this issue how many extra parsing steps we do https://github.com/cosmos/ibc-go/issues/6221

I think with typed trace:
if recv v1
   parse the path string into token type

if send v1
   glue token type back into path string
   
otherwise always store fully typed trace info

colin-axner: https://github.com/cosmos/ibc-go/issues/6362

modules/apps/transfer/types/token.go
colin-axner: The issue isn't `ibc/hash`, the issue is that chainA's denom validation might be different than chainB. So if chainA wants to use an emoji token for its denom, chainB shouldn't fail validation on this (that's why chainB converts the emoji to a hash)

modules/apps/transfer/types/token.go
colin-axner: unfortunately not, the []string types used are of different values. 

```
t.Trace = []string{"portid/channelid", "portid/channelid"}
identifiers = []string{"portid", "channelid", "portid", "channelid"}
```

It seems there is agreement on https://github.com/cosmos/ibc-go/issues/6362, which should remove this issue

modules/apps/transfer/types/msgs.go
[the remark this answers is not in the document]
colin-axner: I can update this

modules/apps/transfer/types/token.go
colin-axner: https://github.com/cosmos/ibc-go/pull/6364

modules/apps/transfer/types/token.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// GetFullDenomPath returns the full denomination according to the ICS20 specification:
+// tracePath + "/" + baseDenom
+// If there exists no trace then the base denomination is returned.
+func (t Token) GetFullDenomPath() string {
+	trace := strings.Join(t.Trace, "/")
+	if len(trace) == 0 {
+		return t.Denom
+	}
```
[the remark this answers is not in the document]
colin-axner: https://github.com/cosmos/ibc-go/pull/6364

modules/apps/transfer/types/packet.go
colin-axner: https://github.com/cosmos/ibc-go/pull/6364

modules/apps/transfer/types/msgs.go
colin-axner: see my proposed change in https://github.com/cosmos/ibc-go/pull/6364

modules/apps/transfer/types/msgs.go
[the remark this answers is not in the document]
colin-axner: https://github.com/cosmos/ibc-go/pull/6364

modules/apps/transfer/ibc_module.go
```diff
@@ -172,6 +174,37 @@ func (IBCModule) OnChanCloseConfirm(
...
 }
 
+// unmarshalPacketDataBytesToICS20V2 attempts to unmarshal the provided packet data bytes into a FungibleTokenPacketDataV2.
+// The version of ics20 should be provided and should be either ics20-1 or ics20-2.
+func (IBCModule) unmarshalPacketDataBytesToICS20V2(bz []byte, ics20Version string) (types.FungibleTokenPacketDataV2, error) {
+	switch ics20Version {
+	case types.V1:
+		var datav1 types.FungibleTokenPacketData
+		if err := json.Unmarshal(bz, &datav1); err != nil {
+			return types.FungibleTokenPacketDataV2{}, errorsmod.Wrapf(ibcerrors.ErrInvalidType, "cannot unmarshal ICS20-V2 transfer packet data: %s", err.Error())
```
colin-axner, line 184:
```suggestion
			return types.FungibleTokenPacketDataV2{}, errorsmod.Wrapf(ibcerrors.ErrInvalidType, "cannot unmarshal ICS20-V1 transfer packet data: %s", err.Error())
```

modules/apps/transfer/ibc_module.go
```diff
@@ -203,26 +240,26 @@ func (im IBCModule) OnRecvPacket(
 		}
 	}
 
-	eventAttributes := []sdk.Attribute{
-		sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
-		sdk.NewAttribute(sdk.AttributeKeySender, data.Sender),
+	ctx.EventManager().EmitEvent(sdk.NewEvent(
```
colin-axner, line 243: I can readjust this back to do a single event emission like before, didn't realize that was the original behaviour here

modules/apps/transfer/ibc_module.go
```diff
@@ -287,25 +332,34 @@ func (im IBCModule) OnTimeoutPacket(
...
+	ctx.EventManager().EmitEvents(sdk.Events{
 		sdk.NewEvent(
 			types.EventTypeTimeout,
-			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
-			sdk.NewAttribute(types.AttributeKeyRefundReceiver, data.Sender),
-			sdk.NewAttribute(types.AttributeKeyRefundDenom, data.Denom),
-			sdk.NewAttribute(types.AttributeKeyRefundAmount, data.Amount),
+			sdk.NewAttribute(sdk.AttributeKeySender, data.Sender),
+			sdk.NewAttribute(types.AttributeKeyReceiver, data.Receiver),
+			sdk.NewAttribute(types.AttributeKeyTokens, types.Tokens(data.Tokens).String()),
```
colin-axner, line 355: was it intentional to change event keys here? (for refund receiver)

colin-axner (commented):
Almost there, I got 10 files left!

modules/apps/transfer/ibc_module.go
```diff
@@ -121,13 +124,11 @@ func (im IBCModule) OnChanOpenTry(
 		return "", err
 	}
 
-	if counterpartyVersion != types.Version {
-		// Propose the current version
-		im.keeper.Logger(ctx).Debug("invalid counterparty version, proposing current app version", "counterpartyVersion", counterpartyVersion, "version", types.Version)
-		return types.Version, nil
+	if !slices.Contains(types.SupportedVersions, counterpartyVersion) {
+		return types.V2, nil
```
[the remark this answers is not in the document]
colin-axner: whatever is fine by me :+1:

modules/apps/transfer/types/token.go
```diff
@@ -0,0 +1,61 @@
...
+	denominternal "github.com/cosmos/ibc-go/v8/modules/apps/transfer/internal/denom"
+)
+
+// Validate validates a token denomination and trace identifiers.
+func (t Token) Validate() error {
+	if err := sdk.ValidateDenom(t.Denom); err != nil {
+		return errorsmod.Wrap(ErrInvalidDenomForTransfer, err.Error())
+	}
+
+	if len(t.Trace) == 0 {
```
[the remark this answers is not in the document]
colin-axner: I can do this

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -37,59 +41,72 @@ func (TransferAuthorization) MsgTypeURL() string {
...
 		// If the spend limit is set to the MaxUint256 sentinel value, do not subtract the amount from the spend limit.
-		if allocation.SpendLimit.AmountOf(msgTransfer.Token.Denom).Equal(UnboundedSpendLimit()) {
-			return authz.AcceptResponse{Accept: true, Delete: false, Updated: nil}, nil
+		// if there is no unlimited spend, then we need to subtract the amount from the spend limit to get the limit left
+		if allocation.SpendLimit.AmountOf(coin.Denom).Equal(UnboundedSpendLimit()) {
+			continue
 		}
 
-		limitLeft, isNegative := allocation.SpendLimit.SafeSub(msgTransfer.Token)
+		limitLeft, isNegative := a.Allocations[index].SpendLimit.SafeSub(coin)
```
colin-axner, line 78:
```suggestion
		limitLeft, isNegative := allocation.SpendLimit.SafeSub(coin)
```
take it or leave it, just felt odd relative to line 74

modules/apps/transfer/keeper/msg_server_test.go
```diff
@@ -106,27 +177,48 @@ func (suite *KeeperTestSuite) TestMsgTransfer() {
...
-					sdk.NewAttribute(types.AttributeKeyAmount, coin.Amount.String()),
-					sdk.NewAttribute(types.AttributeKeyDenom, coin.Denom),
-					sdk.NewAttribute(types.AttributeKeyMemo, "memo"),
-				),
-			}.ToABCIEvents()
+			events := ctx.EventManager().Events().ToABCIEvents()
 
-			if tc.expPass {
+			var expEvents []abci.Event
+			if tc.multiDenom {
```
colin-axner, line 183: we could probably simplify by just expecting the event attribute to be msg.Tokens

modules/apps/transfer/internal/convert/convert.go
```diff
@@ -0,0 +1,54 @@
...
+				Trace:  trace,
+			},
+		},
+		Sender:   packetData.Sender,
+		Receiver: packetData.Receiver,
+		Memo:     packetData.Memo,
+	}
+}
+
+// extractDenomAndTraceFromV1Denom extracts the base denom and remaining trace from a v1 IBC denom.
```
colin-axner, line 30:
```suggestion
// ExtractDenomAndTraceFromV1Denom extracts the base denom and remaining trace from a v1 IBC denom.
```
looks like it wasn't made private?

modules/apps/transfer/internal/convert/convert.go
```diff
@@ -0,0 +1,54 @@
...
+func ExtractDenomAndTraceFromV1Denom(v1Denom string) (string, []string) {
```
colin-axner, line 31: this function appears to assume the denom is validated, should we add this in the godoc?

Otherwise the parsing logic doesn't account for edge cases like the other [function](https://github.com/cosmos/ibc-go/blob/feat/ics20-v2/modules/apps/transfer/types/trace.go#L104) does

colin-axner: I think it's fine for now, lets revisit after #6362 and #6221 which I think will cleanup all the parsing logic. After that issue we should only parse from full path string to typed trace and then have a function to glue them together again

modules/apps/transfer/keeper/relay.go
```diff
@@ -326,48 +344,52 @@ func (k Keeper) OnAcknowledgementPacket(ctx sdk.Context, packet channeltypes.Pac
...
-	sender, err := sdk.AccAddressFromBech32(data.Sender)
-	if err != nil {
-		return err
-	}
+		// parse the transfer amount
+		transferAmount, ok := sdkmath.NewIntFromString(token.Amount)
+		if !ok {
+			return errorsmod.Wrapf(types.ErrInvalidAmount, "unable to parse transfer amount (%s) into math.Int", transferAmount)
+		}
+		token := sdk.NewCoin(trace.IBCDenom(), transferAmount)
```
colin-axner, line 369: maybe lets rename to `coin` given that there's another variable here also named `token` with a different type and denom value :sweat_smile:

colin-axner (approved):
Alrighty! Wahoo! I approve! Finally made it to the end :rocket: 

Many kudos and excellent work by everyone who contributed to this feature! :clap: :clap: :clap: 

The feature is clean and gives transfer that much needed face lift. I can feel it sets a nice tone to remove some unnecessary complexity under the hood. 

There are a few requested changes we should do **after merge to main**:
- #6221 (I think this will help clean up a lot of logic)
- #6362 
- sending packet data based on channel version (necessary for correctness)

modules/apps/transfer/internal/convert/convert.go
```diff
@@ -0,0 +1,54 @@
...
+	// if the path string is empty, then the base denom is the full native denom.
+	if v1DenomTrace.Path == "" {
+		return v1DenomTrace.BaseDenom, nil
+	}
+
+	splitPath := strings.Split(v1DenomTrace.Path, "/")
+
+	// this condition should never be reached.
+	if len(splitPath)%2 != 0 {
+		panic("pathSlice length is not even")
```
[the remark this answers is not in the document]
chatton: yes! I had another panic cleanup related concern, will create an issue to clean these up.

Issue #6379, opened 2024-05-24
Consistent terminology for coin/token in transfer.


DimitrisJim:
Opening this after https://github.com/cosmos/ibc-go/pull/6364 and https://github.com/cosmos/ibc-go/pull/6352#discussion_r1611958976

Would be nice to use consistent terminology for coins and the newly added `Token` type so as to not cause confusion when reviewing/changing things. There should be some places where these are used interchangeably (as the review comment pointed out). Some internal private functions could also be renamed to better align with these types.

Can probably be left for until v2 of ics20 is merged.
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Can we already consider this done?

colin-axner:
thanks, was gonna get to this :+1:

Issue #6362, opened 2024-05-23
Use typed Trace in ics20 v2 Token type


colin-axner:
## Summary

see discussion https://github.com/cosmos/ibc-go/pull/6352#discussion_r1611259574

by using a typed trace field, we limit the amount of string parsing to only when we need to convert to and from the v1 packet data

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #6395, opened 2024-05-27
refactor: use typed Trace, remove DenomTrace

colin-axner:
## Description

Initial proposed change. Overview of changes:
* remove `DenomTrace` and associated functions
* replace `DenomTrace` usage with `Token` (with zero amount)
* move convert funcs to types package (to avoid circular dep cycle)
* rename storage usage of `DenomTrace` to `TokenInfo`
* simplify main handler, remove unnecessary string parsing and joining
* remove unnecessary string parsing functions

Notes: 

As structured, the `DenomTrace` is difficult to use and renders the v2 improvements almost useless from an internal point of view. Initially I tried creating a V2 version of `DenomTrace`, but it is the same content that is in the token. Our code would generate both types, which made me feel like we should just be using the token info. I saw the cleanest options as either:
* use the existing Token type and store with zero amount
* generate a new type which is the denom + trace and use that in the token

for ease, I went with the first option, but open to changing. I named the funcs `Get/SetTokenInfo`, very open to alternative names. Can also just be `Get/SetToken`, but might feel misleading if the amount is always 0. 

By removing the DenomTrace type, that requires changing the genesis file and query endpoints. This breaks proto versioning. 

**What do we think about moving genesis proto to v2 and breaking query api?**

closes: #6362
closes: #6221
closes: #6379

---

write a little note why.


modules/apps/transfer/internal/convert/convert.go
```diff
@@ -1,54 +1 @@
 package convert
-
```
colin-axner, line 2: moved because of import cycle. Ideally we only parse v1 denom when handling v1 packets (and there is only 1 string parsing function). The issue is that v1 packet data needs to parse v1 to do basic validation, so if you move v1 denom parsing here, you gain an import cycle

modules/apps/transfer/keeper/encoding.go
```diff
@@ -1,36 +0,0 @@
-package keeper
-
-import (
-	"github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
-)
-
-// UnmarshalDenomTrace attempts to decode and return an DenomTrace object from
-// raw encoded bytes.
-func (k Keeper) UnmarshalDenomTrace(bz []byte) (types.DenomTrace, error) {
```
colin-axner, line 9: removed file as I felt this was an unnecessary layer of wrapping

modules/apps/transfer/keeper/genesis.go
```diff
@@ -12,10 +12,12 @@ import (
 func (k Keeper) InitGenesis(ctx sdk.Context, state types.GenesisState) {
 	k.SetPort(ctx, state.PortId)
 
-	for _, trace := range state.DenomTraces {
-		k.SetDenomTrace(ctx, trace)
-		k.setDenomMetadata(ctx, trace)
-	}
+	/*
```
colin-axner, line 15: will fixup later if we are in agreement of the approach proposed

modules/apps/transfer/keeper/grpc_query.go
```diff
@@ -21,6 +17,7 @@ import (
 
 var _ types.QueryServer = (*Keeper)(nil)
 
+/*
 // DenomTrace implements the Query/DenomTrace gRPC method
```
colin-axner, line 21: ditto for waiting on agreed approach

modules/apps/transfer/keeper/relay.go
```diff
@@ -87,30 +86,18 @@ func (k Keeper) sendTransfer(
 	var tokens []types.Token
 
 	for _, coin := range coins {
-		// NOTE: denomination and hex hash correctness checked during msg.ValidateBasic
```
colin-axner, line 90: I felt this file had really nice simplification

modules/apps/transfer/keeper/relay.go
```diff
@@ -214,30 +193,24 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
...
+		if token.ReceiverChainIsSource(packet.GetSourcePort(), packet.GetSourceChannel()) {
 			// sender chain is not the source, unescrow tokens
 
 			// remove prefix added by sender chain
-			voucherPrefix := types.GetDenomPrefix(packet.GetSourcePort(), packet.GetSourceChannel())
-			unprefixedDenom := fullDenomPath[len(voucherPrefix):]
-
-			// coin denomination used in sending from the escrow address
-			denom := unprefixedDenom
+			token.Trace = token.Trace[1:]
```
colin-axner, line 200: way nicer way to add/remove the token prefix

modules/apps/transfer/keeper/relay.go
```diff
@@ -214,30 +193,24 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
...
 
-			// The denomination used to send the coins is either the native denom or the hash of the path
-			// if the denomination is not native.
-			denomTrace := types.ParseDenomTrace(unprefixedDenom)
-			if !denomTrace.IsNativeDenom() {
-				denom = denomTrace.IBCDenom()
+			// obtain the sdk coin for bank sends
+			coin, err := token.Coin()
```
colin-axner, line 203: I think I'll likely remove this function since the amount is parsed above in this file

modules/apps/transfer/keeper/relay.go
```diff
@@ -266,39 +239,36 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 
 		// since SendPacket did not prefix the denomination, we must prefix denomination here
 		prefixedDenom := types.GetPrefixedDenom(packet.GetDestPort(), packet.GetDestChannel(), fullDenomPath)
+		token.Trace = append([]types.Trace{types.NewTrace(packet.GetDestPort(), packet.GetDestChannel())}, token.Trace...)
```
colin-axner, line 242: will cleanup so it reads nicer

modules/apps/transfer/types/trace.go
```diff
@@ -14,142 +13,34 @@ import (
...
-// into a DenomTrace type.
-//
-// Examples:
-//
-// - "portidone/channel-0/uatom" => DenomTrace{Path: "portidone/channel-0", BaseDenom: "uatom"}
-// - "portidone/channel-0/portidtwo/channel-1/uatom" => DenomTrace{Path: "portidone/channel-0/portidtwo/channel-1", BaseDenom: "uatom"}
-// - "portidone/channel-0/gamm/pool/1" => DenomTrace{Path: "portidone/channel-0", BaseDenom: "gamm/pool/1"}
-// - "gamm/pool/1" => DenomTrace{Path: "", BaseDenom: "gamm/pool/1"}
-// - "uatom" => DenomTrace{Path: "", BaseDenom: "uatom"}
-func ParseDenomTrace(rawDenom string) DenomTrace {
```
colin-axner, line 31: removed :tada:

modules/apps/transfer/types/trace.go
```diff
@@ -14,142 +13,34 @@ import (
...
-}
-
-// IsNativeDenom returns true if the denomination is native, thus containing no trace history.
-func (dt DenomTrace) IsNativeDenom() bool {
-	return dt.Path == ""
-}
-
-// extractPathAndBaseFromFullDenom returns the trace path and the base denom from
-// the elements that constitute the complete denom.
-func extractPathAndBaseFromFullDenom(fullDenomItems []string) ([]string, string) {
```
colin-axner, line 87: in v1.go

modules/apps/transfer/types/trace.go
```diff
@@ -14,142 +13,34 @@ import (
...
+		return errorsmod.Wrapf(err, "invalid portID")
 	}
-
-	baseDenom := strings.Join(baseDenomSlice, "/")
-
-	return pathSlice, baseDenom
-}
-
-// validateTraceIdentifiers validates the correctness of the trace associated with a particular base denom.
-func validateTraceIdentifiers(identifiers []string) error {
```
colin-axner, line 118: can be removed after migrations

modules/apps/transfer/types/trace.go
```diff
@@ -14,142 +13,34 @@ import (
...
-		}
-		if err := host.ChannelIdentifierValidator(identifiers[i+1]); err != nil {
-			return errorsmod.Wrapf(err, "invalid channel ID at position %d", i)
-		}
-	}
 	return nil
 }
 
-// Validate performs a basic validation of the DenomTrace fields.
-func (dt DenomTrace) Validate() error {
```
colin-axner, line 136: can be removed after migrations

modules/apps/transfer/types/trace.go
```diff
@@ -185,35 +76,6 @@ func (t Traces) Sort() Traces {
...
 }
 
-// ValidatePrefixedDenom checks that the denomination for an IBC fungible token packet denom is correctly prefixed.
-// The function will return no error if the given string follows one of the two formats:
-//
-//   - Prefixed denomination: '{portIDN}/{channelIDN}/.../{portID0}/{channelID0}/baseDenom'
-//   - Unprefixed denomination: 'baseDenom'
-//
-// 'baseDenom' may or may not contain '/'s
-func ValidatePrefixedDenom(denom string) error {
```
colin-axner, line 195: moved to v1.go and greatly simplified

crodriguezvega (commented):
I did a quick scan through the changes and they are looking great.

I don't feel strongly about reusing the existing `Token` type or creating a new one with just denom and trace. Choosing the easiest that will cause the least confusion moving forward sounds good to me.

I just wanted, though, to share some extra renaming ideas to gather early feedback. In general, every time I have to do something with transfer code I am always confused about what is base denom, what is full denom with trace information, what is denom with hash, etc 😅, so maybe taking the opportunity to clean that up now would also be nice (could also be done later in a separate PR to reduce the diffs). Here's the proposal:

For the `Token` type:
- Rename `IsNativeDenom` to just `IsNative`.
- Rename `IBCDenom` function to `OnChainDenom`, since this the actual on-chain denomination of the tokens.
- Rename `GetFullDenomPath` to `GetDenom` or simply `Denom` (I think this would be nice because in `FungibleTokenPacketData` v1 the `Denom` field is the trace + base denom, so then the terminology would be more consistent). We also use the term `PrefixedDenom` in some places (mostly tests actually) for trace + base denom, so it could also be an option to rename the function to `GetPrefixedDenom`.
- Rename `Denom` field to `BaseDenom` or `NativeDenom`.

What do y'all think?

Issue #6413, opened 2024-05-29
Condense full path string parsing to single location/function


colin-axner:
## Summary

Currently we have multiple entries for parsing a v1 transfer packet. We should have a single function for this purpose

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #6424, opened 2024-05-29
refactor: condense v1 denom parsing to single function

colin-axner:
## Description

closes: #6413 

---

write a little note why.

## Summary by CodeRabbit

- **Refactor**
  - Simplified the extraction of denominations and traces in various functions by using a new, dedicated method.
  - Improved code clarity and modularity by delegating extraction logic to a separate function.

- **Bug Fixes**
  - Enhanced the validation process for prefixed denominations, ensuring more accurate and reliable handling of token denominations.

### DIFF
--- a/modules/apps/transfer/types/trace.go
+++ b/modules/apps/transfer/types/trace.go
@@ -95,7 +95,7 @@ func ExtractDenomFromFullPath(fullPath string) Denom {
 		// as an IBC denomination. The hash used to store the token internally on our chain
 		// will be the same value as the base denomination being correctly parsed.
 		if i < length-1 && length > 2 && channeltypes.IsValidChannelID(denomSplit[i+1]) {
-			trace = append(trace, denomSplit[i], denomSplit[i+1])
+			trace = append(trace, denomSplit[i]+"/"+denomSplit[i+1])
 		} else {
 			baseDenomSlice = denomSplit[i:]
 			break
--- a/modules/apps/transfer/types/trace.go
+++ b/modules/apps/transfer/types/trace.go
@@ -28,19 +28,10 @@ import (
 // - "gamm/pool/1" => DenomTrace{Path: "", BaseDenom: "gamm/pool/1"}
 // - "uatom" => DenomTrace{Path: "", BaseDenom: "uatom"}
 func ParseDenomTrace(rawDenom string) DenomTrace {
-	denomSplit := strings.Split(rawDenom, "/")
-
-	if denomSplit[0] == rawDenom {
-		return DenomTrace{
-			Path:      "",
-			BaseDenom: rawDenom,
-		}
-	}
-
-	pathSlice, baseDenom := extractPathAndBaseFromFullDenom(denomSplit)
+	denom := ExtractDenomFromFullPath(rawDenom)
 	return DenomTrace{
-		Path:      strings.Join(pathSlice, "/"),
-		BaseDenom: baseDenom,
+		Path:      strings.Join(denom.Trace, "/"),
+		BaseDenom: denom.Base,
 	}
 }
 
@@ -76,15 +67,23 @@ func (dt DenomTrace) GetFullDenomPath() string {
 	return dt.GetPrefix() + dt.BaseDenom
 }
 
-// extractPathAndBaseFromFullDenom returns the trace path and the base denom from
-// the elements that constitute the complete denom.
-func extractPathAndBaseFromFullDenom(fullDenomItems []string) ([]string, string) {
+// ExtractDenomFromFullPath returns the denom from the full path.
+// Used to support v1 denoms.
+func ExtractDenomFromFullPath(fullPath string) Denom {
+	denomSplit := strings.Split(fullPath, "/")
+
+	if denomSplit[0] == fullPath {
+		return Denom{
+			Base: fullPath,
+		}
+	}
+
 	var (
-		pathSlice      []string
+		trace          []string
 		baseDenomSlice []string
 	)
 
-	length := len(fullDenomItems)
+	length := len(denomSplit)
 	for i := 0; i < length; i += 2 {
 		// The IBC specification does not guarantee the expected format of the
 		// destination port or destination channel identifier. A short term solution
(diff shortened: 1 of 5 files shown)


modules/apps/transfer/types/trace.go
```diff
@@ -76,15 +67,23 @@ func (dt DenomTrace) GetFullDenomPath() string {
 	return dt.GetPrefix() + dt.BaseDenom
 }
 
-// extractPathAndBaseFromFullDenom returns the trace path and the base denom from
-// the elements that constitute the complete denom.
-func extractPathAndBaseFromFullDenom(fullDenomItems []string) ([]string, string) {
+// ExtractDenomFromFullPath returns the denom from the full path.
+// Used to support v1 denoms.
+func ExtractDenomFromFullPath(fullPath string) Denom {
```
colin-axner, line 72: this should be the only function we need to go from: path -> Denom

modules/apps/transfer/types/trace.go
```diff
@@ -95,17 +94,20 @@ func extractPathAndBaseFromFullDenom(fullDenomItems []string) ([]string, string)
 		// will be incorrectly parsed, but the token will continue to be treated correctly
 		// as an IBC denomination. The hash used to store the token internally on our chain
 		// will be the same value as the base denomination being correctly parsed.
-		if i < length-1 && length > 2 && channeltypes.IsValidChannelID(fullDenomItems[i+1]) {
-			pathSlice = append(pathSlice, fullDenomItems[i], fullDenomItems[i+1])
+		if i < length-1 && length > 2 && channeltypes.IsValidChannelID(denomSplit[i+1]) {
+			trace = append(trace, denomSplit[i]+"/"+denomSplit[i+1])
```
colin-axner, line 98: will be typed soon enough

modules/apps/transfer/types/trace.go
```diff
@@ -149,26 +151,20 @@ func (dt DenomTrace) Validate() error {
...
 	}
 
-	path := strings.Join(pathSlice, "/")
-	if path == "" {
-		// NOTE: base denom contains slashes, so no base denomination validation
-		return nil
+	path := strings.Join(denom.Trace, "/")
+	if len(denom.Trace) != 0 {
+		identifiers := strings.Split(path, "/")
+		return validateTraceIdentifiers(identifiers)
```
colin-axner, line 163: soon `denom.Trace.Validate()`

modules/apps/transfer/types/trace.go
crodriguezvega: Maybe `FullPathToDenom` ([ref](https://github.com/cosmos/ibc-go/pull/6426#discussion_r1619414270)) or if we move this function to `denom.go` then maybe `FromFullPath` or `NewDenom`.

modules/apps/transfer/types/trace.go
```diff
@@ -149,26 +151,20 @@ func (dt DenomTrace) Validate() error {
...
-func ValidatePrefixedDenom(denom string) error {
-	denomSplit := strings.Split(denom, "/")
-	if denomSplit[0] == denom && strings.TrimSpace(denom) != "" {
-		// NOTE: no base denomination validation
-		return nil
-	}
-
-	pathSlice, baseDenom := extractPathAndBaseFromFullDenom(denomSplit)
-	if strings.TrimSpace(baseDenom) == "" {
+func ValidatePrefixedDenom(fullPath string) error {
```
crodriguezvega, line 154: I know this has already been like this for a while, but we also use `PrefixedDenom` and `FullPath` for trace + base denom. It would also be nice to converge now into a single term. I don't particularly like either, but what about `IBCDenom` (if we use on-chain denom for `ibc/<hash>`) or `ICS20Denom`?

crodriguezvega (approved):

modules/apps/transfer/types/trace.go
colin-axner: going to leave as is unless there's consensus otherwise

modules/apps/transfer/types/trace.go
colin-axner: I think it's good to open an issue and take discussion there. I'm happy to see the function renamed

chatton (changes requested):
Everything looks good except the edge case I mentioned in the review, would be nice to make sure we have a test case covering this.

modules/apps/transfer/types/trace.go
```diff
@@ -76,15 +67,23 @@ func (dt DenomTrace) GetFullDenomPath() string {
...
+// ExtractDenomFromFullPath returns the denom from the full path.
+// Used to support v1 denoms.
+func ExtractDenomFromFullPath(fullPath string) Denom {
+	denomSplit := strings.Split(fullPath, "/")
+
+	if denomSplit[0] == fullPath {
+		return Denom{
+			Base: fullPath,
+		}
+	}
```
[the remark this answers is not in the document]
colin-axner: https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/trace_test.go#L19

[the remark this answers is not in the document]
colin-axner: This case is caught by the `baseDenomSlice` construction https://github.com/cosmos/ibc-go/pull/6424/files#diff-ff4125d8c59b9ab8785f0ae9e950dc2a59a79428482e3f36b7b1695f57d5e39fR105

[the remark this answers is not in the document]
chatton: thinking now, it's possible that the correct behaviour ends up happening later in the function here

This actually ends up being handled correctly here `baseDenomSlice = denomSplit[i:]`

I feel like we should probably still explicitly check this at the start of the fn `if denomSplit[0] == fullPath` to be more explicit. I think this edge case could slip through if the for loop code gets changed in the future. We should at least add a test like

```golang
func TestFoo(t *testing.T) {
	res := types.ExtractDenomFromFullPath("atom/")
	require.Equal(t, "atom/", res.Base)
	require.Empty(t, res.Trace)
}
```

or similar WDYT?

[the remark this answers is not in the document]
colin-axner: When we remove `ParseDenomTrace` we can move the test cases over to directly test the extraction function (I believe it didn't have unit tests since it was a private function and written before we made use of `export_test.go`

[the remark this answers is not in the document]
chatton: ah yeah I hadn't thought of additional slashes inside, I guess it's not possible to do a check at the start, maybe we can add some additional tests for this function. Happy to do this in a follow up issue. It makes me a bit uneasy to have this happening in that for loop but I think there might be no alternative :thinking:

chatton (approved):
Leaving approval since the edge case I was concerned about is actually covered. We can add additional test coverage for this in a follow up :+1:

modules/apps/transfer/types/trace.go
colin-axner: this function will be deleted in a future pr

PR #6444, opened 2024-06-02
fix(testing/simapp): append to slice with non-zero initialized length

BabyHalimao:
## Description
This PR fixes the unexpected behavior that append to slice `keys` with non-zero initialized length

See example:
https://go.dev/play/p/GWGG_cnd9vV

## Summary by CodeRabbit

- **Refactor**
  - Optimized memory allocation in the `GetStoreKeys` method across multiple modules to improve performance.

- **Tests**
  - Renamed `TestRandomizedGenState` to `TestRandomizedGenState1` for better clarity.


testing/simapp/app.go
```diff
@@ -934,7 +934,7 @@ func (app *SimApp) GetKey(storeKey string) *storetypes.KVStoreKey {
 
 // GetStoreKeys returns all the stored store keys.
 func (app *SimApp) GetStoreKeys() []storetypes.StoreKey {
-	keys := make([]storetypes.StoreKey, len(app.keys))
+	keys := make([]storetypes.StoreKey, 0, len(app.keys))
```
crodriguezvega, line 937: @coderabbitai I suspect there might be other occurrences in the codebase where slices are initialised with a non-zero length and then we append items to it. Can you check please?

crodriguezvega: If both @DimitrisJim with his grep-fu skills and @coderabbitait with its AI powers agree on the answer, then I think we are good. :)

crodriguezvega (approved):
Thanks, @BabyHalimao

testing/simapp/app.go
crodriguezvega: uh, wait a minute, @DimitrisJim and @coderabbitai: check out [this commit](https://github.com/cosmos/ibc-go/pull/6444/commits/ae3f860db663f63701599bb382bce6185101286e), isn't these a couple of places where we should set zero length also?

Issue #6558, opened 2024-06-10
Simplify reverts of forwarding state on acknowledgement


DimitrisJim:
Investigate if we can simplify revert logic by doing regular refund packet tokens first and then the logic for revert forwarding changes.

Surfaced during internal walkthrough of ics20 v2 forwarding (commit [efcfa5d](https://github.com/cosmos/ibc-go/tree/efcfa5d877567731f6b4ff8dbd063a34ede5de00))
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

crodriguezvega:
Looks like it should be possible. Opened https://github.com/cosmos/ibc/pull/1110 to update the spec.

srdtrk:
I can take this if others agree

srdtrk:
closed by #6574

PR #6574, opened 2024-06-12
Fix and simplify reverts of forwarding state

srdtrk:
## Description

closes: #6558

---

write a little note why.


modules/apps/transfer/keeper/relay.go
```diff
@@ -320,67 +319,47 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
...
+	case *channeltypes.Acknowledgement_Error:
+		// We refund the tokens from the escrow address to the sender
+		if err := k.refundPacketTokens(ctx, packet, data); err != nil {
+			return err
+		}
+		if !isForwarded {
+			return nil
+		}
+
+		return k.onForwardedPacketErrorAck(ctx, prevPacket, data)
```
srdtrk, line 342: I can see a case for reverting this logic everywhere so I'd like to know what others think, i.e.
```suggestion
		if isForwarded {
			return k.onForwardedPacketErrorAck(ctx, prevPacket, data)
		}

		return nil
```

gjermundgaraba: This is more readable, for sure. I prefer this too.

I started commenting the same on all the places where it is like this.
To not create too much noise I will just write it here: 
I think all the places with `if !isForwarded` would read and flow better with `if isForwarded` :)

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+
+// writes acknowledgement for a forwarded packet asyncronously
+func (k Keeper) acknowledgeForwardedPacket(ctx sdk.Context, packet channeltypes.Packet, ack channeltypes.Acknowledgement) error {
+	capability, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(packet.DestinationPort, packet.DestinationChannel))
+	if !ok {
+		return errorsmod.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
+	}
+
+	return k.ics4Wrapper.WriteAcknowledgement(ctx, capability, packet, ack)
+}
```
crodriguezvega, line 53: I am wondering if we can structure the code in a different way to have fewer helper functions, since some of them are pretty thin and only called once, so not sure if they really add a lot of value... I was thinking of having one helper function like this (naming, as always is up for debate):

```go
func (k Keeper) acknowledgeForwardedErrorAck(ctx sdk.Context, prevPacket channeltypes.Packet, nextPacketData types.FungibleTokenPacketDataV2, prevPacketAck channeltypes.Acknowledgement_Error) error {
  // the forwarded packet has failed, thus the funds have been refunded to the intermediate address.
  // we must revert the changes that came from successfully receiving the tokens on our chain
  // before propogating the error acknowledgement back to original sender chain
  if err := k.revertInFlightChanges(ctx, prevPacket, nextPacketData); err != nil {
    return err
  }

  capability, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(prevPacket.DestinationPort, prevPacket.DestinationChannel))
  if !ok {
    return errorsmod.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
  }

  return k.ics4Wrapper.WriteAcknowledgement(ctx, capability, prevPacket, prevPacketAck)
}
```

And then wherever is called, either on error ack or timeout, we do something like this (adjusting the message in the ack accordingly):

```go
forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet failed"))
return acknowledgeForwardedErrorAck(ctx, prevPacket, data, forwardAck)
```

What do people think?

modules/apps/transfer/keeper/relay.go
crodriguezvega: Yes, I think I personally prefer it like this, with the `return nil` last...

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+
+// revertInFlightChanges reverts the logic of receive packet that occurs in the middle chains during a packet forwarding.
+// If an error occurs further down the line, the state changes on this chain must be reverted before sending back the error
+// acknowledgement to ensure atomic packet forwarding.
+func (k Keeper) revertInFlightChanges(ctx sdk.Context, prevPacket channeltypes.Packet, failedPacketData types.FungibleTokenPacketDataV2) error {
```
gjermundgaraba, line 58: The more times I read this name, the more I keep thinking that the name is a little off. When I read "inFlight" I imagine we are reverting a change that has been done during the context of our current call/context. Maybe we can find a name that makes it clearer that this actually reverts packets that occours in a middle chain of packet forwarding? 
A couple of suggestions, but I am not sure :D
- `revertForwardedPacket`
- `revertForwarding`
- `rollbackMiddleChain` (sounds off, but I wish I could grasp "reverts the logic of receive packet that occurs in the middle chains during a packet forwarding" from the name :/

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
+package keeper
```
gjermundgaraba, line 1: I love moving this to a separate file, btw.

Would naming this to `packet_forwarding` make it even easier to quickly grasp what it does? Maybe even just `forwarding` is also clearer - but I personally like packet_forwarding for clarity ^^

crodriguezvega (commented):
Nice work with this refactoring, @srdtrk! I left a few suggestions and one thing that I think would be good to is on naming of the packets that are received and sent in the middle hop. I think at the moment we use both `prevPacket` and `forwardedPacket` for the packet that is received in the middle hop, but sometimes I think we also use forwarded packet for the packet that is sent? I think we are also not very consistent with that naming in the spec either, so would be good to agree on the terminology... What about simply `prevPacket`or `recvPacket` for the packet received and `nextPacket` or `sentPacket` for the packet sent?

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+			2. Burning voucher tokens if the funds are foreign
+	*/
+
+	intermediateSenderAddr := types.GetForwardAddress(prevPacket.DestinationPort, prevPacket.DestinationChannel)
+	escrow := types.GetEscrowAddress(prevPacket.DestinationPort, prevPacket.DestinationChannel)
+
+	// we can iterate over the received tokens of prevPacket by iterating over the sent tokens of failedPacketData
+	for _, token := range failedPacketData.Tokens {
+		// parse the transfer amount
+		transferAmount, ok := sdkmath.NewIntFromString(token.Amount)
```
gjermundgaraba, line 74: I noticed that we do this conversion from token to coin (L74-L78) a few places. Would it be an idea to make that into a convenience method on Token? If so, we could add a separate task for that.

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+		if !ok {
+			return errorsmod.Wrapf(types.ErrInvalidAmount, "unable to parse transfer amount (%s) into math.Int", transferAmount)
+		}
+		coin := sdk.NewCoin(token.Denom.IBCDenom(), transferAmount)
+
+		// check if the packet we received was a native token
```
gjermundgaraba, line 80: Bit of a personal nit for me on comments. It feels like we should be able to make the code say this. What if we renamed the `token` to `receivedToken`? Then the next line would read:
`if receivedToken.Denom.IsNative()`

wdyt?

modules/apps/transfer/keeper/relay_forwarding_test.go
```diff
@@ -572,13 +572,10 @@ func (suite *KeeperTestSuite) TestAcknowledgementFailureScenario5Forwarding() {
 	err = suite.chainB.GetSimApp().TransferKeeper.OnAcknowledgementPacket(suite.chainB.GetContext(), packetRecv, data, ack)
 	suite.Require().NoError(err)
 
-	// Check that Escrow B has been refunded amount
-	// NOTE This is failing. The revertInFlightsChanges sohuld mint back voucher to chainBescrow
-	// but this is not happening. It may be a problem related with how we're writing async acks.
-	//
+	// Check that Escrow B is empty the revertInFlightsChanges should burn the entire token since it is not native to Chain B.
```
sangier, line 575: Mmm, I think that at this point amount should have been refunded to Escrow B.

We start the test with 2 simple transfers.

A -> B
B -> C
At the end of this two transfer we verify at lines 460-470 that states are:

EscrowB = Amount (denom AB)
UserC = Amount (denom ABC)
The states we see here are the ones we should expect to see at the end of the test.
Indeed then we start the forwarding transfer C -> B -> A and we force this transfer to fail.
So we should expect that all the states are reverted back to the point before the forwarding transfer happen. Which means we should go back just right after the simple transfer B->C has finalised. So final states should be:

EscrowB = Amount (denom AB)
UserC = Amount (denom ABC)

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+		if token.Denom.IsNative() {
+			// then send it back to the escrow address
+			if err := k.escrowCoin(ctx, intermediateSenderAddr, escrow, coin); err != nil {
+				return err
+			}
+
+			continue
+		}
+
+		// otherwise burn it
```
gjermundgaraba, line 90: As much as I like early returns and avoiding if else and nesting.
Here we basically just write out the else in a comment. I wonder if it read just as nice with if else here:
if token.Denom.IsNative() 
  escrow
else 
  burn
  
I don't feel very strongly about this, but would love to hear opinions on this.

modules/apps/transfer/keeper/relay.go
```diff
@@ -320,67 +319,47 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
...
-	if found {
-		channelCap, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(packet.SourcePort, packet.SourceChannel))
-		if !ok {
-			return errorsmod.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
-		}
-
-		if err := k.revertInFlightChanges(ctx, packet, prevPacket, data); err != nil {
-			return err
-		}
+	if err := k.refundPacketTokens(ctx, packet, data); err != nil {
```
gjermundgaraba, line 352: Is this correct? Maybe I am not reading it correctly, but it looks like the logic here got changed to always calling refundPacketTokens, which it didnt before?

modules/apps/transfer/keeper/forward.go
chatton, line 58: Maybe we can rename this to something like `revertForwardingInFlightChanges` just so we're explicit about it's purpose.  WDYT?

gjermundgaraba:
Just wanted to also say that this change is great! Moving it to a separate file and making the logic much more readable

modules/apps/transfer/keeper/forward.go
crodriguezvega: I agree, maybe we could rename `refundPacketTokens` to `refundSentTokens` and this `revertInFlightChanges` could be named `refundReceivedTokens` (although in the case where tokens are burnt is not really a refund...). Maybe instead of `refund...` we can use `revert...`?

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+		if !ok {
+			return errorsmod.Wrapf(types.ErrInvalidAmount, "unable to parse transfer amount (%s) into math.Int", transferAmount)
+		}
+		coin := sdk.NewCoin(token.Denom.IBCDenom(), transferAmount)
+
+		// check if the packet we received was a native token
+		if token.Denom.IsNative() {
+			// then send it back to the escrow address
+			if err := k.escrowCoin(ctx, intermediateSenderAddr, escrow, coin); err != nil {
+				return err
```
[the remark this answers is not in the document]
chatton: we don't really do any additional wrapping with errors returned from `refundPacketTokens` so I don't think it's strictly necessary to add an additional error type here

modules/apps/transfer/keeper/forward.go
chatton: I think I prefer this suggestion to reduce the number of small helper fns

modules/apps/transfer/keeper/forward.go
chatton: +1 for `revertForwardedPacket`

modules/apps/transfer/keeper/forward.go
chatton: I don't actually think the comment provides any value, I think we can either remove the comment and leave the code as is, or implement the else suggestion and remove the continue.

I think I'd be happy just removing the comment

modules/apps/transfer/keeper/relay.go
chatton: in favour of this change also :+1:

chatton (commented):
really nice job @srdtrk , logic is much easier to follow and reason about. Left a few small comments but nothing major, thanks for doing this refactor :muscle:

modules/apps/transfer/keeper/forward.go
chatton: main thing for me is when I have access to a keeper instance, the filename that contains the function definition doesn't actually help

e.g. 

image


I think it would be nice to have a bit more of a descriptive name.

modules/apps/transfer/keeper/forward.go
srdtrk: Hm, it seems this would only get rid of `acknowledgeForwardedPacket` function. I'd rather not inline this because `GetCapability` hurts readability a lot in my opinion. I want to see `GetCapability` in as few places as possible. Could you reconsider this?

modules/apps/transfer/keeper/forward.go
srdtrk: I like `forwarding.go` more. I'll move forward with this, can change later/discuss if others disagree.

modules/apps/transfer/keeper/forward.go
srdtrk: Yes, please create a new issue for this

modules/apps/transfer/keeper/forward.go
srdtrk: These tokens are both sent and received. So this is confusing. But its actually more accurate to say sent. I changed this whole godoc anyway

modules/apps/transfer/keeper/forward.go
srdtrk: I'll remove the comment. This function follows the structure of `refundPacketTokens`. If we decide to change this, we should change that function too

modules/apps/transfer/keeper/relay.go
srdtrk: Yes, the logic did change, this is not simply a refactor

modules/apps/transfer/keeper/forward.go
srdtrk: Duplicate https://github.com/cosmos/ibc-go/pull/6574#discussion_r1636019471. I am renaming this to `revertForwardedPacket`

srdtrk:
> Nice work with this refactoring, @srdtrk! I left a few suggestions and one thing that I think would be good to is on naming of the packets that are received and sent in the middle hop. I think at the moment we use both `prevPacket` and `forwardedPacket` for the packet that is received in the middle hop, but sometimes I think we also use forwarded packet for the packet that is sent? I think we are also not very consistent with that naming in the spec either, so would be good to agree on the terminology... What about simply `prevPacket`or `recvPacket` for the packet received and `nextPacket` or `sentPacket` for the packet sent?

I agree more consistent naming would be great, and I tried to improve this in this PR with the use of `prevPacket` and `failedPacket`, however, I'm still not very happy with it. I still don't like all the other suggested names since they are all very easy to misinterpret at different points in the packet life cycle. Maybe we should create a new issue about this?

modules/apps/transfer/keeper/relay.go
srdtrk: done

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+	sdkmath "cosmossdk.io/math"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	channeltypes "github.com/cosmos/ibc-go/v8/modules/core/04-channel/types"
+	host "github.com/cosmos/ibc-go/v8/modules/core/24-host"
+)
+
+// reverts the receive packet logic that occurs in the middle chain and asyncronously acknowledges the prevPacket
```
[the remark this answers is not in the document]
srdtrk: done

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+func (k Keeper) onForwardedPacketErrorAck(ctx sdk.Context, prevPacket channeltypes.Packet, failedPacketData types.FungibleTokenPacketDataV2) error {
+	// the forwarded packet has failed, thus the funds have been refunded to the intermediate address.
+	// we must revert the changes that came from successfully receiving the tokens on our chain
+	// before propogating the error acknowledgement back to original sender chain
```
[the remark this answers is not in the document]
srdtrk: done

modules/apps/transfer/keeper/forward.go
```diff
@@ -0,0 +1,96 @@
...
+	/*
+		Recall that RecvPacket handles an incoming packet depending on the denom of the received funds:
+			1. If the funds are native, then the amount is sent to the receiver from the escrow.
+			2. If the funds are foreign, then a voucher token is minted.
+		We revert it in this function by:
+			1. Sending funds back to escrow if the funds are native.
+			2. Burning voucher tokens if the funds are foreign
+	*/
+
+	intermediateSenderAddr := types.GetForwardAddress(prevPacket.DestinationPort, prevPacket.DestinationChannel)
```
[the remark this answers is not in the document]
srdtrk: done

modules/apps/transfer/keeper/forward.go
srdtrk: done

modules/apps/transfer/keeper/forward.go
srdtrk: Comment removed

modules/apps/transfer/keeper/forward.go
crodriguezvega: We could probably handle success/error/timeout all in one function. Something like this:

```go
func (k Keeper) acknowledgeForwardedAck(ctx sdk.Context, prevPacket channeltypes.Packet, nextPacketData types.FungibleTokenPacketDataV2, prevPacketAck channeltypes.Acknowledgement) error {
  if prevPacketAck.Response.(*channeltypes.Acknowledgement_Error) != nil {
    if err := k.revertInFlightChanges(ctx, prevPacket, nextPacketData); err != nil {
      return err
    }
  }

  capability, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(prevPacket.DestinationPort, prevPacket.DestinationChannel))
  if !ok {
    return errorsmod.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
  }

  return k.ics4Wrapper.WriteAcknowledgement(ctx, capability, prevPacket, prevPacketAck)
}
```

And then getting the capability is in a single place.

srdtrk: This doesn't address the timeout case. I am strongly against this suggestion because coupling acknowledgement, error handling, and reversion into a single function overcomplicates the code imo. I'd rather have multiple easy to understand helper functions.

crodriguezvega: I might be wrong, but I think it would cover it. In `OnTimeoutPacket` You would do something like:

```
forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet timed out"))
return k.acknowledgeForwardedAck(ctx, prevPacket, packet, forwardAck)
```

srdtrk: Ah ok, it does handle it. Yeah, lets kick this down the road, I do like functions with simple responsibilities.

Although it may seem ok to merge them now, we might want to perform additional logic in the future depending on whether it was a timeout or not, such as emitting events. Then having these separated could be helpful.

modules/apps/transfer/keeper/forward.go
gjermundgaraba: #6581

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+		return err
+	}
+
+	forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet failed"))
+	return k.acknowledgeForwardedPacket(ctx, prevPacket, forwardAck)
+}
+
+// onForwardedPacketResultAck asynchronously acknowledges the prevPacket
+func (k Keeper) onForwardedPacketResultAck(ctx sdk.Context, prevPacket channeltypes.Packet) error {
+	forwardAck := channeltypes.NewResultAcknowledgement([]byte("forwarded packet succeeded"))
```
damiannolan, line 31: Is this the expected structure for result acknowledgements? should this differ from the single byte which we write on ordinary transfer packet acks?

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+// onForwardedPacketErrorAck reverts the receive packet logic that occurs in the middle chain and asynchronously acknowledges the prevPacket
+func (k Keeper) onForwardedPacketErrorAck(ctx sdk.Context, prevPacket channeltypes.Packet, failedPacketData types.FungibleTokenPacketDataV2) error {
+	// the forwarded packet has failed, thus the funds have been refunded to the intermediate address.
+	// we must revert the changes that came from successfully receiving the tokens on our chain
+	// before propagating the error acknowledgement back to original sender chain
+	if err := k.revertForwardedPacket(ctx, prevPacket, failedPacketData); err != nil {
+		return err
+	}
+
+	forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet failed"))
```
damiannolan, line 25: All of the error acknowledgements written here will not have any error code or codespace associated with them if we just use `errors.New()`, since string error info is stripped in `NewErrorAckwnoledgement` in 04-channel.

We should consider adding `errorsmod.Wrap(someRegisteredErr)`

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+func (k Keeper) acknowledgeForwardedPacket(ctx sdk.Context, packet channeltypes.Packet, ack channeltypes.Acknowledgement) error {
+	capability, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(packet.DestinationPort, packet.DestinationChannel))
+	if !ok {
+		return errorsmod.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
+	}
+
+	return k.ics4Wrapper.WriteAcknowledgement(ctx, capability, packet, ack)
+}
+
+// revertForwardedPacket reverts the logic of receive packet that occurs in the middle chains during a packet forwarding.
```
damiannolan, line 55: nit: are we going with "middle" chain or "intermediary" chain, I think the latter might already be widely used terminology, but I have no real preference, just raising for consistency purposes

modules/apps/transfer/keeper/relay.go
```diff
@@ -572,3 +493,20 @@ func createPacketDataBytesFromVersion(appVersion, sender, receiver, memo string,
 
 	return packetDataBytes
 }
+
+// This function sends coins from the account to the transfer module account and then burn them.
+// We do this because bankKeeper.BurnCoins only works with a module account in SDK v0.50,
+// the next version of the SDK will allow burning coins from any account.
+// TODO: remove this function once we switch forwarding address to a module account (#6561)
+func (k Keeper) burnCoin(ctx sdk.Context, account sdk.AccAddress, coin sdk.Coin) error {
```
damiannolan, line 501: another godoc nit, not sure if linter complains about unexported symbols godocs, but normally for exported symbols it says "should start with func name" - maybe we should do it here also:

e.g. `burnCoins sends coins from the account to the transfer module account ...`

damiannolan (commented):
Awesome work @srdtrk! This is a really nice improvement 🙏🏻 

I just left some nits on godocs phasing, but also raised some points on acknowledgements. Happy to get it merged once there's some resolution to those (either in this PR or following up later with issues).

modules/apps/transfer/keeper/forwarding.go
damiannolan, line 1: potentially could name `onForwardedPacketXYZ` funcs for easier understanding... just to throw out some suggestions.

- `ackForwardPacketSuccess`
- `ackForwardPacketError`
- `ackForwardPacketTimeout`

unsure if this reads a little better or not, happy to keep as is if there's alignment there 👍🏻 

this can also be addressed later if needs be, no stress for now

modules/apps/transfer/keeper/forwarding.go
srdtrk: I agree @damiannolan, this is from Stefano's code. Since the tests are passing, ibc-go likely ignores these bytes anyways

modules/apps/transfer/keeper/forwarding.go
srdtrk: Should we make this a separate issue? I'm trying to keep this PR as a fix and refactor.

gjermundgaraba:
> I agree more consistent naming would be great, and I tried to improve this in this PR with the use of `prevPacket` and `failedPacket`, however, I'm still not very happy with it. I still don't like all the other suggested names since they are all very easy to misinterpret at different points in the packet life cycle. Maybe we should create a new issue about this?

I would agree. There is also some talk about these names in the protocol PR. I think it would be better to get this code merged and clean up the names separately.

modules/apps/transfer/keeper/forwarding.go
gjermundgaraba: I think that is a good idea. I would love to see this merged sooner rather than later to keep momentum going (and it would let more people pitch in on those separate issues) :)

damiannolan: https://github.com/cosmos/ibc-go/issues/6586

modules/apps/transfer/keeper/forwarding.go
damiannolan: I will create another issue for this

damiannolan (approved):
LGTM and create issues for anything that needs to be followed up

modules/apps/transfer/keeper/forwarding.go
srdtrk: I haven't seen "intermediary" used anywhere in the repo. I'll go with whatever others agree.

modules/apps/transfer/keeper/forwarding.go
srdtrk: I've adopted these names

modules/apps/transfer/keeper/relay.go
srdtrk: done

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+func (k Keeper) onForwardedPacketTimeout(ctx sdk.Context, prevPacket channeltypes.Packet, timeoutPacketData types.FungibleTokenPacketDataV2) error {
+	if err := k.revertForwardedPacket(ctx, prevPacket, timeoutPacketData); err != nil {
+		return err
+	}
+
+	forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet timed out"))
+	return k.acknowledgeForwardedPacket(ctx, prevPacket, forwardAck)
+}
+
+// acknowledgeForwardedPacket writes acknowledgement for a forwarded packet asynchronously
```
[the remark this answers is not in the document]
srdtrk: done

modules/apps/transfer/keeper/forwarding.go
damiannolan: https://github.com/cosmos/ibc-go/issues/6587

modules/apps/transfer/keeper/forwarding.go
damiannolan: Fair enough, maybe just a spoken term with Aditya then

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+	if err := k.revertForwardedPacket(ctx, prevPacket, timeoutPacketData); err != nil {
+		return err
+	}
+
+	forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet timed out"))
+	return k.acknowledgeForwardedPacket(ctx, prevPacket, forwardAck)
+}
+
+// acknowledgeForwardedPacket writes the async acknowledgement for packet
+func (k Keeper) acknowledgeForwardedPacket(ctx sdk.Context, packet channeltypes.Packet, ack channeltypes.Acknowledgement) error {
```
colin-axner, line 46: happy as is, but this could also be `writeAcknowledgement`? Since it is just a wrapper grabbing the channel capability

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+	// before propagating the error acknowledgement back to original sender chain
+	if err := k.revertForwardedPacket(ctx, prevPacket, failedPacketData); err != nil {
+		return err
+	}
+
+	forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet failed"))
+	return k.acknowledgeForwardedPacket(ctx, prevPacket, forwardAck)
+}
+
+// ackForwardPacketSuccess writes the async acknowledgement the prevPacket
```
colin-axner, line 29:
```suggestion
// ackForwardPacketSuccess writes a successful async acknowledgement for the prevPacket
```

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+	return k.acknowledgeForwardedPacket(ctx, prevPacket, forwardAck)
+}
+
+// ackForwardPacketTimeout reverts the receive packet logic that occurs in the middle chain and writes the async ack for the prevPacket
+func (k Keeper) ackForwardPacketTimeout(ctx sdk.Context, prevPacket channeltypes.Packet, timeoutPacketData types.FungibleTokenPacketDataV2) error {
+	if err := k.revertForwardedPacket(ctx, prevPacket, timeoutPacketData); err != nil {
+		return err
+	}
+
+	forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet timed out"))
```
colin-axner, line 41:
```suggestion

        // the timeout is converted into an error acknowledgement in order to propagate the failed packet forwarding
        // back to the original sender
	forwardAck := channeltypes.NewErrorAcknowledgement(errors.New("forwarded packet timed out"))
```

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+	return k.acknowledgeForwardedPacket(ctx, prevPacket, forwardAck)
+}
+
+// ackForwardPacketSuccess writes the async acknowledgement the prevPacket
+func (k Keeper) ackForwardPacketSuccess(ctx sdk.Context, prevPacket channeltypes.Packet) error {
+	forwardAck := channeltypes.NewResultAcknowledgement([]byte("forwarded packet succeeded"))
+	return k.acknowledgeForwardedPacket(ctx, prevPacket, forwardAck)
+}
+
+// ackForwardPacketTimeout reverts the receive packet logic that occurs in the middle chain and writes the async ack for the prevPacket
```
colin-axner, line 35:
```suggestion
// ackForwardPacketTimeout reverts the receive packet logic that occurs in the middle chain and writes a failed async ack for the prevPacket
```

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+	capability, ok := k.scopedKeeper.GetCapability(ctx, host.ChannelCapabilityPath(packet.DestinationPort, packet.DestinationChannel))
+	if !ok {
+		return errorsmod.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
+	}
+
+	return k.ics4Wrapper.WriteAcknowledgement(ctx, capability, packet, ack)
+}
+
+// revertForwardedPacket reverts the logic of receive packet that occurs in the middle chains during a packet forwarding.
+// If an error occurs further down the line, the state changes on this chain must be reverted before sending back the error
```
colin-axner, line 56: it's unclear to me what is intended by "further down the line". I guess this implies an error occurring at a later hop

modules/apps/transfer/keeper/forwarding.go
gjermundgaraba: Maybe something we can take up in the team. Would be good to have a good name for this for docs and code

gjermundgaraba (approved):
LGTM 🚀

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -0,0 +1,97 @@
...
+		// parse the transfer amount
+		transferAmount, ok := sdkmath.NewIntFromString(token.Amount)
+		if !ok {
+			return errorsmod.Wrapf(types.ErrInvalidAmount, "unable to parse transfer amount (%s) into math.Int", transferAmount)
+		}
+		coin := sdk.NewCoin(token.Denom.IBCDenom(), transferAmount)
+
+		// check if the token we received was a source token
+		// note that the DestinationPort and DestinationChannel of prevPacket are the
+		// SourcePort and SourceChannel of failed packet
```
colin-axner, line 82: I found this slightly confusing, as it draws my attention to the packet forward step, but now we are focusing on the packet reverse. It's hard to explain concisely though

```suggestion
		// check if the token we received originated on the sender
		// given that the packet is being reversed, we check the DestinationChannel and DestinationPort 
		// of the prevPacket to see if a hop was added to the trace during the receive step
```

modules/apps/transfer/keeper/forwarding.go
colin-axner, line 56:
```suggestion
// If the packet fails to be forwarded all the way to the final destination, the state changes on this chain must be reverted before sending back the error
```

colin-axner (approved):
Fantastic! :100: 

This is a great cleanup! Very easy to follow and reason about. I left suggestions on documentation. You may apply the suggestions as you wish. Thanks for the quick implementation and followups!

modules/apps/transfer/keeper/forwarding.go
colin-axner: left a suggested change https://github.com/cosmos/ibc-go/pull/6574#discussion_r1637840702

chatton (approved):
Really nice cleanup @srdtrk LGTM

modules/apps/transfer/keeper/forwarding.go
srdtrk: Yes, it is difficult to explain, I'll use your suggestion, this can be changed in the future if too many people are confused.

modules/apps/transfer/keeper/forwarding.go
srdtrk: done

modules/apps/transfer/keeper/forwarding.go
srdtrk: I think `writeAcknowledgement` can be confusing since I don't foresee this function being used as a general utility function. I think we should change its name if there is ever another use case for the function.

Issue #6546, opened 2024-06-10
Path forwarding: delete previous packet after packet lifecycle completes for next packet


crodriguezvega:
This is [a change](https://github.com/cosmos/ibc/pull/1090#discussion_r1627279389) that should be coming to the spec, but I think we can already implement it ourselves.

Issue #4062, opened 2023-07-11
Generate distinct Channel/Connection identifiers when running tests


DimitrisJim:
## Summary

Subtle bugs can slip through when we use the same channel/connection (client?) identifiers since many of the verification functions expect those to verify the given proofs. These will work in our testing environment but would fail in the most common real-life case where identifiers are unique.

Case in point: https://github.com/cosmos/ibc-go/pull/4052. A similar bug was also found in UpgradeCancel after forcing distinct identifiers for channels.

One possible hacky solution is forcing a channel/connection init in the coordinators functions before actually setting up the connection/channel (hack inspiration https://github.com/cosmos/ibc-go/pull/4052#issuecomment-1631065794). This was we ensure distinct connection/channel identifiers for our internal tests.
____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

DimitrisJim:
we currently make assumptions about channel/connection ids in the tests so maybe not do this for all tests atm.

crodriguezvega:
Maybe this is too hacky, but sounds simple to me: before the connection and channel handshakes complete, on one of the chains we write in the `nextConnectionSequence` and `nextChannelSequence` keys the value 1, so that when the connection and channel identifiers are generated in the handshake they use that number instead of 0.

chatton:
Since the channelIDs are generated dynamically in core ibc and we parse them from events in our test, it sounds like @crodriguezvega 's solution seems reasonable to me. A little hacky but I think it is worth it to eliminate this class of error.

PR #6621, opened 2024-06-17
feat: delete forwarded packet when it is not needed anymore 

bznein:
## Description

closes: #6546 

---

write a little note why.


modules/apps/transfer/keeper/keeper.go
```diff
@@ -328,3 +328,16 @@ func (k Keeper) GetForwardedPacket(ctx sdk.Context, portID, channelID string, se
...
 	return storedPacket, true
 }
+
+// deleteForwardedPacket deletes the forwarded packet from the store.
+func (k Keeper) deleteForwardedPacket(ctx sdk.Context, portID, channelID string, sequence uint64) error {
+	store := ctx.KVStore(k.storeKey)
+	packetKey := types.PacketForwardKey(portID, channelID, sequence)
+	if !store.Has(packetKey) {
+		return fmt.Errorf("packet with PortID %s, channelID %s and sequence %d not found in the store", portID, channelID, sequence)
+	}
```
colin-axner, line 338: I think logically this shouldn't be possible given that we only end up in `acknowledgeForwardedPacket` if the packetKey existed. Typically delete functions in our codebase don't return an error as the function call is never expected to fail. I have a very minor preference for not returning an error (skipping the has check), but I'm very happy to go with either decision. 

If we decide to return an error, it should use a registered error type otherwise the SDK will discard the error message with an internal error when returning to the relay message (I believe in this [else](https://github.com/cosmos/cosmos-sdk/blob/main/errors/abci.go#L65) statement)

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -49,7 +49,11 @@ func (k Keeper) acknowledgeForwardedPacket(ctx sdk.Context, packet channeltypes.
 		return errorsmod.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
 	}
 
-	return k.ics4Wrapper.WriteAcknowledgement(ctx, capability, packet, ack)
+	if err := k.ics4Wrapper.WriteAcknowledgement(ctx, capability, packet, ack); err != nil {
+		return err
+	}
+
+	return k.deleteForwardedPacket(ctx, packet.SourcePort, packet.SourceChannel, packet.Sequence)
```
colin-axner, line 56: Could we add checks in our testing for deletion of the right packet?

colin-axner, line 56:
```suggestion
	return k.deleteForwardedPacket(ctx, packet.DestinationPort, packet.DestinationChannel, packet.Sequence)
```

This confused me during the walkthrough a bit. During a middlehop, we forward a new packet. In storage, we make a mapping from the portID, channelID we send that packet from to the packet which caused its creation (creating a pointer in the reverse direction)

Thus when processing the ack/timeout on a middlehop, core IBC gives us the packet we created as the result of a forward. We use that source port/channel id to get the reverse direction packet. Thus when we delete the mapping to the reverse direction packet (`packet` arg in this function is `prevPacket` (maybe should be renamed)) we should use the destination port and channel as that matches the source port and channel used in the original set

colin-axner: lets add an in-line code comment to explain this. I'm sure it'll take many code readers a second to parse

colin-axner (commented):
Looking good! I left a few comments. I believe the incorrect port/channel pair is being referenced. It's possible tests don't catch the issue since we tend to reuse the same starting channelid across testing chains ref #4062

modules/apps/transfer/keeper/keeper.go
bznein: > If we decide to return an error, it should use a registered error type otherwise the SDK will discard the error message with an internal error when returning to the relay message (I believe in this [else](https://github.com/cosmos/cosmos-sdk/blob/main/errors/abci.go#L65) statement)

Ah, thanks for the pointer!


I don't feel strongly on the error. I see that `store.Delete()` doesn't return any error and just panics if there is no valid key. I'm happy to skip the `has` check.

modules/apps/transfer/keeper/forwarding.go
bznein: Done for the `OnAcknowledgment`, for the timeout case I have #6596 in review and I'll ad it to that test once we merge this.

crodriguezvega: > we should use the destination port and channel as that matches the source port and channel used in the original set (wording here was a bit long winded, we can maybe fine a more concise way to explain)

Sorry if I am missing something, but as @colin-axner says the `packet` used here is `prevPacket` (the packet that we receive in the middle hop) and this packet [is stored using the source port ID and channel ID of the next packet](https://github.com/cosmos/ibc-go/blob/c546f69b7ad0e312fe0102c172462b3669a7e650/modules/apps/transfer/keeper/relay.go#L283) (the packet we send from the middle chain), right?

And the destination port ID/channel ID of `prevPacket` does not necessarily match the source port ID/channel ID of the next packet, right? So shouldn't we be using here instead the source port ID/channel ID of the packet that was sent from the middle chain?

Again apologies if I am getting this wrong.

colin-axner: Sorry I was incorrect in my above explanation! The packet received in a middlehop has the source port/channel from the chain that sent the packet and the dest port/channel which is connected to that sender chain. We then send a new packet on the first forwarding hop which is distinctly different from the destination port/channel

If we try to use `prevPacket` we need to unmarshal the data. I would argue we should not be storing the entire packet, but rather just the packetID info. Therefore, I would recommend instead, we break the existing `acknowledgeForwardSuccess` and `acknowledgePacketForwardError` to take in the packetID of the forwarded packet so you can use the source port/channel from that. Makes the api less clean, but I don't see another performant approach at the moment

colin-axner: @crodriguezvega you are correct. It looks like we were writing our comments in parallel

chatton: @colin-axner just talking with @bznein  now and trying to clarify exactly what needs to be done here.

Using as an example a three chain setup of A -> B -> C and ack is happening on B.

In this case, the forwarded packet is keyd under chain C's port/channel ID. And so we can get this either from

- the packet sent from A -> B's forwarding hops (from the packet data) [here](https://github.com/cosmos/ibc-go/blob/f9c1e70cab870be0f0736f848ee483cd257abdc0/modules/apps/transfer/keeper/forwarding.go#L47) 
 or
- the source port/channel ID in the packet received in the ack fn [here](https://github.com/cosmos/ibc-go/blob/60a6b9945beaf03820f6490c7853304c40601b2e/modules/apps/transfer/keeper/relay.go#L282).

Your suggestion is to not store the full packet, but it looks like we still need the `prevPacket` in some places like [here](https://github.com/cosmos/ibc-go/blob/f9c1e70cab870be0f0736f848ee483cd257abdc0/modules/apps/transfer/keeper/forwarding.go#L47)

Don't we still need to store the full packet for this to work? (technically we don't need the packet data, just the ports/channel IDS, is this what you meant by `packetID` info?)

colin-axner: Thanks for the followup @chatton. See [packetID](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/types/packet.go#L132) type

>  the forwarded packet is keyd under chain C's port/channel ID.

yes, but more specifically the channel endpoint on chain B which connects to chain C

> but it looks like we still need the prevPacket in some places like [here](https://github.com/cosmos/ibc-go/blob/f9c1e70cab870be0f0736f848ee483cd257abdc0/modules/apps/transfer/keeper/forwarding.go#L47)

My recommendation is to only store the `packetID` https://github.com/cosmos/ibc-go/issues/6636 on the set you linked

> Don't we still need to store the full packet for this to work?

No. Only the packetID is necessary. We aren't acting on the previous data (we know the previous data since we constructed a packet from it). Once https://github.com/cosmos/ibc-go/issues/6635 is addressed, we won't need the full packet stored

See spec changes https://github.com/cosmos/ibc/pull/1119, I think our code abstractions actually made this a bit more complicated than it needs to be

colin-axner: You can pass the full packet through no problem. I am just leaning towards the changes which would be implemented in https://github.com/cosmos/ibc-go/issues/6635, but happy to change things at that time period

bznein: Thanks for the explanation Colin! 

I went with the solution of passing the packet just to be able to land this PR. I can happily take #6635 after that, but it would require me a bit to understand :)

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -63,27 +63,33 @@ func (k Keeper) ackForwardPacketError(ctx sdk.Context, prevPacket channeltypes.P
...
 }
 
 // ackForwardPacketTimeout reverts the receive packet logic that occurs in the middle chain and writes a failed async ack for the prevPacket
-func (k Keeper) ackForwardPacketTimeout(ctx sdk.Context, prevPacket channeltypes.Packet, timeoutPacketData types.FungibleTokenPacketDataV2) error {
+func (k Keeper) ackForwardPacketTimeout(ctx sdk.Context, prevPacket channeltypes.Packet, forwardedPacket channeltypes.Packet, timeoutPacketData types.FungibleTokenPacketDataV2) error {
 	if err := k.revertForwardedPacket(ctx, prevPacket, timeoutPacketData); err != nil {
 		return err
 	}
 
+
```
colin-axner, line 75:
```suggestion
```

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -48,13 +48,13 @@ func (k Keeper) forwardPacket(ctx sdk.Context, data types.FungibleTokenPacketDat
 }
 
 // ackForwardPacketSuccess writes a successful async acknowledgement for the prevPacket
-func (k Keeper) ackForwardPacketSuccess(ctx sdk.Context, prevPacket channeltypes.Packet) error {
+func (k Keeper) ackForwardPacketSuccess(ctx sdk.Context, prevPacket channeltypes.Packet, forwardedPacket channeltypes.Packet) error {
```
colin-axner, line 51:
```suggestion
func (k Keeper) ackForwardPacketSuccess(ctx sdk.Context, prevPacket, forwardedPacket channeltypes.Packet) error {
```
signature could be updated to drop the double type declaration, but fine as is

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -63,27 +63,33 @@ func (k Keeper) ackForwardPacketError(ctx sdk.Context, prevPacket channeltypes.P
...
 	if !ok {
 		return errorsmod.Wrap(channeltypes.ErrChannelCapabilityNotFound, "module does not own channel capability")
 	}
 
-	return k.ics4Wrapper.WriteAcknowledgement(ctx, capability, packet, ack)
+	if err := k.ics4Wrapper.WriteAcknowledgement(ctx, capability, packet, ack); err != nil {
+		return err
+	}
+
+	k.deleteForwardedPacket(ctx, forwardedPacket.DestinationPort, forwardedPacket.DestinationChannel, forwardedPacket.Sequence)
```
colin-axner, line 91:
```suggestion
	k.deleteForwardedPacket(ctx, forwardedPacket.SourcePort, forwardedPacket.SourceChannel, forwardedPacket.Sequence)
```

modules/apps/transfer/keeper/relay_forwarding_test.go
```diff
@@ -561,6 +561,10 @@ func (suite *KeeperTestSuite) TestAcknowledgementFailureScenario5Forwarding() {
 	err = path1.EndpointB.AcknowledgePacket(packetFromBtoA, errorAckOnA.Acknowledgement())
 	suite.Require().NoError(err)
 
+	// Check that B deleted the forwarded packet.
+	_, found = suite.chainB.GetSimApp().TransferKeeper.GetForwardedPacket(suite.chainB.GetContext(), forwardedPacket.DestinationPort, forwardedPacket.DestinationChannel, packet.Sequence)
```
colin-axner, line 565:
```suggestion
	_, found = suite.chainB.GetSimApp().TransferKeeper.GetForwardedPacket(suite.chainB.GetContext(), forwardedPacket.SourcePort, forwardedPacket.SourceChannel, forwardedPacket.Sequence)
```

modules/apps/transfer/keeper/relay_forwarding_test.go
```diff
@@ -561,6 +561,10 @@ func (suite *KeeperTestSuite) TestAcknowledgementFailureScenario5Forwarding() {
...
+	suite.Require().False(found, "chain B should have deleted the packet")
```
colin-axner, line 566:
```suggestion
	suite.Require().False(found, "chain B should have deleted the forwarded packet mapping")
```

modules/apps/transfer/keeper/forwarding.go
bznein: Forgot about this possibility, done!

modules/apps/transfer/keeper/forwarding.go
bznein: Done

modules/apps/transfer/keeper/forwarding.go
bznein: Damn, right, thanks for catching

modules/apps/transfer/keeper/relay_forwarding_test.go
bznein: Done

modules/apps/transfer/keeper/relay_forwarding_test.go
bznein: Done!

modules/apps/transfer/keeper/forwarding.go
chatton: Thanks for the explanation @colin-axner and @bznein for pushing it forward :muscle:

modules/apps/transfer/keeper/relay_forwarding_test.go
```diff
@@ -561,6 +561,10 @@ func (suite *KeeperTestSuite) TestAcknowledgementFailureScenario5Forwarding() {
 	err = path1.EndpointB.AcknowledgePacket(packetFromBtoA, errorAckOnA.Acknowledgement())
 	suite.Require().NoError(err)
 
+	// Check that B deleted the forwarded packet.
```
crodriguezvega, line 564: We can probably add this check to some other tests as well (like the happy path).

modules/apps/transfer/keeper/keeper.go
```diff
@@ -328,3 +328,11 @@ func (k Keeper) GetForwardedPacket(ctx sdk.Context, portID, channelID string, se
 
 	return storedPacket, true
 }
+
+// deleteForwardedPacket deletes the forwarded packet from the store.
+func (k Keeper) deleteForwardedPacket(ctx sdk.Context, portID, channelID string, sequence uint64) {
```
crodriguezvega, line 333: There is some improvement we can do with the naming since "forwarded packet" here in the name of the function refers to the packet that was received by the middle chain, but in the places where this function is called "forwarded packet" refers to the packet that was sent from the middle chain. Anyway, no need to change it for now, we can make a proposal for other naming in a separate PR.

crodriguezvega (approved):
Looks good to me, @bznein!

modules/apps/transfer/keeper/relay_forwarding_test.go
bznein: Good idea, done!

modules/apps/transfer/keeper/keeper.go
bznein: Yeah I agree. I chose a name which is consistent with what we use in the rest of the file (which in my understanding refer to the same concept), but I'm happy to find a better name.

PR #6622, opened 2024-06-17
test(transfer): last chain in forwarding packet is ICS20 v1

gjermundgaraba:
## Description

closes: #6557 

---

write a little note why.


e2e/testsuite/testsuite.go
```diff
@@ -172,9 +169,17 @@ func (s *E2ETestSuite) newInterchain(ctx context.Context, r ibc.Relayer, chains
 	for i := 0; i < len(chains)-1; i++ {
 		pathName := s.generatePathName()
 		pathNames = append(pathNames, pathName)
+
+		chain1 := chains[i]
+		chain2 := chains[i+1]
+		channelOptions := defaultChannelOpts(chains)
+		if channelOpts != nil {
+			channelOpts(&channelOptions, chain1, chain2)
```
gjermundgaraba, line 177: This way we can actually set up the channel options depending on the chains we are connecting.

e2e/tests/transfer/forwarding_test.go
```diff
@@ -91,3 +89,78 @@ func (s *TransferForwardingTestSuite) TestThreeChainSetup() {
 		s.Require().Equal(expected, actualBalance.Int64())
 	})
 }
+
+func (s *TransferForwardingTestSuite) TestForwarding_WithLastChainBeingICS20v1_Succeeds() {
+	ctx := context.TODO()
+	t := s.T()
+
+	threeChainSetup := testsuite.ThreeChainSetup()
+	chainCChainID := "chainC-1" // Setting manually to ensure chainID is set to chainC-1
```
gjermundgaraba, line 98: I want opinion on this: The other option I considered for this was to export the chain C chain ID from testsuite (the one that `ThreeChainSetup` uses), but it felt a bit weird because we have `func (tc TestConfig) GetChainAID() string` and `func (tc TestConfig) GetChainBID() string` in testconfig.go, and I wasn't sure how to make that smooth. Any ideas are welcome, and I can make this a cleaner, more reusable approach.

gjermundgaraba:
Putting this in draft while waiting for #6629

e2e/testsuite/testsuite.go
```diff
@@ -177,52 +177,60 @@ func (s *E2ETestSuite) SetupTest() {
 	s.SetupPath(ibc.DefaultClientOpts(), defaultChannelOpts(s.GetAllChains()))
 }
 
-// SetupPath creates a path between the chains using the provided client and channel options.
+// SetupPath creates paths between the all chains using the provided client and channel options.
 func (s *E2ETestSuite) SetupPath(clientOpts ibc.CreateClientOptions, channelOpts ibc.CreateChannelOptions) {
```
gjermundgaraba, line 181: I was tempted to rename this to `SetupPaths` so the method below could be named `SetupPath`. Would like feedback if we agree on that (and the addition of the method below, obviously) and I can do that in a separate PR.

e2e/tests/transfer/forwarding_test.go
```diff
@@ -95,3 +99,69 @@ func (s *TransferForwardingTestSuite) TestThreeChainSetup() {
...
+	ctx := context.TODO()
+	t := s.T()
+
+	relayer, chains := s.GetRelayer(), s.GetAllChains()
+
+	chainA, chainB, chainC := chains[0], chains[1], chains[2]
+
+	channelAtoB := s.GetChainAChannel()
+
+	// Creating a new path between chain B and chain C with a ICS20-v1 channel
```
gjermundgaraba, line 113: I went a bit back and forth on how to do this and in the end I thought the cleanest way was to just create a new path for this test because the alternative would be to move this to a separate TestSuite which I didn't like very much.

e2e/testsuite/testsuite.go
```diff
@@ -177,52 +177,60 @@ func (s *E2ETestSuite) SetupTest() {
...
 	allChains := s.GetAllChains()
 	for i := 0; i < len(allChains)-1; i++ {
 		chainA, chainB := allChains[i], allChains[i+1]
-		pathName := s.generatePathName()
-		s.T().Logf("establishing path between %s and %s on path %s", chainA.Config().ChainID, chainB.Config().ChainID, pathName)
+		_, _ = s.CreateNewPath(ctx, chainA, chainB, clientOpts, channelOpts)
+	}
+}
+
+func (s *E2ETestSuite) CreateNewPath(ctx context.Context, chainA ibc.Chain, chainB ibc.Chain, clientOpts ibc.CreateClientOptions, channelOpts ibc.CreateChannelOptions) (ibc.ChannelOutput, ibc.ChannelOutput) {
```
bznein, line 192:
```suggestion
func (s *E2ETestSuite) CreateNewPath(ctx context.Context, chainA, chainB ibc.Chain, clientOpts ibc.CreateClientOptions, channelOpts ibc.CreateChannelOptions) (ibc.ChannelOutput, ibc.ChannelOutput) {
```

not feeling too strong about this, but I think it makes the signature a bit easier to read.

e2e/tests/transfer/forwarding_test.go
```diff
@@ -95,3 +99,69 @@ func (s *TransferForwardingTestSuite) TestThreeChainSetup() {
...
+
+	chainAWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+	chainAAddress := chainAWallet.FormattedAddress()
+	chainADenom := chainA.Config().Denom
+
+	chainCWallet := s.CreateUserOnChainC(ctx, testvalues.StartingTokenAmount)
+	chainCAddress := chainCWallet.FormattedAddress()
+
+	t.Run("IBC transfer from A to C with forwarding through B", func(t *testing.T) {
+		inFiveMinutes := time.Now().Add(5 * time.Minute).UnixNano()
```
bznein, line 127: [q] can we use here https://github.com/cosmos/ibc-go/blob/feat%2Fics20-v2-path-forwarding/testing/chain.go#L634 instead?

bznein (approved):
Great stuff, just a couple of small questions!

e2e/testsuite/testsuite.go
gjermundgaraba: Yeah, it used to create only one path always, and still does whenever there are only two chains. But it now creates paths such that all chains are connected in a line A->B->C->etc.

e2e/testsuite/testsuite.go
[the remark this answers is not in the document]
gjermundgaraba: That would make it clearer, agreed! Will do that :)

gjermundgaraba: I am OK with both options. I am undecided on which approach I usually prefer. I guess it depends on the signature. In this case I think I agree that it would make it a bit less verbose since it anyway has so many arguments.

e2e/tests/transfer/forwarding_test.go
gjermundgaraba: Unfortunately no. I have not found a nice way to do this. The best option I had was a function that queries the latest header and gets the current block time from that. But since these are anyway running in real time in the background, I think just using current time will always work. 

Two options I see to improve this:
- Implement the query of latest block header or status from a node
- Make a little helper function for getting a reasonable timeout (just putting this in a function, basically)

e2e/testsuite/testsuite.go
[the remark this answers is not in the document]
gjermundgaraba: Done. But I also ended up putting back the type for chainA because the signature got so long that I line split it.

Issue #6589, opened 2024-06-13
Allow authz granters to specify forwarding information for token transfers


crodriguezvega:
With the addition of `forwarding` to `MsgTransfer`, we should allow authz granters to specify if all or a subset of forwarding routes are allowed when grantees send IBC transfers on their behalf.

As discussed, we can add a new field to [the `Allocation` type](https://github.com/cosmos/ibc-go/blob/2b887d7f1efabd42256464aa501f440a981e4eed/modules/apps/transfer/types/authz.pb.go#L30):

```diff
@@ -21,7 +22,12 @@ message Allocation {
   repeated string allow_list = 4;
   // allow list of memo strings, an empty list prohibits all memo strings;
   // a list only with "*" permits any memo string
-  repeated string allowed_packet_data = 5;
+  repeated string allowed_packet_data     = 5;
+  repeated Hops   allowed_forwarding_hops = 6;
+}
+
+message Hops {
+  repeated Hop hops = 1 [(gogoproto.nullable) = false];
 }
```

So this field is a slice of `Hops`, which is itself a slice of `Hop`, which is [a port ID/channel ID pair](https://github.com/cosmos/ibc-go/blob/ae2046a71ddaa853c77f9f0175cfbf8e8a89409f/proto/ibc/applications/transfer/v1/transfer.proto#L32). 

The granter can specify different combinations of hops that the grantee can use in their `MsgTransfer`. It's bit more problematic how to specify that all routes are allowed, though, since it's not straightforward to use a wildcard `*` character in the `allowed_forwarding_hops` field. I am open to suggestions from the rest of the team.

Then in [`Accept`](https://github.com/cosmos/ibc-go/blob/ba7c4a87a8b3efd8ae30193e13c4586139d1eaf9/modules/apps/transfer/types/transfer_authorization.go#L44) we need to validate the hops in `MsgTransfer` against the`allowed_forwarding_hops` slices in the `Allocation`.

Let's discuss her in the issue is people agree with this proposal and in the naming of things. If we end up reusing `Trace` instead of `Hop`, that should be reflected here as well.

chatton:
@bznein gonna assign this one to you, I think it will be a nice one we can hop on a call to discuss and/or pair on it :muscle:

crodriguezvega:
We decided to not use any wildcard to allow all.

bznein:
@crodriguezvega 

```proto
+  repeated Hops   allowed_forwarding_hops = 6;
+}
+
[...]

Issue #6624, opened 2024-06-18
Allow users to specify automatic unwinding of tokens to origin chain


crodriguezvega:
As discussed with @womensrights for path forwarding requirements.

---

## Summary

With path forwarding we can forward tokens to a final destination through multiple intermediary chains, but we can also send tokens to their originating chain with a single transaction if the forwarding hops specify the trace that the tokens have travelled. It would be nice for users of `MsgTransfer` to have a simple way to indicate that they want the tokens automatically unwound to the origin chain.

## Proposal

Add a `unwind` boolean to `Forwarding` message. It would probably be better to choose for https://github.com/cosmos/ibc-go/issues/6623 the approach of keeping a separate `Forwarding` message for `MsgTransfer` that would have only `hops` and `unwind`:

```proto
message Forwarding {
  bool     unwind   = 1;
  repeated Hop hops = 2 [(gogoproto.nullable) = false];
}
```

(the name of the message might need to change if we cannot have to messages with the same name).

If the `unwind` flag is set, then we would fill in the `hops` of the packet data the necessary hops to unwind the tokens (which we know since we have that information in the denomination of the tokens) and append any hops that the user may have added additionally in the `hops` field to forward the tokens after completing the unwinding.

We should validate that when a user chooses automatic unwind the `tokens` field contains only one token.

Adding this flag also has impact on transfer authorisations (https://github.com/cosmos/ibc-go/issues/6589 cc: @bznein). We should have also an `allowed_unwind` flag in the `Allocation` so that the granter can set it to `true` if it allows grantees to make use of the flag in `MsgTransfer`. For this reason it would probably be better, instead of having a single `allowed_forwarding_hops` field, to have a struct that contains both the allowed hops and the `allowed_unwind` boolean:

```diff
@@ -21,7 +22,12 @@ message Allocation {
   repeated string allow_list = 4;
   // allow list of memo strings, an empty list prohibits all memo strings;
   // a list only with "*" permits any memo string
-  repeated string allowed_packet_data = 5;
+  repeated string allowed_packet_data = 5;
+  repeated Hops   allowed_forwarding = 6;
+}
+
+message AllowedForwarding {
+  repeated Hops allowed_hops = 1;
+  boolean  allowed_unwind    = 2;
+}
+message Hops {
+  repeated Hop hops = 1 [(gogoproto.nullable) = false];
 }
```

[...]

PR #6627, opened 2024-06-18
feat: allow authz granters to specify forwarding info for token transfer

bznein:
## Description

closes: #6589

---

write a little note why.


bznein:
@crodriguezvega this draft PR targets issue #6589, I'm not sure if we want to merge this already and then do the other changes (possibly naming, and those in #6624) on top, or if we want to hold off on this one and merge it when we decided on everything.

bznein:
Discussed it offline with Carlos and agreed on having this out for review and then modify the code as needed later.

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -389,6 +389,84 @@ func (suite *TypesTestSuite) TestTransferAuthorizationAccept() {
...
+					},
+				}
+			},
+			func(res authz.AcceptResponse, err error) {
+				suite.Require().Error(err)
+				suite.Require().False(res.Accept)
+			},
+		},
+		{
+			"failure - Allocation specify hops but msgTransfer does not have hops",
```
bznein, line 454: I'm not sure I interpreted the requirements correctly. Is this a right case or should the `Accept` succeed if the allocation specifies allowed hops and the forwarding field in the `msgTransfer` doesn't have it?

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -171,6 +176,21 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
 	return false
 }
 
+// areHopsAllowed returns whether the provided slice of Hop matches one of the allowed ones.
+func areHopsAllowed(provided []Hop, allowed []*Hops) bool {
+	if len(allowed) == 0 {
```
damiannolan, line 181: shouldn't this be checking len of provided (i.e. msgTransfer hops)? 

If the length of the msg hops is zero, then it should by default be allowed. Otherwise, we must ensure that the provided hops are specified within the authz allocation.

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -171,6 +176,21 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
 	return false
 }
 
+// areHopsAllowed returns whether the provided slice of Hop matches one of the allowed ones.
+func areHopsAllowed(provided []Hop, allowed []*Hops) bool {
```
damiannolan, line 180: nit, maybe would go for something like `isAllowedForwarding`.. food for thought. Happy to hear input from others as well!

modules/core/errors/errors.go
```diff
@@ -60,4 +60,7 @@ var (
 
 	// ErrNotFound defines an error when requested entity doesn't exist in the state.
 	ErrNotFound = errorsmod.Register(codespace, 16, "not found")
+
+	// ErrUnauthorizedHops defines an error when the hops provided are not authorized.
+	ErrUnauthorizedHops = errorsmod.Register(codespace, 17, "unauthorized hops")
```
damiannolan, line 65: Would suggest to maintain this error within `transfer/types` pkg

proto/ibc/applications/transfer/v1/authz.proto
```diff
@@ -22,6 +23,14 @@ message Allocation {
   // allow list of memo strings, an empty list prohibits all memo strings;
   // a list only with "*" permits any memo string
   repeated string allowed_packet_data = 5;
+  // a list of Allowed hops
+  repeated Hops allowed_forwarding_hops = 6;
+}
+
+
+// Hops represent a list of Hop
+message Hops {
```
damiannolan, line 32: Is there any specific reason to introduce a new type here wrapping `Hop`? couldn't we use a `repeated Hop` field within the wrapper inside of `Allocation`?

modules/apps/transfer/types/transfer_authorization.go
bznein: I think that makes sense, more in line with the rest on the function

modules/core/errors/errors.go
bznein: Done

proto/ibc/applications/transfer/v1/authz.proto
bznein: I'm not sure I follow: `allowed_forwarding_hops` is a list of list of Hops. So we can't replace it with a `repeated Hop`. It is a `repeated repeated Hop`.

But let me know if I didn't get your comment!

modules/apps/transfer/types/transfer_authorization.go
bznein: Two things here

* Since `allowed_forwarding_hops` is not a mandatory field, we check whether it's empty because, in that case, all hops are allowed (in my understanding

* > If the length of the msg hops is zero, then it should by default be allowed.
That's the part I'm not sure and I asked the same here: https://github.com/cosmos/ibc-go/pull/6627/files#r1645684477 
If `allowed_forwarding_hops` is not empty, but the msg contains no hops, should we accept it?

gjermundgaraba: I would like a definitive answer on this as well, but my understanding here was that if the allowed hops are set, then those hops needs to be part of the message. Meaning, you shouldn't be allowed to send through any other paths (including one without hops). To me, the allowed_forwarding_hops guarantees that when you send a token, it will end up on the expected chain with the expected denom. If that is not the case, it would be good to get that clarified :)

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -389,6 +389,84 @@ func (suite *TypesTestSuite) TestTransferAuthorizationAccept() {
 				suite.Require().Nil(res.Updated)
 			},
 		},
+		{
+			"success - allowed forwarding hops",
+			func() {
+				msgTransfer.Forwarding = types.NewForwarding("", types.Hop{PortId: "1", ChannelId: "channel-1"}, types.Hop{PortId: "2", ChannelId: "channel-2"})
```
gjermundgaraba, line 395: nit: would it make sense to use the same port as the other tests for consistency `ibctesting.MockPort`?

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -389,6 +389,84 @@ func (suite *TypesTestSuite) TestTransferAuthorizationAccept() {
...
+					},
+				}
+			},
+			func(res authz.AcceptResponse, err error) {
+				suite.Require().NoError(err)
+				suite.Require().True(res.Accept)
+			},
+		},
+		{
+			"failure - allowed forwarding hops is longer",
```
crodriguezvega, line 411: Maybe something like this?

```suggestion
			"failure - allowed forwarding hops contains more hops",
```

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -389,6 +389,84 @@ func (suite *TypesTestSuite) TestTransferAuthorizationAccept() {
...
+					},
+				}
+			},
+			func(res authz.AcceptResponse, err error) {
+				suite.Require().Error(err)
+				suite.Require().False(res.Accept)
+			},
+		},
+		{
+			"failure - One of the hops is different",
```
crodriguezvega, line 433:
```suggestion
			"failure - allowed forwarding hops contains one different hop",
```

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -389,6 +389,84 @@ func (suite *TypesTestSuite) TestTransferAuthorizationAccept() {
...
+							{PortId: "2", ChannelId: "channel-2"},
+						},
+					},
+				}
+			},
+			func(res authz.AcceptResponse, err error) {
+				suite.Require().Error(err)
+				suite.Require().False(res.Accept)
+			},
+		},
```
crodriguezvega, line 469: We could also add an extra failure test where `allowed_forwarding_hops` is empty and `hops` in `MsgTransfer` is not empty.

crodriguezvega (commented):
Thank you, @bznein! Sorry if the requirements were not clear, I hope I have answered your questions and it is now clearer.

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -52,6 +53,10 @@ func (a TransferAuthorization) Accept(goCtx context.Context, msg proto.Message)
 		return authz.AcceptResponse{}, errorsmod.Wrapf(ibcerrors.ErrNotFound, "requested port and channel allocation does not exist")
 	}
 
+	if !isAllowedForwarding(msgTransfer.Forwarding.Hops, a.Allocations[index].AllowedForwardingHops) {
+		return authz.AcceptResponse{}, errorsmod.Wrapf(ErrUnauthorizedHops, "forwarding hops are not authorized")
```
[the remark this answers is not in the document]
bznein: Good idea, done!

modules/apps/transfer/types/transfer_authorization.go
```diff
@@ -171,6 +176,21 @@ func isAllowedAddress(ctx sdk.Context, receiver string, allowedAddrs []string) b
 	return false
 }
 
+// isAllowedForwarding returns whether the provided slice of Hop matches one of the allowed ones.
+func isAllowedForwarding(provided []Hop, allowed []*Hops) bool {
```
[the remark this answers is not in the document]
bznein: Done!

modules/apps/transfer/types/transfer_authorization.go
bznein: Done (I didn't add the `if len(allowed) == 0`) because that's implicit (if `provided` is not empty, but `allowed` is, we directly skip the for loop and return false. 

I've also added a test case to cover it.

Thanks for the clarification!

modules/apps/transfer/types/transfer_authorization_test.go
```diff
@@ -389,6 +389,84 @@ func (suite *TypesTestSuite) TestTransferAuthorizationAccept() {
 				suite.Require().Nil(res.Updated)
 			},
 		},
+		{
+			"success - allowed forwarding hops",
```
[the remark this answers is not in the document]
bznein: Done!

modules/apps/transfer/types/transfer_authorization_test.go
bznein: Yeah good idea, done :)

modules/apps/transfer/types/transfer_authorization_test.go
bznein: Done!

modules/apps/transfer/types/transfer_authorization_test.go
bznein: Done!

modules/apps/transfer/types/transfer_authorization_test.go
bznein: Added!

modules/apps/transfer/types/transfer_authorization.go
gjermundgaraba: The only potential risk there is that would allow someone to spoof the address (e.g. a fake chain) and send it there. The only way to truly enforce that you can only send to the correct chain would then be to use both port/channel (to avoid the case above) and hops, right? 

I can see both sides of this, and neither is 100% intuitive (to me). So I think we just need to carefully document the behavior so nobody gets surprised :)

bznein:
I made a git mess :) pls ignore this PR until I fix it :)

Issue #6496, opened 2024-06-05
Change `MerklePath` to use `bytes` instead of `string`


colin-axner:
## Summary

Change the [`MerklePath.KeyPath`](https://github.com/cosmos/ibc-go/blob/main/proto/ibc/core/commitment/v1/commitment.proto#L29) field to use `repeated bytes` instead of `repeated strings`. 

## Problem Definition

According to the [protobuf specification](https://protobuf.dev/programming-guides/proto3/#scalar):

string
> A string must always contain UTF-8 encoded or 7-bit ASCII text, and cannot be longer than 2^32.

bytes
> May contain any arbitrary sequence of bytes no longer than 2^32.

In go, neither strings nor bytes are required to be valid utf8 characters, [ref](https://go.dev/blog/strings):

> It’s important to state right up front that a string holds arbitrary bytes. It is not required to hold Unicode text, UTF-8 text, or any other predefined format. As far as the content of a string is concerned, it is exactly equivalent to a slice of bytes.

Thus trying to prove a key which includes an address (arbitrary bytes) will fail.

## Use cases

As requested by @benluelo, it is desirable to verify the following [state](https://github.com/berachain/polaris/blob/main/cosmos/x/evm/plugins/state/plugin.go#L409) at this [path](https://github.com/berachain/polaris/blob/main/cosmos/x/evm/plugins/state/keys.go#L52) which is stored under arbiraty bytes (not guaranteed to be valid utf8). 

## Proposal

Change the commitment proto `MerklePath` type to use `bytes`. This is a breaking change to the proto type and thus would require a bump in the package version. This type is only imported in proto by `QueryVerifyMembershipProof` which was recently added. I recommend we reserve the existing field number and add a new field number with the new type, reusing the `merkle_path` field in that query. 

It should be noted that contracts relying on this field will need to be updated to use the new field, but as it was recently added, we should be able to inform any users before they deploy contracts using the old value. 

I believe this will be a common issue as it is quite lucky that all ibc keys happen to be valid utf8 characters. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

srdtrk:
This issue would also help with `ibc-lite` tests so I can take this :).

[...]

PR #6633, opened 2024-06-18
refactor!: adapt merkle path to use repeated bytes in favour of strings

damiannolan:
## Description

Initial PR. Will break the API of `NewMerklePath` in a following PR.

- Change `MerklePath` to use `repeated bytes` in favour of `repeated string`.

ref: #6496 

---

write a little note why.


colin-axner (approved):
Excellent! :heart:

crodriguezvega (approved):
Nice this refactor is started, @damiannolan!

I am wondering if it would help readability to have a type alias for `[][]byte`?

Should we also mention in migration docs that `KeyPath` type is changed from string to `[][]byte`?

damiannolan:
> I am wondering if it would help readability to have a type alias for `[][]byte`?

I mentioned something similar to @colin-axner yesterday in dms. I can explore that when the bulk of the code is updated.

> Should we also mention in migration docs that `KeyPath` type is changed from string to `[][]byte`?

Yes! I'll add it to the issue 👍🏻

damiannolan:
I'll add a changelog to this before merging

proto/ibc/core/commitment/v1/commitment.proto
```diff
@@ -26,7 +26,7 @@ message MerklePrefix {
 // arbitrary structured object (defined by a commitment type).
 // MerklePath is represented from root-to-leaf
 message MerklePath {
-  repeated string key_path = 1;
+  repeated bytes key_path = 1;
```
erwanor, line 29: This is a proto breaking change, what's the tradeoff here vs. field deprecation?

damiannolan: Hey @erwanor! Does this raise any concern in particular for you and do you depend on the message type? 

This is a bit of a protobuf grey area. Myself and @colin-axner looked into this quite a bit to ensure it was okay to make this change in order to support proving kvs stored under non utf8 encoded keys (as requested by union).

Thus far we've considered this a go api breaking but not actually proto encoding breaking. Backwards compatibility should be maintained with this change. I wrote a quick test to validate this, introducing a temp `LegacyMerklePath` type with `repeated string` (old diff).

```go
func (suite *MerkleTestSuite) TestMerklePathEncoding() {
	encodingCfg := moduletestutil.MakeTestEncodingConfig()

	keyPath := []string{"ibc", "commitments/ports/transfer/channels/channel-0/sequences/1"}
	merklePath := types.NewMerklePath(keyPath...)

	bz, err := encodingCfg.Codec.Marshal(&merklePath)
	suite.Require().NoError(err)

	var legacy types.LegacyMerklePath
	err = encodingCfg.Codec.Unmarshal(bz, &legacy)
	suite.Require().NoError(err)

	suite.Require().Equal(keyPath, legacy.KeyPath)
}
```

Some of the resources which helped us come to this conclusion:

https://protobuf.dev/programming-guides/encoding/#structure
> The binary version of a message just uses the field’s number as the key – the name and declared type for each field can only be determined on the decoding end by referencing the message type’s definition (i.e. the .proto file).

https://protobuf.dev/programming-guides/proto2/#updating
> `string` and `bytes` are compatible as long as the bytes are valid UTF-8.

It's possible that we may have overlooked something wrt to rust users/tooling, please let us know if you have any concerns!

damiannolan: The biggest concern for me here was that when encoding to JSON you now end up with a `key_path` field which is a base64 encoded byte string.

damiannolan: Also happy to introduce a `v2` proto pkg and adjust ibc-go code as necessary. If it makes things easier to keep the old type around!

PR #6656, opened 2024-06-20
feat(transfer): add unwinding ability

DimitrisJim:
## Description

A first stab at implementation, tests pending, based on proto refactor branch.

closes: #6624 

---

write a little note why.


bznein:
@DimitrisJim after talking with Carlos, I am adding some commits here. I added the `allow_unwind` field to the `Allocation` proto, together with validation and some testing.

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -59,3 +66,33 @@ func (k Keeper) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams)
...
+	if token.Denom.IsNative() {
+		return nil, errorsmod.Wrap(types.ErrInvalidForwarding, "cannot unwind a native token")
+	}
+	var unwindHops []types.Hop
+	for _, trace := range token.Denom.Trace[1:] {
+		unwindHops = append(unwindHops, types.Hop{PortId: trace.PortId, ChannelId: trace.ChannelId}) //nolint: gosimple
+	}
+
+	// Update message fields.
+	msg.SourcePort, msg.SourceChannel = token.Denom.Trace[0].PortId, token.Denom.Trace[0].ChannelId
```
crodriguezvega, line 93: Since we are overriding here the `SourcePort` and `SourceChannel`, maybe it's possible to relax [the validation of these fields in `ValidateBasic`](https://github.com/cosmos/ibc-go/blob/35181d87c6aabd10a89346586b9f67cee21acd69/modules/apps/transfer/types/msgs.go#L70-L75)? If `unwind` is `true`, then skip the validation and allow them to be empty, for example.

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -59,3 +66,33 @@ func (k Keeper) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams)
...
+	}
+
+	token, err := k.tokenFromCoin(ctx, coins[0])
+	if err != nil {
+		return nil, err
+	}
+
+	if token.Denom.IsNative() {
+		return nil, errorsmod.Wrap(types.ErrInvalidForwarding, "cannot unwind a native token")
+	}
```
crodriguezvega, line 86: Should we also move this to `ValidateBasic`?

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -59,3 +66,33 @@ func (k Keeper) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams)
 
 	return &types.MsgUpdateParamsResponse{}, nil
 }
+
+// unwindHops unwinds the hops present in the tokens denomination and returns the message modified to reflect
+// the unwound path to take. It ensures only a single token must be present in the tokens list
+// and that the token is not native to the chain.
+func (k Keeper) unwindHops(ctx sdk.Context, msg *types.MsgTransfer) (*types.MsgTransfer, error) {
```
crodriguezvega, line 73: If we pass the `msg` as a pointer, could we modify in place instead of returning it again?

proto/ibc/applications/transfer/v1/authz.proto
```diff
@@ -23,9 +23,17 @@ message Allocation {
...
+}
+
+// AllowedForwarding defines which options are allowed for forwarding.
+message AllowedForwarding {
   // a list of allowed source port ID/channel ID pairs through which the packet is allowed to be forwarded until final
   // destination
-  repeated Hops allowed_forwarding_hops = 6 [(gogoproto.nullable) = false];
+  repeated Hops allowed_hops = 1 [(gogoproto.nullable) = false];
+  // Whether to allow automatic unwinding of tokens to the original chain.
+  bool allow_unwind = 2;
```
crodriguezvega, line 36: I would switch the fields around to [match the order in `Forwarding` message for `MsgTransfer`](https://github.com/cosmos/ibc-go/blob/6619821c739891ff132e42eb1c1a49f5e0791e7c/proto/ibc/applications/transfer/v1/transfer.proto#L25-L30)

crodriguezvega (commented):
Thanks for starting this work, @DimitrisJim, and thanks, @bznein, for picking it up. I left some comments: I think it would be good to explore if some of the validations can be moved to `ValidateBasic`. And then it should be easy to test all the different possibilities.

modules/apps/transfer/keeper/msg_server.go
bznein: I personally think it's ok to leave it as it is. It makes it easier it's a different message and makes some code more readable, but I'm open to changing it if you feel strongly about it

modules/apps/transfer/keeper/msg_server.go
bznein: I wasn't able to achieve it since it's consume states from the keeper (`k.tokenFromCoin`) open to suggestions on how to implement it if it's possible though

modules/apps/transfer/keeper/msg_server.go
bznein: Done

modules/apps/transfer/keeper/msg_server.go
crodriguezvega: Ah, yes, that's true.

modules/apps/transfer/types/msgs.go
```diff
@@ -113,6 +117,13 @@ func (msg MsgTransfer) ValidateBasic() error {
 		}
 	}
 
+	if msg.Forwarding.Unwind {
+		// When unwinding, we must have at must one token.
```
crodriguezvega, line 121:
```suggestion
		// When unwinding, we must have at most one token.
```

modules/apps/transfer/types/msgs.go
```diff
@@ -113,6 +117,13 @@ func (msg MsgTransfer) ValidateBasic() error {
...
+		if len(msg.GetCoins()) > 1 {
+			return errorsmod.Wrap(ErrInvalidForwarding, "cannot unwind more that one tokens")
```
crodriguezvega, line 123:
```suggestion
			return errorsmod.Wrap(ErrInvalidForwarding, "cannot unwind more that one token")
```

modules/apps/transfer/types/msgs.go
```diff
@@ -69,11 +69,15 @@ func NewMsgTransfer(
...
 // the chain is not known to IBC.
 func (msg MsgTransfer) ValidateBasic() error {
-	if err := host.PortIdentifierValidator(msg.SourcePort); err != nil {
-		return errorsmod.Wrap(err, "invalid source port ID")
-	}
-	if err := host.ChannelIdentifierValidator(msg.SourceChannel); err != nil {
-		return errorsmod.Wrap(err, "invalid source channel ID")
+	// Source ports and channel are overwritten when unwinding, so there's no need
+	// to validate them.
+	if !msg.Forwarding.Unwind {
```
crodriguezvega, line 74: It might be good to enforce that if `unwind` is true, then the source port and channel should be empty, because the validation that is done with `PortIdentifierValidator` and `ChannelIdentifierValidator` also checks the length. So this means that somebody could set `unwind` to true and put a ridiculously long string in the source port and channel and then we have again issues with spamming.

modules/apps/transfer/types/msgs.go
chatton, line 121:
```suggestion
		// When unwinding, we must have at most one token.
```

modules/apps/transfer/types/msgs.go
chatton, line 123:
```suggestion
			return errorsmod.Wrap(ErrInvalidForwarding, "cannot unwind more that one token")
```
Also, I'm thinking that instead of `ErrInvalidForwarding`, we could use `ErrInvalidCoins`, it's not the forwarding info that's bad, but the tokens that are bade based on the configured forwarding.

modules/apps/transfer/keeper/msg_server_test.go
```diff
@@ -238,3 +239,108 @@ func (suite *KeeperTestSuite) TestUpdateParams() {
...
+			func() {
+				hexHash := denom.IBCDenom()[len(types.DenomPrefix+"/"):]
+
+				hash, err := types.ParseHexHash(hexHash)
+				suite.Require().NoError(err)
+				suite.chainA.DeleteKey(hash)
+			},
+			func(modified *types.MsgTransfer, err error) {
+				suite.Require().Error(err)
+			},
```
chatton, line 294: I don't think this test is testing what we want it to!

A few things,

- in order to get the `hexHash` we can juse use `denom.Hash()` which returns what you are parsing.
- We don't actually set the denom before this test runs, so we don't need to delete the denom, since it not there to begin with. e.g., if you add `suite.chainA.GetSimApp().TransferKeeper.SetDenom(suite.chainA.GetContext(), denom)`  to the beginning of the malleate, this test will fail.
- The key we want to delete is actually in a prefixed store, not the `ibc` store. 

Sorry I led you down the `DeleteKey` route but actually we need to use a prefixed store for the transfer module and delete it there. This should do the trick

```golang
			func() {
				suite.chainA.GetSimApp().TransferKeeper.SetDenom(suite.chainA.GetContext(), denom)

				storeKey := suite.chainA.GetSimApp().GetKey(types.StoreKey) // types.StoreKey == "transfer"
				kvStore := suite.chainA.GetContext().KVStore(storeKey)

				store := prefix.NewStore(kvStore, types.DenomKey)
				store.Delete(denom.Hash())
			},
```

chatton (commented):
Looking great! Found one issue with a test but the main implementation is looking great

modules/apps/transfer/keeper/msg_server_test.go
bznein: I think it's actually my fault for moving some tests stuff around after our call :)

But I think since we don't set the denom for each test now, we can simply avoid all of this. Please take a look at the new implementation and LMK if it looks ok. I als ochanged the suite.Require to test for the exact error type

modules/apps/transfer/types/msgs.go
bznein: Yeah that makes sense! Added a test as well

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -59,3 +66,29 @@ func (k Keeper) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams)
...
+	for _, trace := range token.Denom.Trace[1:] {
+		unwindHops = append(unwindHops, types.Hop{PortId: trace.PortId, ChannelId: trace.ChannelId}) //nolint: gosimple
+	}
+
+	// Update message fields.
+	msg.SourcePort, msg.SourceChannel = token.Denom.Trace[0].PortId, token.Denom.Trace[0].ChannelId
+	msg.Forwarding.Hops = append(unwindHops, msg.Forwarding.Hops...)
+	msg.Forwarding.Unwind = false
+
+	return msg, msg.ValidateBasic()
```
chatton, line 93: maybe we can return `nil` if `msg.ValidateBasic()` returns an error, otherwise we are returning a modified and invalid message.

modules/apps/transfer/keeper/msg_server_test.go
```diff
@@ -238,3 +239,102 @@ func (suite *KeeperTestSuite) TestUpdateParams() {
...
+				)
+				suite.Require().Equal(*msg, *modified, "expected msg and modified msg are different")
+			},
+		},
+		{
+			"failure: no denom set on keeper",
+			func() {},
+			func(modified *types.MsgTransfer, err error) {
+				suite.Require().ErrorIs(err, types.ErrDenomNotFound)
+			},
```
chatton, line 288: :+1:

chatton (approved):
LGTM! Just one small suggestion but looks great overall, thank you @DimitrisJim / @bznein

modules/apps/transfer/keeper/msg_server.go
bznein: I think that if an error is returned, the caller should make no assumptions of the other values returned (and shouldn't attempt using it!) but I will make the change anyway :)

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -59,3 +66,32 @@ func (k Keeper) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams)
 
 	return &types.MsgUpdateParamsResponse{}, nil
 }
+
+// unwindHops unwinds the hops present in the tokens denomination and returns the message modified to reflect
+// the unwound path to take. It ensures only a single token must be present in the tokens list
+// and that the token is not native to the chain.
+func (k Keeper) unwindHops(ctx sdk.Context, msg *types.MsgTransfer) (*types.MsgTransfer, error) {
+	coins := msg.GetCoins()
+	token, err := k.tokenFromCoin(ctx, coins[0])
```
gjermundgaraba, line 75: I know the length has already been verified in ValidateBasic, but I wonder if we should comment or change the godoc because at first glance I was unsure if this actually ensures that a single token is used or just cuts the list if there are multiple - but I guess we working from the assumption that ValidateBasic catches the error of multiple coins.

gjermundgaraba (approved):
Looks good!

modules/apps/transfer/keeper/msg_server.go
bznein: I'll add a comment here for clarity and change the godoc :)

crodriguezvega (approved):
Great work, @bznein. I just left one comment to increase test coverage confidence, let me know if you also think it makes sense. Otherwise it looks great.

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -59,3 +66,32 @@ func (k Keeper) UpdateParams(goCtx context.Context, msg *types.MsgUpdateParams)
...
+	if err != nil {
+		return nil, err
+	}
+
+	if token.Denom.IsNative() {
+		return nil, errorsmod.Wrap(types.ErrInvalidForwarding, "cannot unwind a native token")
+	}
+	var unwindHops []types.Hop
+	for _, trace := range token.Denom.Trace[1:] {
+		unwindHops = append(unwindHops, types.Hop{PortId: trace.PortId, ChannelId: trace.ChannelId}) //nolint: gosimple
```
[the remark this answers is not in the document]
bznein: Yeah that's actually a very good point, added

Issue #6634, opened 2024-06-18
Configure relayers to watch only channels associated with an individual test #2


chatton:
## Summary

~Depends on #6629~

In order to be able to run tests in parallel, we need to be able to ensure a relayer in a test is only watching events on a specific channel / test.

In order to do this for hermes, we can configure packet filtering [ref](https://github.com/informalsystems/hermes/blob/master/config.toml#L380).

In our tests, we would need to modify the config file before starting the relayer. The config struct in interchaintest is [here](https://github.com/cosmos/ibc-go/blob/67b04280c8fc356e80e69e70fead921cf74b1494/modules/apps/27-interchain-accounts/controller/keeper/relay.go#L45-L47)

It may need to be updated to have these new fields.

It may be possible instead to perform a raw modification of the bytes of the config file without depending on a struct with [this](https://github.com/strangelove-ventures/interchaintest/blob/main/relayer/docker.go#L152) function. 

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #6685, opened 2024-06-24
Configure relayers to watch only channels associated with an individual test

chatton:
## Description

This PR replaces the single relayer associated with a test suite, and replaces it with a pool of relayers so each test can have its own.

At the moment, relayers need to be specified before the chain is created. For this reason it does not seem possible to lazily create a relayer when a test needs one. I created [this issue](https://github.com/strangelove-ventures/interchaintest/issues/1153) which when resolved, we should be able to create relayers as needed, rather than pre-loading them.

Each of these relayers has been configured to only watch the channel(s) for a single test with the caveat that it is only configured for hermes (no-op for other relayers currently)

The tests will still be running 1 test per host, however the relayer will be unique to that test, as well as configured to watch only the channels created for that test.

There is an exception for ICA channels at the moment, a wild card is used which means some additional work will need to be done to update the packet filter while the relayer is running. This might be tricky as I believe we only get the ica channelID after the relayer has performed the channel handshake :thinking: 

closes: #6634

ref: https://github.com/strangelove-ventures/interchaintest/issues/1153

---

write a little note why.


e2e/relayer/relayer.go
```diff
@@ -51,6 +57,87 @@ func New(t *testing.T, cfg Config, logger *zap.Logger, dockerClient *dockerclien
...
+		// [chains.packet_filter]
+		//	# policy = 'allow'
+		//	# list = [
+		//	#   ['ica*', '*'],
+		//	#   ['transfer', 'channel-0'],
+		//	# ]
+
+		// TODO(chatton): explicitly enable watching of ICA channels
+		// this will ensure the ICA tests pass, but this will need to be modified to make sure
+		// ICA tests will succeed in parallel.
```
chatton, line 102: I won't create another issue for this since it will be implicitly required to fix before our tests will pass in parallel

e2e/relayer/relayer.go
```diff
@@ -51,6 +57,87 @@ func New(t *testing.T, cfg Config, logger *zap.Logger, dockerClient *dockerclien
...
+			"policy": "allow",
+			"list":   channelIDs,
+		}
+
+		return nil
+	})
+}
+
+// modifyHermesConfigFile reads the hermes config file, applies a modification function and returns an error if any.
+func modifyHermesConfigFile(ctx context.Context, h *hermes.Relayer, modificationFn func(map[string]interface{}) error) error {
```
gjermundgaraba, line 116: Not sure if it would be any better or not, but interchaintest has some utility functions for modifying Toml files:
https://github.com/strangelove-ventures/interchaintest/blob/main/testutil/toml.go

gjermundgaraba (approved):
This is a great improvement for the e2e tests!

e2e/testsuite/testsuite.go
```diff
@@ -208,20 +229,17 @@ func (s *E2ETestSuite) SetupPath(clientOpts ibc.CreateClientOptions, channelOpts
...
-		channelsB, err := r.GetChannels(ctx, s.GetRelayerExecReporter(), chainB.Config().ChainID)
-		s.Require().NoError(err)
+		for _, c := range []ibc.Chain{chainA, chainB} {
+			channels, err := r.GetChannels(ctx, s.GetRelayerExecReporter(), c.Config().ChainID)
+			s.Require().NoError(err)
 
-		if s.channels[s.T().Name()] == nil {
-			s.channels[s.T().Name()] = make(map[ibc.Chain][]ibc.ChannelOutput)
-		}
+			s.channels[s.T().Name()][c] = channels
```
chatton, line 238: just realised this is wrong! We are currently fetching all channels associated with the chain, not just the one that was created. It means every relayer is still watching all channels, we need to identify the exact channel for this test. I think it is `channels[len(channels)-1]`

e2e/relayer/relayer.go
chatton: good call, I think I reinvented the wheel here!

chatton: https://github.com/strangelove-ventures/interchaintest/issues/1154

damiannolan (approved):
Awesome work! Thanks for diving into the depths again! 🫡 

Happy to approve and move forward with this if it will improve things for us wrt to runners 🙏🏻

e2e/relayer/relayer.go
```diff
@@ -51,6 +57,87 @@ func New(t *testing.T, cfg Config, logger *zap.Logger, dockerClient *dockerclien
...
+				chain = c
+				break
+			}
+		}
+
+		if chain == nil {
+			return fmt.Errorf("failed to find chain with id %s", chainID)
+		}
+
+		var channelIDs [][]string
```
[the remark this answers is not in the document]
chatton: yeah good call, it's not actually the ids :D

e2e/testsuite/testconfig.go
```diff
@@ -557,6 +558,9 @@ func DefaultChainOptions() ChainOptions {
 
 	return ChainOptions{
 		ChainSpecs: []*interchaintest.ChainSpec{chainASpec, chainBSpec},
+		// arbitrary number that will not be required if https://github.com/strangelove-ventures/interchaintest/issues/1153 is resolved.
+		// It can be overridden in individual test suites in SetupSuite if required.
+		RelayerCount: 10,
```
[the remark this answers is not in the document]
chatton: yeah, we definitely can improve this arbitrary number.

Issue #6561, opened 2024-06-10
Use a transfer module account as an intermediary address for forwards


DimitrisJim:
Reference code:

https://github.com/cosmos/ibc-go/blob/efcfa5d877567731f6b4ff8dbd063a34ede5de00/modules/apps/transfer/types/keys.go#L86-L97

Surfaced during internal walkthrough of ics20 v2 forwarding (commit [efcfa5d](https://github.com/cosmos/ibc-go/tree/efcfa5d877567731f6b4ff8dbd063a34ede5de00))

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

srdtrk:
I recommend we switch to a module account. In SDK v0.50, bank keeper's [`BurnCoins`](https://github.com/cosmos/cosmos-sdk/blob/v0.50.7/x/bank/keeper/keeper.go#L47) method only works on module accounts. This is [extended to any account](https://github.com/cosmos/cosmos-sdk/blob/78327f40c5ad6e31a056c662e2046cbf1b6ed691/x/bank/keeper/keeper.go#L48) in the next version of the SDK.

The simplifications I'm doing in https://github.com/cosmos/ibc-go/issues/6558 would benefit from the ability to use `BurnCoins` directly

colin-axner:
Yes we should use SDK module accounts (for forwarding escrow) :+1: 

Context on why transfer doesn't use module accounts already is that transfer was developed before module accounts was a feature. Migrating all escrow addresses is quite sensitive (and not necessarily beneficial given that the current construction is safe). But this historical baggage should not inform our current decisions when possible. We also use module accounts in [ics29](https://github.com/cosmos/ibc-go/blob/9a34c317cf31130c3b2cb2dfe069b4f600f49088/modules/apps/29-fee/keeper/escrow.go#L29)

colin-axner:
Did some extra thinking about needing to create isolated buckets based on the hop port/channel, but it doesn't seem to provide any security benefits. Transfer escrow addresses being indexed by destination port/channel is useful since it creates fault tolerance. If connection to chainB goes malicious, chainB cannot steal funds in escrow for chainC

Given that the only funds moving through the forward address go through the normal transfer logic, I don't see how a malicious chain could generate some sequence of steps to steal from funds deposited through the forward address. In addition, funds should never be stored in the forward address after the transfer message completes execution. 

[...]

PR #6688, opened 2024-06-24
chore: use module account instead of custom forward address

gjermundgaraba:
## Description

These changes triggered some issues with the transfer module account being blocked by the bank transfers. I solved this by using bankkeeper.SendCoins directly and do a blocked check where used instead.

I have two questions:
1. Would it make sense to put the module addr on the transfer keeper itself to avoid having to look it up all the time?
2. Should the module account + block address check be done in a convenience method instead to avoid all the extra checks everywhere?

closes: #6561

---

write a little note why.


modules/apps/transfer/keeper/relay.go
```diff
@@ -209,7 +216,7 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 
 			coin := sdk.NewCoin(token.Denom.IBCDenom(), transferAmount)
 
-			if k.bankKeeper.BlockedAddr(receiver) {
+			if !receiver.Equals(moduleAddr) && k.bankKeeper.BlockedAddr(receiver) {
```
chatton, line 219: yeah I think something like a `k.IsBlockedAddr(addr)` or `k.IsAllowedToSend(addr)` to handle this check

modules/apps/transfer/keeper/forwarding.go
```diff
@@ -19,8 +19,8 @@ func (k Keeper) forwardPacket(ctx sdk.Context, data types.FungibleTokenPacketDat
 		nextForwardingPath = types.NewForwarding(false, data.Forwarding.Hops[1:]...)
 	}
 
-	// sending from the forward escrow address to the original receiver address.
-	sender := types.GetForwardAddress(packet.DestinationPort, packet.DestinationChannel)
+	// sending from module account (used as a temporary forward escrow) to the original receiver address.
+	sender := k.authKeeper.GetModuleAddress(types.ModuleName)
```
chatton, line 23: looks like `k.authKeeper.GetModuleAddress(types.ModuleName)` is already a map lookup, are you thinking of something like `k.GetModuleAddress()` as a utility to delegate to the authKeeper?

modules/apps/transfer/keeper/relay.go
```diff
@@ -128,8 +129,9 @@ func (k Keeper) sendTransfer(
 			labels = append(labels, telemetry.NewLabel(coretypes.LabelSource, "false"))
 
 			// transfer the coins to the module account and burn them
-			if err := k.bankKeeper.SendCoinsFromAccountToModule(
-				ctx, sender, types.ModuleName, sdk.NewCoins(coin),
+			moduleAccountAddr := k.authKeeper.GetModuleAddress(types.ModuleName) // TODO: Change
+			if err := k.bankKeeper.SendCoins(
+				ctx, sender, moduleAccountAddr, sdk.NewCoins(coin),
```
chatton, line 134: what's the purpose of this change?

chatton (commented):
looking good! I think we should definitely make some sort of helper for the checking of the address, would lean towards a helper fn that delegates to the auth keeper to fetch the address rather than storing a field.

modules/apps/transfer/keeper/relay.go
gjermundgaraba: Yeah, I don't think this was actually needed, will remove

modules/apps/transfer/keeper/relay.go
gjermundgaraba: Done

modules/apps/transfer/keeper/forwarding.go
chatton: happy to have it called directly in the utility fn!

chatton (approved):
LGTM! Seems a bit cleaner using a module address like this :muscle:

modules/apps/transfer/keeper/keeper.go
```diff
@@ -336,3 +336,14 @@ func (k Keeper) deleteForwardedPacket(ctx sdk.Context, portID, channelID string,
...
 	store.Delete(packetKey)
 }
+
+// IsBlockedAddr checks if the given address is allowed to send or receive tokens.
+// The module account is always allowed to send and receive tokens.
+func (k Keeper) IsBlockedAddr(addr sdk.AccAddress) bool {
+	moduleAddr := k.authKeeper.GetModuleAddress(types.ModuleName)
+	if !addr.Equals(moduleAddr) && k.bankKeeper.BlockedAddr(addr) {
+		return true
+	}
```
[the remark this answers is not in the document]
gjermundgaraba: Ah yes, that is much cleaner! Will do 👍

Issue #6696, opened 2024-06-25
Comments after alpha audit of path forwarding


crodriguezvega:
From internal alpha audit of ICS20 v2 path forwarding:

- [ ] Add a comment in `unwindHops` where we loop over the trace elements except the first one.
- [ ] Move all forwarding validation together to the same helper function if possible (including the check for empty source port and channel if `unwind` is true).
- [ ] Run packet data validation after it is constructed in `createPacketDataBytesFromVersion` both for V1 and V2. Change signature of function to also return `error` .
- [ ] Pass only `hops` to `sendTransfer`. Change `EmitTransferEvent` to accept only hops as well.
- [ ] `getReceiverFromPacketData` switch around the if statements and add comment about why we do that: If the forwarding is set, then do not process the data.Receiver field and instead just shortcircuit to create receiver from forwardingInfo
- [ ] Switch to if-else blocks in OnRecvPacket instead of using `continue`
- [ ] Maybe replace `SenderChainIsSource` and `ReceiverChainIsSource` with function `IsTracePrefixed` (like the one used in the ICS20 spec). Open a PR and see which one is easier to reason about.
- [ ] Add hops to the ack error while propogating back to source so we know which chain the packet failed or timed out on [could also do counter]. Include also the deterministic error string from the error acknowledgement written by the chain where the forwarding failed.
- [ ] Authz:
  - Allow explicit list of allowed hops. Do not allow unwind boolean
  - Use `repeated AllowedForwarding allowed_forwarding` with `AllowedForwarding` with `repeated Hop hops` . We choose to go with this option. Granter will have to specify the list of hops to unwind.
- [ ] Replace deep equal in authz with something else. Check if `slices.Contains` works to get rid of the loop

Thank you @chatton for leading the walkthrough and the rest of the team for the feedback.

damiannolan:
> Add hops to the ack error while propogating back to source so we know which chain the packet failed or timed out on [could also do counter]. Include also the deterministic error string from the error acknowledgement written by the chain where the forwarding failed

[...]

PR #6705, opened 2024-06-25
imp: replace `Sender/ReceiverChainIsSource` with `HasPrefix`

crodriguezvega:
## Description

ref: #6696

---

write a little note why.


modules/apps/transfer/keeper/relay.go
```diff
@@ -115,7 +115,7 @@ func (k Keeper) sendTransfer(
 		// chain inside the packet data. The receiving chain will perform denom
 		// prefixing as necessary.
 
-		if token.Denom.SenderChainIsSource(sourcePort, sourceChannel) {
+		if !token.Denom.HasPrefix(sourcePort, sourceChannel) {
```
colin-axner, line 118: I think if we do this change, we should flip flop the if statement. I think there's a bit of unnecessary mental gymnastics to figure out what a non prefix set means. I feel like it's easier to write a comment which states:
```go
// if the port and channel we are sending the token on was prefixed to this denom
// then we must be returning the token back to the chain they originated from
```
then I think the burn coins logic is more natural

modules/apps/transfer/keeper/relay.go
```diff
@@ -201,7 +201,7 @@ func (k Keeper) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet, data t
 		// NOTE: We use SourcePort and SourceChannel here, because the counterparty
 		// chain would have prefixed with DestPort and DestChannel when originally
 		// receiving this coin as seen in the "sender chain is the source" condition.
-		if token.Denom.ReceiverChainIsSource(packet.GetSourcePort(), packet.GetSourceChannel()) {
+		if token.Denom.HasPrefix(packet.GetSourcePort(), packet.GetSourceChannel()) {
```
colin-axner, line 204: I personally don't feel like this reads easier, but I'm okay with the change. I think either restructing the comments or flip flopping the if statements could make it better, but as is requires me to reparse all the logic to figure out which chain is source and which port and channel to be passing in

colin-axner: maybe you could update the comment:
```go
// if the token we are receiving has the packet source port and channel prefixed
// then the sending chain must have prefixed this hop to the token when it originally
// received these tokens from us. Given that we are receiving back the tokens
// we should remove the prefix added by the counterparty and unescrow the tokens. 
```

I think the previous comment is more relevant to the previous function and even references it ""sender chain is source" condition"

modules/apps/transfer/keeper/relay.go
```diff
@@ -347,7 +347,7 @@ func (k Keeper) refundPacketTokens(ctx sdk.Context, packet channeltypes.Packet,
 			return err
 		}
 
-		if token.Denom.SenderChainIsSource(packet.GetSourcePort(), packet.GetSourceChannel()) {
```
colin-axner, line 350: I have a feeling that if we are changing to `HasPrefix`, the logic will always be easier to parse if the first if statement is not negated

crodriguezvega:
@colin-axner I think I have addressed your comments. Please take a look when you have time. I agree it makes sense to not negate the condition. :)

modules/apps/transfer/keeper/relay.go
```diff
@@ -356,29 +357,29 @@ func (k Keeper) refundPacketTokens(ctx sdk.Context, packet channeltypes.Packet,
...
+		// if the token we must refund is prefixed by the source port and channel
+		// then the tokens were burnt on the sending chain and we must mint new tokens
```
colin-axner, line 361:
```suggestion
		// then the tokens were burnt when the packet was sent and we must mint new tokens
```

colin-axner (approved):
Thank you for updating the if else, I think it's more natural to reason about

chatton (approved):
LGTM! This reads a good bit better now i think, thanks for the cleanup :muscle:

PR #6716, opened 2024-06-27
feat: add a method to check if the scope module already exists in the capability keeper

Pantani:
## Description

Add method `HasModule(moduleName string)` to check if the scope module already exists in the capability keeper. This method can help avoid creating two scope keepers by mistake and panicking the chain, and it helps the workaround to make IBC work with the app wiring chains.


damiannolan (approved):
Thanks @Pantani! 🙏🏻

modules/capability/keeper/keeper_test.go
```diff
@@ -76,6 +76,12 @@ func (suite *KeeperTestSuite) TestSeal() {
 	})
 }
 
+func (suite *KeeperTestSuite) TestHasModule() {
+	_ = suite.keeper.ScopeToModule(bankModuleName)
+	suite.Require().True(suite.keeper.HasModule(bankModuleName), "bank module not exist")
+	suite.Require().False(suite.keeper.HasModule("invalid"), "invalid module exist")
```
bznein, line 82:
```suggestion
	suite.Require().True(suite.keeper.HasModule(bankModuleName), "bank module does not exist")
	suite.Require().False(suite.keeper.HasModule("invalid"), "invalid module exists")
```

bznein (approved):

PR #6748, opened 2024-07-03
(test): generate different channelIDs in testing.

bznein:
## Description

closes: #4062

Rationale: I've worked on this with @chatton and we decided to go with having a global variable that is incremented everytime `Endpoint.ChanOpenInit` and `Endpoint.ChanOpenTry` are called. In this way we have unique channel IDs everywhere in tests.

This broke several tests that just relied on the fact that all channels are equal. We went on and fixed them, which mostly meant one of the following:

* Replacing `ibctesting.FirstChannelID` with the proper channel ID (usually `pathX.EndpointY.ChannelID`)
* Fixed wrong path endpoint used
* Fixed wrong path used

Plus a couple more complex cases, which I commented on the PR itself.

There are also a few tests that take as input some JSON files with hardcoded values. We couldn't find a neat way to fix those, so we added an opt-out function to rely on the old behaviour, that is currently only used for those tests. We can discuss in the engineering call how to deal with them (they are all very old, mostly untouched since 2021, so we'll need someone with context on them)

---

write a little note why.


testing/endpoint.go
```diff
@@ -323,8 +325,19 @@ func (endpoint *Endpoint) QueryConnectionHandshakeProof() (
...
+var sequenceNumber int
+
+func (endpoint *Endpoint) IncrementNextChannelSequence() {
+	if endpoint.disableUniqueChannelIDs {
+		return
+	}
+	sequenceNumber++
+	endpoint.Chain.App.GetIBCKeeper().ChannelKeeper.SetNextChannelSequence(endpoint.Chain.GetContext(), uint64(sequenceNumber))
+}
+
```
bznein, line 337: This is the core of the change. We force the increment of the `NextChannelSequence` entry so that every chan has a different ID.

testing/path.go
```diff
@@ -66,6 +66,12 @@ func (path *Path) SetChannelOrdered() {
 	path.EndpointB.ChannelConfig.Order = channeltypes.ORDERED
 }
 
+func (path *Path) DisableUniqueChannelIDs() *Path {
+	path.EndpointA.disableUniqueChannelIDs = true
+	path.EndpointB.disableUniqueChannelIDs = true
+	return path
+}
+
```
bznein, line 74: This is the opt-out function, currently used by only one test suite, and ideally it should not be used anymore.

modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
```diff
@@ -88,6 +88,8 @@ func NewICAPath(chainA, chainB *ibctesting.TestChain, encoding string, ordering
 
 // SetupICAPath invokes the InterchainAccounts entrypoint and subsequent channel handshake handlers
 func SetupICAPath(path *ibctesting.Path, owner string) error {
+	path.EndpointA.IncrementNextChannelSequence()
```
bznein, line 91: `SetupICAPath`  manually calls `chanOpenTry` but not `chanOpenInit`, so we force here the increase of the sequence number.

modules/apps/29-fee/keeper/grpc_query_test.go
```diff
@@ -201,9 +203,17 @@ func (suite *KeeperTestSuite) TestQueryIncentivizedPacketsForChannel() {
...
 					QueryHeight: 0,
 				}
+				identifiedFees1 := types.NewIdentifiedPacketFees(channeltypes.NewPacketID(ibctesting.MockFeePort, path.EndpointA.ChannelID, 1), packetFees.PacketFees)
+				identifiedFees2 := types.NewIdentifiedPacketFees(channeltypes.NewPacketID(ibctesting.MockFeePort, path.EndpointA.ChannelID, 2), packetFees.PacketFees)
+				identifiedFees3 := types.NewIdentifiedPacketFees(channeltypes.NewPacketID(ibctesting.MockFeePort, path.EndpointA.ChannelID, 3), packetFees.PacketFees)
+
+				expIdentifiedPacketFees = append([]*types.IdentifiedPacketFees{}, &identifiedFees1, &identifiedFees2, &identifiedFees3)
+				for _, identifiedPacketFees := range expIdentifiedPacketFees {
+					suite.chainA.GetSimApp().IBCFeeKeeper.SetFeesInEscrow(suite.chainA.GetContext(), identifiedPacketFees.PacketId, types.NewPacketFees(identifiedPacketFees.PacketFees))
+				}
```
bznein, line 216: This is a bit ugly, but down in the main test logic we use `FirstChannelID` and we can't do it now as the channelID depends on the path being setup

modules/apps/transfer/ibc_module_test.go
```diff
@@ -392,16 +394,16 @@ func (suite *TransferTestSuite) TestOnRecvPacket() {
 			seq := uint64(1)
 			packet = channeltypes.NewPacket(packetData.GetBytes(), seq, path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID, path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID, clienttypes.ZeroHeight(), suite.chainA.GetTimeoutTimestamp())
 
-			module, _, err := suite.chainA.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(suite.chainA.GetContext(), ibctesting.TransferPort)
+			ctx := suite.chainB.GetContext()
+			module, _, err := suite.chainB.App.GetIBCKeeper().PortKeeper.LookupModuleByPort(ctx, ibctesting.TransferPort)
 			suite.Require().NoError(err)
 
-			cbs, ok := suite.chainA.App.GetIBCKeeper().PortKeeper.Route(module)
+			cbs, ok := suite.chainB.App.GetIBCKeeper().PortKeeper.Route(module)
```
bznein, line 401: we should be receiving on chainB, not chainA (see line 395 where we set the packet source and dest)

modules/apps/transfer/keeper/mbt_relay_test.go
```diff
@@ -309,8 +309,8 @@ func (suite *KeeperTestSuite) TestModelBasedRelay() {
 		}
 
 		suite.SetupTest()
-		pathAtoB := ibctesting.NewTransferPath(suite.chainA, suite.chainB)
-		pathBtoC := ibctesting.NewTransferPath(suite.chainB, suite.chainC)
+		pathAtoB := ibctesting.NewTransferPath(suite.chainA, suite.chainB).DisableUniqueChannelIDs()
+		pathBtoC := ibctesting.NewTransferPath(suite.chainB, suite.chainC).DisableUniqueChannelIDs()
```
bznein, line 313: Opt-out from this new behaviour as we weren't able to find a nice way to fix these

modules/core/04-channel/keeper/upgrade_test.go
```diff
@@ -2715,14 +2717,18 @@ func (suite *KeeperTestSuite) TestWriteErrorReceipt() {
...
 		},
 		{
 			"failure: channel not found",
 			func() {
 				suite.chainA.DeleteKey(host.ChannelKey(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID))
 			},
-			errorsmod.Wrap(types.ErrChannelNotFound, "port ID (mock) channel ID (channel-0)"),
+			func() error {
+				return errorsmod.Wrap(types.ErrChannelNotFound, fmt.Sprintf("port ID (mock) channel ID (%s)", path.EndpointA.ChannelID))
+			},
```
bznein, line 2731: I don't particularly like changing this to a function, but I needed a way to be able to access `path.EndpointA.ChannelID` and in this way the function gets evaluated after `path.setup` has been called

modules/apps/29-fee/keeper/grpc_query_test.go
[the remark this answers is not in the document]
bznein: Thanks, I made it much cleaner now!

modules/apps/transfer/keeper/grpc_query_test.go
```diff
@@ -247,7 +248,7 @@ func (suite *KeeperTestSuite) TestEscrowAddress() {
 			func() {
 				req = &types.QueryEscrowAddressRequest{
 					PortId:    ibctesting.TransferPort,
-					ChannelId: ibctesting.FirstChannelID,
```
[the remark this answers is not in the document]
bznein: This is still used in a lot of places. And I think in some of them still represents a valid identifier. I only changed tests that broke with the new functionality. I agree we should revisit it but it might be a longer work than it seems.

modules/apps/29-fee/keeper/grpc_query_test.go
```diff
@@ -690,17 +700,18 @@ func (suite *KeeperTestSuite) TestQueryFeeEnabledChannels() {
...
+					channelID := channeltypes.FormatChannelIdentifier(uint64(i + int(nextSeq)))
 					suite.chainA.GetSimApp().IBCFeeKeeper.SetFeeEnabled(suite.chainA.GetContext(), ibctesting.MockFeePort, channelID)
 
 					expChannel := types.FeeEnabledChannel{
 						PortId:    ibctesting.MockFeePort,
 						ChannelId: channelID,
 					}
 
-					if i < 5 { // add only the first 5 channels, as our default pagination limit is 5
+					if i < 4 { // add only the first 5 channels, as our default pagination limit is 5
```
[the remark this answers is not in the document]
bznein: I was equally puzzled :)

crodriguezvega (approved):
Great work, @bznein and @chatton! Scary to see that a good amount of tests needed to be fixed to work with the changes, though...  😓

modules/apps/transfer/keeper/grpc_query_test.go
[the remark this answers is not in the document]
bznein: I created #6765 , please feel free to edit to make it clearer, I wasn't sure how to phrase it. I haven't done any estimation yet on how much works is needed

testing/endpoint.go
```diff
@@ -35,6 +35,8 @@ type Endpoint struct {
 	ClientConfig     ClientConfig
 	ConnectionConfig *ConnectionConfig
 	ChannelConfig    *ChannelConfig
+
+	disableUniqueChannelIDs bool
```
chatton, line 39: maybe no harm to add a docstring saying what this is.

chatton (approved):
Looking great! Thanks for digging into all this stuff, adds a lot of confidence now in writing future tests that they will fail if we mis-wire them :muscle:

testing/endpoint.go
```diff
@@ -323,8 +328,22 @@ func (endpoint *Endpoint) QueryConnectionHandshakeProof() (
...
+var sequenceNumber int
+
+// IncrementNextChannelSequence incrementes the value "nextChannelSequence" in the store,
+// which is used to determine the next channel ID.
+// This guarantees that we'll have always different IDs while running tests.
+func (endpoint *Endpoint) IncrementNextChannelSequence() {
+	if endpoint.disableUniqueChannelIDs {
+		return
+	}
+	sequenceNumber++
```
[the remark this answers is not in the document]
chatton: We actually thought of a few different approaches like this. Initializing B at some higher value, maintaining some mapping on the endpoints or paths (since channel IDs need to just be unique between chains on a single path within the test) etc.

For example something like this

```golang
func NewPath(chainA, chainB *TestChain) *Path {
	endpointA := NewDefaultEndpoint(chainA)
	endpointB := NewDefaultEndpoint(chainB)

	// initialize chainB with a higher sequence number to avoid duplicate channel IDs within tests.
	chainB.App.GetIBCKeeper().ChannelKeeper.SetNextChannelSequence(chainB.GetContext(), uint64(1000))

	endpointA.Counterparty = endpointB
	endpointB.Counterparty = endpointA

	return &Path{
		EndpointA: endpointA,
		EndpointB: endpointB,
	}
}

```

But this ends up failing on this test https://github.com/cosmos/ibc-go/blob/b61a35a4d25f6c71983db814ab03ab8f91efc349/modules/core/keeper/msg_server_test.go#L1587-L1624

because of the re-use of channel-1000

(a bunch of tests fail, this is just used as an example)

This test passes if we go back to using a global to track number of paths created

```golang
const channelOffset = 100

var createdPaths = 0

// NewPath constructs an endpoint for each chain using the default values
// for the endpoints. Each endpoint is updated to have a pointer to the
// counterparty endpoint.
func NewPath(chainA, chainB *TestChain) *Path {
	createdPaths++

	endpointA := NewDefaultEndpoint(chainA)
	endpointB := NewDefaultEndpoint(chainB)

	// initialize chainB with a higher sequence number to avoid duplicate channel IDs within tests.
	chainB.App.GetIBCKeeper().ChannelKeeper.SetNextChannelSequence(chainB.GetContext(), uint64(createdPaths*channelOffset))

	endpointA.Counterparty = endpointB
	endpointB.Counterparty = endpointA

	return &Path{
		EndpointA: endpointA,
		EndpointB: endpointB,
	}
}
```

This one actually does pass all but the reworked happy path case of `TestQueryIncentivizedPacketsForChannel`  (and the ModelBasedRelay tests )

I think we can remove the potential issues of this number going out of sync if we modify it here right? 

We still need the package level variables AFAIK unless we want to re-work effected tests.

I think a package level variable here that is used in NewPath is safer than one used the way it is now in the endpoint helper fns.

WDYT @colin-axner  / @bznein

Issue #6642, opened 2024-06-19
Add v8 to v9 upgrade tests


crodriguezvega:
We should add an e2e chain upgrade test for v8 to v9 to check that the automatic migrations execute fine and make sure everything is set correctly. This could be done in two different tests (and PRs):

- [x] Test a chain upgrade from v8 to v9 and then check that migration [`MigrateDenomTraceToDenom`](https://github.com/cosmos/ibc-go/blob/9a34c317cf31130c3b2cb2dfe069b4f600f49088/modules/apps/transfer/keeper/migrations.go#L76) runs and works as expected.
- [ ] Test a chain upgrade from v8 to v9 and then run a channel upgrade from `ics20-1` to `ics20-2` and do a transfer for multi-denom (do people think we should also do a forwarding transfer?)

PR #6791, opened 2024-07-09
test: e2e upgrade tests for v8 -> v9

gjermundgaraba:
## Description

Related to #6642

First two tests for v8->v9 upgrade

---

write a little note why.


e2e/testsuite/query/grpc_query.go
```diff
@@ -18,6 +18,11 @@ func GRPCQuery[T any](ctx context.Context, chain ibc.Chain, req proto.Message, o
 		return nil, err
 	}
 
+	return GRPCQueryWithMethod[T](ctx, chain, req, path, opts...)
+}
+
+// GRPCQueryWithMethod queries the chain with a query request with a specific method (grpc path) and deserializes the response to T
+func GRPCQueryWithMethod[T any](ctx context.Context, chain ibc.Chain, req proto.Message, method string, opts ...grpc.CallOption) (*T, error) {
```
gjermundgaraba, line 25: I had to add this, because the v2 query path is not standard and I think it is better to send in the path rather than have a bunch of ifs here.

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -80,14 +86,17 @@ func (s *UpgradeTestSuite) UpgradeChain(ctx context.Context, chain *cosmos.Cosmo
...
-	height, err := chain.Height(ctx)
-	s.Require().NoError(err, "error fetching height before upgrade")
-
 	timeoutCtx, timeoutCtxCancel := context.WithTimeout(ctx, time.Minute*2)
 	defer timeoutCtxCancel()
 
-	err = test.WaitForBlocks(timeoutCtx, int(haltHeight-height)+1, chain)
-	s.Require().Error(err, "chain did not halt at halt height")
+	err = test.WaitForCondition(time.Minute*2, time.Second*2, func() (bool, error) {
+		status, err := chain.GetNode().Client.Status(timeoutCtx)
```
chatton, line 93: unsure if we need to pass in the `timeoutCtx` to the `Status` call, if I understand it correctly, it means each status call will have 2 minutes to execute but the `WaitForCondition` fn constructs a context based on the polling interval and poll period provided.

Don't think it actually matters, but I think this could just be a `context.TODO()` instead.

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -1031,6 +1040,251 @@ func (s *UpgradeTestSuite) TestV8ToV8_1ChainUpgrade_ChannelUpgrades() {
 	})
 }
 
+func (s *UpgradeTestSuite) TestV8ToV9ChainUpgrade() {
```
chatton, line 1043: not something for today, but it seems like it could be a good idea to parameterize these upgrade tests. It looks like a lot of them are 90% the same.

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -1031,6 +1040,251 @@ func (s *UpgradeTestSuite) TestV8ToV8_1ChainUpgrade_ChannelUpgrades() {
...
+
+	s.Require().NoError(test.WaitForBlocks(ctx, 5, chainA), "failed to wait for blocks")
+
+	t.Run("upgrade chain", func(t *testing.T) {
+		govProposalWallet := s.CreateUserOnChainA(ctx, testvalues.StartingTokenAmount)
+		s.UpgradeChain(ctx, chainA.(*cosmos.CosmosChain), govProposalWallet, testCfg.UpgradeConfig.PlanName, testCfg.ChainConfigs[0].Tag, testCfg.UpgradeConfig.Tag)
+	})
+
+	t.Run("query denoms after upgrade", func(t *testing.T) {
+		resp, err := query.GRPCQueryWithMethod[transfertypes.QueryDenomsResponse](ctx, chainA, &transfertypes.QueryDenomsRequest{}, "/ibc.applications.transfer.v2.QueryV2/Denoms")
```
chatton, line 1105: I wonder, do you think it's better to create some sort of wrapper function or at least extract a const of `"/ibc.applications.transfer.v2.QueryV2/Denoms"`. Feel like this could get cumbersome if used in other tests.

Maybe even make the `GRPCQueryWithMethod` private and provide a wrapper. 

I don't suspect we will use this method heavily anyway, so happy to leave as is if you don't think it's a big deal.

e2e/tests/upgrades/upgrade_test.go
```diff
@@ -1031,6 +1040,251 @@ func (s *UpgradeTestSuite) TestV8ToV8_1ChainUpgrade_ChannelUpgrades() {
...
+
+func (s *UpgradeTestSuite) TestV8ToV9ChainUpgrade_Localhost() {
+	t := s.T()
+	testCfg := testsuite.LoadConfig()
+	ctx := context.Background()
+
+	testName := t.Name()
+
+	_, channelA := s.CreateUpgradeTestPath(testName)
+	channelVersion := channelA.Version
```
chatton, line 1133: do we need to actually create a path? I believe we just need a version. Can we just do

`channelVersion := transfertypes.V2`

chatton (commented):
upgrade tests are looking great! 

We need to make sure we add them to [this](https://github.com/cosmos/ibc-go/blob/65f0ec7749f0b585aede5018acf95e42c002924c/.github/workflows/e2e-upgrade.yaml#L162-L161) workflow though.

Currently they are running on this PR because the upgrade test files have changed, but if we want to explicitly trigger upgrade tests, we'll need to add them to that file also!

e2e/tests/upgrades/upgrade_test.go
gjermundgaraba: Haha, you are of course right :D Will fix that

e2e/tests/upgrades/upgrade_test.go
gjermundgaraba: Yeah, wrapper seems like a better option! Will change to that and use const so we can easily just add more if we ever need to query the other v2 queries.

e2e/tests/upgrades/upgrade_test.go
gjermundgaraba: Oh yeah, you are right. It was left over from the previous version there, it is not necessary anymore.

e2e/tests/upgrades/upgrade_test.go
chatton: one of the things I've been thinking about a bit is basically extracting re-usable fns from our existing E2Es and making the actual test cases super short, something like

```golang
   PerformTransferV1(chainA, chainB)  
   UpgradeChain(chainA, newVersion)
   PerformTransferV2(chainA, chainB)
```

Think it will be possible to reduce a huge amount of duplication in the E2Es.

gjermundgaraba:
> upgrade tests are looking great!
> 
> We need to make sure we add them to [this](https://github.com/cosmos/ibc-go/blob/65f0ec7749f0b585aede5018acf95e42c002924c/.github/workflows/e2e-upgrade.yaml#L162-L161) workflow though.
> 
> Currently they are running on this PR because the upgrade test files have changed, but if we want to explicitly trigger upgrade tests, we'll need to add them to that file also!

I've added them now, let me know if it looks correct

chatton (approved):
LGTM! Great job with these :muscle:

damiannolan (approved):
Nice, good stuff! 🙏🏻

Issue #3801, opened 2023-06-09
update method signatures of `verify` functions in 03-connection for readability


charleenfei:
## Summary of Bug

some of the methods (`VerifyNextSequenceRecv`, `VerifyPacketReceiptAbsence`, `VerifyPacketAcknowledgement`, `VerifyPacketCommitment`, `VerifyChannelState`) in `03-connection` `verify.go` have the `portID` and `channelID` params ordered at the end. 

we should try to be consistent by having `portID` and `channelID` after the `ctx` param, for readability
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

DimitrisJim:
aren't these currently part of our public API? I agree with the issue (and I've found other places where, e.g unused vars exist) but have reservations because in theory someone might be depending on these.

charleenfei:
You make a good point. I guess channel upgradeability would anyway have other API breaking changes though, so it could all go into the release...but maybe you're right that it's worth looking into if other projects might be using this proof verification logic.

colin-axner:
These functions are used for the core IBC handlers. There is a world in which they are not public, so I think it'd be okay to make these changes along with a major version release, but I agree, it isn't high priority. If a release has a lot of breaking changes, it might make sense not to bundle it into that release (the more changes, the more headaches our users will have)

chandiniv1:
@charleenfei can I work on this issue?

crodriguezvega:
See [here](https://github.com/cosmos/ibc-go/pull/6896#issuecomment-2242345530) for reasoning to close the issue.

PR #6896, opened 2024-07-19
update method signatures of verify functions

chandiniv1:
closes: #3801


gjermundgaraba (approved):

crodriguezvega:
We could merge and release in v9, although we have a lot of API-breaking changes already. But maybe these functions are not really that much used outside of ibc-go? However, with the IBC eureka/v10 release these functions might also change probably (and some of them, like the handshake and upgrade ones, might be completely removed), so alternatively we could also just merge to main and not back port to v9.0.x. What do people think?

crodriguezvega:
> fine with not backporting unless someone for some reason asks (which I doubt would be the case)

Yeah, I doubt it: this is a year-old issue and nobody has requested the changes so far, so...

crodriguezvega:
Before merging I will add migration docs though.

crodriguezvega (approved):
Thank you, @chandiniv1.

damiannolan:
Hey @chandiniv1, thank you very much for picking up this issue and opening the PR.

I'd like to share my thoughts on the original issue and proposed changes in https://github.com/cosmos/ibc-go/issues/3801.
I feel hesitant to move forward with these changes and generally disagree with the proposal to move the portID and channelID arguments. 

These state verification in 03-connection have the args listed like so:

```
func VerifyType(
  ctx: the sdk context, store access, block time/height etc
  connection: the connection which we are proving state through, gives access to client ID
  height: the height of the counterparty's consensus state root which we are proving against
  proof: proof bytes to calculate root
  keyPath arguments: e.g. connectionID or portID, channelID etc
  value bytes: leaf value bytes from tree node we are proving
)
```

In my opinion this becomes more of a problem, or creates a disconnect of these arguments when moving portID and channelID to the beginning of the argument list.

```golang
func (k *Keeper) VerifyPacketCommitment(
	ctx sdk.Context,
	connection types.ConnectionEnd,
	height exported.Height,
	proof []byte,
	portID,
	channelID string,
	sequence uint64,
	commitmentBytes []byte,
)
```

Take the function above as an example, if we move portID and channelID to the beginning of the argument list the the sequence part which makes up the keypath sits alone, and all of sudden things the argument list doesn't actually make that much sense.
I would recommend closing the issue and not moving forward with the PR.

crodriguezvega:
Following on @damiannolan's comment above, we just had a discussion internally and we agreed that it's best to close the linked issue and this PR. Thanks again for taking the time to work on this; I am sure we have other issues for which we would appreciate your contribution!

Issue #6971, opened 2024-07-28
Introduce Version field into the Packet struct


AdityaSripal:
Add a new field to allow applications to include a version string into the Packet struct.

We need to do this in a non-breaking way to ensure that it doesn't affect ibc-classic

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #6972, opened 2024-07-28
Implement new packet hashing for IBC Eureka


AdityaSripal:
The packet hashing must be changed to accomodate the IBC Eureka protocol. Any information that is prenegotiated through the channel handshake must be explicitly hashed into the packet commitment.

This includes the destination port and the version.

In `ibc-lite`, we hashed the destination port **and** destination channel for consistency.

We must also hash the version into the packet commitment. This of course requires the version to be included in the Packet, thus this issue is dependent on #6971

PR #6986, opened 2024-07-29
Add versions to packet and create Eureka packet commitment

AdityaSripal:
## Description

TODO: 

- [x] Use IBC Version to switch between the two hashing functions
- [ ] Get feedback from port router team on new hashing function
- [x] Merge into single entry point for packet msg handler
- [ ] Parameterize the commitment function. This should be done in a separate PR after some discussion and is not necessary for initial release

closes: #6971 #6972 

---

write a little note why.


proto/ibc/core/channel/v1/channel.proto
```diff
@@ -122,6 +129,10 @@ message Packet {
   ibc.core.client.v1.Height timeout_height = 7 [(gogoproto.nullable) = false];
   // block timestamp (in nanoseconds) after which the packet times out
   uint64 timeout_timestamp = 8;
+  // which IBC version to use to commit this packet
+  IBCVersion ibc_version = 9;
```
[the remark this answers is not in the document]
AdityaSripal: There are a couple reasons I like having the IBC Version in the packet.

One minor reason is that we could use the same client for Eureka and Classic. Of course, we could disable this if we choose to in other ways.

The major reason is that I think having these versions explicitly defined in the Packet will make it easier for us to upgrade in the future. If we have ProtocolVersion stored on a per-client basis, the entire chain and all applications on it have to make a synchronous decision to move to the next protocol. If it's a per-packet basis, one application can decide to upgrade if the sender supports the new version. If the receiver doesn't then it just doesn't get received and we time out. i think in general having version in the packet and being permissive to version-related failures to relay will make the system more easily upgradable over time than a prenegotiated version.

cc: @colin-axner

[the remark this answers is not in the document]
AdityaSripal: Also this is an enum that is not actually being hashed into the commitment. So it is really taking a small amount of space in the packet. The packet is also only in the transaction bytes, it is not stored on chain except for a couple edge-cases.

modules/core/04-channel/types/packet.go
```diff
@@ -2,22 +2,42 @@ package types
...
+	}
+}
+
+// CommitClassicPacket returns the classic packet commitment bytes. The commitment consists of:
 // sha256_hash(timeout_timestamp + timeout_height.RevisionNumber + timeout_height.RevisionHeight + sha256_hash(data))
 // from a given packet. This results in a fixed length preimage.
+// NOTE: A fixed length preimage is ESSENTIAL to prevent relayers from being able
+// to malleate the packet fields and create a commitment hash that matches the original packet.
 // NOTE: sdk.Uint64ToBigEndian sets the uint64 to a slice of length 8.
-func CommitPacket(cdc codec.BinaryCodec, packet Packet) []byte {
```
AdityaSripal, line 20: The codec was never used in a version of IBC. Think it was used in early pre-launch implementations but got changed before launch. the function signature never changed tho

proto/ibc/core/channel/v1/channel.proto
```diff
@@ -90,6 +90,21 @@ enum Order {
...
 
+// IBCVersion defines the version that the IBC packet intends to use.
+// This allows asynchronous upgrades over time as unsupported IBC Versions
+// will simply be rejected by the receiver and timed out on sender.
+enum IBCVersion {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // zero-value for IBC version.
+  // This will be treated as a classic packet
+  IBC_VERSION_UNSPECIFIED = 0;
```
colin-axner, line 101: Hmm, should we assign classic to 0? Would be encoding compliant with the existing structure, I'm fairly sure (empty fields not encoded), but it doesn't matter so much since this is just go api

proto/ibc/core/channel/v1/channel.proto
```diff
@@ -122,6 +137,10 @@ message Packet {
   ibc.core.client.v1.Height timeout_height = 7 [(gogoproto.nullable) = false];
   // block timestamp (in nanoseconds) after which the packet times out
   uint64 timeout_timestamp = 8;
+  // which IBC version to use to commit this packet
+  IBCVersion protocol_version = 9;
```
colin-axner, line 141: The idea here being rather than introspecting the channel-id, we have the sender specify a version?

For other reviewers, you should think of the packet type as our go api and the commitment structure as the spec requirement. 2 ibc implementations could use different internal packet types and still be spec compliant

proto/ibc/core/channel/v1/channel.proto
```diff
@@ -122,6 +137,10 @@ message Packet {
...
+  // version which application should use to process the packet data
+  string app_version = 10;
```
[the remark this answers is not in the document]
AdityaSripal: I think we can do that in a multi packet data version

proto/ibc/core/channel/v1/channel.proto
AdityaSripal: Proto makes me suffix this enum with UNSPECIFIED. idk if it can be overridden.

Agreed, this is just inside the chain doesn't go over the protocol so it doesn't matter

modules/core/04-channel/types/packet.go
```diff
@@ -35,6 +55,51 @@ func CommitPacket(cdc codec.BinaryCodec, packet Packet) []byte {
...
+// from a given packet. This results in a fixed length preimage.
+// NOTE: A fixed length preimage is ESSENTIAL to prevent relayers from being able
+// to malleate the packet fields and create a commitment hash that matches the original packet.
+// NOTE: sdk.Uint64ToBigEndian sets the uint64 to a slice of length 8.
+func CommitEurekaPacket(packet Packet) []byte {
+	timeoutHeight := packet.GetTimeoutHeight()
+
+	timeoutBuf := sdk.Uint64ToBigEndian(packet.GetTimeoutTimestamp())
+
+	// only hash the timeout height if it is non-zero. This will allow us to remove it entirely in the future
```
sangier, line 69: If we want to use the fixed-size preimage, which I think is a good thing, can we use here a sort of "padding" bytes to reach the fixed-size when the height is zero?

sangier (approved):
I can't spot anything wired, LGTM!

modules/core/04-channel/types/packet.go
```diff
@@ -35,6 +55,51 @@ func CommitPacket(cdc codec.BinaryCodec, packet Packet) []byte {
...
+	buf := timeoutHash[:]
+
+	// hash the destination identifiers since we can no longer retrieve them from the channelEnd
+	portHash := sha256.Sum256([]byte(packet.GetDestPort()))
+	buf = append(buf, portHash[:]...)
+	destinationHash := sha256.Sum256([]byte(packet.GetDestChannel()))
+	buf = append(buf, destinationHash[:]...)
+
+	// hash the version only if it is nonempty
+	if packet.AppVersion != "" {
```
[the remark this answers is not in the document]
AdityaSripal: I don't see any version validation in the handshakes.

I'm not sure if we should be allowing it to be empty. Maybe we shouldn't, but i was thinking a blank version would imply use your default version. So if transfer receives a blank version, it tries to unmarshal data into ics20-v1 packet for example

modules/core/04-channel/types/msgs_test.go
```diff
@@ -64,6 +64,7 @@ var (
 	disabledTimeout   = clienttypes.ZeroHeight()
 	validPacketData   = []byte("testdata")
 	unknownPacketData = []byte("unknown")
+	validVersion      = "ics-20"
```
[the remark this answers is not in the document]
AdityaSripal: in packet_test.go, just put the vars in one place rather than splitting them across files. We already use these vars in that test file

proto/ibc/core/channel/v1/channel.proto
```diff
@@ -90,6 +90,21 @@ enum Order {
...
+enum IBCVersion {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // zero-value for IBC version.
+  // This will be treated as a classic packet
+  IBC_VERSION_UNSPECIFIED = 0;
+  // IBC version 1 implements the Classic protocol
+  IBC_VERSION_CLASSIC = 1;
+  // IBC version 2 implements the Eureka protocol
+  IBC_VERSION_EUREKA = 2;
```
colin-axner, line 105: looked over this with @chatton since it overlaps with our work. He made a good point that Eureka name is leaking into the codebase. I think long term, referring to this as v1 and v2 packet commitment structures would be more intuitive. Otherwise you keep adding codenames for every version (which I find to be very confusing)

proto/ibc/core/channel/v1/channel.proto
colin-axner: https://github.com/cosmos/ibc-go/blob/4a0da8f37d309ea52e089b67c6eda929ccad35ae/proto/ibc/core/channel/v1/channel.proto#L85

modules/core/04-channel/types/packet.go
```diff
@@ -2,22 +2,42 @@ package types
...
-// CommitPacket returns the packet commitment bytes. The commitment consists of:
+// CommitPacket returns the packet commitment bytes based on
+// the ProtocolVersion specified in the Packet. The commitment
+// must commit to all fields in the packet apart from the source port
+// source channel and sequence (which will be committed to in the packet commitment key path)
+// and the ProtocolVersion which is defining how to hash the packet fields.
+// NOTE: The commitment MUSTs be a fixed length preimage to prevent relayers from being able
+// to malleate the packet fields and create a commitment hash that matches the original packet.
+func CommitPacket(packet Packet) []byte {
+	switch packet.ProtocolVersion {
```
colin-axner, line 24: if we aren't proving the protocol version a relayer could provide the opposite version that a packet was sent on

AdityaSripal: It would only matter if there was a hash collision between classic packets and eureka packets. Since we changed the commitment algorithm, I don't think that's possible here.

The same exact packet fields with the only thing changed being the ProtocolVersion would lead to different commitments, since the commitment algorithm is different even for the existing fields (and we newly hash dest identifiers)

And additionally two different packets hashing to the same hash by swapping out the ProtocolVersion would break the sha256 preimage collision so we don't need to worry about that.

I have thought about the issue of future maintainer adding a third version. And this third version commits packets in the exact same way as v2. In this case, if the future maintainer is not careful, they can allow relayers to modify the version on receive.

I could signal the possibility of that footgun here by explicitly hashing protocol version. But that would only be a hint for right behavior, since future maintainers may not do it (if they don't change commitment hashes in between the next versions, they will need to hash in the protocol version)

proto/ibc/core/channel/v1/channel.proto
AdityaSripal: Agreed, switched to v2 for Eureka and v1 for classic. With this context, I think it still makes sense to leave the 0 enum as UNSPECIFIED but treat it as a v1 packet. What do you think?

I prefer that to zero indexing the IBC_Version_1 enum

proto/ibc/core/channel/v1/channel.proto
AdityaSripal: https://github.com/cosmos/ibc-go/pull/6986#discussion_r1699154395

modules/core/04-channel/types/packet.go
```diff
@@ -58,6 +123,28 @@ func NewPacket(
 		DestinationChannel: destinationChannel,
 		TimeoutHeight:      timeoutHeight,
 		TimeoutTimestamp:   timeoutTimestamp,
+		ProtocolVersion:    IBC_VERSION_CLASSIC,
```
srdtrk, line 126: Why is protocol version contained in the ack commitment but not the packet commitment?

modules/core/04-channel/types/packet.go
```diff
@@ -35,6 +55,51 @@ func CommitPacket(cdc codec.BinaryCodec, packet Packet) []byte {
...
+	timeoutHash := sha256.Sum256(timeoutBuf)
+	buf := timeoutHash[:]
+
+	// hash the destination identifiers since we can no longer retrieve them from the channelEnd
+	portHash := sha256.Sum256([]byte(packet.GetDestPort()))
+	buf = append(buf, portHash[:]...)
+	destinationHash := sha256.Sum256([]byte(packet.GetDestChannel()))
+	buf = append(buf, destinationHash[:]...)
+
+	// hash the version only if it is nonempty
```
srdtrk, line 89: why? What's wrong with hashing the version even if it is empty

modules/core/04-channel/types/packet.go
AdityaSripal: protocol version is not in ack commitment?

modules/core/04-channel/types/packet.go
AdityaSripal: We could do this, thought to save on the compute if hash is empty. More relevant for VM implementations

AdityaSripal:
Merging to not block future PRs. I believe I've answered people's questions. If there are followups, please open new issues or discuss on the eureka channel

modules/core/04-channel/types/packet.go
srdtrk: Ah, sorry

modules/core/04-channel/types/packet.go
srdtrk: I think it is better not to have branching in the code when not necessary. A single hash isn't gonna make or break it. So can we hash even if it is empty?

PR #7052, opened 2024-08-05
feat(tests): add helper functions, keeper test suite.

DimitrisJim:
## Description

should probably do trick for keeper funcs for time being, testing package should need additional tweaks in future

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/core/packet-server/keeper/keeper.go
+++ b/modules/core/packet-server/keeper/keeper.go
@@ -8,14 +8,14 @@ import (
 
 type Keeper struct {
 	cdc           codec.BinaryCodec
-	channelKeeper types.ChannelKeeper
-	clientKeeper  types.ClientKeeper
+	ChannelKeeper types.ChannelKeeper
+	ClientKeeper  types.ClientKeeper
 }
 
 func NewKeeper(cdc codec.BinaryCodec, channelKeeper types.ChannelKeeper, clientKeeper types.ClientKeeper) *Keeper {
 	return &Keeper{
 		cdc:           cdc,
-		channelKeeper: channelKeeper,
-		clientKeeper:  clientKeeper,
+		ChannelKeeper: channelKeeper,
+		ClientKeeper:  clientKeeper,
 	}
 }
--- a/modules/core/packet-server/keeper/keeper_test.go
+++ b/modules/core/packet-server/keeper/keeper_test.go
@@ -38,13 +38,13 @@ func (suite *KeeperTestSuite) SetupTest() {
 // TODO: Remove, just testing the testing setup.
 func (suite *KeeperTestSuite) TestCreateEurekaClients() {
 	path := ibctesting.NewPath(suite.chainA, suite.chainB)
-	path.SetupEurekaClients()
+	path.SetupV2()
 
 	// Assert counterparty set and creator deleted
-	_, found := suite.chainA.App.GetIBCKeeper().ClientKeeper.GetCounterparty(suite.chainA.GetContext(), path.EndpointA.ClientID)
+	_, found := suite.chainA.App.GetPacketServer().ClientKeeper.GetCounterparty(suite.chainA.GetContext(), path.EndpointA.ClientID)
 	suite.Require().True(found)
 
 	// Assert counterparty set and creator deleted
-	_, found = suite.chainB.App.GetIBCKeeper().ClientKeeper.GetCounterparty(suite.chainB.GetContext(), path.EndpointB.ClientID)
+	_, found = suite.chainB.App.GetPacketServer().ClientKeeper.GetCounterparty(suite.chainB.GetContext(), path.EndpointB.ClientID)
 	suite.Require().True(found)
 }
--- a/modules/core/packet-server/types/expected_keepers.go
+++ b/modules/core/packet-server/types/expected_keepers.go
@@ -16,7 +16,7 @@ type ChannelKeeper interface {
 	GetPacketCommitment(ctx sdk.Context, portID string, channelID string, sequence uint64) []byte
 
 	// DeletePacketCommitment deletes the packet commitment hash under the commitment path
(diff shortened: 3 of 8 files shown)


testing/path.go
```diff
@@ -162,6 +162,20 @@ func (path *Path) SetupClients() {
 	}
 }
 
+// SetupEurekaClients is a helper function to create clients supporting ibc-eureka on both
+// chains. It assumes the caller does not anticipate any errors.
+func (path *Path) SetupEurekaClients() {
```
AdityaSripal, line 167: Rather than doing this all in one step, let's keep SetupClients unchanged, and **add** a new function called `SetupCounterparties` that does the provide counterparty for each side.

Then in our tests, we can setup like so:

```go
path.SetupClients()
path.SetupCounterparties()
```

AdityaSripal (commented):
Restructuring suggestion.

Also let's remove reference to Eureka in function names and use V2 instead

AdityaSripal:
Also we need the Packet-server keeper to be wired up in the app and made a public field so we can call it from `suite.chainA` for example.

We can modify `IBCKeeper` later, but let's keep it a top level field of the testing app for now

testing/path.go
colin-axner: eventually `path.Setup()` will just do both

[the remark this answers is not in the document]
colin-axner: I like naming suggestion below `SetupCounterparties`, nice to avoid introducing `Eureka` when possible

testing/endpoint.go
```diff
@@ -165,6 +165,18 @@ func (endpoint *Endpoint) UpdateClient() (err error) {
 	return endpoint.Chain.sendMsgs(msg)
 }
 
+// ProvideCounterparty will construct and execute a MsgProvideCounterparty on the associated endpoint.
+func (endpoint *Endpoint) ProvideCounterparty() (err error) {
+	merklePath := commitmenttypes.NewMerklePath([]byte("ibc"), []byte(""))
```
colin-axner, line 170: `Endpoint` should probably be modified to have the merkle prefix now

modules/core/packet-server/keeper/keeper_test.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// SetupTest creates a coordinator with 2 test chains.
+func (suite *KeeperTestSuite) SetupTest() {
+	suite.coordinator = ibctesting.NewCoordinator(suite.T(), 2)
+	suite.chainA = suite.coordinator.GetChain(ibctesting.GetChainID(1))
+	suite.chainB = suite.coordinator.GetChain(ibctesting.GetChainID(2))
+
+	// commit some blocks so that QueryProof returns valid proof (cannot return valid query if height <= 1)
+	suite.coordinator.CommitNBlocks(suite.chainA, 2)
```
colin-axner, line 34: Is this necessary? I think it might be legacy code

testing/path.go
AdityaSripal: created a SetupV2 that does both

testing/endpoint.go
AdityaSripal: Agreed, let's create an issue and do this later.

Think it should be added as a new field but this should be the default value that can then be malleated later by test writer

modules/core/packet-server/keeper/keeper.go
```diff
@@ -8,14 +8,14 @@ import (
 
 type Keeper struct {
 	cdc           codec.BinaryCodec
-	channelKeeper types.ChannelKeeper
-	clientKeeper  types.ClientKeeper
+	ChannelKeeper types.ChannelKeeper
```
AdityaSripal, line 11: made this public so they were retrievable in testchains through the packetserver.

We already do this with IBCKeeper

modules/core/packet-server/types/expected_keepers.go
```diff
@@ -16,7 +16,7 @@ type ChannelKeeper interface {
 	GetPacketCommitment(ctx sdk.Context, portID string, channelID string, sequence uint64) []byte
 
 	// DeletePacketCommitment deletes the packet commitment hash under the commitment path
-	DeletePacketCommitment(ctx sdk.Context, portID string, channelID string, sequence uint64)
+	// DeletePacketCommitment(ctx sdk.Context, portID string, channelID string, sequence uint64)
```
AdityaSripal, line 19: removed this for now. Can be readded in future PR that needs it

testing/simapp/app.go
```diff
@@ -370,6 +372,9 @@ func NewSimApp(
 		appCodec, keys[ibcexported.StoreKey], app.GetSubspace(ibcexported.ModuleName), ibctm.NewConsensusHost(app.StakingKeeper), app.UpgradeKeeper, scopedIBCKeeper, authtypes.NewModuleAddress(govtypes.ModuleName).String(),
 	)
 
+	// Setup packet server to call on Eureka tests
+	app.PacketServer = packetserverkeeper.NewKeeper(appCodec, app.IBCKeeper.ChannelKeeper, app.IBCKeeper.ClientKeeper)
```
AdityaSripal, line 376: Needed this so that we can call into these functions from the TestChain struct

AdityaSripal: I suppose we could also always create the PacketServer.Keeper struct on the fly using the IBCKeeper fields...

testing/path.go
```diff
@@ -162,6 +169,18 @@ func (path *Path) SetupClients() {
 	}
 }
 
+// SetupCounterparties is a helper function to set the counterparties supporting ibc-eureka on both
+// chains. It assumes the caller does not anticipate any errors.
+func (path *Path) SetupCounterparties() {
```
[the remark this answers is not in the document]
AdityaSripal: Hmm... yea I did it in case you wanted to `SetupClients`, test X, Y, Z, then call `SetCounterparties` after and do some more testing

PR #7058, opened 2024-08-06
refactor: migrate from sdk.Context to Context.Context

tac0turtle:
## Description

This pr aims to remove sdk.Context in ibcKeeper. This pr will also aim to replace many sdk imports with services from core to allow reducing the dependency graph on the sdk 

Example commit messages:

fix: skip emission of unpopulated memo field in ics20
deps: updating sdk to v0.46.4
chore: removed unused variables
e2e: adding e2e upgrade test for ibc-go/v6
docs: ics27 v6 documentation updates
feat(api)!: this is an api breaking feature
fix(statemachine)!: this is a statemachine breaking fix
-->

---

write a little note why.

### DIFF
--- a/modules/core/02-client/keeper/keeper.go
+++ b/modules/core/02-client/keeper/keeper.go
@@ -1,10 +1,12 @@
 package keeper
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"strings"
 
+	corestoretypes "cosmossdk.io/core/store"
 	errorsmod "cosmossdk.io/errors"
 	"cosmossdk.io/log"
 	"cosmossdk.io/store/prefix"
@@ -24,7 +26,7 @@ import (
 // Keeper represents a type that grants read and write permissions to any client
 // state information
 type Keeper struct {
-	storeKey       storetypes.StoreKey
+	storeService   corestoretypes.KVStoreService
 	cdc            codec.BinaryCodec
 	router         *types.Router
 	legacySubspace types.ParamSubspace
@@ -32,13 +34,13 @@ type Keeper struct {
 }
 
 // NewKeeper creates a new NewKeeper instance
-func NewKeeper(cdc codec.BinaryCodec, key storetypes.StoreKey, legacySubspace types.ParamSubspace, uk types.UpgradeKeeper) *Keeper {
+func NewKeeper(cdc codec.BinaryCodec, storeService corestoretypes.KVStoreService, legacySubspace types.ParamSubspace, uk types.UpgradeKeeper) *Keeper {
 	router := types.NewRouter()
 	localhostModule := localhost.NewLightClientModule(cdc, key)
 	router.AddRoute(exported.Localhost, localhostModule)
 
 	return &Keeper{
-		storeKey:       key,
+		storeService:   storeService,
 		cdc:            cdc,
 		router:         router,
 		legacySubspace: legacySubspace,
@@ -136,9 +138,12 @@ func (k *Keeper) SetClientConsensusState(ctx sdk.Context, clientID string, heigh
 }
 
 // GetNextClientSequence gets the next client sequence from the store.
-func (k *Keeper) GetNextClientSequence(ctx sdk.Context) uint64 {
-	store := ctx.KVStore(k.storeKey)
-	bz := store.Get([]byte(types.KeyNextClientSequence))
+func (k *Keeper) GetNextClientSequence(ctx context.Context) uint64 {
+	store := k.storeService.OpenKVStore(ctx)
+	bz, err := store.Get([]byte(types.KeyNextClientSequence))
+	if err != nil {
+		panic(err)
+	}
 	if len(bz) == 0 {
 		panic(errors.New("next client sequence is nil"))
 	}
@@ -147,8 +152,8 @@ func (k *Keeper) GetNextClientSequence(ctx sdk.Context) uint64 {
 }
 
 // SetNextClientSequence sets the next client sequence to the store.
-func (k *Keeper) SetNextClientSequence(ctx sdk.Context, sequence uint64) {
-	store := ctx.KVStore(k.storeKey)
+func (k *Keeper) SetNextClientSequence(ctx context.Context, sequence uint64) {
+	store := k.storeService.OpenKVStore(ctx)
 	bz := sdk.Uint64ToBigEndian(sequence)
 	store.Set([]byte(types.KeyNextClientSequence), bz)
 }
@@ -156,8 +161,8 @@ func (k *Keeper) SetNextClientSequence(ctx sdk.Context, sequence uint64) {
 // IterateConsensusStates provides an iterator over all stored consensus states.
 // objects. For each State object, cb will be called. If the cb returns true,
 // the iterator will close and stop.
-func (k *Keeper) IterateConsensusStates(ctx sdk.Context, cb func(clientID string, cs types.ConsensusStateWithHeight) bool) {
-	store := ctx.KVStore(k.storeKey)
+func (k *Keeper) IterateConsensusStates(ctx context.Context, cb func(clientID string, cs types.ConsensusStateWithHeight) bool) {
+	store := k.storeService.OpenKVStore(ctx)
 	iterator := storetypes.KVStorePrefixIterator(store, host.KeyClientStorePrefix)
 
 	defer sdk.LogDeferred(ctx.Logger(), func() error { return iterator.Close() })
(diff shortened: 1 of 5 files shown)


tac0turtle:
opening this for review so the other prs dont get blocked and build up a backlog

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -35,13 +39,13 @@ type Keeper struct {
 
 // NewKeeper creates a new 29-fee Keeper instance
 func NewKeeper(
-	cdc codec.BinaryCodec, key storetypes.StoreKey,
+	cdc codec.BinaryCodec, key corestore.KVStoreService,
```
damiannolan, line 42: nit on `key` -> `storeService`

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -112,43 +117,57 @@ func (k Keeper) EscrowAccountHasBalance(ctx sdk.Context, coins sdk.Coins) bool {
...
 }
 
 // IsFeeEnabled returns whether fee handling logic should be run for the given port. It will check the
 // fee enabled flag for the given port and channel identifiers
-func (k Keeper) IsFeeEnabled(ctx sdk.Context, portID, channelID string) bool {
-	store := ctx.KVStore(k.storeKey)
-	return store.Has(types.KeyFeeEnabled(portID, channelID))
+func (k Keeper) IsFeeEnabled(ctx context.Context, portID, channelID string) bool {
+	store := k.storeService.OpenKVStore(ctx)
+	has, err := store.Has(types.KeyFeeEnabled(portID, channelID))
```
damiannolan, line 159: are the new `err` returns just propagations of errors that would surfaced before as panics? is there any risk here for actually getting an err back or would things need to be corrupted in the db somehow?

damiannolan: there is also an error returned from `Get` now? how does that behave? 

if the key doesn't exist is there an error is it just `nil` bytes? Assuming the latter based on other code diffs

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -168,28 +187,39 @@ func (k Keeper) GetAllFeeEnabledChannels(ctx sdk.Context) []types.FeeEnabledChan
...
 		return "", false
 	}
 
-	return string(store.Get(key)), true
+	bz, err := store.Get(key)
+	if err != nil {
+		panic(err)
+	}
+
+	return string(bz), true
```
damiannolan, line 207: Assuming these could be refactored to a single `Get` call and if `len(bz) == 0` then return `false` for found bool. 

Think this would save gas.

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -212,29 +242,38 @@ func (k Keeper) GetAllPayees(ctx sdk.Context) []types.RegisteredPayee {
...
 		return "", false
 	}
 
-	addr := string(store.Get(key))
-	return addr, true
+	addr, err := store.Get(key)
+	if err != nil {
+		panic(err)
+	}
+	return string(addr), true // TODO: why this cast?
```
damiannolan, line 269: ditto on refactor to single `Get` call and check len bytes.

regarding the TODO - its stored plain string because its an address belonging to a counterparty chain and "in theory" could be non cosmos-sdk addr.

e.g. relayer submits MsgRecvPacket to chainB and looks up a stored address to encode into the acknowledgement. Ack is relayed back to source(chainA) and address is parsed from acknowledgement data and fee is paid out to the account on source chain.

damiannolan (commented):
So far so good, left a couple of comments to begin with, just nits mostly. Made my way through ica and fee code, will continue reviewing tomorrow.

Thanks @tac0turtle!

modules/light-clients/06-solomachine/proof.go
```diff
@@ -26,7 +26,7 @@ func VerifySignature(pubKey cryptotypes.PubKey, signBytes []byte, sigData signin
 		if err := pubKey.VerifyMultisignature(func(signing.SignMode) ([]byte, error) {
 			return signBytes, nil
 		}, data); err != nil {
-			return errorsmod.Wrapf(ErrSignatureVerificationFailed, err.Error())
+			return errorsmod.Wrapf(ErrSignatureVerificationFailed, "failed to verify multisignature: %s", err.Error())
```
damiannolan, line 29: noticed another unrelated error formatting diff (transfer authz) but seems fine, shouldn't have been using `Wrapf` previously...

damiannolan (commented):
Made my way through the rest of the diffs. Would like to clarify the api differences in corestore vs storetypes before merging. We now need to handle a bunch of error returns that were previously not there, just wondering how/what will trigger them to be surfaced(see other comment).

Assuming there will be parity available through other service types which we can consume later, in order fully remove sdk.Context and address all the TODOs? Some of the main things that come to mind are:
- ChainID
- BlockTime
- Gas Consumption
- Tx exec mode

Otherwise lgtm! Thank you @tac0turtle. I pushed a couple of consistency and nit follow up commits

modules/apps/29-fee/keeper/keeper.go
tac0turtle: we are now returning the panic we were doing internally. it is up to the caller to decide what to do with the error. Handling errors is a breaking change that i would like to avoid in these prs as panics to error handling can have hidden consequences.

modules/apps/29-fee/keeper/keeper.go
damiannolan: I believe the reasoning was likely around string usage being more human readable, rather than passing around the raw bytes of a foreign address. @AdityaSripal can probably speak to it more tbh

modules/apps/callbacks/ibc_middleware.go
```diff
@@ -102,7 +103,8 @@ func (im IBCMiddleware) SendPacket(
 	// packet is created without destination information present, GetSourceCallbackData does not use these.
 	packet := channeltypes.NewPacket(data, seq, sourcePort, sourceChannel, "", "", timeoutHeight, timeoutTimestamp)
 
-	callbackData, err := types.GetSourceCallbackData(ctx, im.app, packet, im.maxCallbackGas)
+	sdkCtx := sdk.UnwrapSDKContext(ctx) // TODO: remove after sdk.Context is removed from core IBC
```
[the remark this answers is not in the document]
tac0turtle: i will remove in subsequent PRs, this is only the first of many. I can still open or append it to the open issue. Which would you prefer?

modules/apps/29-fee/keeper/keeper.go
damiannolan: Addressed in: https://github.com/cosmos/ibc-go/pull/7058/commits/3cd2a56eec88630bb058d3ec2d424105215fa7ee

modules/apps/callbacks/ibc_middleware.go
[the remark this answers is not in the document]
damiannolan: I updated the comments to link the issue so its easier to grep for

modules/core/03-connection/keeper/keeper.go
```diff
@@ -22,35 +25,36 @@ type Keeper struct {
...
+	sdkCtx := sdk.UnwrapSDKContext(ctx) // TODO: remove when sdk.Context is removed
+	return sdkCtx.Logger().With("module", "x/"+exported.ModuleName+"/"+types.SubModuleName)
 }
 
 // GetCommitmentPrefix returns the IBC connection store prefix as a commitment
 // Prefix
-func (k *Keeper) GetCommitmentPrefix() exported.Prefix {
-	return commitmenttypes.NewMerklePrefix([]byte(k.storeKey.Name()))
+func (*Keeper) GetCommitmentPrefix() exported.Prefix {
+	return commitmenttypes.NewMerklePrefix([]byte("ibc")) // TODO: import store key directly
```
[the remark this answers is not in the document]
damiannolan: Hmm okay.. 🤔 

I've updated to use `exported.StoreKey` (== "ibc") for now. Discussed with the team and the general opinion is that nobody actually customises this in cosmos chains. A counterparty running in a different chain runtime could have a different store prefix tho, maybe denoting some smart contract or other isolated store space.

damiannolan (approved):
LGTM pending CI. Cheers @tac0turtle

modules/core/03-connection/keeper/keeper.go
```diff
@@ -202,10 +231,14 @@ func (k *Keeper) addConnectionToClient(ctx sdk.Context, clientID, connectionID s
...
-	bz := store.Get([]byte(types.ParamsKey))
-	if bz == nil { // only panic on unset params and not on empty params
+func (k *Keeper) GetParams(ctx context.Context) types.Params {
+	store := k.storeService.OpenKVStore(ctx)
+	bz, err := store.Get([]byte(types.ParamsKey))
+	if err != nil {
+		panic(err)
+	}
+
+	if len(bz) == 0 { // only panic on unset params and not on empty params
```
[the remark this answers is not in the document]
damiannolan: should stay nil based on dm discussion I believe, good catch

if someone sets empty params then bz are not nil but len(bz) are 0... weird scenario, not sure if it ever causes any issue but fine with maintaining previous behaviour

modules/apps/29-fee/keeper/keeper.go
```diff
@@ -168,28 +187,35 @@ func (k Keeper) GetAllFeeEnabledChannels(ctx sdk.Context) []types.FeeEnabledChan
...
+	store := k.storeService.OpenKVStore(ctx)
 	key := types.KeyPayee(relayerAddr, channelID)
 
-	if !store.Has(key) {
+	bz, err := store.Get(key)
+	if err != nil {
+		panic(err)
+	}
+
+	if len(bz) == 0 || bz == nil {
```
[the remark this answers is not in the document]
damiannolan: ah nice one, I missed this one earlier!

Issue #6949, opened 2024-07-25
Make `MsgTransfer` a convenience wrapper around SendPacket rpc


colin-axner:
## Summary

With the introduction of the `SendPacket` api, `MsgTransfer` should become a convenience wrapper which translates its user inputted arguments into a ics20 transfer packet. All necessary checks for ics20 should be moved into `OnSendPacket`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

PR #7068, opened 2024-08-07
chore(api)!: move checks from Transfer to OnSendPacket

bznein:
## Description

closes: #6949

---

write a little note why.


modules/apps/transfer/keeper/keeper.go
```diff
@@ -391,7 +391,7 @@ func (k Keeper) iterateForwardedPackets(ctx sdk.Context, cb func(packet types.Fo
 
 // IsBlockedAddr checks if the given address is allowed to send or receive tokens.
 // The module account is always allowed to send and receive tokens.
-func (k Keeper) isBlockedAddr(addr sdk.AccAddress) bool {
+func (k Keeper) IsBlockedAddr(addr sdk.AccAddress) bool {
```
bznein, line 394: This was needed in order to be able to call it from `OnSendPacket`

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -33,27 +29,6 @@ func (k Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types.
 		return nil, errorsmod.Wrapf(types.ErrSendDisabled, err.Error())
```
bznein, line 29: This check was left in this function as the logic is not easy to implement on `OnSendPacket`, we would have to export `BankKeeper` but also add a lot of logic to convert `coins` back to `tokens`

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -169,21 +169,6 @@ func (suite *KeeperTestSuite) TestSendTransfer() {
 		// 	},
 		// 	channeltypes.ErrInvalidPacket,
 		// },
-		{
```
bznein, line 172: This test now doesn't succeed due to the following logic:

- `Transfer` does not check for the version/forwarding combination anymore
- the function that creates the packet data will ignore everything in forwarding when the version is V1.
Note that there is still a check on `OnSendPacket` but it will not be triggered because Hops will not be present.


I'll open an issue to add tests specifically for `OnSendPacket` since we currently do not have them

modules/apps/transfer/ibc_module.go
```diff
@@ -182,6 +190,17 @@ func (im IBCModule) OnSendPacket(
 		return err
 	}
 
+	if ics20Version == types.V1 {
+		// ics20-1 only supports a single coin, so if that is the current version, we must only process a single coin.
+		if len(data.Tokens) > 1 {
+			return errorsmod.Wrapf(ibcerrors.ErrInvalidRequest, "cannot transfer multiple coins with %s", types.V1)
+		}
+
+		if len(data.Forwarding.Hops) > 1 {
```
colin-axner, line 199:
```suggestion
		if len(data.Forwarding.Hops) > 0 {
```
we should have a test for this

modules/apps/transfer/keeper/msg_server.go
colin-axner: This check is necessary to be moved. We can move it into `keeper.OnSendPacket`. There's already logic to convert tokens to coin there

colin-axner: Note: going forward, users can invoke `OnSendPacket` without using `MsgTransfer`, so it's unsafe to have any necessary checks remain in `MsgTransfer`

modules/apps/transfer/ibc_module.go
bznein: Yeah I wanted to create a new issue with tests, bu better to do it in this PR!

modules/apps/transfer/keeper/relay.go
```diff
@@ -347,6 +347,10 @@ func (k Keeper) refundPacketTokens(ctx sdk.Context, packet channeltypes.Packet,
 				return err
 			}
 
+			if k.IsBlockedAddr(sender) {
+				return errorsmod.Wrapf(ibcerrors.ErrUnauthorized, "%s is not allowed to send funds", sender)
+			}
+
```
colin-axner, line 353: this is an unnecessary addition?

modules/apps/transfer/keeper/relay.go
```diff
@@ -430,7 +434,7 @@ func createPacketDataBytesFromVersion(appVersion, sender, receiver, memo string,
 	case types.V1:
 		// Sanity check, tokens must always be of length 1 if using app version V1.
 		if len(tokens) != 1 {
-			return nil, errorsmod.Wrapf(ibcerrors.ErrInvalidRequest, "cannot transfer multiple coins with %s", types.V1)
+			return nil, errorsmod.Wrapf(types.ErrInvalidVersion, "length of tokens must be equal to 1 if using %s version", types.V1)
```
colin-axner, line 437: would recommend doing changes like this against main since it is unrelated to the port router changes

bznein: this was the result of a weird merge, because we used to `panic` here so I added a similar change but not identical, will revert

modules/apps/transfer/keeper/relay.go
bznein: No idea how this got in, sorry :/

modules/apps/transfer/keeper/relay_test.go
colin-axner: It should? I wonder if this was not working because of the `1` -> `0` change?

bznein: ouch thanks, yeah it does. I'm trying to give up coffee starting today and you can see the results :(

colin-axner (approved):
Nice work! As a side note, I noticed #7080 which came from the initial refactor, but as it was not necessary for this work I opened a separate issue

bznein:
> Nice work! As a side note, I noticed #7080 which came from the initial refactor, but as it was not necessary for this work I opened a separate issue

Ah! We noticed the same with Cian and were about to open an issue :D thanks!

modules/apps/transfer/keeper/msg_server.go
damiannolan, line 1: there is some commented out code on L54 that can be removed now!

```
// packetData := types.NewFungibleTokenPacketData(
// 	fullDenomPath, msg.Token.Amount.String(), sender.String(), msg.Receiver, msg.Memo,
// )
```

damiannolan (approved):

PR #7081, opened 2024-08-07
p2

tac0turtle:
## Description

second pr of migrating the repository to use context in public apis 

Example commit messages:

fix: skip emission of unpopulated memo field in ics20
deps: updating sdk to v0.46.4
chore: removed unused variables
e2e: adding e2e upgrade test for ibc-go/v6
docs: ics27 v6 documentation updates
feat(api)!: this is an api breaking feature
fix(statemachine)!: this is a statemachine breaking fix
-->

---

write a little note why.

### DIFF
--- a/modules/apps/27-interchain-accounts/controller/keeper/keeper.go
+++ b/modules/apps/27-interchain-accounts/controller/keeper/keeper.go
@@ -7,11 +7,13 @@ import (
 	"fmt"
 	"strings"
 
+	corestore "cosmossdk.io/core/store"
 	errorsmod "cosmossdk.io/errors"
 	"cosmossdk.io/log"
 	storetypes "cosmossdk.io/store/types"
 
 	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/cosmos/cosmos-sdk/runtime"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	capabilitytypes "github.com/cosmos/ibc-go/modules/capability/types"
@@ -27,7 +29,7 @@ import (
 
 // Keeper defines the IBC interchain accounts controller keeper
 type Keeper struct {
-	storeKey       storetypes.StoreKey
+	storeService   corestore.KVStoreService
 	cdc            codec.Codec
 	legacySubspace icatypes.ParamSubspace
 	ics4Wrapper    porttypes.ICS4Wrapper
@@ -45,7 +47,7 @@ type Keeper struct {
 
 // NewKeeper creates a new interchain accounts controller Keeper instance
 func NewKeeper(
-	cdc codec.Codec, key storetypes.StoreKey, legacySubspace icatypes.ParamSubspace,
+	cdc codec.Codec, storeService corestore.KVStoreService, legacySubspace icatypes.ParamSubspace,
 	ics4Wrapper porttypes.ICS4Wrapper, channelKeeper icatypes.ChannelKeeper, portKeeper icatypes.PortKeeper,
 	scopedKeeper exported.ScopedKeeper, msgRouter icatypes.MessageRouter, authority string,
 ) Keeper {
@@ -54,7 +56,7 @@ func NewKeeper(
 	}
 
 	return Keeper{
-		storeKey:       key,
+		storeService:   storeService,
 		cdc:            cdc,
 		legacySubspace: legacySubspace,
 		ics4Wrapper:    ics4Wrapper,
@@ -79,12 +81,13 @@ func (k Keeper) GetICS4Wrapper() porttypes.ICS4Wrapper {
 }
 
 // Logger returns the application logger, scoped to the associated module
-func (Keeper) Logger(ctx sdk.Context) log.Logger {
-	return ctx.Logger().With("module", fmt.Sprintf("x/%s-%s", exported.ModuleName, icatypes.ModuleName))
+func (Keeper) Logger(ctx context.Context) log.Logger {
+	sdkCtx := sdk.UnwrapSDKContext(ctx) //TODO: remove after sdk.Context is removed from core IBC
+	return sdkCtx.Logger().With("module", fmt.Sprintf("x/%s-%s", exported.ModuleName, icatypes.ModuleName))
 }
 
 // GetConnectionID returns the connection id for the given port and channelIDs.
-func (k Keeper) GetConnectionID(ctx sdk.Context, portID, channelID string) (string, error) {
+func (k Keeper) GetConnectionID(ctx context.Context, portID, channelID string) (string, error) {
 	channel, found := k.channelKeeper.GetChannel(ctx, portID, channelID)
 	if !found {
 		return "", errorsmod.Wrapf(channeltypes.ErrChannelNotFound, "port ID (%s) channel ID (%s)", portID, channelID)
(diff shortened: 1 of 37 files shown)


tac0turtle:
opening for preliminary review as i make sure tests are passing

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -136,19 +142,23 @@ func (k Keeper) GetAppVersion(ctx context.Context, portID, channelID string) (st
...
 	key := icatypes.KeyActiveChannel(portID, connectionID)
 
-	if !store.Has(key) {
+	bz, err := store.Get(key)
+	if len(bz) == 0 {
 		return "", false
 	}
+	if err != nil {
+		panic(err)
+	}
```
damiannolan, line 155: error handling and panic before `len(bz)` check?

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -136,19 +142,23 @@ func (k Keeper) GetAppVersion(ctx context.Context, portID, channelID string) (st
...
 
-	return string(store.Get(key)), true
+	return string(bz), true // todo: why the cast?
```
damiannolan, line 157: value is stored as `[]byte(channelID)` where channelID is string

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -202,32 +212,38 @@ func (k Keeper) GetAllActiveChannels(ctx sdk.Context) []genesistypes.ActiveChann
...
 	key := icatypes.KeyOwnerAccount(portID, connectionID)
 
-	if !store.Has(key) {
+	bz, err := store.Get(key)
+	if len(bz) == 0 {
 		return "", false
 	}
+	if err != nil {
+		panic(err)
+	}
```
damiannolan, line 239: ditto

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -202,32 +212,38 @@ func (k Keeper) GetAllActiveChannels(ctx sdk.Context) []genesistypes.ActiveChann
...
 
-	return string(store.Get(key)), true
+	return string(bz), true // todo: why the cast?
```
damiannolan, line 241: address value is stored as string... similar reasoning to the ics29 question you had on counterparty address, could store address bytes but string is more human readable

modules/apps/27-interchain-accounts/host/keeper/keeper.go
```diff
@@ -151,19 +159,28 @@ func (k Keeper) getAppMetadata(ctx sdk.Context, portID, channelID string) (icaty
...
 		return "", false
 	}
 
-	return string(store.Get(key)), true
+	bz, err := store.Get(key)
+	if err != nil {
+		panic(err)
+	}
+
+	return string(bz), true
```
damiannolan, line 179: I think we can optimise this for gas consumption similarly to one or two funcs on the other PR. cc @DimitrisJim

modules/apps/27-interchain-accounts/host/keeper/keeper.go
```diff
@@ -201,32 +218,38 @@ func (k Keeper) GetAllActiveChannels(ctx sdk.Context) []genesistypes.ActiveChann
...
 	key := icatypes.KeyOwnerAccount(portID, connectionID)
 
-	if !store.Has(key) {
+	bz, err := store.Get(key)
+	if len(bz) == 0 {
 		return "", false
 	}
+	if err != nil {
+		panic(err)
+	}
```
damiannolan, line 245: ditto

modules/core/02-client/keeper/migrations.go
```diff
@@ -3,7 +3,7 @@ package keeper
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
-	"github.com/cosmos/ibc-go/v9/modules/core/02-client/migrations/v7"
+	v7 "github.com/cosmos/ibc-go/v9/modules/core/02-client/migrations/v7"
```
damiannolan, line 6: This is annoying. My editor will always put this in place as well and then the linter complains about a redundant import alias... happens with vim-go and vscode extensions... 🫠

damiannolan (commented):
Nice one, diffs look good. Left a few nits, myself and @DimitrisJim can address them.

Looks like there is a couple of failing tests in ICA which we'll have to look into before merging

modules/core/02-client/keeper/migrations.go
tac0turtle: ill play with the linter settings

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
tac0turtle: moved it as the logic was wrong above

modules/apps/27-interchain-accounts/controller/keeper/handshake_test.go
```diff
@@ -294,11 +294,14 @@ func (suite *KeeperTestSuite) TestOnChanOpenInit() {
...
-				chanCap, err = suite.chainA.App.GetScopedIBCKeeper().NewCapability(suite.chainA.GetContext(), host.ChannelCapabilityPath(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID))
-				suite.Require().NoError(err)
+				channelID := channeltypes.FormatChannelIdentifier(suite.chainA.GetSimApp().IBCKeeper.ChannelKeeper.GetNextChannelSequence(suite.chainA.GetContext()))
+				path.EndpointA.ChannelID = channelID
 
 				tc.malleate() // malleate mutates test data
 
+				chanCap, err = suite.chainA.App.GetScopedIBCKeeper().NewCapability(suite.chainA.GetContext(), host.ChannelCapabilityPath(path.EndpointA.ChannelConfig.PortID, path.EndpointA.ChannelID))
+				suite.Require().NoError(err)
+
```
damiannolan, line 304: I find it weird that I had to do this to fix failing tests here given the diffs, I can try to provide more info later, need more reasoning on it

damiannolan: Let me try to explain this a little further.

The channel capability is generated in 04-channel (core IBC) and passed to application callback handlers (e.g. like this one under test - ica `OnChanOpenInit`).
This test calls `keeper.OnChanOpenInit` directly, rather than delivering a msg e2e, testing core integration points with the app - thus, the capability is generated directly in the test code here and passed to app callback.

From what I can see, this test as it existed, was creating a new capability with an empty channel ID. This is because the `path.EndpointA.ChannelID` field is not actually fulfilled here until after the `INIT` step of the handshake. 

In the diffs, I get the next channel sequence nonce (monotonically increasing) and format a new channel ID with it. Then use that to create the new capability before invoking the app callbacks directly on the keeper.
It's moved under the call to `tc.malleate()` as some test cases create a channel and set it to closed, in order to test regaining control of an interchain account on a closed channel, i.e. without moving it we fix some tests, and cause failures in others surfaced from capability already being claimed for a particular channelID.

Hope that makes sense!

damiannolan: The changes which caused the test failures to be raised in this PR are from `GetActiveChannelID` which previously allowed storing an empty string and have it being treated as a "found" or "truthy" value.

e.g. in testcase `failure: different ordering from previous channel`, there is a call to `SetActiveChannelID` in the `malleate()` func. It's actually setting an empty channel id as the active channel id on main, while `store.Has()` treats it as true.

damiannolan (approved):
LGTM! Thank you @tac0turtle, and @DimitrisJim for fixing the other test!

Issue #7115, opened 2024-08-09
Add CLI to submit tx with `MsgProvideCounterparty`


crodriguezvega:
I guess it would be come handy to have a CLI that submits the tx with `MsgProvideCounterparty`. The CLI should accept the client identifier, the counterparty client identifier and the counterparty merkle path prefix.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7118, opened 2024-08-10
wip

vishal-kanna:
closes: #7115

### DIFF
--- a/modules/core/02-client/client/cli/tx.go
+++ b/modules/core/02-client/client/cli/tx.go
@@ -450,3 +450,42 @@ func newScheduleIBCUpgradeProposalCmd() *cobra.Command {
 
 	return cmd
 }
+
+// client identifier, the counterparty client identifier and the counterparty merkle path prefix
+func newMsgProvideCounterpartycmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "",
+		Args:  cobra.ExactArgs(3),
+		Short: "",
+		Long:  ``,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+			signer := clientCtx.FromAddress.String()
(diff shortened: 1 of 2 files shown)


modules/core/02-client/client/cli/tx.go
```diff
@@ -450,3 +451,34 @@ func newScheduleIBCUpgradeProposalCmd() *cobra.Command {
...
+		Long:  `Provide the counterparty to an IBC client specified by its client ID.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			clientIdentifier := args[0]
+			counterpartyClientIdentifier := args[1]
+			counterpartyMerklePathPrefix := v2.NewMerklePath([]byte(args[2]))
```
crodriguezvega, line 470: Since the merle path prefix consists of a slice of `[]byte`, I am wondering that maybe the best way to input this information would be as a comma-separated list of bases-64 encoded strings. Then we can iterate over the list, base64 decode them into `[]byte` and construct the `MerklePath`.

crodriguezvega (commented):
Thank you for the PR, @vishal-kanna. 

As an improvement for future PRs: when you open one, would you mind to open it with a descriptive title of what the PR tries to achieve? Thank you!

modules/core/02-client/client/cli/tx.go
crodriguezvega: Correct if I'm wrong, but I don't think my comment is addressed yet, right? 

The way the code it is right now, it will just convert whatever string into is bytes representation, but if we want to use base64 encoded strings, we should first decode them, right?

vishal-kanna: Yes,if we use base64 encoded string we need to decode them first

crodriguezvega (approved):
I am leaving my approval even though I contributed to the PR.

Issue #6975, opened 2024-07-29
Alias new channel IDs to the client IDs


AdityaSripal:
Allow multiple aliases to map to the same client-id. This should be implemented as a many-to-one layer in the channel keeper. When doing packet processing, retrieve the client id by resolving the channel id to the underlying client using the alias map.

The alias map can be added to using genesis, migration, or through governance.

This will allow a backwards compatible transition from ibc-classic to ibc-eureka since we do not need to migrate over channel ids and transfer escrows. It will also allow governance to use this alias map in the future for more readable channel-ids.

colin-axner:
Delay periods are the only reason I see not to do an automatic migration. Perhaps we could add logic:
* if delay period is non zero on the connection, skip
* gov proposal to update alias mapping

In addition, we should consider spreading the migration across several blocks since it doesn't need to all be done in one go.

colin-axner:
Ah, connection counterparties may vary even when using the same local client id. I believe you will need to alias this as well. Essentially, we would have a map from channelID -> {local_client_id, counterparty_connection_id} and in the counterparty map, connection_id -> {counterparty_client_id, prefix}. I believe things would break if you tried changing the counterparty value in existing connections as the counterparty would make an assertion on the dest_channel_id which would be incorrect

colin-axner:
Thinking about this some more, we would need to store the counterparty alias as well, this would be utilized in [RecvPacket](https://github.com/cosmos/ibc-go/blob/0743c090f7ff67dd73b691c5b70fd88de69c9aaa/modules/core/packet-server/keeper/keeper.go#L132) when checking the source information. Given that this would then turn into an alias map which is just the channel/connection layers with condensed information, rather than duplicating this information, lets rely on the channel/connection layers (even if it is two store reads rather than one) in the meantime? 

Once we plan to remove the connection/channel structs, then an alias map seems useful, but I believe the best motivation to support existing channels to use eureka would be to access the app version? 

I can open a pr with resolving channel ids to client ids to add support for eureka flow with existing channels. Then maybe as a separate, nice-to-have issue, we could attempt to migrate connection structs to this alias map. The map would be:

alias map
`channelID -> {local_client_id, counterparty_channel_id}`
with the counterparty map adding
`counterparty_channel_id -> {counterparty_client_id, prefix}`

It's unclear to me if the connection IDs need to be referenced at all?

colin-axner:
Actually, I'm lost on this safety of this issue. 

[...]

PR #7174, opened 2024-08-14
Draft: Aliasing the v1 channels to v2 counterparty so that v1 channels can support v2 packets

AdityaSripal:
## Description

closes: #6975 

The following is the simplest possible aliasing that preserves two different code paths for v1 and v2 packets.

The channel is converted into a v2 counterparty for the purpose of handling v2 packets. In this way, you can consider the connection and channel handshake as being very complicated ways to map the identifier `(portID, channelID)` to the client, and `(cpPortId, cpChannelId)` to the client counterparty.

This will allow v1 channels to immediately start using v2 packets while retaining the identifiers that they already use. This is critical for immediately enabling atomtic multi-packet data on existing transfer channels. As well as allowing existing channels to very easily upgrade their versions by simply sending a v2 packet with a new version rather than going through channel upgradability

- [ ] Wire the ability to alias channels by converting Channel v1 into a channel v2 in a one-time fashion (write into state)
- [ ] Test the aliasing ability specifically against transfer because this is the only important app to alias. Ensure denoms don't change when moving from v1 packet to v2 packet with same channel id
- [ ] Decide on what to do with in-flight channel upgrades (ensure both sides have eureka enabled before disabling channel upgrades)
- [ ] Rename Counterparty struct to Channel

---

write a little note why.


modules/core/04-channel/keeper/keeper.go
```diff
@@ -95,6 +96,30 @@ func (k *Keeper) SetChannel(ctx sdk.Context, portID, channelID string, channel t
 	store.Set(host.ChannelKey(portID, channelID), bz)
 }
 
+// GetV2Counterparty returns a version 2 counterparty for the given port and channel ID
+// by converting the channel into a version 2 counterparty
+func (k *Keeper) GetV2Counterparty(ctx sdk.Context, portID, channelID string) (clienttypes.Counterparty, bool) {
```
AdityaSripal, line 101: This effectively resolves the portID and channel ID into a counterparty. So the portID and channelID are in effect the clientID here. It's just a different arbitrary string identifier.

The one annoying thing is that the portID here is both acting as part of the v2 "client-id" (though we could move away from this with a different storage format since channel ids are unique) and also selecting the application to use.

This is fine for packets that are only sending one app data.

For MULTI_PACKET_DATA, we can deal with this by leaving the top portID as is, with the plan to eventually deprecate it. And then still allowing different ports in the internal packet data list

crodriguezvega (commented):
Thank you, @AdityaSripal.

Taking transfer packets as an example, with the channel aliasing it will be possible to send a packet from an ibc-go v10 chain to another ibc-go v10 chain either on a existing transfer channel that was already set up or over eureka, right? Are we going to leave the decision of over which channel to send the packet to the user? If that's the case, would we need to add a protocol version field to [`MsgSendPacket`](https://github.com/cosmos/ibc-go/blob/6327b580a29da4c7f585a21313fed21ea0c56614/proto/ibc/core/channel/v1/tx.proto#L214-L226)? And the transfer application, which [also sends a `MsgSendPacket`](https://github.com/cosmos/ibc-go/blob/6327b580a29da4c7f585a21313fed21ea0c56614/modules/apps/transfer/keeper/msg_server.go#L54-L67), should specify decide if the packet should be sent over a classic or an eureka channel, right?  

It would be nice if there was a way by which we could internally determine if a counterparty chain supports eureka and then we prioritise sending the packet over a eureka channel instead of the existing classic channel. But I don't think that's going to be easy, right?

modules/core/packet-server/keeper/keeper.go
```diff
@@ -125,7 +125,7 @@ func (k Keeper) RecvPacket(
 	// Lookup counterparty associated with our channel and ensure that it was packet was indeed
 	// sent by our counterparty.
 	// Note: This can be implemented by the current keeper
-	counterparty, ok := k.ClientKeeper.GetCounterparty(ctx, packet.DestinationChannel)
+	counterparty, ok := k.GetCounterparty(ctx, packet.DestinationPort, packet.DestinationChannel)
```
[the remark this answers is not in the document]
colin-axner: No, the `DestinationChannel` will continue to be the existing channel identifier (that's the purpose of this adjustment, to keep fungibility but allow existing channels to send v2 packets)

modules/core/02-client/client/cli/tx.go
```diff
@@ -455,40 +455,40 @@ func newScheduleIBCUpgradeProposalCmd() *cobra.Command {
...
-				Signer:       clientCtx.GetFromAddress().String(),
-			}
-			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), &msg)
-		},
-	}
-
-	flags.AddTxFlagsToCmd(cmd)
-	return cmd
-}
+// func newProvideCounterpartyCmd() *cobra.Command {
```
crodriguezvega, line 458: I commented out this for now because in this PR `MsgProvideCounterparty` has changed and needs more information from the CLI to fill it all in.

modules/core/02-client/keeper/grpc_query.go
```diff
@@ -427,8 +427,8 @@ func (q *queryServer) Client(ctx context.Context, req *types.QueryClientRequest)
 	creator, _ := q.GetCreator(sdkCtx, req.ClientId)
 	res.Creator = creator
 
-	counterparty, _ := q.GetCounterparty(sdkCtx, req.ClientId)
-	res.Counterparty = counterparty
+	// counterparty, _ := q.GetCounterparty(sdkCtx, req.ClientId)
+	// res.Counterparty = counterparty
```
crodriguezvega, line 431: Commented this also out because there is a circular dependency in the pb.go file between 02-client and packet-server types.

proto/ibc/core/client/v1/query.proto
```diff
@@ -255,6 +255,6 @@ message QueryClientRequest {
 
 // QueryClientRequest is the response type for the Query/Client RPC method
 message QueryClientResponse {
-  string       creator      = 1;
-  Counterparty counterparty = 2 [(gogoproto.nullable) = false];
+  string creator = 1;
+  // ibc.core.packetserver.v1.Counterparty counterparty = 2 [(gogoproto.nullable) = false];
```
crodriguezvega, line 259: Commented this out to prevent the circular dependency I mentioned above. We might need to move `Counterparty` to a different place.

crodriguezvega (commented):
A lot of changes had been done on the PR branch before it was synced with the feature branch, so I took me a bit of time to fix the conflicts and sync. I had to comment out some things (for which we can open an issue to keep track). I hope I didn't miss anything while fixing the conflicts. 🤞

proto/ibc/core/packetserver/v1/counterparty.proto
```diff
@@ -0,0 +1,22 @@
...
+
+option go_package = "github.com/cosmos/ibc-go/v9/modules/core/packet-server/types";
+
+import "gogoproto/gogo.proto";
+import "ibc/core/commitment/v2/commitment.proto";
+
+// Counterparty defines the counterparty for a light client to implement IBC eureka protocol
+// this value is stored under our side's channel ID. which we will use to write all packet messages
+// sent to counterparty
+message Counterparty {
```
colin-axner, line 13: Think it'd be nice to see this `Counterparty` type as something used by historical situations (channel/port), but not necessarily only for that

proto/ibc/core/packetserver/v1/tx.proto
```diff
@@ -0,0 +1,36 @@
...
+// MsgProvideCounterparty defines the message used to provide the counterparty client
+// identifier. Can only be invoked one time by the signer of MsgCreateClient if the counterparty
+// client identifier was not provided in the initial MsgCreateClient message.
+message MsgProvideCounterparty {
+  option (cosmos.msg.v1.signer) = "signer";
+
+  option (gogoproto.goproto_getters) = false;
+
+  // unique identifier we will use to write all packet messages sent to counterparty
+  string channel_id = 1;
```
colin-axner, line 28:
```suggestion
  string id = 1;
```
similar naming suggestion

modules/core/keeper/msg_server.go
```diff
@@ -143,10 +145,10 @@ func (k *Keeper) IBCSoftwareUpgrade(goCtx context.Context, msg *clienttypes.MsgI
 }
 
 // ProvideCounterparty defines a rpc handler method for MsgProvideCounterparty.
-func (k *Keeper) ProvideCounterparty(goCtx context.Context, msg *clienttypes.MsgProvideCounterparty) (*clienttypes.MsgProvideCounterpartyResponse, error) {
+func (k *Keeper) ProvideCounterparty(goCtx context.Context, msg *packetservertypes.MsgProvideCounterparty) (*packetservertypes.MsgProvideCounterpartyResponse, error) {
```
colin-axner, line 148: Lets initialize the packet sequence when setting the counterparty? This would allow us to remove the if's in send packet. I feel like calling `ProvideCounterparty` should be viewed as finalizing the communication details

modules/core/packet-server/types/keys.go
```diff
@@ -1,3 +1,11 @@
...
 
-const SubModuleName = "packetserver"
+const (
+	// SubModuleName defines the IBC packet server name
+	SubModuleName string = "packetserver"
+
+	// CounterpartyKey is the key used to store counterparty in the client store.
+	// the counterparty key is imported from types instead of host because
+	// the counterparty key is not a part of the ics-24 host specification
+	CounterpartyKey = "counterparty"
```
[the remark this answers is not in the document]
colin-axner: On second thought, if we are reusing the channel store, fine as is

colin-axner:
I think we should consider breaking up this pr for digestibility, I think there are some changes that should have been left for a followup.
1. Moving `Counterparty` type to packet server
2. Moving packet handlers out of keeper.go
3. Changing `Counterparty` structure
4. Implementing `GetV2Counterparty` usage
5. Ensuring no v1 packet code path disruption (channel upgrades)

proto/ibc/core/packetserver/v1/counterparty.proto
```diff
@@ -0,0 +1,22 @@
...
+  // the client identifier of the counterparty chain
+  // client id of the counterparty stored on our chain
+  string client_id = 1;
```
colin-axner, line 16: I think this can be removed now. I see this mapping as:

key: my identifier (v1: channel/port, v2: clientid)
value: counterparty id, merkle path

The client id used to verify the counterparty is in the key, or can be resolved in the v1 case. It seems redundant atm, I would guess in the actual implementation it would only be referenced by the grpc, but we should probably create a new struct for grpc, like we did with the other types `IdentifiedClientState` for example

colin-axner: really we have a mapping from: our reserved key path -> counterparty reserved key path

resolving the client id responsible for the key path, should happen outside of this mapping?

colin-axner: Before I forget, I went down the thought rabbit hole on this. The proposed construction works perfectly fine, but there's just some minor considerations one could have:

The current construction embeds the clientID into the `Counterparty`. When we handle v1 cases, the clientID is resolved and set into the counterparty type. In the v2 case, the value is redundant as it is in the key setting the value. The main benefit is code structure as it bundles all the necessary components into a single type. One could also return the clientID in the get counterparty function to achieve the same functionality. 

The main reason to keep this approach, would be for future scenarios where we delete the channel/connection mappings and you want to migrate such that all v1 mappings are directly embedded into the counterparty type. One could have two mappings, as I hint at in the above comment. One to obtain the counterparties keyspace which is set for all v1/v2 and one to obtain the light client either directly in v2 case, or via a lookup in v1 case. 

The reason why I went this rabbit hole was primarily due to naming. `counterparty.clientID` referring to your local client identifier, and not a client identifier on the counterparty, could be confusing. Alternatively, if you renamed the `Counterparty` to something like `Connection` or `Channel`, then `connection.clientId` and `connection.counterpartyID` could make sense. Either way, just a minor consideration.

AdityaSripal:
Closed and superceded by the feature branch with the merge of #7358

Issue #7190, opened 2024-08-20
Use expected errors instead of boolean values for 29fee


bznein:
ref: #7175

PR #7191, opened 2024-08-20
(chore): use expected errors in 29fee

bznein:
## Description

closes: #7190

---

write a little note why.

### DIFF
--- a/modules/apps/29-fee/ibc_middleware_test.go
+++ b/modules/apps/29-fee/ibc_middleware_test.go
@@ -2,7 +2,9 @@ package fee_test
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"strings"
 
 	errorsmod "cosmossdk.io/errors"
 	sdkmath "cosmossdk.io/math"
@@ -17,6 +19,7 @@ import (
 	channeltypes "github.com/cosmos/ibc-go/v9/modules/core/04-channel/types"
 	porttypes "github.com/cosmos/ibc-go/v9/modules/core/05-port/types"
 	host "github.com/cosmos/ibc-go/v9/modules/core/24-host"
+	ibcerrors "github.com/cosmos/ibc-go/v9/modules/core/errors"
 	"github.com/cosmos/ibc-go/v9/modules/core/exported"
 	ibctesting "github.com/cosmos/ibc-go/v9/testing"
 	ibcmock "github.com/cosmos/ibc-go/v9/testing/mock"
@@ -33,44 +36,44 @@ func (suite *FeeTestSuite) TestOnChanOpenInit() {
 	testCases := []struct {
 		name         string
 		version      string
-		expPass      bool
 		isFeeEnabled bool
+		expErr       error
 	}{
 		{
 			"success - valid fee middleware and mock version",
 			string(types.ModuleCdc.MustMarshalJSON(&types.Metadata{FeeVersion: types.Version, AppVersion: ibcmock.Version})),
 			true,
-			true,
+			nil,
 		},
 		{
 			"success - fee version not included, only perform mock logic",
 			ibcmock.Version,
-			true,
 			false,
-		},
-		{
-			"invalid fee middleware version",
-			string(types.ModuleCdc.MustMarshalJSON(&types.Metadata{FeeVersion: "invalid-ics29-1", AppVersion: ibcmock.Version})),
-			false,
-			false,
-		},
-		{
-			"invalid mock version",
-			string(types.ModuleCdc.MustMarshalJSON(&types.Metadata{FeeVersion: types.Version, AppVersion: "invalid-mock-version"})),
-			false,
-			false,
-		},
-		{
-			"mock version not wrapped",
-			types.Version,
-			false,
-			false,
+			nil,
 		},
 		{
 			"passing an empty string returns default version",
 			"",
 			true,
-			true,
+			nil,
+		},
+		{
+			"invalid fee middleware version",
+			string(types.ModuleCdc.MustMarshalJSON(&types.Metadata{FeeVersion: "invalid-ics29-1", AppVersion: ibcmock.Version})),
+			false,
+			types.ErrInvalidVersion,
+		},
+		{
+			"invalid mock version",
+			string(types.ModuleCdc.MustMarshalJSON(&types.Metadata{FeeVersion: types.Version, AppVersion: "invalid-mock-version"})),
+			false,
+			errors.New("incorrect mock version"),
+		},
+		{
+			"mock version not wrapped",
+			types.Version,
+			false,
+			errors.New("incorrect mock version"),
 		},
 	}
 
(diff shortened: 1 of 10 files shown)


modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -33,44 +36,44 @@ func (suite *FeeTestSuite) TestOnChanOpenInit() {
...
 			true,
-			true,
+			nil,
 		},
 		{
 			"success - fee version not included, only perform mock logic",
 			ibcmock.Version,
-			true,
 			false,
+			nil,
```
bznein, line 52: Moved this up as it's a success case

modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -135,7 +138,7 @@ func (suite *FeeTestSuite) TestOnChanOpenInit() {
 
 					suite.Require().NoError(err, "unexpected error from version: %s", tc.version)
 				} else {
-					suite.Require().Error(err, "error not returned for version: %s", tc.version)
+					suite.Require().True(errors.Is(err, tc.expErr) || strings.Contains(err.Error(), tc.expErr.Error()), "error no returned for version: %s or error %s is not %s", tc.version, err, tc.expErr)
```
bznein, line 141: This line will come up several times in the PR. It's ugly, but the issue is that some of the errors that we get (especially from `sdk`) are not wrapped, so we can't check for a specific error but rather we have to rely on the error string. It's not ideal, but the other option would be to catch every usage of those errors and wrap them ourselves with something else

modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -325,19 +328,16 @@ func (suite *FeeTestSuite) TestOnChanCloseInit() {
...
 		},
 		{
-			"application callback fails", func() {
-				suite.chainA.GetSimApp().FeeMockModule.IBCApp.OnChanCloseInit = func(
-					ctx sdk.Context, portID, channelID string,
-				) error {
-					return fmt.Errorf("application callback fails")
-				}
-			}, false,
+			"fee module is not enabled", func() {
```
bznein, line 337: I moved stuff around here to have success cases before failures

modules/apps/29-fee/keeper/grpc_query_test.go
```diff
@@ -24,7 +24,7 @@ func (suite *KeeperTestSuite) TestQueryIncentivizedPackets() {
 	testCases := []struct {
 		name     string
 		malleate func()
-		expPass  bool
+		errMsg   string
```
bznein, line 27: This file relies entirely on error messages as all the errors we get from GRPC queries are not wrapped

modules/apps/29-fee/keeper/grpc_query_test.go
```diff
@@ -643,16 +643,61 @@ func (suite *KeeperTestSuite) TestQueryCounterpartyPayee() {
...
 				suite.Require().Equal(expCounterpartyPayeeAddr.String(), res.CounterpartyPayee)
 			} else {
-				suite.Require().Error(err)
+				suite.Require().ErrorContains(err, tc.errMsg)
 			}
 		})
 	}
 }
 
+func (suite *KeeperTestSuite) TestQueryFeeEnabledChannelsWithPagination() {
```
bznein, line 656: This was really interesting. 

In this test, that was before a test case of another one, we create 10 extra channel, then get the first 5 from pagination and compare them.  However, when building them, we loop over an integer and create them. 

BUT, the logic that handles pagination paginates them based on their *lexicographical order*.
This means that if our channels are all after 10 (and before 100, so as to have 2 digits), the two orders are equal.

But in our case, depending on how many tests were ran before that, we could have the case in which our channels were 1,2,3,4,5,6,7,8,9,10, but the pagination would return 1, *10*, 3,4,5,6,7,8,9, making the test fail!

Thanks @chatton for helping me debug this!

modules/apps/29-fee/keeper/msg_server_test.go
```diff
@@ -21,90 +25,92 @@ func (suite *KeeperTestSuite) TestRegisterPayee() {
...
+			ibcerrors.ErrUnauthorized,
 		},
 	}
 
 	for _, tc := range testCases {
 		tc := tc
 
-		suite.SetupTest()
-		suite.path.Setup()
+		suite.Run(tc.name, func() {
```
bznein, line 69: There are a few instances in this PR where I introduced `t.Run` or `suite.Run`. It helped me debug things but I also think it's better to have it this way, even though it increases the diff. Happy to revert if people find it would be better suited for another PR

modules/apps/29-fee/types/fee_test.go
```diff
@@ -198,31 +201,33 @@ func TestPacketFeeValidation(t *testing.T) {
...
-		if tc.expPass {
-			require.NoError(t, err, tc.name)
-		} else {
-			require.Error(t, err, tc.name)
-		}
+			if tc.expErr == nil {
+				require.NoError(t, err, tc.name)
+			} else {
+				require.True(t, errors.Is(err, tc.expErr) || strings.Contains(err.Error(), tc.expErr.Error()), err.Error())
+			}
```
chatton, line 230: I wonder would it be worth while turning this into some helper within the tests, seems like we need it in a good few places where errors aren't wrapped.

chatton (approved):
LGTM! Thanks for going through all these, it ended up being quite a bit :1st_place_medal: 

Only suggestion is maybe adding a helper for that error checking condition, otherwise looks great!

modules/apps/29-fee/types/fee_test.go
bznein: yeah makes sense, done!

modules/apps/29-fee/ibc_middleware_test.go
```diff
@@ -135,7 +137,7 @@ func (suite *FeeTestSuite) TestOnChanOpenInit() {
 
 					suite.Require().NoError(err, "unexpected error from version: %s", tc.version)
 				} else {
-					suite.Require().Error(err, "error not returned for version: %s", tc.version)
+					ibctesting.RequireErrorIsOrContains(suite.T(), err, tc.expErr, "error no returned for version: %s or error %s is not %s", tc.version, err, tc.expErr)
```
crodriguezvega, line 140: You forgot maybe to push this new function?

crodriguezvega:
```suggestion
					ibctesting.RequireErrorIsOrContains(suite.T(), err, tc.expErr, "error not returned for version: %s or error %s is not %s", tc.version, err, tc.expErr)
```

modules/apps/29-fee/keeper/grpc_query_test.go
crodriguezvega: Very interesting! And thank you both for debugging and fixing it!

crodriguezvega (commented):
Looks like we are missing the function `RequireErrorIsOrContains`. Other than that it looks good; thanks again for continuing with these refactors!

crodriguezvega (approved):
🎉

Issue #7154, opened 2024-08-13
Use client and counterparty terminology in packet server rather than channel


colin-axner:
## Summary

The packet server for eureka uses channel terminology when it is actually referencing clients. Maybe this will be more useful when channel aliasing is added, but I would recommend renaming `ErrChannelNotFound` to `ErrCounterpartyNotFound` and `sourceChannel` to be `sourceClient`. The latter may be affected by channel aliasing

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chandiniv1:
@colin-axner can I work on this issue?

crodriguezvega:
Hi @chandiniv1. Yes, you can take it. Thank you!

Please use `feat/ibc-eureka` and the packet server that Colin refers to it's the folder `packet-server` in core. `ErrChannelNotFound` is part of 04-channel, so instead of renaming it I would use instead `ErrClientNotFound` that already exists in 02-client.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7195, opened 2024-08-21
Use client and counterparty terminology in packet server 

chandiniv1:
closes: #7154

### DIFF
--- a/modules/core/packet-server/keeper/keeper.go
+++ b/modules/core/packet-server/keeper/keeper.go
@@ -60,7 +60,7 @@ func (k Keeper) SendPacket(
 	// Lookup counterparty associated with our source channel to retrieve the destination channel
 	counterparty, ok := k.ClientKeeper.GetCounterparty(ctx, sourceChannel)
 	if !ok {
-		return 0, channeltypes.ErrChannelNotFound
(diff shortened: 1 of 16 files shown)


crodriguezvega:
Just took a quick look and there are unrelated changes in this PR. Can you try to sync with the feature branch?

modules/core/packet-server/keeper/keeper.go
```diff
@@ -59,7 +59,7 @@ func (k Keeper) SendPacket(
 	// Lookup counterparty associated with our source channel to retrieve the destination channel
 	counterparty, ok := k.ClientKeeper.GetCounterparty(ctx, sourceChannel)
 	if !ok {
-		return 0, channeltypes.ErrChannelNotFound
+		return 0, errorsmod.Wrap(clienttypes.ErrCounterpartyNotFound, sourceChannel)
```
crodriguezvega, line 62: We can add a more descriptive error string here when we decide what to do with channel aliasing. For now I kept it simple.

crodriguezvega (approved):
Thank you, @chandiniv1. 

When I was reviewing the PR I thought it would make more sense to add a new error for counterparty not found, so I pushed that commit.

Issue #7150, opened 2024-08-13
Address logger todo


colin-axner:
## Summary

See [TODO](https://github.com/cosmos/ibc-go/blob/2a07e6e060c8ba8778297528e9a826b6045e295e/modules/core/packet-server/keeper/keeper.go#L38)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned

chandiniv1:
@colin-axner can I work on this issue?

crodriguezvega:
Thank you, @chandiniv1! I think a submodule name like `packetserver` would work for now.

crodriguezvega:
Closed by #7196

PR #7196, opened 2024-08-21
add submodule to logger

chandiniv1:
closes: #7150

### DIFF
--- a/modules/core/packet-server/types/merkle.go
+++ b/modules/core/packet-server/types/merkle.go
@@ -5,6 +5,8 @@ import (
 	commitmenttypesv2 "github.com/cosmos/ibc-go/v9/modules/core/23-commitment/types/v2"
 )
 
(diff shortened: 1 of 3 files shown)


modules/core/packet-server/types/merkle.go
```diff
@@ -5,6 +5,8 @@ import (
 	commitmenttypesv2 "github.com/cosmos/ibc-go/v9/modules/core/23-commitment/types/v2"
 )
 
+const SubModuleName = "packetserver"
```
crodriguezvega, line 8: We could add a `keys.go` file in `types` and move this there. I think that would be a better place for now.

crodriguezvega (commented):
Thank you, @chandiniv1. Just left one comment!

crodriguezvega (approved):
Thank you for addressing the feedback!

Issue #7119, opened 2024-08-12
Encapsulate logic to retrieve handler and module based on protocol version


crodriguezvega:
In the packet rpc handlers we have a [case statement that determines the packet handler and module based on the protocol version](https://github.com/cosmos/ibc-go/blob/2a07e6e060c8ba8778297528e9a826b6045e295e/modules/core/keeper/msg_server.go#L569-L585). This code is pretty much the same in the rpc functions, so maybe there is an opportunity to encapsulate it into a helper function somehow.

chandiniv1:
@crodriguezvega can I work on this issue?

crodriguezvega:
Thank you, @chandiniv1. I had missed your message!

PR #7198, opened 2024-08-21
add helper to retrieve handler and module based on protocol version

chandiniv1:
closes: #7119


crodriguezvega (approved):
I pushed some changes, so it looks good to me now. This code might change again in the next weeks, but I think for now this is good enough. Thanks!

PR #7303, opened 2024-09-17
add height field to gRPC requests

noot:
## Description

Adds `height` field to relevant gRPC requests and `latest_height` bool to override height and query at the latest height.

Following up to my previous PR #5685 - this came up as an issue again when testing IBC with the hermes relayer, as only proofs at the latest height were returned, but this is not the state root we were expecting to compare against, causing verification errors. 

Let me know any thoughts/feedback on this approach, if it looks good I can update the rest of the code.

closes: https://github.com/cosmos/ibc-go/issues/149

---

write a little note why.


damiannolan:
Hi @noot, thanks for opening the PR. I have a couple of questions:
- How would these fields intend to be used inside of the ibc-go implementation code?
- Are you using a different implementation of ibc where you are seeing the errors mentioned?

> this came up as an issue again when testing IBC with the hermes relayer, as only proofs at the latest height were returned, but this is not the state root we were expecting to compare against, causing verification errors.

I think I mentioned in the previous PR that the intended approach for querying state at historical heights is to the [`x-cosmos-block-height`](https://github.com/cosmos/cosmos-sdk/blob/main/types/grpc/headers.go#L5). It is up to the implementation of the application to consume this correctly, and load a version of the db store at the given height within this header.

erwanor:
Unsurprisingly, we ran into this (non-Cosmos SDK). And it was a massive pain point.

erwanor:
@damiannolan I think it would be much better to not have to stuff the gRPC headers with secret API arguments. Instead, we should change the messages so that they can model a useful RPC protocol. That's how the `QueryConsensusStateRequest` messages are already laid out:

```protobuf
/// QueryConsensusStateRequest is the request type for the Query/ConsensusState
// RPC method. Besides the consensus state, it includes a proof and the height
// from which the proof was retrieved.
message QueryConsensusStateRequest {
  // client identifier
  string client_id = 1;
  // consensus state revision number
  uint64 revision_number = 2;
  // consensus state revision height
  uint64 revision_height = 3;
  // latest_height overrides the height field and queries the latest stored
  // ConsensusState
  bool latest_height = 4;
}
```

It is very useful to be able to obtain commitment proofs for an arbitrary height, and not just the latest one. This is not just a source of bugs, but also unlocks future work. Is there anything we could do to get this reviewed and merged?

damiannolan:
Thanks for input @erwanor. And apologies that this PR has fallen off the radar for so long. I will try to prioritise getting eyes on this from the team. 

Personally, I'm happy to add the additional arguments to requests if it relieves pain points. It will need to be clearly documented that fulfilling the fields has no affect for cosmos-sdk based nodes. The baseapp in cosmos-sdk controls loading the versioned tree state before the req hits the intended rpc handler. 

IMHO the best long term approach to standardising the protobuf definitions is moving them outside of ibc-go and into cosmos/ibc.

erwanor:
Thanks, appreciate it.

>IMHO the best long term approach to standardizing the protobuf definitions is moving them outside of ibc-go and into cosmos/ibc.

Yes, agreed. I think this would be better, because we are slightly past an ecosystem zero-to-one moment in terms of having multiple real world production IBC implementation existing. So this will become more important to do as time goes on. Happy to help push for this and support it with review time, or otherwise.

rnbguy:
For situations like this, at ibc-rs, we maintain the domain types with an optional height field. When we convert ibc-proto types to our domain types -- we just keep it `None` and follow ibc-go's implementation. ([ref](https://docs.rs/ibc-query/latest/src/ibc_query/core/channel/types/request.rs.html#138-156))

This year, when `ibc-rs` team worked on IBC integration for sovereign-sdk, we [used these domain types directly](https://github.com/informalsystems/sovereign-ibc/blob/83b5c91a/crates/modules/sov-ibc/src/rpc/methods.rs#L55) to expose the height as an argument.

DimitrisJim:
note: eureka grpcs would need to add these too.

erwanor:
@DimitrisJim good idea, where does eureka dev happen?

DimitrisJim:
@erwanor currently in https://github.com/cosmos/ibc-go/pull/7505 most grpcs for channel/V2 have been added at this point.

I'll open up an issue to not lose track of this!

erwanor:
I don't understand how the reaction to "this is a massive pain point" is to let this PR bitrot not once, but twice, and then close it?

Issue #7366, opened 2024-10-01
Add store functions to the ChannelKeeper V2 that use the new provable paths


chatton:
The `ChannelKeeper` V2 should include new functions which fetch and write to the store using the new paths outlined in the spec.

e.g. the equivalent of `channelkeeperv1.SetPacketCommitment`

ref https://github.com/cosmos/ibc/pull/1148

chatton:
Linked issue has been merged, we should be good to close!

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7378, opened 2024-10-02
Add new keys and getter / setter functions

### LINKED ISSUE #7132: Change packet key to support efficient iteration in v2
## Summary

Change packet keys to use big endian encodings on the sequence to support efficient iteration

## Problem Definition

Packet keys which have sequences which are lexographically ordered are not efficient to iterate and cause issues later on which are hard to workaround. See https://github.com/cosmos/ibc-go/issues/5413 for example. This will also be problematic if we tried to create a pruning solution as we cannot easily iterate. We would need to do individual deletions or queries to the store. 

## Proposal

The "Eureka" version of IBC introduces a V2 to the packet commitment structure. Given that we are changing how we are committing to packets, we might as well change how we are storing these packet commitments, so that we can future proof the protocol and make future features, like pruning, much easier to implement

I propose changing the packet key structure to be:

`clients/<client-ID/sequences/<big-endian-encoded-sequence>`

Currently we use:
`ports/<port-ID>/channels/<channel-id>/sequences/<sequence>`

I think the port ID should be moved to the commitment value (this would be in-line with multi-packetdata) and I think we should be more explicit in the future approach that we are using clientIDs not channelIDs, but I haven't given these additional points much thought, so I am happy to exclude them. The big endian encoding of the sequence should definitely be done

____

#### For Admin Use

- [ ] Not duplicate issue

chatton:
## Description

NOTE: explicitly not closing linked issue, as these fields are subject to change pending final spec.

With these we can continue with implementations of new handler functions.

ref: #7366

closes #7132

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/v2/keeper/keeper.go
+++ b/modules/core/04-channel/v2/keeper/keeper.go
@@ -66,7 +66,8 @@ func (k *Keeper) GetCounterparty(ctx context.Context, clientID string) (types.Co
 // GetPacketReceipt returns the packet receipt from the packet receipt path based on the sourceID and sequence.
 func (k *Keeper) GetPacketReceipt(ctx context.Context, sourceID string, sequence uint64) (string, bool) {
 	store := k.storeService.OpenKVStore(ctx)
-	bz, err := store.Get(hostv2.PacketReceiptKey(sourceID, sequence))
+	bigEndianBz := sdk.Uint64ToBigEndian(sequence)
+	bz, err := store.Get(hostv2.PacketReceiptKey(sourceID, bigEndianBz))
 	if err != nil {
 		panic(err)
 	}
@@ -80,7 +81,8 @@ func (k *Keeper) GetPacketReceipt(ctx context.Context, sourceID string, sequence
 // This is a public path that is standardized by the IBC V2 specification.
 func (k *Keeper) SetPacketReceipt(ctx context.Context, sourceID string, sequence uint64) {
 	store := k.storeService.OpenKVStore(ctx)
-	if err := store.Set(hostv2.PacketReceiptKey(sourceID, sequence), []byte{byte(1)}); err != nil {
+	bigEndianBz := sdk.Uint64ToBigEndian(sequence)
+	if err := store.Set(hostv2.PacketReceiptKey(sourceID, bigEndianBz), []byte{byte(1)}); err != nil {
 		panic(err)
 	}
 }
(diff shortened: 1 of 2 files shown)


modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"github.com/cosmos/ibc-go/v9/modules/core/04-channel/v2/types"
 	host "github.com/cosmos/ibc-go/v9/modules/core/24-host"
+	hostv2 "github.com/cosmos/ibc-go/v9/modules/core/24-host/v2"
```
chatton, line 18: new keys have been added under a host/v2

modules/core/24-host/v2/packet_keys.go
```diff
@@ -0,0 +1,24 @@
+package v2
+
+import "fmt"
+
+// PacketReceiptKey returns the store key of under which a packet
+// receipt is stored
+func PacketReceiptKey(sourceID string, sequence uint64) []byte {
+	return []byte(fmt.Sprintf("receipts/channels/%s/sequences/%d", sourceID, sequence))
```
chatton, line 8: we can introduce utility functions if needed, but I didn't think it was necessary here

modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -61,3 +62,96 @@ func (k *Keeper) GetCounterparty(ctx context.Context, clientID string) (types.Co
 	k.cdc.MustUnmarshal(bz, &counterparty)
 	return counterparty, true
 }
+
+// GetPacketReceipt returns the packet receipt from the packet receipt path based on the sourceID and sequence.
+func (k *Keeper) GetPacketReceipt(ctx context.Context, sourceID string, sequence uint64) (string, bool) {
```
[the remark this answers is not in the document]
sangier: Sequences should be in big endian as specified by @DimitrisJim. Is there a way to specify that in the go code? Which is the default option in go?

modules/core/24-host/v2/packet_keys.go
damiannolan, line 1: With new keys we should probably consider how we format the sequence in order to support effecient iteration, right?
I think this was an issue before, as formatting ended up in lexographical ordering rather than numerical(?)

E.g. format sequence using big endian bytes rather than decimal string

damiannolan (commented):
Nice, looks super clean! I thought it would be good to raise the question around sequence formatting here and see what folks think

modules/core/24-host/v2/packet_keys.go
damiannolan: Should've refreshed my browser before publishing review... 😆

modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -61,3 +62,103 @@ func (k *Keeper) GetCounterparty(ctx context.Context, clientID string) (types.Co
...
+		panic(err)
+	}
+
+	return found
+}
+
+// GetPacketCommitment returns the packet commitment hash under the commitment path.
+func (k *Keeper) GetPacketCommitment(ctx context.Context, sourceID string, sequence uint64) (string, bool) {
+	store := k.storeService.OpenKVStore(ctx)
+	bigEndianBz := sdk.Uint64ToBigEndian(sequence)
```
damiannolan, line 115: don't want to be nitting too hard but might be nicer to do this inside `host` pkg. API looks friendlier with a uint64 sequence, and the formatting is a `host` responsibility imo 🤷🏻

damiannolan (approved):
ACK! Thanks @chatton!

Issue #7369, opened 2024-10-01
Implement MessageServerV2 Acknowledgement on ChannelKeeperV2


chatton:
depends on: https://github.com/cosmos/ibc-go/issues/7352

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7412, opened 2024-10-08
feat: add acknowledgePacket handler to channel/v2

damiannolan:
## Description

- Adds `acknowledgePacket` handler to channel/v2
- Refactored commitment funcs to `commitment.go` in channel/v2/types

closes: #7369 

---

write a little note why.


modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -199,3 +199,15 @@ func (k *Keeper) AliasV1Channel(ctx context.Context, portID, channelID string) (
 	}
 	return counterparty, true
 }
+
+// getV1Counterparty attempts to retrieve a v1 channel from the channel keeper if it exists, then converts it
```
damiannolan, line 203: moved from relay.go to keeper.go

modules/core/04-channel/v2/keeper/packet.go
damiannolan, line 1: renamed relay.go to packet.go

modules/core/04-channel/v2/types/commitment.go
damiannolan, line 1: created commitment.go to house commitment funcs

modules/core/04-channel/v2/keeper/packet.go
```diff
@@ -173,6 +162,62 @@ func (k Keeper) recvPacket(
...
+
+	packetCommitment := channeltypesv2.CommitPacket(packet)
+
+	// verify we sent the packet and haven't cleared it out yet
+	if !bytes.Equal(commitment, packetCommitment) {
+		return errorsmod.Wrapf(channeltypes.ErrInvalidPacket, "commitment bytes are not equal: got (%v), expected (%v)", packetCommitment, commitment)
+	}
+
+	path := hostv2.PacketAcknowledgementKey(packet.DestinationId, sdk.Uint64ToBigEndian(packet.Sequence))
+	merklePath := types.BuildMerklePath(counterparty.MerklePathPrefix, path)
```
damiannolan, line 198: `BuildMerklePath`, used here and other funcs in this file is imported from the packet-server package.

We should move this to the channel/v2 types pkg afaik. We should not depend on anything in packet-server (which will ultimately be removed).

I'd also like some insight into the implementation of `BuildMerklePath`.
Some questions I have are:
Why are we not using the `MerklePrefix` type for the "ibc" prefix used by counterparties - instead we're using `MerklePath`. 
Secondly, implementation looks complex and kinda funky, haven't stepped through to see exactly why that is, was hoping someone could explain it for me.

modules/core/04-channel/v2/keeper/packet.go
```diff
@@ -173,6 +162,62 @@ func (k Keeper) recvPacket(
 	return nil
 }
 
+func (k Keeper) acknowledgePacket(ctx context.Context, packet channeltypesv2.Packet, acknowledgement channeltypesv2.Acknowledgement, proof []byte, proofHeight exported.Height) error {
+	// Lookup counterparty associated with our channel and ensure
+	// that the packet was indeed sent by our counterparty.
+	counterparty, ok := k.GetCounterparty(ctx, packet.SourceId)
```
damiannolan, line 168: this will become a v2 channelEnd, and below `clientID := counterparty.ClientID` will read better... right now its slightly confusing. 

The channelEnd.ClientID would suggest to me that it's the clientID of the client representing the counterparty state machine. Rather than a counterparty.ClientID which would suggest to me its the clientID on the counterparty which repesents _this_ chain (self host where this code is being executed)

damiannolan (commented):
Leaving some comments for the reader!

modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -75,17 +75,17 @@ func (k *Keeper) GetCounterparty(ctx context.Context, clientID string) (types.Co
 }
 
 // GetPacketReceipt returns the packet receipt from the packet receipt path based on the sourceID and sequence.
-func (k *Keeper) GetPacketReceipt(ctx context.Context, sourceID string, sequence uint64) (string, bool) {
+func (k *Keeper) GetPacketReceipt(ctx context.Context, sourceID string, sequence uint64) []byte {
```
chatton, line 78: why the removal of the `bool` return value?

modules/core/04-channel/v2/keeper/packet.go
chatton: yeah the naming is a bit all over the place at the moment with regards to counterparty/channel/ids etc, hopefully we'll be able to sort this out quite soon!

modules/core/04-channel/v2/keeper/packet.go
```diff
@@ -173,6 +162,62 @@ func (k Keeper) recvPacket(
...
+		channeltypesv2.CommitAcknowledgement(acknowledgement),
+	); err != nil {
+		return errorsmod.Wrapf(err, "failed packet acknowledgement verification for client (%s)", clientID)
+	}
+
+	k.DeletePacketCommitment(ctx, packet.SourceId, packet.Sequence)
+
+	k.Logger(ctx).Info("packet acknowledged", "sequence", strconv.FormatUint(packet.GetSequence(), 10), "source_id", packet.GetSourceId(), "destination_id", packet.GetDestinationId())
+
+	EmitAcknowledgePacketEvents(ctx, packet, acknowledgement)
```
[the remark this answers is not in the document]
chatton: I think that would make the most sense yeah, can do in a follow up though I think!

modules/core/04-channel/v2/keeper/keeper.go
chatton: Looking at the usage later, I guess technically i provides no added value as we can do a len check in line, and also we would never be storing an empty bz array so that case does not need to be differentiated. I also see we do it this way in the channel v1 stuff, will leave the comment here in case anyone else was wondering!

modules/core/04-channel/v2/keeper/packet.go
chatton: re dependencies: I think we can do a pass at the end and make sure there is no dependency on v1 anything within the channel/v2 package.

chatton (approved):
LGTM :saluting_face:

modules/core/04-channel/v2/keeper/keeper.go
damiannolan: Yeah, might change this back to be honest. As v1 has the bool return value for receipts (but not commitments - which I also changed). 

Actually had a todo which I forgot to address here where the usage should be `HasReceipt` I think. I can push an update before merging!

modules/core/04-channel/v2/keeper/packet.go
damiannolan: would be nice to get this rename done and dusted sooner rather than later methinks!

modules/core/04-channel/v2/keeper/packet.go
damiannolan: cc. @AdityaSripal @DimitrisJim on this func

modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -75,16 +75,27 @@ func (k *Keeper) GetCounterparty(ctx context.Context, clientID string) (types.Co
...
 	if len(bz) == 0 {
-		return "", false
+		return nil, false
 	}
-	return string(bz), true
+	return bz, true
+}
+
+// HasPacketRceipt returns true if the packet receipt exists, otherwise false.
+func (k *Keeper) HasPacketReceipt(ctx context.Context, sourceID string, sequence uint64) bool {
```
bznein, line 91: Broader question but I take this line to bring it up: given https://github.com/cosmos/ibc-go/pull/7428, do we want to also use `sourceChannel` and `destinationChannel` in all function arguments?

damiannolan: Indeed we do, will update when others are merged

Issue #7356, opened 2024-10-01
Implement v2 `Acknowledgement` rpc handler


damiannolan:
Requires v2 `IBCModule` design and clarity regarding new acknowledgement type and commitment structure

bznein:
> done in [#7452](https://github.com/cosmos/ibc-go/pull/7452) it seems? Re-open if mistaken!

Yeah I keep forgetting that "closes" keyword doesn't work for PRs not targeting main

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7452, opened 2024-10-14
chore: Implement v2 Acknowledgement rpc handler 

bznein:
## Description

closes: #7356

---

write a little note why.


testing/utils.go
```diff
@@ -87,6 +87,7 @@ func UnmarshalMsgResponses(cdc codec.Codec, data []byte, msgs ...codec.ProtoMars
 // RequireErrorIsOrContains verifies that the passed error is either a target error or contains its error message.
 func RequireErrorIsOrContains(t *testing.T, err, targetError error, msgAndArgs ...interface{}) {
 	t.Helper()
+	require.Error(t, err)
```
chatton, line 90: think this was removed in the recv PR, we don't need this since `errors.Is` accounts for a `nil` value.

modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -66,19 +66,20 @@ func (k *Keeper) Acknowledgement(ctx context.Context, msg *channeltypesv2.MsgAck
...
-	//	err := cbs.OnSendPacket(ctx, msg.SourceId, sequence, msg.TimeoutTimestamp, pd, signer)
-	//	if err != nil {
-	//		return nil, err
-	//	}
-	// }
+	for _, pd := range msg.Packet.Data {
+		cbs := k.Router.Route(pd.SourcePort)
+		err := cbs.OnAcknowledgementPacket(ctx, msg.Packet.SourceChannel, msg.Packet.DestinationChannel, pd, recvResults[pd.DestinationPort].Acknowledgement, relayer)
+		if err != nil {
+			return nil, err
```
chatton, line 78: maybe we should wrap this error with source/dest channel ID and source port to specify which app failed.

modules/core/04-channel/v2/keeper/msg_server_test.go
```diff
@@ -337,3 +337,119 @@ func (suite *KeeperTestSuite) TestProvideCounterparty() {
...
+			res, err := path.EndpointA.Chain.SendMsgs(msgSendPacket)
+			suite.Require().NoError(err)
+			suite.Require().NotNil(res)
+
+			suite.Require().NoError(path.EndpointB.UpdateClient())
+			recvPacket = channeltypesv2.NewPacket(1, path.EndpointA.ChannelID, path.EndpointB.ChannelID, timeoutTimestamp, mockv2.NewMockPacketData(mockv2.ModuleNameA, mockv2.ModuleNameB))
+			// get proof of packet commitment from chainA
+			packetKey := hostv2.PacketCommitmentKey(recvPacket.SourceChannel, recvPacket.Sequence)
+			proof, proofHeight := path.EndpointA.QueryProof(packetKey)
+			msgRecvPacket := channeltypesv2.NewMsgRecvPacket(recvPacket, proof, proofHeight, suite.chainB.SenderAccount.GetAddress().String())
```
chatton, line 420: [nit] maybe add some spacing in between the different sections, send/recv/ack etc.

modules/core/04-channel/v2/keeper/msg_server_test.go
```diff
@@ -337,3 +337,119 @@ func (suite *KeeperTestSuite) TestProvideCounterparty() {
...
+		expError error
+	}{
+		{
+			name:     "success",
+			malleate: func() {},
+		},
+		{
+			name: "success: NoOp",
+			malleate: func() {
+				suite.chainA.App.GetIBCKeeper().ChannelKeeperV2.SetPacketCommitment(suite.chainA.GetContext(), recvPacket.SourceChannel, recvPacket.Sequence, channeltypesv2.CommitPacket(recvPacket))
```
chatton, line 359: might be nice to verify app callback wasn't executed, maybe not executed. We could do that by toggling a bool in the ack app callback, but would require a slight refactor to the test to allow custom assertions post test, will leave it to you if you think it's worth it.

testing/utils.go
bznein: The problem is that if `errors.Is` fails, we go to the second part of the `||` and we try to call `.Error()` on `err`, causing a panic.

This only happens when the test is broken and we are checking for an error that doesn't happen, so unlikely to be checked in on `main`, but I think it's still worth having it.

modules/core/04-channel/v2/keeper/msg_server_test.go
bznein: Interestingly, this makes the test fail! I think it's either my setpacketcommitment that is wrong, or (maybe) an error in the code, thanks for the idea!

bznein: Instead of changing the structure of the code, I modified the callback to return an error, so that we can simply check that an error is not returned, WDYT? (the comment explains it better :D)

Issue #7390, opened 2024-10-03
Check that V2 Packet Data has length of 1


bznein:
Surfaced [here](https://github.com/cosmos/ibc-go/pull/7389/files#r1785964485) by @chatton:

> I think we want this to be the final state for validate basic for the initial release, but in order for mock applications to be testable with multiple packet datas (which I still think is worthwhile doing ), maybe we can leave this check commented out and create an issue to revert it. (or we could just fail on == 0 but create an issue to change)

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7484, opened 2024-10-21
chore: limit packet data to length of one.

DimitrisJim:
## Description

in the rest of the validation funcs

closes: #7390 

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/v2/types/msgs_test.go
+++ b/modules/core/04-channel/v2/types/msgs_test.go
@@ -176,6 +176,13 @@ func (s *TypesTestSuite) TestMsgSendPacketValidateBasic() {
 		{
 			name: "failure: invalid length for payload",
 			malleate: func() {
+				msg.Payloads = []types.Payload{{}, {}}
+			},
+			expError: types.ErrInvalidPayload,
+		},
+		{
+			name: "failure: invalid packetdata",
+			malleate: func() {
 				msg.Payloads = []types.Payload{}
 			},
 			expError: types.ErrInvalidPayload,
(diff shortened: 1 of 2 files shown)


modules/core/04-channel/v2/types/msgs.go
```diff
@@ -164,6 +168,10 @@ func (msg *MsgAcknowledgement) ValidateBasic() error {
 		return errorsmod.Wrap(commitmenttypesv1.ErrInvalidProof, "cannot submit an empty acknowledgement proof")
 	}
 
+	if len(msg.Packet.Data) != 1 {
+		return errorsmod.Wrapf(ErrInvalidPayload, "packet data must be of length 1, got %d instead", len(msg.Packet.Data))
+	}
```
chatton, line 173: [q] why not just have this check in `Packet.ValidateBasic()`? we don't do other validation on the packet  directly here.

Just checked there actually, we already do this in the packet fn, the tests should still pass without these checks I think

chatton: ah right I get you, yeah I guess I was thinking we would move the check to the packet.ValidateBasic(), and tests should pass once that is done. I think it's okay with the varargs as long as validate basic is != 1 and then eventually > 0.

chatton: you can still always construct an invalid packet with the constructor, so I wouldn't be too concerned about allowing the construction of one with an empty payload list if it gets validated in ValidateBasic

damiannolan (approved):

PR #7518, opened 2024-10-25
chore(docs): delete not exist file path

odyslaeth:
Hi there, the file is not exist. so I delete the path

### DIFF
--- a/docs/versioned_docs/version-v7.8.x/03-light-clients/03-localhost/03-client-state.md
+++ b/docs/versioned_docs/version-v7.8.x/03-light-clients/03-localhost/03-client-state.md
@@ -33,6 +33,7 @@ It is possible to disable the localhost client by removing the `09-localhost` en
 
 The latest height is updated periodically through the ABCI [`BeginBlock`](https://docs.cosmos.network/v0.47/building-modules/beginblock-endblock) interface of the 02-client submodule in core IBC.
 
+[See `BeginBlocker` in abci.go.](https://github.com/cosmos/ibc-go/blob/v7.8.0/modules/core/02-client/abci.go#L12)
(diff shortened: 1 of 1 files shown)


docs/versioned_docs/version-v7.8.x/03-light-clients/03-localhost/03-client-state.md
```diff
@@ -33,8 +33,6 @@ It is possible to disable the localhost client by removing the `09-localhost` en
 
 The latest height is updated periodically through the ABCI [`BeginBlock`](https://docs.cosmos.network/v0.47/building-modules/beginblock-endblock) interface of the 02-client submodule in core IBC.
 
-[See `BeginBlocker` in abci.go.](https://github.com/cosmos/ibc-go/blob/09-localhost/modules/core/02-client/abci.go#L12)
```
damiannolan, line 36: Can we change the link to https://github.com/cosmos/ibc-go/blob/v7.8.0/modules/core/02-client/abci.go#L12

damiannolan (changes requested):
Hi thanks for the PR @odyslaeth! 

Given that this is versioned documentation, for previous release lines, it would be great if we could update the link to reflect the correct version. It looks like it previously was pointing to a feat branch. I've linked the correct code link in the PR.

Note this is only relevant to older release lines of ibc-go. As the localhost client is now stateless and does not require tracking updates via the BeginBlocker lifecycle handler.

odyslaeth:
NP Done thanks for the catching @damiannolan

damiannolan (approved):

damiannolan:
Thank you @odyslaeth!

Issue #7337, opened 2024-09-24
Enforce Max timeout value for all packets


DimitrisJim:
Raised during aliasing walkthrough.

To not potentially have lingering packets for too long IIRC. Note mentions a minimum might make sense too.

Constant in ibc-go
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7521, opened 2024-10-28
Add MaxTimeout condition

sangier:
## Description

Add MaxTimeout condition in the SendPacket function as specified in [004](https://github.com/cosmos/ibc/blob/d234b5be813f412f8fb47e234ccb1d542f0995a6/spec/core/v2/ics-004-channel-and-packet-semantics/README.md#:~:text=//%20disallow%20packet%20with%20timeoutTimestamp%20less%20than%20currentTimestamp%20and%20timeoutTimestamp%20value%20bigger%20than%20currentTimestamp%20%2B%20MaxTimeoutDelta%20%0A%20%20%20%20assert(currentTimestamp()%20%3C%20timeoutTimestamp%20%3C%20currentTimestamp()%20%2B%20MAX_TIMEOUT_DELTA))

closes: https://github.com/cosmos/ibc-go/issues/7337

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/v2/keeper/msg_server.go
+++ b/modules/core/04-channel/v2/keeper/msg_server.go
@@ -3,6 +3,7 @@ package keeper
 import (
 	"context"
 	"slices"
+	"time"
 
 	errorsmod "cosmossdk.io/errors"
 
@@ -71,14 +72,15 @@ func (k *Keeper) SendPacket(ctx context.Context, msg *channeltypesv2.MsgSendPack
 	}
 
 	// Note, the validate basic function in sendPacket does the timeoutTimestamp != 0 check and other stateless checks on the packet.
-	// timeoutTimestamp must be less than current block time + MaxTimeoutDelta
-	if msg.TimeoutTimestamp > uint64(sdkCtx.BlockTime().Unix())+channeltypesv2.MaxTimeoutDelta {
-		return nil, errorsmod.Wrap(channeltypesv2.ErrMaxTimeoutDeltaExceeded, "timeout exceeds the maximum expected value")
+	// timeoutTimestamp must be greater than current block time
+	timeout := time.Unix(int64(msg.TimeoutTimestamp), 0)
+	if timeout.Before(sdkCtx.BlockTime()) {
+		return nil, errorsmod.Wrap(channeltypesv2.ErrTimeoutTooLow, "timeout is less than the current block timestamp")
 	}
 
-	// timeoutTimestamp must be greater than current block time
-	if uint64(sdkCtx.BlockTime().Unix()) > msg.TimeoutTimestamp {
-		return nil, errorsmod.Wrap(channeltypesv2.ErrTimeoutTooLow, "timeout is less than the current block timestamp")
+	// timeoutTimestamp must be less than current block time + MaxTimeoutDelta
+	if timeout.After(sdkCtx.BlockTime().Add(channeltypesv2.MaxTimeoutDelta)) {
+		return nil, errorsmod.Wrap(channeltypesv2.ErrMaxTimeoutDeltaExceeded, "timeout exceeds the maximum expected value")
 	}
 
 	signer, err := sdk.AccAddressFromBech32(msg.Signer)
--- a/modules/core/04-channel/v2/keeper/msg_server.go
+++ b/modules/core/04-channel/v2/keeper/msg_server.go
@@ -63,10 +63,16 @@ func (k *Keeper) RegisterCounterparty(goCtx context.Context, msg *channeltypesv2
 // SendPacket defines a rpc handler method for MsgSendPacket.
 func (k *Keeper) SendPacket(ctx context.Context, msg *channeltypesv2.MsgSendPacket) (*channeltypesv2.MsgSendPacketResponse, error) {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
 	sequence, destChannel, err := k.sendPacket(ctx, msg.SourceChannel, msg.TimeoutTimestamp, msg.Payloads)
 	if err != nil {
 		sdkCtx.Logger().Error("send packet failed", "source-channel", msg.SourceChannel, "error", errorsmod.Wrap(err, "send packet failed"))
 		return nil, errorsmod.Wrapf(err, "send packet failed for source id: %s", msg.SourceChannel)
+	}
+
+	// Note, the validate basic function in sendPacket does the TimeoutTimestamp != 0 check and other stateless checks on the packet.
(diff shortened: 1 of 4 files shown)


modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -63,12 +63,18 @@ func (k *Keeper) RegisterCounterparty(goCtx context.Context, msg *channeltypesv2
...
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
 	sequence, destChannel, err := k.sendPacket(ctx, msg.SourceChannel, msg.TimeoutTimestamp, msg.Payloads)
 	if err != nil {
 		sdkCtx.Logger().Error("send packet failed", "source-channel", msg.SourceChannel, "error", errorsmod.Wrap(err, "send packet failed"))
 		return nil, errorsmod.Wrapf(err, "send packet failed for source id: %s", msg.SourceChannel)
 	}
 
+	// Note, the validate basic function in sendPacket does the TimeoutTimestamp != 0 check and other stateless checks on the packet.
+	if uint64(sdkCtx.BlockTime().Unix()) > msg.TimeoutTimestamp || msg.TimeoutTimestamp > uint64(sdkCtx.BlockTime().Unix())+channeltypesv2.MaxTimeout {
```
AdityaSripal, line 74: These are separate conditions. In the first case the msg timeout timestamp is too low. But in the second case, it is too high. But the error case only says it exceeded maximum value. Split these up so you have correct error message in each case

modules/core/04-channel/v2/keeper/msg_server_test.go
```diff
@@ -124,6 +134,14 @@ func (suite *KeeperTestSuite) TestMsgSendPacket() {
 			},
 			expError: channeltypesv1.ErrTimeoutElapsed,
 		},
+		{
+			name: "failure: timeout timestamp exceeds max allowed input",
```
AdityaSripal, line 137: Let's also add an error case for when its too low

modules/core/04-channel/v2/types/msgs.go
```diff
@@ -13,6 +13,8 @@ import (
 	ibcerrors "github.com/cosmos/ibc-go/v9/modules/core/errors"
 )
 
+const MaxTimeout uint64 = 86400 // 24h in seconds
```
AdityaSripal, line 16: Should we call this `MaxTimeoutDelta` instead? So that it is not interpreted as an absolute value?

Also might be better to be explicit:

```suggestion
const MaxTimeoutDelta uint64 = 24 * 60 * 60 // 24h in seconds
```

AdityaSripal (changes requested):

AdityaSripal (approved):

modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -63,12 +63,24 @@ func (k *Keeper) RegisterCounterparty(goCtx context.Context, msg *channeltypesv2
...
+	// Note, the validate basic function in sendPacket does the timeoutTimestamp != 0 check and other stateless checks on the packet.
+	// timeoutTimestamp must be less than current block time + MaxTimeoutDelta
+	if msg.TimeoutTimestamp > uint64(sdkCtx.BlockTime().Unix())+channeltypesv2.MaxTimeoutDelta {
+		return nil, errorsmod.Wrap(channeltypesv2.ErrMaxTimeoutDeltaExceeded, "timeout exceeds the maximum expected value")
+	}
+
+	// timeoutTimestamp must be greater than current block time
+	if uint64(sdkCtx.BlockTime().Unix()) > msg.TimeoutTimestamp {
+		return nil, errorsmod.Wrap(channeltypesv2.ErrTimeoutTooLow, "timeout is less than the current block timestamp")
+	}
```
damiannolan, line 82: since sdk block time is a `time.Time`, we could leverage the `time.After()` / `time.Before()` apis from std lib here, by creating a `time.Time` with the absolute timeout. Would imagine it to be something like this:

```golang
// ------------------------------------------------
const timeoutDelta time.Duration = 24 * time.Hour
// ------------------------------------------------

timeout := time.Unix(int64(msg.TimeoutTimestamp), 0)
if timeout.Before(sdkCtx.BlockTime()) {
	// err timeout less than current block timestamp
}

if timeout.After(sdkCtx.BlockTime().Add(timeoutDelta)) {
	// err timeout exceeds max exp value
}
```

I'm fine with using uint64 absolute values too. But it might look a little less verbose to leverage the stdlib apis. Interested on input from others!

damiannolan (approved):
I left a suggestion but happy to go either way. Would like to hear what others think!

modules/core/04-channel/v2/keeper/msg_server.go
AdityaSripal: i think we should use the time package as well when possible

sangier: Find out that using the time package is the idiomatic choice in go. My bad. Changes applied as suggested!

modules/core/04-channel/v2/types/msgs.go
```diff
@@ -13,6 +15,8 @@ import (
 	ibcerrors "github.com/cosmos/ibc-go/v9/modules/core/errors"
 )
 
+const MaxTimeoutDelta time.Duration = 24 * time.Hour
```
womensrights, line 18: why was 24 hours chosen?

sangier: This decision came after discussion with Adi. While we have a median time for MsgReceivePacket of 17s and for Ack 11s, we decided to set the timeout to 24h to take in consideration potential relayers/Sp1 network downtimes. Especially relayers in some cases have gone down for periods of time of 2-4 hours. 

In case you have a different opinion/guidance on why this should be different we can revisit this choice.

DimitrisJim: there's https://github.com/cosmos/ibc-go/issues/7530 that is pertinent too

womensrights: I think 24 hours is excessive, its the time your tokens can be left in limbo and you can't really do anything about it, I would say something more like 6 hours is even being generous

DimitrisJim: note that this is the max value a user could allow for a packet. Think we can easily change to six though, I have no insight into what value is typically used for timeouts atm

sangier: Yeah, I initially proposed a 4h too, but we kept it in 24h to be conservative. Anyway, as Jim says, that's easy to change. So let's agree on a concrete value.

PR #7524, opened 2024-10-29
Ibcmodulev2 transfer application spike

bznein:
## Description

This Spike was done by Cian & Nikolas to help inform the IBCModuleV2 interface definition and to get a PoC of what a transfer v2 module looks like and ensures that funigibility between v1 and v2 can be retained. 

A few notes about what we did here before anything goes into the feature branch.

- We created a new Keeper under `transfer/v2`. This keeper embeds the existing transfer/v1 keeper, and overrides functions required. E.g. the new OnRecv, Ack, Timeout etc.
- Forwarding is **not supported** in this PoC. This will require further discussion and spec changes.
- Some types/functions were made public to be accessible from the new transfer/v2 Keeper. This can be cleaned up / refactored before merging to not need to do this.
- A new IBCModule (V2) implementation for transfer was created, and added to the V2 router in our testing simapp.
- Events / telemetry currently disgregarded.
- The sequence was added to the `OnRecv` application callback, as this can be required by some applications. I think we probably need to add this to Ack/Timeout as well, but for now have not done so.
- We added some basic tests for all the new transfer module at the message server layer. A simple happy path test and some test cases where the transfer application fails.
- We added a test which does the following
  * Sets up 3 chains
  * Sends tokens from A -> B using MsgTransfer (protocol v1)
  * Sends same tokens from B -> C using MsgSendPacket (protocol v2)
  * Sends tokens back from C -> B using MsgSendPacket (protocol v2)
  * Sends tokens from B -> A using MsgTransfer (protocol V1)
  
This test ensures that fungibility remains and tokens can be sent back and forth using either the v1 or v2 router.

### Questions

1. What does forwarding look like in transfer v2? In v1 we stored the forwarded packets in state, we no longer provide the entire packet to the application callbacks. 
2. Are we happy with the general idea of re-using the transfer keeper and extending/overriding with new logic where appropriate.
3.  If we are going to include the sequence, should we also include the timeout timestamp? Previously we included the full packet in the application callbacks, but now we are providing a subset of that information. Are there use cases where this extra info should be needed?   I guess there is nothing stopping an application from including the timestamp in the packet data if required. I think it's probably fine to omit.

closes: #XXXX

---

write a little note why.


gjermundgaraba:
> Are we happy with the general idea of re-using the transfer keeper and extending/overriding with new logic where appropriate.

To me, this depends a lot on how long we expect v1 to live in our codebase. Short-term, this might be OK, I don't love the idea of a v2 depending on v1 and having these convoluted paths that might end up with weird bugs later. For one release, I might be OK :)

> I guess there is nothing stopping an application from including the timestamp in the packet data if required. I think it's probably fine to omit.

But the application would not know if the timestamp is the same as in the packet, unless only the application is allowed to construct new packets on all implementations?

AdityaSripal (commented):
I looked mainly at the high level interfaces. We want the sequence in ack and timeout as well.

modules/core/api/module.go
```diff
@@ -27,6 +27,7 @@ type IBCModule interface {
 		ctx context.Context,
 		sourceChannel string,
 		destinationChannel string,
+		sequence uint64,
```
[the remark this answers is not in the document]
chatton: yep makes sense!

chatton:
> > Are we happy with the general idea of re-using the transfer keeper and extending/overriding with new logic where appropriate.
> 
> To me, this depends a lot on how long we expect v1 to live in our codebase. Short-term, this might be OK, I don't love the idea of a v2 depending on v1 and having these convoluted paths that might end up with weird bugs later. For one release, I might be OK :)
> 

Thinking a bit more about this, I was thinking it might be possible to create some type that is a subset of of the transfer keeper that has all the functionality that is shared (e.g. code regarding denoms/ native vs ibc etc. )

Extracting some interfaces/ functions into a the transfer v1 module, then having (potentially a new go mod for) transfer v2 import from v1 the required components/functions and implement all the new logic in a new keeper while re-using a subset from v1.

Kind of similar to what is happening in this PR, but this is a bit messier just shoving it all in a sub directory.

gjermundgaraba:
For some context: this PR is being split up in such a way that forwarding support will happen in a separate PR.

modules/apps/transfer/types/packet.go
```diff
@@ -210,6 +210,8 @@ func (ftpd FungibleTokenPacketDataV2) HasForwarding() bool {
 // UnmarshalPacketData attempts to unmarshal the provided packet data bytes into a FungibleTokenPacketDataV2.
 // The version of ics20 should be provided and should be either ics20-1 or ics20-2.
 func UnmarshalPacketData(bz []byte, ics20Version string) (FungibleTokenPacketDataV2, error) {
+	// TODO: in transfer ibc module V2, we need to respect he encoding value passed via the payload, some hard coded assumptions about
```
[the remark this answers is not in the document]
bznein: I actually have the commit ready to be added to this PR (if we want to!) or I can add it in a separate one! I just wanted to go over it w/ Cian before

modules/apps/transfer/v2/keeper/msg_server_test.go
```diff
@@ -0,0 +1,597 @@
...
+			}
+
+			ftpd := transfertypes.NewFungibleTokenPacketDataV2(tokens, suite.chainA.SenderAccount.GetAddress().String(), suite.chainB.SenderAccount.GetAddress().String(), "", transfertypes.ForwardingPacketData{})
+			bz := suite.chainA.Codec.MustMarshal(&ftpd)
+
+			timestamp := suite.chainA.GetTimeoutTimestampSecs()
+			// TODO: note, encoding field currently not respected in the implementation. encoding is determined by the version.
+			// ics20-v1 == json
+			// ics20-v2 == proto
+			payload = channeltypesv2.NewPayload(transfertypes.ModuleName, transfertypes.ModuleName, transfertypes.V2, "json", bz)
```
AdityaSripal, line 74: Shouldn't this be "proto"?

testing/mock/v2/ibc_app.go
```diff
@@ -9,8 +9,8 @@ import (
...
-	OnTimeoutPacket         func(ctx context.Context, sourceChannel string, destinationChannel string, data channeltypesv2.Payload, relayer sdk.AccAddress) error
-	OnAcknowledgementPacket func(ctx context.Context, sourceChannel string, destinationChannel string, data channeltypesv2.Payload, acknowledgement []byte, relayer sdk.AccAddress) error
+	OnSendPacket            func(ctx context.Context, sourceChannel string, destinationChannel string, sequence uint64, payload channeltypesv2.Payload, signer sdk.AccAddress) error
+	OnRecvPacket            func(ctx context.Context, sourceChannel string, destinationChannel string, sequence uint64, payload channeltypesv2.Payload, relayer sdk.AccAddress) channeltypesv2.RecvPacketResult
+	OnTimeoutPacket         func(ctx context.Context, sourceChannel string, destinationChannel string, sequence uint64, payload channeltypesv2.Payload, relayer sdk.AccAddress) error
+	OnAcknowledgementPacket func(ctx context.Context, sourceChannel string, destinationChannel string, sequence uint64, payload channeltypesv2.Payload, acknowledgement []byte, relayer sdk.AccAddress) error
```
AdityaSripal, line 15: This fn signature looks different than the fn outside the testing package

AdityaSripal (approved):
LGTM, i have a couple suggestions but not blocking ack. I'm also a bit confused with how the app encoding logic

modules/apps/transfer/v2/ibc_module.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if !im.keeper.GetParams(ctx).SendEnabled {
+		return transfertypes.ErrSendDisabled
+	}
+
+	if im.keeper.IsBlockedAddr(signer) {
+		return errorsmod.Wrapf(ibcerrors.ErrUnauthorized, "%s is not allowed to send funds", signer)
+	}
+
+	data, err := transfertypes.UnmarshalPacketData(payload.Value, payload.Version)
```
AdityaSripal, line 44: Ahh because you aren't using the `payload.Encoding` here

modules/apps/transfer/v2/keeper/msg_server_test.go
bznein: Yes it should! I actually encountered some errors due to this while working on using the encoding! Not worth fixing it now since I have already a commit (which might end up here on in a separate PR) that fixes it

PR #7546, opened 2024-11-10
fix: packet keys corrupted

gjermundgaraba:
## Description

I think we will change the format to use bytes later (?), but the current setup gave problems because the byte array for big endian didn't stringify very well. So until we get the packet keys sorted, this will at least work.

closes: #XXXX

---

write a little note why.

### DIFF
--- a/modules/core/24-host/v2/packet_keys.go
+++ b/modules/core/24-host/v2/packet_keys.go
@@ -2,24 +2,22 @@ package v2
 
 import (
 	"fmt"
-
-	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 // PacketReceiptKey returns the store key of under which a packet
 // receipt is stored
 func PacketReceiptKey(channelID string, sequence uint64) []byte {
-	return []byte(fmt.Sprintf("receipts/channels/%s/sequences/%s", channelID, sdk.Uint64ToBigEndian(sequence)))
+	return []byte(fmt.Sprintf("receipts/channels/%s/sequences/%d", channelID, sequence))
 }
 
 // PacketAcknowledgementKey returns the store key of under which a packet acknowledgement is stored.
 func PacketAcknowledgementKey(channelID string, sequence uint64) []byte {
-	return []byte(fmt.Sprintf("acks/channels/%s/sequences/%s", channelID, sdk.Uint64ToBigEndian(sequence)))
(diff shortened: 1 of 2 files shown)


modules/core/24-host/v2/packet_keys.go
```diff
@@ -2,24 +2,22 @@ package v2
...
 	"fmt"
-
-	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 // PacketReceiptKey returns the store key of under which a packet
 // receipt is stored
 func PacketReceiptKey(channelID string, sequence uint64) []byte {
-	return []byte(fmt.Sprintf("receipts/channels/%s/sequences/%s", channelID, sdk.Uint64ToBigEndian(sequence)))
+	return []byte(fmt.Sprintf("receipts/channels/%s/sequences/%d", channelID, sequence))
```
DimitrisJim, line 10: wdym by corrupted? String value for key is only intermediate repr before casting to bytes, no? Is there missing context whereby we care about string value for these?

Looking at spec, these will continue using BigEndian encoding for seqs to keep sorted order.

gjermundgaraba: This is what the string ends up looking like: `"acks/channels/channel-0/sequences/\x00\x00\x00\x00\x00\x00\x00\x01` 
and it creates some issues with our relayer. It should technically not be a huge problem, but I can't imagine this being the idea of what is should look like at least :)

chatton: I think the main reason for doing this was to make it so iteration over channel IDs/ sequences etc happens in an expected order, rather than lexicographically. We are doing this also in the PacketForwardKey fn.

DimitrisJim: yup, specifically https://github.com/cosmos/ibc-go/issues/7132

gjermundgaraba: Alright, so we want it to look like this? If so, I can go back and figure out a way to  deal with those characters in the operator cli

DimitrisJim: yea. Can't you just decode the byte slice into the number where needed?

gjermundgaraba:
Close in favor of https://github.com/cosmos/ibc-go/commit/00505238557a6d9eab1013e07e43a75f401680f8

PR #7555, opened 2024-11-12
Simplify Acknowledgement structure

AdityaSripal:
## Description

This is a proposal that follows the current spec and I believe is cleaner than the current approach.
closes: #XXXX

---

write a little note why.


modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -147,23 +147,26 @@ func (k *Keeper) RecvPacket(ctx context.Context, msg *types.MsgRecvPacket) (*typ
...
-		ack.AcknowledgementResults = append(ack.AcknowledgementResults, types.AcknowledgementResult{
-			AppName:          pd.DestinationPort,
-			RecvPacketResult: res,
-		})
+		if res.Status == types.PacketStatus_Async {
+			// Set packet acknowledgement to async if any of the acknowledgements are async.
+			isAsync = true
+			// Return error if there is more than 1 payload
+			// TODO: Handle case where there are multiple payloads
+			if len(msg.Packet.Payloads) > 1 {
```
AdityaSripal, line 155: Currently just error for multipayload in async ack case

modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -203,14 +206,10 @@ func (k *Keeper) Acknowledgement(ctx context.Context, msg *types.MsgAcknowledgem
 		return nil, errorsmod.Wrap(err, "acknowledge packet verification failed")
 	}
 
-	recvResults := make(map[string]types.RecvPacketResult)
```
AdityaSripal, line 206: This puts the list into a map which will remove any subsequent acknowledgements from the same app. This is an error since we want to support sending multiple packets with the same portIDs.

Moreover it is unnecessary, the acknowledgement list is meant to match 1-1 to the list of payloads so we can associate each individual acknowledgement to the correct payload by looking at the portIDs in the matching payload in the payload list

proto/ibc/core/channel/v2/packet.proto
```diff
@@ -39,13 +39,7 @@ message Payload {
 
 // Acknowledgement contains a list of all ack results associated with a single packet.
 message Acknowledgement {
-  repeated AcknowledgementResult acknowledgement_results = 1 [(gogoproto.nullable) = false];
-}
-
-// AcknowledgementResult of identified acknowledgement. Correlating IBC application name to opaque ack bytes.
-message AcknowledgementResult {
```
AdityaSripal, line 46: Unnecessary struct since the acknowledgment side only ever uses the Acknowledgement bytes inside the `RecvPacketResult`.

The `status` in RecvPacketResult is **only** used on the receive side. Thus we do not need to transmit it over the wire back to the sending chain. Instead we can just package each individual opaque app acknowledgement into the Acknowledgment struct and send it over.

Moreover, the commitment isn't using the status so it isn't even secure to use it on the sending side (which we don't)

AdityaSripal (commented):
Use RecvPacketResult purely as a type internal to the ibc-go implementation to package the app acknowledgement with other useful information for the receiver chain.

However, when packaging the acknowledgement to send over the wire; the protocol only needs the list of app acknowledgements. Thus I've simplified the types here.

modules/core/04-channel/v2/keeper/msg_server.go
chatton: good catch, and yeah makes sense that we can just use the index.

chatton (approved):
Yeah, changes make sense! Thanks for putting together this PR, clears things up a bit.

DimitrisJim (approved):
lgtm too, appears this also unblocks https://github.com/cosmos/ibc-go/pull/7554 now

Issue #4697, opened 2023-09-18
Simplify E2E Configuration


chatton:
## Summary 

Depends on [this issue](https://github.com/cosmos/ibc-go/issues/3981) for more granular relayer configution.

Currently we configure E2E tests with both environmental variables, and with a [configuration file](https://github.com/cosmos/ibc-go/blob/4ef91861eec3685fa460ce4c70df679244b5069e/e2e/sample.config.yaml#L1)

The environment variables take precedence over the config files and are used to dynamically specify things like docker tags, relayer type etc.

This works, but can be simplified quite a lot. The only values that actually 100% need to be dynamic are `CHAIN_A_TAG`, `CHAIN_B_TAG` and `ACTIVE_RELAYER`. Everything else can actually be static.

Because of this, I suggest we create 1 configuration file per type of test, and override only the bare minimum of values. This will reduce the likely hood of error when passing fields throughout workflows, and also create a single source of truth for things such as the relayer image/version.

Currently there is some defaulting going on in the [workflows](https://github.com/cosmos/ibc-go/blob/fe52d2a147592885c6aa189d9325b792c46c7b86/.github/workflows/e2e-manual-simd.yaml#L69-L73) and also in the [code](https://github.com/cosmos/ibc-go/blob/4ef91861eec3685fa460ce4c70df679244b5069e/e2e/relayer/relayer.go#L52-L54).

We can move all defaults into config files instead of the code.

A sample config might look like

```yaml
---
chains:
  # the entry at index 0 corresponds to CHAIN_A
- chainId: chain-a
  numValidators: 4
  numFullNodes: 1
  image: ghcr.io/cosmos/ibc-go-simd
  tag: main # override with CHAIN_A_TAG
  binary: simd

  # the entry at index 1 corresponds to CHAIN_B
- chainId: chain-b
  numValidators: 4
  numFullNodes: 1
  image: ghcr.io/cosmos/ibc-go-simd # override with CHAIN_IMAGE
  tag: main # override with CHAIN_B_TAG
  binary: simd

relayers:
  activeRelayer: hermes # ACTIVE_RELAYER 
  configurations:
  - relayer: hermes
    image: "ghcr.io/informalsystems/hermes"
    tag: 1.4.0
  - relayer: rly
    id: rly 
    image: "ghcr.io/cosmos/relayer"
    tag: latest 
cometbft:
  logLevel: info

debug:
  dumpLogs: false

upgrade:
  planName: ""
  tag: ""

```

[...]

PR #7568, opened 2024-11-14
Simplify E2E Configuration

chatton:
## Description

This PR modifies the way configurations are passed to E2E tests.

While every field can still be overridden by environment variables, the E2Es now have their own config file that is used, and it is possible to remove some special casing around tests being in CI vs running locally.

The TestConfig struct has been updated to behave in a similar way as the relayers, in that we provide a list of all upgrade plans, then just specify which one we want by name.

closes: #4697

---

write a little note why.


e2e/ci-e2e-config.yaml
```diff
@@ -0,0 +1,68 @@
...
+    binary: simd # override with CHAIN_BINARY
+
+    # the entry at index 1 corresponds to CHAIN_B
+  - chainId: chainB-1
+    numValidators: 4
+    numFullNodes: 1
+    image: ghcr.io/cosmos/ibc-go-simd # override with CHAIN_IMAGE
+    binary: simd # override with CHAIN_BINARY
+
+activeRelayer: hermes # override with RELAYER_ID
```
bznein, line 34: I assume this has to be one of the IDs in the list below, maybe worth specifying it in the comment?

e2e/testsuite/testconfig.go
```diff
@@ -207,6 +216,35 @@ func (tc TestConfig) validateGenesisDebugConfig() error {
...
+	// the upgrade plan name specified must match one of the upgrade plans in the upgrade configs.
+	foundPlan := false
+	for _, upgrade := range tc.UpgradeConfigs {
+		if strings.TrimSpace(upgrade.Tag) == "" {
+			return fmt.Errorf("upgrade config missing tag: %+v", upgrade)
+		}
+
+		if strings.TrimSpace(upgrade.PlanName) == "" {
+			return fmt.Errorf("upgrade config missing plan name: %+v", upgrade)
+		}
```
bznein, line 234: [nit] I wonder if it would make more sense to either:

1) do this check only if `upgrade.PlanName == tc.UpgradePlanName` (so a "broken" upgrade config is allowed if it's not used for the test)
2) split this into two different loops. Since we don't care about efficiency here, I think it's a bit more clear if validation and search are split

e2e/testsuite/testconfig.go
```diff
@@ -207,6 +216,35 @@ func (tc TestConfig) validateGenesisDebugConfig() error {
...
+
+		if upgrade.PlanName == tc.UpgradePlanName {
+			foundPlan = true
+		}
```
bznein, line 238: [nit] if we go for either 1 or 2 in the above comment, we can `return nil` here.

Or even:

if we go with 1 above: use `slices.IndexFunc()` to get the index of the right upgrade plan, and then do the validation and return

if we go with 2 above: simply `return slices.ContainsFunc()`

bznein (commented):
Did a quick fly-by review as discuss offline with Cian, it mostly lgtm with a few minor comments

e2e/testsuite/testconfig.go
chatton: I figured it would be best to just explode immediately if anything about config is bad. Don't have a strong preference for multiple loops, will always just be a few elements anyway.

bznein: Up to you! I am slightly more inclined towards having one loop of validation followed by one of search, for clarity, but not a deal breaker

bznein (approved):
SLAM if all tests are green!

gjermundgaraba (approved):

Issue #7591, opened 2024-11-22
Add support for ABI encoding in ICS20Transfer (ibc-go)


gjermundgaraba:
Given the results of https://github.com/cosmos/solidity-ibc-eureka/issues/95, we have decided to implement ABI encoding for ICS20 transfer for ics20-1 payloads.

Definition of done:
* Add support for ABI encoding of ics20-1 payloads
* Good test coverage of new functionality
* Add some level of integration test coverage to make sure both happy- and sad-path cases are covered

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7559, opened 2024-11-13
feat: add custom querier to wasm image

gjermundgaraba:
## Description

This adds the custom querier needed for the union light client to the WASM simapp image.

This change also makes the usual amd64 image unusable on arm64 machines (Mac M1/2/etc.), so I've updated the build WASM image pipeline to build for amd64 and arm64 at the same time.

Unfortunately, the speed of that build has taken a significant hit due to the virtual nature of QEMU that is needed to build for arm64 on an amd64 machine. It can take up to around 25 minutes (https://github.com/cosmos/ibc-go/actions/runs/11817324167).

It is _possible_ to improve this, but it would require a more complex pipeline that builds the images in a separate job with separate machines with separate tags (using the matrix strategy) and then downloads those images and pushes them together with the same tag. Or something very similar using Docker manifest - I haven't fully dived into the details yet (but it would be more complex for sure and take more time to set up). If this would be a big issue, I am open to doing that or finding other solutions.

---

Before we can merge this PR, please make sure that all the following items have been
checked off. If any of the checklist items are not applicable, please leave them but
write a little note why.

- [ ] Targeted PR against the correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#pull-request-targeting)).
- [ ] Linked to GitHub issue with discussion and accepted design, OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/build/building-modules/11-structure.md) and [Go style guide](../docs/dev/go-style-guide.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package).
- [ ] Updated relevant documentation (`docs/`).
- [ ] Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Provide a [conventional commit message](https://github.com/cosmos/ibc-go/blob/main/docs/dev/pull-requests.md#commit-messages) to follow the repository standards.
- [ ] Include a descriptive changelog entry when appropriate. This may be left to the discretion of the PR reviewers. (e.g. chores should be omitted from changelog)
- [ ] Re-reviewed `Files changed` in the GitHub PR explorer.
- [ ] Review `SonarCloud Report` in the comment section below once CI passes.

PR #7592, opened 2024-11-24
feat: add solidity ABI support in ICS20Transfer ics20-1

gjermundgaraba:
## Description

closes: #7591 

---

write a little note why.


.github/workflows/test.yml
```diff
@@ -34,14 +34,14 @@ jobs:
...
             go.mod
             go.sum
-      - name: Build ibc-go
-        run: GOARCH=${{ matrix.go-arch }} LEDGER_ENABLED=false make build
       - name: Install compiler for arm64.
         if: matrix.go-arch == 'arm64'
         run: |
           sudo apt-get update
-          sudo apt-get install -y gcc-aarch64-linux-gnu
+          sudo apt-get install -y gcc-aarch64-linux-gnu g++-aarch64-linux-gnu build-essential
```
gjermundgaraba, line 41: Some changes in dependencies has made this necessary

.golangci.yml
```diff
@@ -40,6 +40,9 @@ issues:
     - linters:
         - staticcheck
       text: "SA1019:" # silence errors on usage of deprecated funcs
+    - text: "G115: integer overflow conversion"
```
gjermundgaraba, line 43: This is not strictly necessary currently, but we will need it as soon as we update to a newer version of golangci-lint. We use a newer one in the solidity repo (and 1.57 seems broken on my Mac, so I am running the same from the solidity repo), which turns on G115 by default. We need convert integers constantly, and there is no way to appease the linter (according to issues I found on gh). So if we just add it now, we'll save ourselves some pain when we upgrade golangci-lint.

modules/apps/transfer/types/packet.go
```diff
@@ -26,6 +27,7 @@ var (
 const (
 	EncodingJSON     = "application/json"
 	EncodingProtobuf = "application/x-protobuf"
+	EncodingABI      = "application/x-solidity-abi"
```
gjermundgaraba, line 30: I was not able to find an actual "standard" or widely accepted MIME type for solidity ABI encoded data. So I made this one... Let me know if you have a better suggestion.

modules/light-clients/08-wasm/testing/simapp/customquery/crypto.go
damiannolan, line 1: when was this file added? is this something needed for eureka? I don't think we have it on main

modules/apps/transfer/types/packet.go
damiannolan: Yeah, I'm not sure what the expectation is here for mime types that don't really exist. 
This would be my hesitation for using mime format rather than just defining our own enum, although, much of a muchness really.. only negative is if at some point a standard mime type is added for abi solidity

go.mod
```diff
@@ -19,12 +19,13 @@ require (
 	github.com/cosmos/cosmos-sdk v0.50.10
 	github.com/cosmos/gogoproto v1.7.0
 	github.com/cosmos/ics23/go v0.11.0
+	github.com/cosmos/solidity-ibc-eureka/abigen v0.0.0-20241124101149-3e2c74b736f4
```
damiannolan, line 22: this is a separate go module, right? 

maybe in future could be nice to have all encoding and protobuf stuff in the ibc repo, but that's probably quite far off right now

gjermundgaraba: Yes, it is generated code from the ABI definitions of the solidity EVM contracts (Solidity -> ABI -> go).

modules/light-clients/08-wasm/testing/simapp/customquery/crypto.go
gjermundgaraba: It was added recently in #7559. It is currently needed for the Union 08-wasm Ethereum light client. I hope to remove this dependency as we move the light client 🙏

modules/apps/transfer/types/packet.go
gjermundgaraba: Yeah, that is true :/ In that case we might just have to support both for a while. Well, this is feedback that @AdityaSripal can think about :)

modules/apps/transfer/types/packet_test.go
```diff
@@ -719,20 +722,57 @@ func TestUnmarshalPacketData(t *testing.T) {
...
 		packetDataBz []byte
 		version      string
+		encoding     string
 	)
 
 	testCases := []struct {
 		name     string
 		malleate func()
-		expError error
+		expError string
```
bznein, line 731: wdyt of using instead an `error` and `RequireErrorIsOrContains` (https://github.com/cosmos/ibc-go/blob/main/testing/utils.go#L88) and either:

- Use the exact error
- Use errors.New() 

you can see an example hre: https://github.com/cosmos/ibc-go/blob/main/modules/apps/29-fee/types/msgs_test.go#L22-L94

modules/core/04-channel/v2/types/commitment_test.go
```diff
@@ -15,32 +18,68 @@ import (
...
+			func() {
+				transferData, err := ics20lib.EncodeFungibleTokenPacketData(ics20lib.ICS20LibFungibleTokenPacketData{
+					Denom:    "uatom",
+					Amount:   big.NewInt(1000000),
+					Sender:   "sender",
+					Receiver: "receiver",
+					Memo:     "memo",
+				})
+				fmt.Printf("transferData: %s\n", string(transferData))
+				fmt.Printf("hex value: %s\n", hex.EncodeToString(transferData))
```
bznein, line 43: leftover code?

gjermundgaraba: lol yes

modules/apps/transfer/types/packet_test.go
gjermundgaraba: I like that! I will try it out :)

modules/apps/transfer/types/packet_test.go
```diff
@@ -747,6 +788,63 @@ func TestUnmarshalPacketData(t *testing.T) {
...
+			func() {
+				packetDataBz = []byte("invalid")
+				encoding = types.EncodingJSON
+			},
+			ibcerrors.ErrInvalidType,
+		},
+		{
+			"failure: invalid type for protobuf",
+			func() {
+				packetDataBz = []byte("invalid")
```
bznein, line 835: [nit] maybe we can extract it somewhere since we use it in at least three places

bznein (approved):
LGTM

go.mod
srdtrk: I actually think that we should have the encoding/decoding logic here (or ibc repo I guess). I'd even rather use manually written code to abi encode/decode the packet data than to use autogenerated code from another repo. But for now, this is probably ok...

gjermundgaraba: There is a bigger discussion to be had on centralizing types and encoding. For now, this is at least relatively convenient and gives a direct link to where the correct definitions are.

Issue #7595, opened 2024-11-25
Create 08-wasm ethereum light client structure


gjermundgaraba:
Find a reasonable place in the ibc-go repo and create a skeleton structure for an 08-wasm light client. This issue does not include any of the actual light client logic, as that will be moved over in #7594.

Definition of done:
* Basic project structure
* Generated types from proto files
* Working 08-wasm light client implementation (in that it can be deployed and will accept the messages from the 08-wasm go light client module)
* Basic CI pipeline that builds and runs unit tests

gjermundgaraba:
Moved and superseded by https://github.com/cosmos/solidity-ibc-eureka/issues/142

PR #7603, opened 2024-11-26
feat: create 08-wasm-light-clients structure

gjermundgaraba:
## Description

closes: #7595 

From the ibc-go devs I am looking mostly for structure feedback. @srdtrk can provide the rust specific feedback.

I played around with a few different options for folder structure, but ended up with the current structure, which has the main benefit of being close to everything light clients in the repo. An alternative I also tried was a root level "rust" folder, which also works, but the naming of the sub-folders felt more natural when it was under the light-clients folder.

Happy to hear other ideas and other naming suggestions in general. 

This PR just aims to set up the structure to start moving code over from the Union light client repo. The 08-wasm light client that is there now is essentially a "mock" light client that will accept anything and always return true.

I also tried to move the api_contract into proto files to get them generated - and it is possible. It does require quite a bit of custom handling, especially for the rust generated code. In the end, I decided it might not be worth it due to both implementation really only needing JSON/serde. We could do this at a later point if we think it might be useful to also publish that contract api as a rust crate for easier implementation. 

---

write a little note why.


modules/light-clients/08-wasm-light-clients/packages/ibc-go-proto/Cargo.toml
srdtrk, line 1: If we don't want to publish and continue to maintain this, we could also use tonic's build macros instead of generating the code ourselves. This also hides all the generated code (similar to how alloy's sol macro works). But I'm fine with this as is.

I understand why we need this, but we should either use ibc-proto in the future, or move ibc-proto here. In the mean time, tonic macros could be used if it works...

modules/light-clients/08-wasm-light-clients/.cargo/config.toml
srdtrk, line 1: I think this file is not needed since we are only using it to keep 1 alias which we can keep in our makefile instead.

modules/light-clients/08-wasm-light-clients/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[workspace]
+resolver = "2"
+members = [
+    "contracts/*",
+    "packages/*",
```
srdtrk, line 5: Instead of using `*`, you probably want to point to the exact directory instead since this is a multi language repo.

```suggestion
    "contracts/ethereum-light-client",
    "packages/ibc-go-proto",
```

modules/light-clients/08-wasm-light-clients/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+[workspace.package]
+edition = "2021"
+repository = "https://github.com/cosmos/ibc-go"
+version = "0.1.0"
+
+[workspace.metadata.scripts]
+optimize = """docker run --rm -v "$(pwd)":/code \
+  --mount type=volume,source="$(basename "$(pwd)")_cache",target=/target \
+  --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+  cosmwasm/optimizer:0.16.0"""
```
srdtrk, line 17:
```suggestion
  cosmwasm/optimizer:0.16.1"""
```

modules/light-clients/08-wasm-light-clients/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+thiserror = { version = "2.0.3" }
+
+# Proto related dependencies
+ics23                  = { version = "0.12.0", default-features = false }
+tendermint-proto = { version = "0.40.0", default-features = false }
+cosmos-sdk-proto = { version = "0.26.1", default-features = false }
+prost                  = { version = "0.13.3", default-features = false, features = ["prost-derive"] }
+
+# dev-dependencies
+cw-multi-test = "2.2.0"
```
srdtrk, line 37: This file could use some clean up, and also we might want to remove the patch versions since I don't think they are very useful anyway.

```suggestion
[workspace.dependencies]
# Local packages
ibc-go-proto = { path = "packages/ibc-go-proto" }

# CosmWasm related dependencies
cosmwasm-schema = "2.1.4"
cosmwasm-std = "2.1.4"
cw-storage-plus = "2.0.0"
serde = { version = "1.0.215", default-features = false, features = ["derive"] }
thiserror = "2.0.3"

# Proto related dependencies
ics23                  = { version = "0.12.0", default-features = false }
tendermint-proto = { version = "0.40.0", default-features = false }
cosmos-sdk-proto = { version = "0.26.1", default-features = false }
prost                  = { version = "0.13.3", default-features = false, features = ["prost-derive"] }

# dev-dependencies
cw-multi-test = "2.2.0"
```

.github/workflows/wasm-light-clients.yml
```diff
@@ -0,0 +1,60 @@
+name: Wasm Light Clients
+# This workflow runs when a PR is opened that targets code that is part of the 08-wasm-light-clients
+on:
+  merge_group:
+  pull_request:
+  push:
+    branches:
+      - main
+      - feat/ibc-eureka # TODO: Remove before merging to main
```
srdtrk, line 9: This PR should have it's own feature branch and not be merged directly to feat/ibc-eureka in my opinion

modules/light-clients/08-wasm-light-clients/contracts/ethereum-light-client/src/error.rs
```diff
@@ -0,0 +1,11 @@
+use cosmwasm_std::StdError;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ContractError {
+    #[error("{0}")]
+    Std(#[from] StdError),
+
+    #[error("Unauthorized")]
+    Unauthorized {},
```
srdtrk, line 10:
```suggestion
    Unauthorized,
```

modules/light-clients/08-wasm-light-clients/contracts/ethereum-light-client/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+pub mod contract;
```
srdtrk, line 1: I think we should enforce documentation in all public API in this crate like we do in ibc-go (since people and future devs might want to read). We can also add stricter linting as follows:

```suggestion
#![deny(clippy::nursery, clippy::pedantic, warnings, missing_docs)]

pub mod contract;
```

Makefile
```diff
@@ -141,6 +141,11 @@ build-docker-wasm:
 
 .PHONY: build-docker-wasm
 
+build-wasm-light-clients:
+	cd modules/light-clients/08-wasm-light-clients && cargo wasm
```
srdtrk, line 145: why don't we have a build script that uses cosmwasm optimizer? I think this is the only one we need since we want reproducible builds

srdtrk (commented):
Looks good. I think this needs to have it's own feature branch since it contains only a template at the moment

modules/light-clients/08-wasm-light-clients/Cargo.toml
gjermundgaraba: It's a multi language repo, but the folder we're in is for the rust code. If we suddenly add anything there, builds will fail and then we can adjust, no?

modules/light-clients/08-wasm-light-clients/contracts/ethereum-light-client/src/lib.rs
gjermundgaraba: Sure! I will do that in a separate task: #7604

Makefile
gjermundgaraba: I was planning to add that when this is together with e2e integration, but it does take quite a bit longer to run, and it is not strictly necessary for CI right now.

modules/light-clients/08-wasm-light-clients/.cargo/config.toml
gjermundgaraba: Yeah, I guess it's not strictly needed. I'll remove it.

gjermundgaraba:
> Looks good. I think this needs to have it's own feature branch since it contains only a template at the moment

I'm keen to hear other people's opinion on this, but my personal take is that I would love to avoid feature branches as much as possible unless there is a very good reason to keep the code out of main or feat/ibc-eureka. I have tried to structure the work in such a way that it can be merged as smaller PRs and not break anything along the way. Getting things merged sooner and more often feels better than having a lot of feature branches all over the place.

modules/light-clients/08-wasm-light-clients/packages/ibc-go-proto/Cargo.toml
gjermundgaraba: My thinking was that this should could replace ibc-proto eventually. Tonic is an option, but to be honest I don't love hidden generated code because it makes it a lot harder to understand underlying types and what is really going on (too much magic for my taste).

.github/workflows/wasm-light-clients.yml
bznein: > If possible to get things merged in smaller increments, it will be easier than having to maintain yet another feature branch. I would prefer to avoid another branch if smaller PRs can do the job

Having been dealing with multiple feature branches for a while now, I 100% agree!

modules/light-clients/08-wasm-light-clients/contracts/ethereum-light-client/src/contract.rs
```diff
@@ -0,0 +1,299 @@
+use cosmwasm_std::entry_point;
```
yihuang, line 1: Do we need to include cosmwasm module to use this?
Or we just use the wasm runtime to execute the code, but not require the module?

damiannolan: There is a hard dependency right now on ibc wasm light clients with cosmwasm (through the 08-wasm go proxy module).

PR #7619, opened 2024-12-04
chore: add channel consensus state rpc

DimitrisJim:
## Description

Note, we don't have a cli for this in main so didn't add one here. Can add it if needed (assume it doesn't exist in main for a reason)

closes: #XXXX

---

write a little note why.


AdityaSripal:
Maybe a stupid question since all queries seem to have this. But why have proof in the response if it is always nil? I understand proof is not retreivable from within the state machine. Will it be filled through a different gateway?

damiannolan:
Agree we normally put it in our msgs because we offered some CLIs which short-circuited grpc by hitting abci directly - in those cases we fulfil proof and proofHeight. But if we don't offer that kind of CLI here then don't see the point in putting it in the proto msg. You can still query abci directly and get a proof back for it. 

One thing to consider is that other consumers of the protobuf messages treat them as canonical apis for different runtime envs (e.g. penumbra etc)... not sure how that affects decision making here tho..

DimitrisJim:
> One thing to consider is that other consumers of the protobuf messages treat them as canonical apis for different runtime envs (e.g. penumbra etc)... not sure how that affects decision making here tho..

think this is good enough for now to keep things as is for time being (and as we historically did). Maybe its a point to be raised with other teams though and possibly rm'ing them after that is resolved.

Issue #7472, opened 2024-10-17
Add constructor/validation for Acknowledgement type


DimitrisJim:
Pending this type being finalized. 

Validation for it should be invoked in `MsgAcknowledge`. Opening issue to keep track.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

AdityaSripal:
Also need to then add this validation into WriteAcknowledgement along with tests. Will add in-line TODO so it doesn't get missed

bznein:
What exactly is the validation we should do here? 

The Acknowledgement type in eureka is just a slice of slice of bytes, what kind of validation can be performed? @AdityaSripal @DimitrisJim

DimitrisJim:
the only one I can currently think of is that the length of the acks must be one (as we do for payloads)

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7621, opened 2024-12-04
Use global success value in acknowledgment

AdityaSripal:
## Description

closes: #7645 
closes: #7472 

---

write a little note why.


modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -131,20 +131,20 @@ func (k *Keeper) RecvPacket(ctx context.Context, msg *types.MsgRecvPacket) (*typ
...
 
-		if res.Status != types.PacketStatus_Failure {
-			// write application state changes for asynchronous and successful acknowledgements
-			writeFn()
-		} else {
+		if res.Status == types.PacketStatus_Failure {
+			// Set RecvSuccess to false
+			ack.RecvSuccess = false
 			// Modify events in cached context to reflect unsuccessful acknowledgement
 			sdkCtx.EventManager().EmitEvents(internalerrors.ConvertToErrorEvents(cacheCtx.EventManager().Events()))
```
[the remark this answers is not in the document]
AdityaSripal: I looked into doing this. It became too complicated because it caused empty acks to be valid which I didn't want to do. Otherwise I need to create a sentinel ack and fill it in for the acks that didn't error which was more complicated than necessary especially since we only support single payload for now

modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -212,7 +217,7 @@ func (k *Keeper) Acknowledgement(ctx context.Context, msg *types.MsgAcknowledgem
 	for i, pd := range msg.Packet.Payloads {
 		cbs := k.Router.Route(pd.SourcePort)
 		ack := msg.Acknowledgement.AppAcknowledgements[i]
-		err := cbs.OnAcknowledgementPacket(ctx, msg.Packet.SourceChannel, msg.Packet.DestinationChannel, msg.Packet.Sequence, ack, pd, relayer)
+		err := cbs.OnAcknowledgementPacket(ctx, msg.Packet.SourceChannel, msg.Packet.DestinationChannel, msg.Packet.Sequence, msg.Acknowledgement.RecvSuccess, ack, pd, relayer)
```
AdityaSripal, line 220: Note: We will need to tweak this eventually since it would be weird to call ack callback with a successful ack bytes but a false recvSuccess but not necessary for single payload change

gjermundgaraba (approved):
This looks good to me! Do we have an issue to make this change in solidity as well?

modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -111,26 +104,45 @@ func (k *Keeper) RecvPacket(ctx context.Context, msg *types.MsgRecvPacket) (*typ
...
 		// append app acknowledgement to the overall acknowledgement
 		ack.AppAcknowledgements = append(ack.AppAcknowledgements, res.Acknowledgement)
-	}
 
-	if len(ack.AppAcknowledgements) != len(msg.Packet.Payloads) {
-		return nil, errorsmod.Wrapf(types.ErrInvalidAcknowledgement, "length of app acknowledgement %d does not match length of app payload %d", len(ack.AppAcknowledgements), len(msg.Packet.Payloads))
+		// if any of the application callbacks return a failure, the entire packet receive is considered a failure
+		// break the acknowledgement loop but keep a single non-nil acknowledgement in the acknowledgement slice
+		// so that the error acknowledgment is not equivalent to the empty acknowledgement
+		if res.Status == types.PacketStatus_Failure {
```
AdityaSripal, line 118: I moved this down so that specific app acknowledgement is filled in even in failure with single packet payload

modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -173,10 +185,24 @@ func (k *Keeper) Acknowledgement(ctx context.Context, msg *types.MsgAcknowledgem
 		return nil, errorsmod.Wrap(err, "acknowledge packet verification failed")
 	}
 
+	// acknowledgement only needs to be filled in if the receive was successful
+	// thus we only validate if receive is successful
+	if msg.Acknowledgement.RecvSuccess {
```
AdityaSripal, line 190: One annoying thing here is I prevent an empty ack in RecvPacket or WriteAck since this could be a developer error

But on AcknowledgePacket, I allow relayers to just pass in empty ack to signal a failed receive since the alternative is to enforce a sentinel non-empty ack that will not be inspected at all by handler

AdityaSripal: Open to feedback on better ways to structure

modules/core/04-channel/v2/types/commitment.go
```diff
@@ -6,6 +6,8 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+var FailedAcknowledgementCommitment = sha256.Sum256([]byte{byte(2), byte(0)})
```
AdityaSripal, line 9: I've left this as a var for now but can change to a string const if preferred.

Need feedback from @srdtrk if this constant err ack hash is acceptable

modules/core/04-channel/v2/types/msgs.go
```diff
@@ -108,15 +108,13 @@ func (msg *MsgAcknowledgement) ValidateBasic() error {
...
-	if err := msg.Acknowledgement.Validate(); err != nil {
-		return err
-	}
-
 	_, err := sdk.AccAddressFromBech32(msg.Signer)
 	if err != nil {
 		return errorsmod.Wrapf(ibcerrors.ErrInvalidAddress, "string could not be parsed as address: %v", err)
 	}
 
+	// NOTE: Acknowledgement is allowed to be empty if the receive failed
```
AdityaSripal, line 116: Note: This annoying point that I do allow empty acks being sent in by relayer as a valid message now for error acks

modules/core/04-channel/v2/types/commitment.go
```diff
@@ -53,13 +55,21 @@ func hashPayload(data Payload) []byte {
...
 	var buf []byte
 	for _, ack := range acknowledgement.GetAppAcknowledgements() {
 		hash := sha256.Sum256(ack)
 		buf = append(buf, hash[:]...)
 	}
 
-	buf = append([]byte{byte(2)}, buf...)
+	// prepend the preimage with the IBC Version and the recv success
+	// to ensure there is no possibility of hash collisions
+	buf = append([]byte{byte(2), byte(1)}, buf...)
```
srdtrk, line 72: This is a consensus breaking change right? i.e. we need to update this in solidity as well. Would it be possible to avoid this by picking a different `FailedAcknowledgementCommitment`?

modules/core/04-channel/v2/types/msgs.go
srdtrk: Why do you allow this? Isn't it emitted in the event? This doesn't make much sense to me since you'd need the proof for it anyway

proto/ibc/core/channel/v2/packet.proto
```diff
@@ -39,7 +39,8 @@ message Payload {
 
 // Acknowledgement contains a list of all ack results associated with a single packet.
 message Acknowledgement {
-  repeated bytes app_acknowledgements = 1;
+  bool           recv_success         = 1;
+  repeated bytes app_acknowledgements = 2;
 }
```
srdtrk, line 44: This is not submitted by the relayer right? This is an internal ack

testing/mock/v2/ibc_module.go
```diff
@@ -42,16 +44,20 @@ func (im IBCModule) OnSendPacket(ctx context.Context, sourceChannel string, dest
...
+func (im IBCModule) OnRecvPacket(ctx context.Context, sourceChannel string, destinationChannel string, sequence uint64, payload channeltypesv2.Payload, relayer sdk.AccAddress) channeltypesv2.RecvPacketResult {
 	if im.IBCApp.OnRecvPacket != nil {
-		return im.IBCApp.OnRecvPacket(ctx, sourceChannel, destinationChannel, sequence, data, relayer)
+		return im.IBCApp.OnRecvPacket(ctx, sourceChannel, destinationChannel, sequence, payload, relayer)
+	}
+	if bytes.Equal(mockv1.MockPacketData, payload.Value) {
+		return MockRecvPacketResult
+	} else {
+		return MockFailRecvPacketResult
 	}
```
srdtrk, line 55: Linter issue, you should't need to use else after an if statement if the if block ends with a return.

srdtrk (commented):
Looking good. Left some comments

AdityaSripal:
Superceded by #7924

PR #7640, opened 2024-12-08
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:


DimitrisJim (approved):
nicely caught! thanks!

Issue #7175, opened 2024-08-15
Use expected errors instead of boolean values


bznein:
## Summary of Bug

This is an umbrella issue to enumerate all the test files that use `tc.expPass` rather than expecting a specific error, extracted with the following command (might not cover ALL of them):

```
internal/validate/validate_test.go
modules/apps/27-interchain-accounts/controller/keeper/keeper_test.go
modules/apps/27-interchain-accounts/controller/ibc_middleware_test.go
modules/apps/27-interchain-accounts/genesis/types/genesis_test.go
modules/apps/27-interchain-accounts/host/keeper/genesis_test.go
modules/apps/27-interchain-accounts/host/keeper/msg_server_test.go
modules/apps/27-interchain-accounts/host/keeper/handshake_test.go
modules/apps/27-interchain-accounts/host/keeper/keeper_test.go
modules/apps/27-interchain-accounts/host/types/codec_test.go
modules/apps/27-interchain-accounts/host/types/msgs_test.go
modules/apps/27-interchain-accounts/host/ibc_module_test.go
modules/apps/27-interchain-accounts/types/account_test.go
modules/apps/27-interchain-accounts/types/codec_test.go
modules/apps/27-interchain-accounts/types/metadata_test.go
modules/apps/27-interchain-accounts/types/packet_test.go
modules/apps/27-interchain-accounts/types/port_test.go
modules/apps/29-fee/keeper/grpc_query_test.go
modules/apps/29-fee/keeper/msg_server_test.go
modules/apps/29-fee/keeper/relay_test.go
modules/apps/29-fee/types/codec_test.go
modules/apps/29-fee/types/fee_test.go
modules/apps/29-fee/types/genesis_test.go
modules/apps/29-fee/types/keys_test.go
modules/apps/29-fee/types/msgs_test.go
modules/apps/29-fee/ibc_middleware_test.go
modules/apps/transfer/keeper/grpc_query_test.go
modules/apps/transfer/keeper/invariants_test.go
modules/apps/transfer/keeper/keeper_test.go
modules/apps/transfer/keeper/msg_server_test.go
modules/apps/transfer/types/codec_test.go
modules/apps/transfer/types/genesis_test.go
modules/apps/transfer/types/msgs_test.go
modules/apps/transfer/types/transfer_authorization_test.go
modules/apps/transfer/types/packet_test.go
modules/capability/types/genesis_test.go
modules/core/02-client/keeper/client_test.go
[...]

PR #7680, opened 2024-12-12
fix tests: update tests in 04-channel/types , use expected errors

likesToEatFish:
## Description

ref: https://github.com/cosmos/ibc-go/issues/7175

Use expected errors in modules/core/04-channel/types/

---

write a little note why.


modules/core/24-host/parse.go
```diff
@@ -27,7 +27,7 @@ func ParseIdentifier(identifier, prefix string) (uint64, error) {
 
 	sequence, err := strconv.ParseUint(splitStr[1], 10, 64)
 	if err != nil {
-		return 0, errorsmod.Wrap(err, "failed to parse identifier sequence")
+		return 0, errorsmod.Wrap(ErrInvalidID, err.Error())
```
gjermundgaraba, line 30: In these scenario (where there is an actual error), we typically don't wrap like this, we wrap the way this was already doing: `errorsmod.Wrap(err, "failed to parse identifier sequence")`

likesToEatFish: -Consistency: Other errors in the same function are using ErrInvalidID instead of wrapping directly from err. This ensures that the error message is always consistent about the error type ErrInvalidID.

-Error structure: ErrInvalidID is used to represent a specific identifier format error. If wrapped with another external err (errorsmod.Wrap), it may not ensure consistency about the error type.

I suggest the following:
`return 0, errorsmod.Wrapf(ErrInvalidID, "failed to parse sequence: %s", err.Error())`

Are you ok?

damiannolan: For this case either is fine imo, but generally prefer to wrap err returned as @gjermundgaraba suggested

likesToEatFish: Thanks u

> For this case either is fine imo, but generally prefer to wrap err returned as @gjermundgaraba suggested

DimitrisJim (approved):
Lgtm, will merge after sdk upgrade pr is merged

Issue #7712, opened 2024-12-17
Migrate to cosmos-sdk `appmodule.Environment` in ibc


damiannolan:
## Summary

Follow up issue from Olympus upgrade (#7223)

With upgrading to cosmos-sdk v0.52 we will leverage the service based `Environment` architecture where possible, rather than relying on the legacy `sdk.Context` approach.

This refactor was already started with icacontroller in order to get the bulk of the Olympus upgrade merged to main. This issue serves to track the remainder of modules and submodules as we migrate to `appmodule.Environment`.

- [x] icacontroller
- [x] icahost
- [x] fee middleware
- [x] transfer
- [x] core
- [x] lightclients (solomachine, tendermint, localhost, wasm)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7732, opened 2024-12-18
refactor: use branch service in 29-fee

damiannolan:
## Description

This PR migrates the 29-fee module to use `env.BranchService` in favour of cached context.

ref: #7712 

---

write a little note why.

### DIFF
--- a/modules/apps/29-fee/keeper/escrow.go
+++ b/modules/apps/29-fee/keeper/escrow.go
@@ -3,6 +3,7 @@ package keeper
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 
 	errorsmod "cosmossdk.io/errors"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/cosmos/ibc-go/v9/modules/apps/29-fee/types"
 	channeltypes "github.com/cosmos/ibc-go/v9/modules/core/04-channel/types"
+	ibcerrors "github.com/cosmos/ibc-go/v9/modules/core/errors"
 )
 
 // escrowPacketFee sends the packet fee to the 29-fee module account to hold in escrow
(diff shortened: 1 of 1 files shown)


modules/apps/29-fee/keeper/escrow.go
```diff
@@ -46,38 +48,39 @@ func (k Keeper) escrowPacketFee(ctx context.Context, packetID channeltypes.Packe
...
-	sdkCtx := sdk.UnwrapSDKContext(ctx) // TODO: https://github.com/cosmos/ibc-go/issues/5917
-	cacheCtx, writeFn := sdkCtx.CacheContext()
+	if err := k.BranchService.Execute(ctx, func(ctx context.Context) error {
+		// forward relayer address will be empty if conversion fails
+		forwardAddr, _ := sdk.AccAddressFromBech32(forwardRelayer)
+
+		for _, packetFee := range packetFees {
+			if !k.EscrowAccountHasBalance(ctx, packetFee.Fee.Total()) {
+				// NOTE: we lock the fee module on error return so that the state changes are persisted
+				return ibcerrors.ErrInsufficientFunds
```
gjermundgaraba, line 60: I understand the concept here, but I have to admit I don't love the use of errors as control flow. It feels a lot like using exceptions as control flow and while it works, it is quite ugly. Another potential way to do this would be to not shadow the ctx variable so that you could write to state directly here and then return an error. Of course, that has some downsides too as it makes it more likely to accidentally use the wrong context inside the here (and it feels weird to write to state inside the branch callback).

Not sure which I prefer, they are both ugly to me :D

gjermundgaraba: Could another potential solution be to just do this loop before branching? We only really need it for the last line, right?

gjermundgaraba (approved):
It looks correct to me. I did leave a comment for discussion on how we can deal with the control flow. Let me know if you think other approaches could be useful here.

modules/apps/29-fee/keeper/escrow.go
damiannolan: I agree with your rational in general. I don't really love the use of errors as control flow per se as well. 

On the other hand this is an exceptional circumstance, and in most code flows we will likely not be needing partial state changes and such, so I would be fine to keep it like this. 

I kinda like the idea of not shadowing the ctx var and seeing how that behaves - in theory I think it should(?) work. But I also think its less explicit than what we are doing here already and having clear and readable code is more preferable imo when dealing with funds. 

> Could another potential solution be to just do this loop before branching? We only really need it for the last line, right?

This could potentially be an opt also maybe. But I'd like to stray from spending too much time on this rn given how much this module is used and its lifespan going forward.

DimitrisJim (approved):

Issue #7706, opened 2024-12-17
Handle events for transfer module


DimitrisJim:
As mentioned in a todo:

https://github.com/cosmos/ibc-go/blob/a5b69ec73687dec86e1f6ab5a2aebd9c667a5f3e/modules/apps/transfer/v2/keeper/keeper.go#L78-L79

Potentially needed in other places too, should be double checked.

possibly blocked on these being spec'ed out too as was done for core.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

Sruthi2357:
Shall I work on this? @DimitrisJim

gjermundgaraba:
> Shall I work on this? [@DimitrisJim](https://github.com/DimitrisJim)

Thank you for the interest, @Sruthi2357, but we will take this in the team as it requires some more discussion and possible spec input before we can complete it.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7782, opened 2024-12-21
chore: emit events

Sruthi2357:
Closes: #7706


gjermundgaraba:
Thank you for the PR, @Sruthi2357. We will take this on in the team as it might require some spec updates and updates to the events themselves.

PR #7784, opened 2024-12-22
fix: typos in documentation files

vtjl10:
Corrected `acccounts` to `accounts`


gjermundgaraba (approved):
Would you look at that, thank you! :)

PR #7797, opened 2024-12-24
refactor: use address.Codec to replace deprecated function

1ch0:
## Description

This PR replaces the deprecated address-related functions with the new `address.Codec` implementations:

* Replace `sdk.AccAddressFromBech32()` with `address.Codec  StringToBytes()`
* Replace `sdk.AccAddress.String()` with address.Codec BytesToString()`

These changes align with the latest SDK address handling practices and remove usage of deprecated functions.

### DIFF
--- a/modules/light-clients/08-wasm/client/cli/tx.go
+++ b/modules/light-clients/08-wasm/client/cli/tx.go
@@ -4,7 +4,9 @@ import (
 	"encoding/hex"
 	"fmt"
 	"os"
+	"sync"
 
+	"github.com/hashicorp/golang-lru/simplelru"
 	"github.com/spf13/cobra"
 
 	govcli "cosmossdk.io/x/gov/client/cli"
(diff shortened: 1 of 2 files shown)


gjermundgaraba (changes requested):
Hi! Thanks for the PR! I think there is a simpler way of doing this with the SDK. If you look at the bank module cli as an example: https://github.com/cosmos/cosmos-sdk/blob/release/v0.52.x/x/bank/client/cli/tx.go

I am not sure if this has been set up in our simapps yet, but it might also require some upgrade docs to make sure the client context always have this set up (especially since we don't support  depinject).

1ch0:
> Hi! Thanks for the PR! I think there is a simpler way of doing this with the SDK. If you look at the bank module cli as an example: https://github.com/cosmos/cosmos-sdk/blob/release/v0.52.x/x/bank/client/cli/tx.go
> 
> I am not sure if this has been set up in our simapps yet, but it might also require some upgrade docs to make sure the client context always have this set up (especially since we don't support depinject).

Thank you so much for taking the time to review my PR and for pointing me to the relevant example! I’ve only recently started learning about Cosmos, so I really appreciate your feedback and guidance. I’m excited to keep contributing to this project and to learn more from the community. Looking forward to collaborating more in the future!

1ch0:
> Hi! Thanks for the PR! I think there is a simpler way of doing this with the SDK. If you look at the bank module cli as an example: https://github.com/cosmos/cosmos-sdk/blob/release/v0.52.x/x/bank/client/cli/tx.go
> 
> I am not sure if this has been set up in our simapps yet, but it might also require some upgrade docs to make sure the client context always have this set up (especially since we don't support depinject).

#7802  Changed by reference example: https://github.com/cosmos/cosmos-sdk/blob/release/v0.52.x/x/bank/client/cli/tx.go

PR #7803, opened 2024-12-27
fix: typo bug-report.md

futreall:
**elegible - eligible**


gjermundgaraba (approved):
Thank you! This should have been caught in our spell-checker, but since you found this I was able to fix the issue so that we'll catch this next time 🙏

Issue #7825, opened 2025-01-06
query handlers are ignoring accumulate flag


beer-1:
## Summary of Bug

Affected Query Handlers: 
* [02-client grpc_query #1](https://github.com/cosmos/ibc-go/blob/decc8ec9ae8eeda9cf3791d45d3005a6e929a990/modules/core/02-client/keeper/grpc_query.go#L80C88-L80C98)
* [02-client grpc_query #2](https://github.com/cosmos/ibc-go/blob/decc8ec9ae8eeda9cf3791d45d3005a6e929a990/modules/core/02-client/keeper/grpc_query.go#L171)
* [02-client grpc_query #3](https://github.com/cosmos/ibc-go/blob/decc8ec9ae8eeda9cf3791d45d3005a6e929a990/modules/core/02-client/keeper/grpc_query.go#L213)
* [04-channel grpc_query](https://github.com/cosmos/ibc-go/blob/decc8ec9ae8eeda9cf3791d45d3005a6e929a990/modules/core/04-channel/keeper/grpc_query.go#L111)

These query handlers are not properly considering the accumulate flag, which leads to incorrect pagination results. Specifically:

When pagination.limit = 1 is set, the response includes 2 items instead of 1.
When pagination.count_total = true is enabled alongside pagination.limit = 1, the response incorrectly includes all items instead of limiting to just 1.
## Expected Behaviour

Should return `pagination.limit` of items

## Version

* main
* v9
* v8
* v7

## Steps to Reproduce

Do query with pagination setting

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged/assigned
- [ ] Estimate provided

charymalloju:
@beer-1 I would like to work on this issue.

beer-1:
yea sure please feel free to take and make PR!

gjermundgaraba:
Hi :) Normally the team would assign the issue, when we agree about the issue and solution, but it's all good. I'll assign @charymalloju to this and will drop a review in a moment.

I added a couple of things to the issue:
* There is one more place where this seems to be a potential issue, so I added a link to that for 04-channel grpc_query.go
* Looking through our current supported versions, this issue is likely a problem there too. I will add the backport tag to the connected PR which will open new PRs to those release branches (and hopefully there will be no conflicts 🤞)

gjermundgaraba:
We've decided to wait with this as 0.52 is on hold

PR #7831, opened 2025-01-09
fix: query handlers are ignoring accumulate flag

charymalloju:
## Description

closes: #7825 

---

write a little note why.

- [x] Updated relevant documentation (`docs/`).
- [x] Include a descriptive changelog entry when appropriate. This may be left to the discretion of the PR reviewers. (e.g. chores should be omitted from changelog)
- [x] Review `SonarCloud Report` in the comment section below once CI passes.


modules/core/02-client/keeper/grpc_query_test.go
```diff
@@ -116,13 +116,15 @@ func (suite *KeeperTestSuite) TestQueryClientStates() {
 		msg      string
 		malleate func()
 		expErr   error
+		expTotal int
```
gjermundgaraba, line 119: The name is a little confusing, since what you are checking is actually the count. So maybe this should be renamed to expCount.

gjermundgaraba: Also, please move this up to the beginning of the function (similar to `expClientStates`) and update it in malleate.

modules/core/02-client/keeper/grpc_query_test.go
```diff
@@ -157,6 +160,31 @@ func (suite *KeeperTestSuite) TestQueryClientStates() {
 				}
 			},
 			nil,
+			2,
+		},
+		{
+			"success request with pagination limit",
```
gjermundgaraba, line 166: Can you please add a tests case to all the other queries you fixed in this PR?

gjermundgaraba (changes requested):
Thank you for this PR! This will be nice to get fixed.

I think the solution itself is correct, but I added some comments in the tests.

I have linked to the issue for the PR, in which I added one more location that has the same issue, so please fix that one and the test cases.

AdityaSripal:
Hi, thanks for the pr! do you have a link to another filtered paginate query that does something similar so we can compare to an existing query

gjermundgaraba:
@charymalloju, did you get a chance to look at the comment from Aditya above?

Issue #7739, opened 2024-12-19
Aliasing new approach


AdityaSripal:
We internally discussed different approaches to aliasing v1 channels to v2 and weighed the potential complexities in ibc-go with the potential complexity of retaining a channel layer for new users of IBC.

We decided that having a channel layer was unnecessarily complex for new users, and that conceptually the idea of creating two clients and then directly connecting them together was much simpler. Thus, we are moving back to the `CounterpartyInfo` struct that we had discussed before which will be keyed directly against the clientID.

There will thus, only be two clients and counterparty information signalling the path to verify against. The packet handlers will remain and will directly retrieve and use these structs for processing.

We will also retain the idea that there is only a single identifier to create a Eureka connection between two chains in the standard case (rather than having a client identifier **and** a separate channel identifier)

In order to support existing channels, we will create an exceptional case handling in ibc-go that will create a direct map from the channel-id to the clientID to retreive the underlying client and `counterpartyInfo`

The counterpartyInfo itself should have only the `merklePathPrefix` and the `counterpartyId`

The `counterpartyId` is to be interpreted as the provable path under which the counterparty will write all outgoing packet messages toward us.

Thus, in the aliasing case; the `counterpartyId` should be the counterparty's original channel-id since the counterpartyId need not be strictly linked to the clientID.

Furthermore, given the new key formats introduced in https://github.com/cosmos/ibc/pull/1155 and the changes we made to supporting broader tree types; we should be able to have just a single field of the counterpartyId which should be explored by the implementor writing up this approach.

damiannolan:
Is the intended approach here to keep the existing msg server apis in 04-channel/v2?

This may just be more refactoring but it might be better in this case to not use 04-channel/v2 as a place to maintain the new code. If the intended desire is to not cause confusion and complexity around maintaining the concept of a channel in ibc/v2. 

I think the concept of a v2 channel being represented as the pairing of light clients on opposing chains was easily grokable but maybe to some this is confusing. 

If we move away from the concept of a v2 channel entirely then IMO it might be best to split the ibc/v2 code out completely away from 04-channel. Start with a flat enough go pkg structure and organise things as needed. This will likely take more dev time but in the long run I would like to see ibc/v2 as a new, separate go module from ibc-go classic. I think its a great opportunity to develop a more robust set of go modules which make up ibc/v2 and its apis.

[...]

PR #7842, opened 2025-01-13
Move channel back to counterparty

AdityaSripal:
## Description

closes: #7739 

---

write a little note why.


proto/ibc/core/client/v2/counterparty.proto
```diff
@@ -0,0 +1,43 @@
...
+// MsgRegisterCounterparty defines a message to register a counterparty on a client
+message MsgRegisterCounterparty {
+  option (cosmos.msg.v1.signer) = "signer";
+
+  option (gogoproto.goproto_getters) = false;
+
+  // client identifier
+  string client_id = 1;
+  // counterparty merkle prefix
+  repeated bytes merkle_prefix = 2;
```
gjermundgaraba, line 35: To avoid confusion, I would also name this counterparty_merkle_prefix.
Or restructure the message like this:
```proto
string client_id = 1;
CounterpartyInfo counterparty = 2;

File: modules/core/04-channel/v2/keeper/keeper.go [hunk 2]
```diff
@@ -252,6 +253,30 @@ func (k *Keeper) aliasV1Channel(ctx context.Context, portID, channelID string) (
 	return channelv2, true
 }
 
+// resolveV2Identifiers returns the client identifier and the counterpartyInfo for the client given the packetId
+// Note: For fresh eureka channels, the client identifier and packet identifier are the same.
+// For aliased channels, the packet identifier will be the original channel ID and the counterpartyInfo will be constructed from the channel
+func (k *Keeper) resolveV2Identifiers(ctx context.Context, portId string, packetId string) (string, clienttypes.CounterpartyInfo, error) {
```
gjermundgaraba, line 259: I think maybe the name could be better here. If I have to read the comment to understand what the function does for me, I consider it a code smell :)

proto/ibc/core/channel/v2/packet.proto
```diff
@@ -14,9 +14,9 @@ message Packet {
   // with a later sequence number.
   uint64 sequence = 1;
   // identifies the sending chain.
-  string source_channel = 2;
+  string source_id = 2;
```
gjermundgaraba, line 17: It will be the last time I bring this up (I hope), but what is the reason we don't want to name source and destination identifier client_id (as in source_client_id)? Can it be anything else?

modules/core/04-channel/v2/keeper/packet.go
```diff
@@ -19,33 +20,26 @@ import (
...
-	// Lookup channel associated with our source channel to retrieve the destination channel
-	channel, ok := k.GetChannel(ctx, sourceChannel)
-	if !ok {
-		// TODO: figure out how aliasing will work when more than one payload is sent.
-		channel, ok = k.convertV1Channel(ctx, payloads[0].SourcePort, sourceChannel)
-		if !ok {
-			return 0, "", errorsmod.Wrap(types.ErrChannelNotFound, sourceChannel)
-		}
+	// lookup counterparty and clientid from packet identifiers
+	clientID, counterparty, err := k.resolveV2Identifiers(ctx, payloads[0].SourcePort, sourceId)
```
gjermundgaraba, line 28: I thought this function was taking in a packet identifier? But it is taking in the source (client) identifier?

modules/core/04-channel/v2/keeper/keeper.go
gjermundgaraba: After seeing it used further down I am even more confused. Is it taking in packetId or a client id?

gjermundgaraba (commented):
Thanks! This is going to be very nice! 🙌

The general structure and logic seems straightforward and understandable. I have some questions on naming that I left in comments.

modules/core/04-channel/v2/keeper/keeper.go
AdityaSripal: Yea this was because it could also be an original channelID for aliasing, in which case I still need to retrieve the original clientiD

modules/apps/transfer/v2/keeper/msg_server_test.go
```diff
@@ -593,7 +593,7 @@ func (suite *KeeperTestSuite) TestV2RetainsFungibility() {
 	denomBtoC := transfertypes.Denom{
 		Base: sdk.DefaultBondDenom,
 		Trace: []transfertypes.Hop{
-			transfertypes.NewHop(transfertypes.ModuleName, pathv2.EndpointB.ChannelID),
+			transfertypes.NewHop(transfertypes.ModuleName, pathv2.EndpointB.ClientID),
 			transfertypes.NewHop(path.EndpointB.ChannelConfig.PortID, path.EndpointB.ChannelID),
```
AdityaSripal, line 597: First hop used IBC Classic, second hop used IBC Eureka

modules/core/04-channel/v2/keeper/grpc_query.go
```diff
@@ -136,19 +60,15 @@ func (q *queryServer) PacketCommitment(ctx context.Context, req *types.QueryPack
...
-	if err := host.ChannelIdentifierValidator(req.ChannelId); err != nil {
+	if err := host.ClientIdentifierValidator(req.ClientId); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	if req.Sequence == 0 {
 		return nil, status.Error(codes.InvalidArgument, "packet sequence cannot be 0")
 	}
 
-	if !q.HasChannel(ctx, req.ChannelId) {
```
AdityaSripal, line 147: This was a pain to find the analog of especially in a world with aliasing. Just decided to remove it since if client (or aliased client) doesn't exist, none of these other auxilliary state will exist either

modules/core/04-channel/v2/keeper/grpc_query_test.go
```diff
@@ -347,31 +344,21 @@ func (suite *KeeperTestSuite) TestQueryPacketCommitment() {
...
-					ChannelId: ibctesting.FirstChannelID,
-					Sequence:  0,
+					ClientId: ibctesting.FirstClientID,
+					Sequence: 0,
 				}
 			},
 			status.Error(codes.InvalidArgument, "packet sequence cannot be 0"),
 		},
-		{
-			"channel not found",
```
AdityaSripal, line 357: Removed these cases. They still error but now equivalent to the queried key not existing

proto/ibc/core/channel/v2/query.proto
```diff
@@ -5,132 +5,60 @@ package ibc.core.channel.v2;
...
-  // ChannelConsensusState queries for the consensus state for the channel associated
-  // with the provided channel identifiers.
-  rpc ChannelConsensusState(QueryChannelConsensusStateRequest) returns (QueryChannelConsensusStateResponse) {
-    option (google.api.http).get = "/ibc/core/channel/v2/channels/{channel_id}/consensus_state";
-  }
-
   // NextSequenceSend returns the next send sequence for a given channel.
   rpc NextSequenceSend(QueryNextSequenceSendRequest) returns (QueryNextSequenceSendResponse) {
-    option (google.api.http).get = "/ibc/core/channel/v2/channels/{channel_id}/next_sequence_send";
+    option (google.api.http).get = "/ibc/core/channel/v2/clients/{client_id}/next_sequence_send";
```
AdityaSripal, line 17: There rpc urls are very ugly since they still contain channel/v2. Should we pull this whole thing back out to its own package `packetserver`? 🫣

proto/ibc/core/client/v2/counterparty.proto
```diff
@@ -0,0 +1,43 @@
...
+
+option go_package = "github.com/cosmos/ibc-go/v9/modules/core/02-client/types";
+
+import "cosmos/msg/v1/msg.proto";
+import "gogoproto/gogo.proto";
+
+// CounterpartyInfo defines the key that the counterparty will use to message our client
+message CounterpartyInfo {
+  // merkle prefix key is the prefix that ics provable keys are stored under
+  repeated bytes merkle_prefix = 1;
```
AdityaSripal, line 13: I decided to just make this `repeated bytes` rather than importing ics23.commitment MerklePath

I think the inclusion of ics23 into the required external structs has caused more confusion than anything else. It is also an unnecessary import here

I convert to ics23 commitment MerklePath internally

gjermundgaraba:
> Now that we have channel completely removed should we move `channel/v2` into a different package? Perhaps back to `packetserver`

I think that would make sense and keep things much cleaner. Not sure what the best name is, but separate package for sure.

gjermundgaraba (commented):
I like where this is going! Left some comments on names only :)

modules/core/keeper/msg_server.go
```diff
@@ -40,9 +41,33 @@ func (k *Keeper) CreateClient(ctx context.Context, msg *clienttypes.MsgCreateCli
...
+	k.ClientKeeper.SetClientCreator(ctx, clientID, sdk.AccAddress(msg.Signer))
+
 	return &clienttypes.MsgCreateClientResponse{ClientId: clientID}, nil
 }
 
+// RegisterCounterparty will register the eureka counterparty info for the given client id
+// it must be called by the same relayer that called CreateClient
+func (k *Keeper) RegisterCounterparty(ctx context.Context, msg *clienttypes.MsgRegisterCounterparty) (*clienttypes.MsgRegisterCounterpartyResponse, error) {
+	creator := k.ClientKeeper.GetClientCreator(ctx, msg.ClientId)
+	if !creator.Equals(sdk.AccAddress(msg.Signer)) {
```
srdtrk, line 54: Should you instead have the creator become a string? Should be consistent with how we do these comparisons elsewhere?

proto/ibc/core/channel/v2/query.proto
srdtrk: I don't mind the urls but up to you

proto/ibc/core/client/v2/counterparty.proto
srdtrk: I think it is ok to have this be `ics23.commitment.MerklePath` but use `repeated bytes` in `MsgRegisterCounterparty` since `CounterpartyInfo` is used internally anyway?

But this is also good :)

srdtrk (approved):
lgtm. Surprised so many lines were deleted. This is the right way it seems.

proto/ibc/core/channel/v2/packet.proto
```diff
@@ -13,10 +13,10 @@ message Packet {
   // with an earlier sequence number must be sent and received before a Packet
   // with a later sequence number.
   uint64 sequence = 1;
-  // identifies the sending chain.
-  string source_channel = 2;
-  // identifies the receiving chain.
-  string destination_channel = 3;
+  // identifies the sending client on the sending chain.
+  string source_client = 2;
```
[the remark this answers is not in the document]
AdityaSripal: I will leave this out for now, since this is analogous to what has existed in ibc-go forever. We can make a decision to change in a different PR/issue

modules/core/keeper/msg_server.go
AdityaSripal: Left as is now, since it that case, we may as well just have the keeper return a string. Maybe we can do later as a small followup if really necessary

proto/ibc/core/channel/v2/packet.proto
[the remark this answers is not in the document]
gjermundgaraba: Created an issue to track this here #7854

Issue #7642, opened 2024-12-09
Clean up and fix BuildMerklePath


gjermundgaraba:
* BuildMerklePath should panic on empty prefix (not return like it does now)
* BuildMerklePath needs to be cleaned up (it ugly!)
* Move it to 23-commitment and to be a method on MerklePath that clones itself and returns a new one
* Document it well

lacsomot:
Hi. Is this still needed ?

DimitrisJim:
this one might be slightly more involved @lacsomot 😬 Maybe best to see if some other good issues pop up soon!

gjermundgaraba:
Closed by #7863

PR #7863, opened 2025-01-21
imp: simplified `BuildMerklePath`

srdtrk:
## Description

closes: #7642

---

write a little note why.


modules/core/04-channel/v2/types/merkle.go
```diff
@@ -1,31 +1,20 @@
...
 
 // BuildMerklePath takes the merkle path prefix and an ICS24 path
 // and builds a new path by appending the ICS24 path to the last element of the merkle path prefix.
-func BuildMerklePath(prefix commitmenttypesv2.MerklePath, path []byte) commitmenttypesv2.MerklePath {
-	if prefix.Empty() {
-		return commitmenttypes.NewMerklePath(path)
+func BuildMerklePath(prefix [][]byte, path []byte) commitmenttypesv2.MerklePath {
+	prefixLength := len(prefix)
+	if prefixLength == 0 {
+		panic("cannot build merkle path with empty prefix")
```
AdityaSripal, line 12: Don't think this should necessarily error. If there's no prefix, the path should just be returned as it was originally. See removed diff

This would be equivalent to a chain storing the ICS24 paths directly under root. Which is potentially possible for a solomachine chain or something equivalent

modules/core/04-channel/v2/types/merkle.go
```diff
@@ -1,31 +1,20 @@
...
 	}
 
-	// avoid mutating the provided prefix
-	prefixKeys := make([][]byte, len(prefix.KeyPath))
-	copy(prefixKeys, prefix.KeyPath)
-
-	lastElement := prefixKeys[len(prefixKeys)-1]
+	// copy prefix to avoid modifying the original slice
+	prefix = append([][]byte(nil), prefix...)
```
AdityaSripal, line 16: Should we name this var to something else since it eventually becomes the full path?

modules/core/04-channel/v2/types/merkle.go
AdityaSripal: What is the reasoning for changing the behavior here?

AdityaSripal: Ahh i see it was decided in the original issue! Ok, i mean i suppose we could require all implementations to store under a prefix

gjermundgaraba: I don't remember the entire context for why we said this, but it was something we mentioned during the alpha audit itself, I believe.

srdtrk: I could take this back. I don't feel strongly about it. What should we do?

modules/core/04-channel/v2/types/merkle.go
gjermundgaraba: Agreed.
Also good to not reassign or change function arguments, so would prefer a new name in any case.

modules/core/04-channel/v2/types/merkle.go
gjermundgaraba: @AdityaSripal can have the final say on it :)

modules/core/04-channel/v2/types/merkle.go
srdtrk: done

modules/core/04-channel/v2/types/merkle.go
AdityaSripal: I think its more likely that someone accidentally leaves prefix empty than having a counterparty that is genuinely storing everything under root.

So lets have the error for now, and if we ever have a user run into this issue and report it we can change the behavior of this function quite easily

AdityaSripal (approved):

PR #7883, opened 2025-01-25
feat(client): migrate block header data to use comet service

DeVikingMark:
This commit implements the first step of migrating away from sdk.Context 
to using comet service for accessing block header information. Specifically:

- Add blockHeader parameter of type RequestBeginBlock
- Use blockHeader.Header for Time and NextValidatorsHash
- Remove TODO comment as the primary goal is implemented
- Update documentation

This is part of a larger effort to reduce dependency on sdk.Context.
Future work will include migrating the remaining sdk.Context usages
to more specific interfaces.

### DIFF
--- a/modules/core/02-client/abci.go
+++ b/modules/core/02-client/abci.go
@@ -2,18 +2,20 @@ package client
 
 import (
 	"context"
+	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	cometabci "github.com/cometbft/cometbft/abci/types"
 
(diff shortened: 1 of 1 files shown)


technicallyty (changes requested):
Purple monkey banana orange juice? Pancakes and waffle cereal. Giraffe skateboarding through treacherous landscapes. Alpacas?

gjermundgaraba:
@DeVikingMark, it seems like you might be trying to run PR runs with some kind of AI agent. Please don't. I would rather not waste our time on reviewing _maybe_ compiling code (FYI, this does not).

Issue #7923, opened 2025-02-06
Refactor 02-client v2 with a standardized msg server and genesis


gjermundgaraba:
02-client is currently incorrectly set up with a message server in `proto/ibc/core/02-client/v2/counterparty.proto`. It is also missing genesis.

There are two ways I see this being wired up correctly:
1. Add a txv2.proto, genesisv2.proto or something similar and implement that in the current 02-client module
  * The question here is what the correct way to do this is when you add new things like this. We've added new state and new tx's and queries, but we already have a v1. Not sure what is the recommended approach.
  * Should it be a new module, or is it OK to wire it up in the existing module?
2. Create and wire up a fully separate module similar to how channel v2 has been set up.

Would like feedback on the best way to do this.

aljo242:
The best path forward here will be to create and wire up a separate module like channel v2 because we need them to implement different msg and query servers

Issue #7875, opened 2025-01-24
Add a query for CounterpartyInfo


srdtrk:
This query should have been added after removing channels from IBC Eureka.

We need a new basic new query in 02-client/v2 that takes in a client_id and returns the CounterpartyInfo.

ArtGinn:
Hi @srdtrk, could you point the line of logic to remove channels? then I can handle this

gjermundgaraba:
Hi @ArtGinn, channels have already been removed for Eureka. The query itself is handled by myself, so help needed on this right now :)

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #7936, opened 2025-02-07
feat: clientv2 module

aljo242:
## Description

closes: #7875
closes: #7923

- move all `Counterparty` keeper and type logic to a `v2` package
- create query implementation 
- create and wire genesis into core module

---

write a little note why.


gjermundgaraba (changes requested):
Similar to the comment in channel-v2. Genesis (including tests) and most of the wiring happens in the core module for v1. v2 should do the same.

modules/core/02-client/v2/keeper/grpc_query.go
```diff
@@ -0,0 +1,40 @@
...
+func NewQueryServer(k *Keeper) types.QueryServer {
+	return &queryServer{
+		Keeper: k,
+	}
+}
+
+// CounterpartyInfo gets the CounterpartyInfo from the store corresponding to the request client ID.
+func (q queryServer) CounterpartyInfo(ctx context.Context, request *types.QueryCounterpartyInfoRequest) (*types.QueryCounterpartyInfoResponse, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
```
AdityaSripal, line 33: We generally do a bit more validation before the grpc directly gets from store. Would be good to follow the same.

See the v1 chanKeeper grpc implementation

modules/core/02-client/v2/keeper/grpc_query.go
AdityaSripal, line 1: Should have a test file for this as well.

modules/core/02-client/v2/keeper/keeper_test.go
```diff
@@ -0,0 +1,112 @@
...
+	keeper         *keeper.Keeper
+	consensusState *ibctm.ConsensusState
+	valSet         *cmttypes.ValidatorSet
+	valSetHash     cmtbytes.HexBytes
+	privVal        cmttypes.PrivValidator
+	now            time.Time
+	past           time.Time
+	solomachine    *ibctesting.Solomachine
+
+	signers map[string]cmttypes.PrivValidator
```
AdityaSripal, line 52: Lot of unnecessary fields here

modules/core/04-channel/v2/types/expected_keepers.go
```diff
@@ -23,6 +23,4 @@ type ClientKeeper interface {
 	GetClientState(ctx context.Context, clientID string) (exported.ClientState, bool)
 	// GetClientConsensusState gets the stored consensus state from a client at a given height.
 	GetClientConsensusState(ctx context.Context, clientID string, height exported.Height) (exported.ConsensusState, bool)
-	// GetClientCounterparty returns the counterpartyInfo given a clientID
-	GetClientCounterparty(ctx context.Context, clientID string) (clienttypes.CounterpartyInfo, bool)
```
AdityaSripal, line 27: I kind of think this should remain but have `clientV2Keeper` implement this interface and only have channelKeeper take in a ClientV2Keeper.

What do you think @gjermundgaraba

modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -37,13 +41,15 @@ func NewKeeper(
 	cdc codec.BinaryCodec,
 	env appmodule.Environment,
 	clientKeeper types.ClientKeeper,
```
AdityaSripal, line 43: Think we should still just have this interface and have `clientV2Keeper` implement it

AdityaSripal (changes requested):
I think I would prefer the channelV2Keeper still only taking a single clientV2Keeper since the fact that they are different is an implementation detail that only exists because ibc-go is implementing both IBC Classic and IBC Eureka

But open to pushback from the other engineers

modules/core/04-channel/v2/types/expected_keepers.go
gjermundgaraba: To me it seems a bit strange to have it on v1, since it is not used there. The only reason to keep it on v1 would be to avoid having to deal with v2. I feel like it would be less indirection and clearer code if we use the v2 keeper directly when needed.

modules/core/04-channel/v2/keeper/keeper.go
aljo242: Do you mean have the clientV2Keeper implement the full `ClientKeeper` interface?

modules/core/04-channel/v2/types/expected_keepers.go
aljo242: @AdityaSripal do you mean to have the ClientV2Keeper basically wrap the client V1 Keeper and extend it with this extra CounterParty functionality?

modules/core/02-client/v2/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,89 @@
...
+			status.Error(codes.InvalidArgument, "identifier cannot be blank: invalid identifier"),
+		},
+		{
+			"counterparty not found",
+			func() {
+				path1 := ibctesting.NewPath(suite.chainA, suite.chainB)
+				path1.SetupClients()
+				// counter party not set up
+
+				expInfo = types.NewCounterpartyInfo(path1.EndpointA.Counterparty.MerklePathPrefix.KeyPath, path1.EndpointA.ClientID)
```
AdityaSripal, line 46: this doesn't need to exist right since we have non-nil err

modules/core/02-client/v2/types/genesis.go
```diff
@@ -0,0 +1,36 @@
...
+// DefaultGenesisState returns the ibc client submodule's default genesis state.
+func DefaultGenesisState() GenesisState {
+	return GenesisState{
+		CounterpartyInfos: []GenesisCounterpartyInfo{},
+	}
+}
+
+// Validate checks the CounterpartyInfos added to the genesis for validity.
+func (gs GenesisState) Validate() error {
+	seenIDs := make(map[string]struct{})
```
AdityaSripal, line 14: why do you do this instead of a bool?

AdityaSripal: It's a good check but don't think we do this elsewhere

AdityaSripal: ahh i see. no its fine keep it as-is

modules/core/04-channel/v2/types/expected_keepers.go
AdityaSripal: Yes, but i will do this in a separate PR probably after the audit

AdityaSripal (approved):

PR #7938, opened 2025-02-09
Add error handling in BeginBlock for IBC client

avorylli:
- Add proper error handling for BeginBlocker in core module
- Wrap error with context for better debugging
- Improve reliability by not silently ignoring potential errors

### DIFF
--- a/modules/core/module.go
+++ b/modules/core/module.go
@@ -197,7 +197,9 @@ func (AppModule) ConsensusVersion() uint64 { return 7 }
 
 // BeginBlock returns the begin blocker for the ibc module.
 func (am AppModule) BeginBlock(ctx context.Context) error {
-	ibcclient.BeginBlocker(sdk.UnwrapSDKContext(ctx), am.keeper.ClientKeeper)
+	if err := ibcclient.BeginBlocker(sdk.UnwrapSDKContext(ctx), am.keeper.ClientKeeper); err != nil {
+		return fmt.Errorf("failed to begin block for IBC client: %w", err)
+	}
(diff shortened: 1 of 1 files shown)


modules/core/module.go
```diff
@@ -197,7 +197,9 @@ func (AppModule) ConsensusVersion() uint64 { return 7 }
 
 // BeginBlock returns the begin blocker for the ibc module.
 func (am AppModule) BeginBlock(ctx context.Context) error {
-	ibcclient.BeginBlocker(sdk.UnwrapSDKContext(ctx), am.keeper.ClientKeeper)
+	if err := ibcclient.BeginBlocker(sdk.UnwrapSDKContext(ctx), am.keeper.ClientKeeper); err != nil {
```
gjermundgaraba, line 200: It's OK to catch the error and log it, but we cannot return an error here as it would not let the block finalize and subsequently halt the chain

gjermundgaraba (changes requested):

modules/core/module.go
```diff
@@ -197,7 +197,10 @@ func (AppModule) ConsensusVersion() uint64 { return 7 }
...
+		// Log the error but do not return it to avoid halting the chain
+		fmt.Printf("non-critical error in IBC BeginBlocker: %v\n", err)
```
gjermundgaraba, line 202: fmt is not the correct way to log errors in our modules. Take a look at how it is done elsewhere to find some examples :)

gjermundgaraba (changes requested):

Issue #7999, opened 2025-02-23
Remove channel upgradability


gjermundgaraba:
With the advent of IBC v2, and the removal of ICS20 v2, there are very few real use cases left for channel upgradability. The only real use left is for ICA channels to change from ORDERED to UNORDERED, but most of the users of that has already done this work by now, and anyone who still needs to do this, can still do this on ibc-go v7 and v8.

The reason for wanting to remove channel upgradability is because it's a large feature with a lot of lines of code that need to be maintained and kept secure. With few use cases left for it, we want to just get rid of it to keep the code base leaner.

To do this, we will need to remove all the code related to upgradability, and introduce a store migration to get rid of all unused state.
In addition, we will need to tests:
- Store migration tests happy path (i.e. no existing channels in the middle of an upgrade)
- Store migration tests with channels in all currently possible states (UNINITIALIZED, INIT, TRYOPEN, OPEN, CLOSED, FLUSHING, FLUSHCOMPLETE) and make sure they have the correct state and values otherwise afterwards
- E2E upgrade tests

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #8008, opened 2025-02-25
feat: remove channel upgradability

gjermundgaraba:
## Description

closes: #7999 

---

write a little note why.


modules/core/04-channel/keeper/handshake_test.go
```diff
@@ -562,132 +561,6 @@ func (suite *KeeperTestSuite) TestChanCloseInit() {
 	}
 }
 
-// TestChanCloseConfirm tests the confirming closing channel ends by calling ChanCloseConfirm
-// on chainB. Both chains will use message passing to setup OPEN channels. ChanCloseInit is
-// bypassed on chainA by setting the channel state in the ChannelKeeper.
-func (suite *KeeperTestSuite) TestChanCloseConfirm() {
```
AdityaSripal, line 568: Why was this test removed. Only upgrade case should be removed

modules/core/04-channel/keeper/migrations.go
```diff
@@ -16,10 +16,6 @@ func NewMigrator(keeper *Keeper) Migrator {
...
 }
 
-// MigrateParams migrates params to the default channel params.
-func (m Migrator) MigrateParams(ctx sdk.Context) error {
-	params := channeltypes.DefaultParams()
-	m.keeper.SetParams(ctx, params)
-	m.keeper.Logger(ctx).Info("successfully migrated ibc channel params")
-	return nil
+func (m *Migrator) Migrate7To8(ctx sdk.Context) error {
+	return v10.MigrateStore(ctx, m.keeper.storeService, m.keeper.cdc, m.keeper)
```
AdityaSripal, line 20: Why are you doing v10 migrate store in this function?

modules/core/04-channel/migrations/v10/store.go
```diff
@@ -0,0 +1,117 @@
...
+func handleChannelMigration(ctx sdk.Context, store corestore.KVStore, cdc codec.BinaryCodec, channelKeeper ChannelKeeper) error {
+	// Remove channel upgrade sequences and set in-upgrade channels back to open
+	iterator := storetypes.KVStorePrefixIterator(runtime.KVStoreAdapter(store), []byte(host.KeyChannelEndPrefix))
+
+	defer sdk.LogDeferred(channelKeeper.Logger(ctx), func() error { return iterator.Close() })
+	for ; iterator.Valid(); iterator.Next() {
+		var channel Channel
+		cdc.MustUnmarshal(iterator.Value(), &channel)
+
+		if channel.State == FLUSHING || channel.State == FLUSHCOMPLETE {
```
AdityaSripal, line 79: This will mess with the channel upgrade procedure. If we have a FLUSHCOMPLETE on one side, we are expecting the upgrade to occur successfully.

So if you downgrade from this point, you risk the other side upgrading successfully using your FLUSHCOMPLETE state and having the channel be invalid.

I would suggest that we just panic migrating this if any channel is in FLUSHING or FLUSHCOMPLETE.

Since these are gated by governance ultimately, we can do the upgrade to v10 when we have no active upgrades in-flight.

Creates much less headache for us thinking through these cases

modules/core/04-channel/keeper/migrations.go
gjermundgaraba: This is to have a migrator for channel, but keep the migration logic out of the keeper itself. We have the same process in 02-client:
```
func (m Migrator) Migrate2to3(ctx sdk.Context) error {
	return v7.MigrateStore(ctx, m.keeper.storeService, m.keeper.cdc, m.keeper)
}
```

modules/core/04-channel/migrations/v10/store.go
gjermundgaraba: Yeah, that makes sense. I was considering that, but figured it might just work out if both sides migrated. I guess panic might be safer in this case. Will fix.

modules/core/04-channel/keeper/handshake_test.go
gjermundgaraba: Huh, good point. Will check that!

modules/core/04-channel/migrations/v10/store.go
gjermundgaraba: It will error instead now

modules/core/04-channel/keeper/handshake_test.go
gjermundgaraba: Added back

modules/core/04-channel/keeper/migrations.go
AdityaSripal: Yes but wouldn't this be called `Migrate8To10`?

AdityaSripal (approved):
LGTM module one nit

modules/core/04-channel/keeper/migrations.go
gjermundgaraba: That number refers to the consensus version of the store. It's a bit of a weird mix. Not sure what the best solution is. But client also mixes it the same way.

PR #8074, opened 2025-02-28
Update IBC client version in quick guide from v6.2.0 to v6.4.1

leopardracer:
This PR updates the IBC client version reference in the quick guide documentation from v6.2.0 to v6.4.1.

The change updates the client.proto URL path to point to the latest v6.4.1 version instead of v6.2.0.

This ensures the documentation stays in sync with the latest released version.


gjermundgaraba:
Hi, thank you for the PR, but we generally don't accept small non-functional changes (such as small text changes) without an accepted issue. This is because we get a lot of PRs and it takes up valuable time of the team without much value added.

PR #8156, opened 2025-03-06
ibc-callbacks: Use custom callback type to satisfy interface

AdityaSripal:
## Description

Currently v2 ibc-callbacks wraps the `recvResult.Acknowledgement` in a channeltypesv2.Acknowledgement. This causes the top-level acknowledgement to be encoded with proto:

https://github.com/cosmos/ibc-go/blob/main/modules/apps/callbacks/v2/ibc_middleware.go#L192
https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/v2/types/acknowledgement.go#L46

This is then passed to the contract by the cosmwasm implementation:
https://github.com/CosmWasm/wasmd/blob/38fd8660b944d8db58435cecb924234d36335a03/x/wasm/ibc.go#L416

However, most contracts are expecting to just get the individual app ack bytes rather than this wrapped type.

So i create a special type here for recv Acknowledgement to get the proper bytes to the contract. We can also always return true in `Success` since error acks do not reach this line at all.

This will fit the contract expectations better than it did previously

closes: #XXXX

---

write a little note why.


gjermundgaraba (approved):
Seems legit. I am down to to do this. Seems to be a minor linter issue only :)

modules/apps/callbacks/v2/ibc_middleware.go
```diff
@@ -14,9 +14,29 @@ import (
...
+
+// Create internal implementation of exported.Acknowledgement
+// to pass the app acknowledgement bytes to contractKeeper IBCReceivePacketCallback
+// interface
+type RecvAcknowledgement []byte
+
+// RecvPacket only passes callback to contract if the acknowledgement
+// is successful. Thus, we can just return true here.
+func (rack RecvAcknowledgement) Success() bool {
+	return true
```
srdtrk, line 33: This is not we did with `channeltypes2.Acknowledgement`? Don't we do this check: https://github.com/cosmos/ibc-go/blob/75c3eed2d57e287fd46d64a6e436cec06c69768b/modules/core/04-channel/v2/types/acknowledgement.go#L42

srdtrk (changes requested):

modules/apps/callbacks/v2/ibc_middleware.go
AdityaSripal: Yes, im just satisfying the interface since it is guaranteed to not be ErrorAcknowledgement here

srdtrk: Yes, but why don't we add this check anyway. If we change the callback behavior in the future (like making callbacks on failures as well), adding this would save us.

AdityaSripal: Ok i can add

srdtrk (approved):
lgtm

PR #8172, opened 2025-03-12
Permission rly v2 per client

AdityaSripal:
## Description

The following PR allows client creators and the chain authority to optionally permission the V2 relaying on an IBC client along with UpdateClient

The configs are set **per** clientID in the v2 keeper and can be set by the creator or authority to a list of relayer addresses.

If the list is empty, this is interpreted as allowing permissionless relaying.

If the list is non-empty, only a relayer that matches an element in the list will be allowed to call `UpdateClient` and the v2 packet messages.

This is to ensure safety in the initial rollout of IBC v2. NOTE: All of the usual IBC client checks are still performed so we are not trusting the relayers for correctness.

We are trusting the relayers for liveness, and should there be an exploit in the new light clients being deployed with IBC V2; then we are trusting that these relayers will not exploit them. This is to ensure a safe and secure rollout for IBC V2.

We intend to remove this restriction and get back to permissionless relaying as the ecosystem gains confidence in the new protocol

TODO:

- [x] Test permissioning
- [x] Query params and creator
- [x] Delete creator msg
- [ ] E2E/manual testing?

closes: #XXXX

---

write a little note why.


modules/core/02-client/keeper/grpc_query.go
```diff
@@ -261,6 +261,24 @@ func (q *queryServer) ClientStatus(goCtx context.Context, req *types.QueryClient
 	}, nil
 }
 
+// ClientCreator implements the Query/ClientCreator gRPC method
+func (q *queryServer) ClientCreator(goCtx context.Context, req *types.QueryClientCreatorRequest) (*types.QueryClientCreatorResponse, error) {
```
AdityaSripal, line 265: Added this query since i think it will be useful now that its being persisted

modules/core/04-channel/v2/keeper/msg_server_test.go
```diff
@@ -473,8 +545,6 @@ func (suite *KeeperTestSuite) TestMsgTimeout() {
 
 			tc.malleate()
 
-			suite.Require().NoError(path.EndpointA.UpdateClient())
```
AdityaSripal, line 476: I did this because i wanted to malleate after update in one case

modules/core/keeper/msg_server.go
```diff
@@ -81,6 +80,15 @@ func (k *Keeper) UpdateClient(goCtx context.Context, msg *clienttypes.MsgUpdateC
 		return nil, err
 	}
 
+	// only check v2 params if this chain is setup with v2 clientKeepr
+	if k.ClientV2Keeper != nil {
+		// check if this relayer is allowed to update if v2 params are set
+		params := k.ClientV2Keeper.GetParams(ctx, msg.ClientId)
+		if !params.IsAllowedRelayer(sdk.MustAccAddressFromBech32(msg.Signer)) {
```
AdityaSripal, line 87: Note if params is unset this will pass

proto/ibc/core/client/v2/params.proto
```diff
@@ -0,0 +1,14 @@
...
+
+package ibc.core.client.v2;
+
+option go_package = "github.com/cosmos/ibc-go/v10/modules/core/02-client/v2/types";
+
+// Params is a **per-client** parameter that sets which relayers are allowed to relay v2 IBC messages
+// for a given client.
+// If it is set, then only relayers in the allow list can send v2 messages
+// If it is not set, then the client allows permissionless relaying of v2 messages
+message Params {
```
srdtrk, line 11: Maybe should be called `ClientParams`. As this can be confused with a cosmos-sdk module param

modules/core/02-client/v2/types/params.go
```diff
@@ -0,0 +1,53 @@
...
+func (p Params) Validate() error {
+	return validateRelayers(p.AllowedRelayers)
+}
+
+// IsAllowedRelayer checks if the given address is registered on the allowlist.
+func (p Params) IsAllowedRelayer(relayer sdk.AccAddress) bool {
+	if len(p.AllowedRelayers) == 0 {
+		return true
+	}
+	for _, r := range p.AllowedRelayers {
```
srdtrk, line 34: You can use slices instead of a for loop. The newer go versions should be recommending this

proto/ibc/core/client/v1/query.proto
```diff
@@ -195,6 +200,20 @@ message QueryClientParamsResponse {
   Params params = 1;
 }
 
+// QueryClientCreatorRequest is the request type for the Query/ClientCreator RPC
+// method.
+message QueryClientCreatorRequest {
```
gjermundgaraba, line 205: Should this be in v2?

modules/core/keeper/msg_server.go
srdtrk: Should something like this be in membership as well?

modules/core/keeper/msg_server.go
```diff
@@ -68,7 +68,6 @@ func (k *Keeper) RegisterCounterparty(goCtx context.Context, msg *clientv2types.
 	// initialize next sequence send to enable packet flow
 	k.ChannelKeeperV2.SetNextSequenceSend(ctx, msg.ClientId, 1)
 
-	k.ClientKeeper.DeleteClientCreator(ctx, msg.ClientId)
```
srdtrk, line 71: Isn't this a problem since the counterparty info can now change at anytime? Which is not just a DOS but also a security concern

modules/core/04-channel/v2/keeper/msg_server.go
```diff
@@ -52,6 +53,12 @@ func (k *Keeper) RecvPacket(goCtx context.Context, msg *types.MsgRecvPacket) (*t
 		return nil, errorsmod.Wrap(err, "invalid address for msg Signer")
 	}
 
+	// check if this client is allowed to update if v2 params are set
+	params := k.clientV2Keeper.GetParams(ctx, msg.Packet.DestinationClient)
```
srdtrk, line 57: Hmm, I guess putting this in membership was hard.

srdtrk (commented):
You stopped deleting the client creator. Doesn't that mean they can change the counterparty info anytime now?

modules/core/keeper/msg_server.go
gjermundgaraba: I think the goal was to add that as a separate message later, but maybe the way to deal with this is to disable sending on a client until the client creator has been deleted (once we add that)?

AdityaSripal: Hmm yes, i forgot this was the main protection against rerunning this message as well. I will instead ensure that you cannot register if it has already been set.

proto/ibc/core/client/v1/query.proto
AdityaSripal: The keys set/get/delete are in v1 keeper which is why i left the query here as well

modules/core/keeper/msg_server.go
gjermundgaraba: Another option might be to either:
1. Don't allow sending on clients that still has a creator set
2. Delete client creator here still, such that setting the client params can only be done between client creation and setting the counterparty.

AdityaSripal: No I want the creator here to be set indefinitely because I want the ability for creators to change the allowed relayer list long after it has been initially set up

gjermundgaraba (approved):

PR #8190, opened 2025-03-14
test: set expError for fake acknowledgement test case

rootulp:
## Context

https://celestia-team.slack.com/archives/C01BZ4XNGFK/p1741950351888399

cc: @gjermundgaraba 

## Problem

[This PR](https://github.com/cosmos/ibc-go/commit/a5b5b929831859be62c5f7dc94970efa868cc84b) added a test case for the most recent security vulnerability but the test case doesn't verify that the error is of type `types.ErrInvalidAcknowledgement`

### DIFF
--- a/modules/core/04-channel/keeper/packet_test.go
+++ b/modules/core/04-channel/keeper/packet_test.go
@@ -699,6 +699,7 @@ func (suite *KeeperTestSuite) TestAcknowledgePacket() {
 		{
 			"fake acknowledgement",
 			func() {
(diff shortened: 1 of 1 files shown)


rootulp:
We don't need to forward port this PR to v8.x or v10.x because the unit tests were refactored between v7.x and v8.x and v8.x and above already assert on the expected error

gjermundgaraba (approved):

gjermundgaraba:
Thank you for the PR 🙏

PR #8396, opened 2025-05-15
test: replace boolean flags with expected errors in 07-tendermint tests

avorylli:
Replace `expPass` boolean values with specific `expError` values in 07-tendermint light client tests. This change improves test clarity by explicitly stating what error is expected instead of using generic pass/fail flags.

Modified files:
- modules/light-clients/07-tendermint/update_test.go
- modules/light-clients/07-tendermint/store_test.go

All tests continue to pass with the updated structure.

closes: #7175


gjermundgaraba:
There are parts here that are wrong. I also think the changes in this case doesn't make a lot of sense. I'll close this PR for now and also close out the issues as I don't think we need to resolve more of these for now.

Issue #8372, opened 2025-05-12
Go style guide refresh


gjermundgaraba:
@desherbojhaa to fill in

PR #8415, opened 2025-05-21
chore: update style guide

DeshErBojhaa:
## Description

This PR will be a discussion place to finalize our updated style guide. With the unleash the builder initiative, and new people joining the team, it's high time we solidify best practices so that the code remains in high standard. 

closes: #8372 

---

write a little note why.

- [ ] ~Include changelog entry when appropriate (e.g. chores should be omitted from changelog).~
- [ ] ~Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package) if relevant.~
- [ ] ~Updated documentation (`docs/`) if anything is changed.~
- [ ] ~Added `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code) if relevant.~
- [x] Self-reviewed `Files changed` in the GitHub PR explorer.

### DIFF
--- a/docs/dev/go-style-guide.md
+++ b/docs/dev/go-style-guide.md
@@ -3,7 +3,7 @@
 
 In order to keep our code looking good with lots of programmers working on it, it helps to have a "style guide", so all the code generally looks quite similar. This doesn't mean there is only one "right way" to write code, or even that this standard is better than your style.  But if we agree to a number of stylistic practices, it makes it much easier to read and modify new code. Please feel free to make suggestions if there's something you would like to add or modify.
 
-We expect all contributors to be familiar with [Effective Go](https://golang.org/doc/effective_go.html) (and it's recommended reading for all Go programmers anyways). Additionally, we generally agree with the suggestions in [Uber's style guide](https://github.com/uber-go/guide/blob/master/style.md) and use that as a starting point.
+We expect all contributors to be familiar with [Effective Go](https://golang.org/doc/effective_go.html) (and it's recommended reading for all Go programmers anyways). Additionally, we generally agree with the suggestions in [Google's style guide](https://google.github.io/styleguide/go/index) and use that as a starting point.
 
 ## Code Structure
 
@@ -65,6 +65,7 @@ type middleware struct {
 
 - Acronyms are all capitalized, like "RPC", "gRPC", "API". "MyID", rather than "MyId".
 - Whenever it is safe to use Go's built-in `error` instantiation functions (as opposed to Cosmos SDK's error instantiation functions), prefer `errors.New()` instead of `fmt.Errorf()` unless you're actually using the format feature with arguments.
+- As a general guideline, prefer to make the methods for a type [either all pointer methods or all value methods.](https://google.github.io/styleguide/go/decisions#receiver-type)
 
 ## Importing libraries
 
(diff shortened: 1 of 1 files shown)


docs/dev/go-style-guide.md
```diff
@@ -117,3 +118,200 @@ sdkerrors.Wrapf(
...
+    // Ommitted
+```
+
+---
+**Naked returns**
+
+Only acceptable in a small function. Once it's a medium-sized function, be explicit with your returned values. Similarly, do not name result parameters just because it enables you to use naked returns. Clarity is always more important than saving a few lines in your function.
+
+---
+**Struct defined outside of the package**
```
gjermundgaraba, line 154: I def agree with this, but is this a common issue you've seen (it very well might be, I just can't remember seeing it)?

docs/dev/go-style-guide.md
```diff
@@ -117,3 +118,200 @@ sdkerrors.Wrapf(
...
+// Good:
+good := []*Type{{ // Cuddled correctly
+    Field: "value",
+}, {
+    Field: "value",
+}}
+```
+
+---
+**Function and method calls should not be separated based solely on line length**
```
gjermundgaraba, line 198: This is probably a bit of a controversial one for some ;)
I am personally good with either option, but happy to do this if laid out clearly in the google style guide. That said, this item should probably have a little more context and link back to the source.

There is also some inconsistency with the items as some are linked back and some are not. Would be good to have consistency in context and links

docs/dev/go-style-guide.md
```diff
@@ -117,3 +118,200 @@ sdkerrors.Wrapf(
...
+
+## Good Practices
+
+**Testing context**
+
+Go 1.24 added a (testing.TB).Context() method. In tests, prefer using (testing.TB).Context() over context.Background() to provide the initial context.Context used by the test. Helper functions, environment or test double setup, and other functions called from the test function body that require a context should have one explicitly passed. [Referance](https://google.github.io/styleguide/go/decisions#contexts)
+
+## Known Anti Patterns
+
+It's strongly recommended [not to create a custom context.](https://google.github.io/styleguide/go/decisions#custom-contexts) But the cosmos sdk has it's own context that is passed around functions. And at this point it's too much of a pain to fix that.
```
gjermundgaraba, line 308:
```suggestion
It's strongly recommended [not to create a custom context.](https://google.github.io/styleguide/go/decisions#custom-contexts). The cosmos sdk has it's own context that is passed around, and we should not try to work against that pattern to avoid confusion.
```

Or something like that.

Also, feels like there should be more anti-patterns? Maybe there aren't that many?

gjermundgaraba (commented):
Nice! I like the Common Mistakes section. We should be sure to update that as we see issues!

I left some comments on a few items that we can disucss :)

I wonder if there are more linters we can use with golangci-lint that would help us conform more closely to the google styleguide (separate PR for that if it exists).

docs/dev/go-style-guide.md
```diff
@@ -117,3 +118,250 @@ sdkerrors.Wrapf(
...
+    packet channeltypes.Packet,
+    relayer sdk.AccAddress,
+) ibcexported.Acknowledgement {
+
+// good
+func (im IBCMiddleware) OnRecvPacket(ctx sdk.Context, channelVersion string, packet channeltypes.Packet, relayer sdk.AccAddress) ibcexported.Acknowledgement {
+```
+
+---
+**Don't Use Get in function/Method names**
```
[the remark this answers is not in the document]
gjermundgaraba: Yeah, we do, but we agree with the style guide that we shouldn't.
At the moment, this is likely not worth changing because it would break too many APIs, but we should move towards that.
We could make a note of this in the style guide to clarify.

gjermundgaraba (approved):

PR #8473, opened 2025-06-04
feat: channel aliasing with migration

AdityaSripal:
## Description

closes IBCGO-18, IBCGO-19, IBCGO-20

---

write a little note why.


modules/apps/transfer/keeper/msg_server.go
```diff
@@ -55,6 +55,7 @@ func (k Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types.
 
 	// if a channel exists with source channel, then use IBC V1 protocol
 	// otherwise use IBC V2 protocol
+	// TODO: Change this
 	channel, isIBCV1 := k.channelKeeper.GetChannel(ctx, msg.SourcePort, msg.SourceChannel)
```
AdityaSripal: At the moment, the Transfer endpoint can't use IBC v2 with a v1 channelID. This is easy to change but pivoting to EVM before doing this

modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -261,3 +261,24 @@ func (k *Keeper) getAllPacketStateForClient(ctx sdk.Context, clientID string, pr
 	}
 	return packets
 }
+
+// SetBaseClient sets the base client ID under the base client key for an alias
+func (k *Keeper) SetBaseClient(ctx sdk.Context, alias string, baseClientID string) {
```
gjermundgaraba: Could this just be named something like "SetClientAlias"?

modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -261,3 +261,24 @@ func (k *Keeper) getAllPacketStateForClient(ctx sdk.Context, clientID string, pr
...
+	store := k.storeService.OpenKVStore(ctx)
+	if err := store.Set(types.BaseClientKey(alias), []byte(baseClientID)); err != nil {
```
gjermundgaraba: and then "ClientAliasKey"?

modules/core/04-channel/v2/keeper/keeper.go
```diff
@@ -261,3 +261,24 @@ func (k *Keeper) getAllPacketStateForClient(ctx sdk.Context, clientID string, pr
...
+		panic(err)
+	}
+}
+
+// GetBaseClient get the base client ID under the base client key for an alias
+func (k *Keeper) GetBaseClient(ctx sdk.Context, alias string) (string, bool) {
```
gjermundgaraba: And perhaps "ResolveClientAlias"?

modules/core/04-channel/v2/keeper/keeper.go
gjermundgaraba: It took me a little time to realize what baseClient actually was, and having alias in the function name might be it clearer while reading the usage of this

modules/core/04-channel/v2/keeper/packet.go
```diff
@@ -54,21 +54,28 @@ func (k *Keeper) sendPacket(
 		return 0, "", errorsmod.Wrapf(types.ErrInvalidPacket, "constructed packet failed basic validation: %v", err)
 	}
 
+	// Before we do client keeper level checks, we first get underlying base clientID
+	clientId := sourceClient
+	baseClientId, ok := k.GetBaseClient(ctx, sourceClient)
+	if ok {
+		clientId = baseClientId
+	}
```
gjermundgaraba: Then this could be something like:
```suggestion
	if resolvedClientId, hasAlias := k.GetBaseClient(ctx, sourceClient); hasAlias {
		clientId = resolvedClientId
	}
```

modules/core/04-channel/v2/keeper/packet.go
```diff
@@ -54,21 +54,28 @@ func (k *Keeper) sendPacket(
 		return 0, "", errorsmod.Wrapf(types.ErrInvalidPacket, "constructed packet failed basic validation: %v", err)
 	}
 
+	// Before we do client keeper level checks, we first get underlying base clientID
```
gjermundgaraba, line 59: What about the check on line 28?

Is there a reason why we wouldn't want the sourceClient to already be resolved before it gets sent to this function?

gjermundgaraba (commented):
Nice work 🚀 The approach generally looks good. It's mostly very straight-forward and easy to reason about, which I like! 
I've left some comments, and depending on the answer, I might need to dive deeper to get a better sense of the flow.

modules/core/04-channel/v2/keeper/keeper.go
AdityaSripal: Yes, though the key here is the alias, and the value is the underlying clientID. So i feel it would be confusing to name it setting alias since that is not what the value is. But open to changing

modules/core/04-channel/v2/keeper/packet.go
AdityaSripal: The counterparties in client keeper are stored on the channel IDs

gjermundgaraba: Oh right. Yeah that makes sense. 
I think it would be good to add a comment on that, just so it's easy to remember in the future when coming back here :)

modules/core/04-channel/v2/keeper/keeper.go
gjermundgaraba: I get that, and this follows the normal pattern we use for setting key value in the store, but base is so ambiguous that it doesn't read too well.
I would like either to use a more descriptive name like `MapAliasToClient`, `RegisterAlias` or if you prefer a `Set...` method it could be something like `SetClientForAlias` or `SetCanonicalClientForAlias`. 
Essentially, because there are no _very_ good names to describe the counterparty to an alias, I think we need a more descriptive name that includes alias in it somehow.

modules/core/04-channel/migrations/v11/store.go
```diff
@@ -0,0 +1,70 @@
...
+
+	ibcKeeper.ChannelKeeper.IterateChannels(ctx, func(ic types.IdentifiedChannel) (stop bool) {
+		// only add counterparty for channels that are OPEN and UNORDERED
+		// set a base client mapping from the channelId to the underlying base client
+		counterparty, ok := ibcKeeper.ChannelKeeper.GetV2Counterparty(ctx, ic.PortId, ic.ChannelId)
+		if ok {
+			ibcKeeper.ClientV2Keeper.SetClientCounterparty(ctx, ic.ChannelId, counterparty)
+			connection, ok := ibcKeeper.ConnectionKeeper.GetConnection(ctx, ic.ConnectionHops[0])
+			if !ok {
+				panic("connection not set")
```
gjermundgaraba, line 46: There are some uncovered cases here. Would it make sense to make sure this behaves as expected in the unit tests too?

modules/apps/transfer/v2/alias_test.go
```diff
@@ -0,0 +1,293 @@
...
+func (suite *TransferTestSuite) TestAliasedTransferChannel() {
+	path := ibctesting.NewTransferPath(suite.chainA, suite.chainB)
+	path.Setup()
+
+	// mock v1 format for both sides of the channel
+	mockV1Format(path.EndpointA)
+	mockV1Format(path.EndpointB)
+
+	// migrate the store for both chains
+	v11.MigrateStore(suite.chainA.GetContext(), runtime.NewKVStoreService(suite.chainA.GetSimApp().GetKey(ibcexported.StoreKey)), suite.chainA.App.AppCodec(), suite.chainA.App.GetIBCKeeper())
```
gjermundgaraba: this returns an error at the moment, I believe?

modules/core/04-channel/migrations/v11/store_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (suite *MigrationsV11TestSuite) TestMigrateStore() {
+	ctx := suite.chainA.GetContext()
+	cdc := suite.chainA.App.AppCodec()
+	ibcKeeper := suite.chainA.App.GetIBCKeeper()
+	storeService := runtime.NewKVStoreService(suite.chainA.GetSimApp().GetKey(ibcexported.StoreKey))
+	store := storeService.OpenKVStore(ctx)
+	numberOfChannels := 100
+
+	for i := range numberOfChannels {
```
gjermundgaraba, line 49: Do we have any channels in this test that are initializing, but not open yet?

modules/core/04-channel/v2/keeper/packet.go
```diff
@@ -298,11 +326,17 @@ func (k *Keeper) timeoutPacket(
 		return errorsmod.Wrapf(clientv2types.ErrInvalidCounterparty, "counterparty id (%s) does not match packet destination id (%s)", counterparty.ClientId, packet.DestinationClient)
 	}
 
+	// Before we do client keeper level checks, we first get underlying base clientID
+	clientId := packet.SourceClient
+	if underlyingClientId, isAlias := k.GetClientForAlias(ctx, packet.SourceClient); isAlias {
+		clientId = underlyingClientId
```
gjermundgaraba, line 332: Is codecov correct that this line is not hit in the tests? 🤔

gjermundgaraba (commented):
Left a few comments.
One thing I'm not seeing here is creating v2 channels when new v1 channels are opened.

modules/core/04-channel/migrations/v11/store.go
AdityaSripal: Hmmm... these should never happen because a channel that exists in state should ALWAYS have an underlying connection.

I could add a test case just for the sake of codecov but i could also just ignore the returned bool since it effectively has to be there or the chain state is deeply corrupted

gjermundgaraba: Yeah, no that's right. No need to test for that :) Just take a look at the other ones and make sure code paths we expect to run in real life are covered :)

modules/core/04-channel/keeper/handshake.go
```diff
@@ -308,6 +324,22 @@ func (k *Keeper) WriteOpenConfirmChannel(
 	k.SetChannel(ctx, portID, channelID, channel)
 	k.Logger(ctx).Info("channel state updated", "port-id", portID, "channel-id", channelID, "previous-state", types.TRYOPEN, "new-state", types.OPEN)
 
+	if channel.Ordering == types.UNORDERED {
```
gjermundgaraba, line 327: Is this tested?

modules/core/04-channel/migrations/v11/store.go
```diff
@@ -0,0 +1,70 @@
...
+const (
+	KeyNextSeqSendPrefix = "nextSequenceSend"
+	KeyChannelEndPrefix  = "channelEnds"
+	KeyChannelPrefix     = "channels"
+	KeyPortPrefix        = "ports"
+)
+
+// NextSequenceSendV1Key returns the store key for the send sequence of a particular
+// channel binded to a specific port.
+func NextSequenceSendKey(portID, channelID string) []byte {
```
gjermundgaraba: Should this be named like the comment suggests?
```suggestion
func NextSequenceSendV1Key(portID, channelID string) []byte {
```

modules/core/04-channel/migrations/v11/store.go
```diff
@@ -0,0 +1,70 @@
...
+	return fmt.Appendf(nil, "%s/%s", KeyNextSeqSendPrefix, channelPath(portID, channelID))
+}
+
+func channelPath(portID, channelID string) string {
```
gjermundgaraba: This is not the first time we've duplicated this function. Maybe we just make it public in 24-host/channel_keys.go instead? I don't care deeply one way or another, but just feels a bit silly to keep copying this around ^^

modules/core/04-channel/keeper/handshake.go
AdityaSripal: Yes it gets tested here: https://github.com/cosmos/ibc-go/pull/8473/files#diff-6986e434ce925556327ce7c25b8457ab0c4db4b50d49a53328c4cbf17be30c1fR685

Where i setup a new v1 channel after the migration. This is mocked as just opening a new channel with the current code. The aliasing still works because the counterparties are set on the handshake OPEN calls.

It is also covered in the codecov report from what i can see, and verified it hits this case with the test i pasted above with a print statement

gjermundgaraba: Perfect, I realized eventually that it probably was covered, but left the comment to make sure :) Thanks!

gjermundgaraba (approved):

PR #8493, opened 2025-06-06
chore: remove all internal uses of legacy params

johnnylarner:
## Description

Removed all legacy param references. Legacy files contain key values as bytes that were referenced. I moved these into the params file.

These changes only affect how we internally build our simapps. The `NewKeeper` remains unchanged and therefore I do not see the need to include this in the CHANGELOG. We may want to classify as a chore in that case.

closes: #XXXX

---

write a little note why.


gjermundgaraba:
I noticed that the genesis e2e test was failing, and when I started looking a bit deeper, I realized we could actually safely remove x/params entirely from our simapps.
The usage of the subspaces in the SDK modules were only for migration purposes, and we don't need that for our simapps.

The only place that I've left any usage of params left is in our e2e tests, and that is only because it is needed for compatibility tests (where we spin up older versions). We will eventually be able to get rid of those too, but we need to at a minimum get rid of ibc-go v7 first (which might take some time still).

gjermundgaraba (approved):
Great work! 🙌 This cleans up the codebase in a very real way 🎉

I left a comment on some updates I did to the PR, sorry for taking over a bit here at the end, I just wanted to get this merged in to test with some other changes :)

Issue #3889, opened 2023-06-18
Remove `MsgSubmitMisbehaviour` and corresponding rpc handler/CLI


crodriguezvega:
With 02-client refactor [`MsgSubmitMisbehaviour`](https://github.com/cosmos/ibc-go/blob/cf519bc4d4285aa0b69511ba08d69b7456744f5d/proto/ibc/core/client/v1/tx.proto#L99) was deprecated.

- [ ] Remove `MsgSubmitMisbehaviour`.
- [ ] Remove [`SubmitMisbehaviour` rpc handler](https://github.com/cosmos/ibc-go/blob/cf519bc4d4285aa0b69511ba08d69b7456744f5d/modules/core/keeper/msg_server.go#L86).
- [ ] Remove [CLI command](https://github.com/cosmos/ibc-go/blob/cf519bc4d4285aa0b69511ba08d69b7456744f5d/modules/core/02-client/client/cli/tx.go#L134).

Adding `needs_discussion` label to discuss in what release we should remove this.

PR #8516, opened 2025-06-10
chore: rm misbehaviour endpoint

johnnylarner:
## Description
Will add to the CHANGELOG and double check the test code again tomorrow.

closes: #3889

---

write a little note why.


modules/core/ante/ante_test.go
```diff
@@ -24,6 +31,64 @@ import (
 	"github.com/cosmos/ibc-go/v10/testing/mock/v2"
 )
 
+const (
+	invalidHashValue = "invalid_hash"
+)
+
+// createMaliciousTMHeader creates a header with the provided trusted height with an invalid app hash.
+func createMaliciousTMHeader(chainID string, blockHeight int64, trustedHeight clienttypes.Height, timestamp time.Time, tmValSet, tmTrustedVals *cmttypes.ValidatorSet, signers []cmttypes.PrivValidator, oldHeader *cmtproto.Header) (*ibctm.Header, error) {
```
gjermundgaraba: super nit, but would prefer this method to below all the tests since it's a helper function. Same with `createMaliciousUpdateClientMessage`

modules/core/ante/ante_test.go
```diff
@@ -504,6 +591,19 @@ func (suite *AnteTestSuite) TestAnteDecoratorCheckTx() {
 			},
 			channeltypes.ErrRedundantTx,
 		},
+		{
```
gjermundgaraba, line 537: Would want @AdityaSripal to confirm this

gjermundgaraba (commented):
I just did an initial review, but it looks very good 👍

modules/core/ante/ante_test.go
johnnylarner: Agree with the header func but `createMaliciousUpdateClientMessage ` one is located with all the other `-> sdk.Msg` functions which is quite coherent imo

modules/core/ante/ante_test.go
johnnylarner: Yep, wasn't clear to me why the the older `Submit...` behaviour affects message redundancy, but judging by the rest of the tests the update client message should behave as this test does

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### API Breaking
 * (apps) [\#8476](https://github.com/cosmos/ibc-go/pull/8476) Remove `ParamSubspace` from all `Keeper` constructors
 * (light-clients/08-wasm) [\#8511](https://github.com/cosmos/ibc-go/pull/8511) Remove deprecated `Checksums` type
+* (core/02-client) [\#8516](https://github.com/cosmos/ibc-go/pull/8516) Remove deprecated `SubmitMisbehaviour` RPC endpoint and types
```
gjermundgaraba:
```suggestion
* (core/02-client) [\#8516](https://github.com/cosmos/ibc-go/pull/8516) Remove deprecated `SubmitMisbehaviour` message handler
```

I think RPC endpoint is not precise in this case. It's sort of that too, but the core of it is a message handler.

modules/core/ante/ante_test.go
```diff
@@ -2,6 +2,7 @@ package ante_test
 
 import (
 	"errors"
+	"fmt"
```
gjermundgaraba: lint errors, I guess?

gjermundgaraba (approved):
Looks mostly good to go. Left a couple of minor comments, and pending answer on the question from Aditya :)

modules/core/ante/ante_test.go
johnnylarner: mb forgot to lint!

modules/core/ante/ante_test.go
AdityaSripal: Hmmm... this just seems like an oversight that happened once we merged the misbehaviour and recvPacket messages into a single endpoint.

Generally speaking, a relayer wouldn't pair a malicious update message with recvpacket messages together, since that would just freeze the client that you wanted to use later.

It could be the case that someone submits an update message that ends up being proof of misbehaviour (i.e. the update client is just a single header but it conflicts with a consensus state that already exists). If this update message was paired with only redundant receive packet messages it would also be ignored.

Perhaps we should check the update message first before rejecting the redundant messages in the antehandler

AdityaSripal: In practice, this is not a huge deal though

AdityaSripal (approved):
LGTM! Yes, the redundant tx antedecorator can be improved to allow misbehavior processing even if the attached packet messages are redundant

But in practice, this is a relatively small concern

modules/core/ante/ante_test.go
gjermundgaraba: What is the most important scenario we should actually test for here in the antehandler?

gjermundgaraba: Or, to put it another way: is this test covering the most important scenario wrt misbehavior? <br><br>Also: I don't really think it makes sense to optimize for a scenario where someone posts misbehavior and recvPacket. I just want to make sure we have the right test case here

AdityaSripal: The most important scenario:

There exists a consensus state in client with height 50 and hash: 0xABC

Relayer submits a new tx with:

- Valid Header Update for height 50 and hash: 0xDEF
- Redundant RecvPacket txs

Note: If the header update was submitted just on its own; IBC would automatically detect the misbehaviour even though it is a regular update because it conflicts with existing consensus state.

My suspicion however, is that this tx would be dropped by the antehandler

Issue #8492, opened 2025-06-06
Support Aliasing V2 Packets with MsgTransfer


AdityaSripal:
Aliasing was added as a core IBC feature in #8473 

However, MsgTransfer still switches on protocol version by checking if a channel exists on the channelID in the msg

This will always then send a v1 packet on a channelID if you use MsgTransfer

We should add some boolean flag that tells `MsgTransfer` to send a v2 packet even if the channelID maps to a v1 channel to support aliasing through the `MsgTransfer` entrypoint

PR #8518, opened 2025-06-10
MsgTransfer supports aliasing

AdityaSripal:
## Description

closes: #8492 

---

write a little note why.


proto/ibc/applications/transfer/v1/tx.proto
```diff
@@ -53,6 +53,11 @@ message MsgTransfer {
   string memo = 8;
   // optional encoding
   string encoding = 9;
+  // boolean flag to indicate if the transfe message
+  // is sent with the IBC v2 protocol.
+  // This only needs to be set if the channel IDs
+  // are V1 channel identifiers.
+  bool is_v2 = 10 [(amino.dont_omitempty) = true];
```
AdityaSripal: This field is ignored if the channel doesn't exist

modules/apps/transfer/types/msgs.go
```diff
@@ -61,6 +61,28 @@ func NewMsgTransfer(
 	}
 }
 
+// NewMsgTransferAliased creates a new MsgTransfer instance
+// with isV2 set to true, indicating that it is using the V2 protocol
+// with v1 channel identifiers.
+func NewMsgTransferAliased(
```
AdityaSripal, line 68: I can add some more validation in `ValidateBasic` to check channel identifiers based on `isV2` boolean flag

proto/ibc/applications/transfer/v1/tx.proto
```diff
@@ -53,6 +53,11 @@ message MsgTransfer {
   string memo = 8;
   // optional encoding
   string encoding = 9;
+  // boolean flag to indicate if the transfe message
```
gjermundgaraba:
```suggestion
  // boolean flag to indicate if the transfer message
```

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -53,21 +53,21 @@ func (k Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types.
 		return nil, errorsmod.Wrapf(err, "failed to validate %s packet data", types.V1)
 	}
 
-	// if a channel exists with source channel, then use IBC V1 protocol
-	// otherwise use IBC V2 protocol
-	// TODO: Change this
-	channel, isIBCV1 := k.channelKeeper.GetChannel(ctx, msg.SourcePort, msg.SourceChannel)
+	// if the channel does not exist, or we are using channel aliasing then use IBC V2 protocol
```
gjermundgaraba, line 56: Should this comment be something like the below?
```suggestion
	// if the channel does not exist, or if `IsV2` is set (in which case the source channel is either a client id or a channel alias) then IBC V2 protocol
```

proto/ibc/applications/transfer/v1/tx.proto
gjermundgaraba: If we add a transfer v2, this will be hella confusing :D Might wanna rename for clarity

gjermundgaraba: Also, I feel like we need to document both source channel and this flag in more detail to make it clear how to send v1 and v2 messages, and more details on how that is chosen, so nobody gets surprised.

modules/apps/transfer/types/msgs.go
gjermundgaraba: Might be a good idea, since we can rely on the format in ibc-go.

modules/apps/transfer/keeper/msg_server.go
AdityaSripal: No because if the sourceChannel is a clientID then we are necessarily using V2. I only require the boolean to be set to true if we are using channelIDs **AND** we want to use V2.

Maybe i should change the name of the field to make this more clear

gjermundgaraba: Right, but would GetChannel return anything if the sourceChannel is clientID?
At least it's not clear to me that the comment is saying the same thing that the code is doing 🤔

faddat:
For some reason I don't like this concept.  I get where it's heading but wonder if there could be a better way?

gjermundgaraba:
> For some reason I don't like this concept. I get where it's heading but wonder if there could be a better way?

If you have a better suggestion, I am all ears :)

proto/ibc/applications/transfer/v1/tx.proto
gjermundgaraba: Maybe renaming to something like `use_channel_aliasing` might make sense?

faddat:
I don't have a better suggestion yet, but I guess I should voice my concern clearly. The whole concept of aliasing as regards IBC worries me because if you can have one alias potentially you can have many and of course we would try and prevent that but it feels not quite right. 

What is motivating this feature?

gjermundgaraba:
Migration paths for existing ibc v1 channels to be able to work using ibc v2 (which doesn't have channels and normally use client identifiers instead), and keep fungibility with existing tokens.

modules/apps/transfer/keeper/msg_server.go
gjermundgaraba: Nice, this looks much easier to grasp 🙏

faddat:
Yeah I understand what you mean, but I am definitely still uncomfortable
with the concept of aliases for this purpose.

Issue #5814, opened 2024-02-05
Middleware Wiring Improvements


srdtrk:
## Summary

The proposed changes aim to simplify the middleware wiring process significantly, making it more intuitive and less error-prone, thereby improving the overall reliability and security of IBC middleware integration. The proposal recommends introducing a new helper called `IBCStackBuilder`.

## Problem

The current mechanism for integrating IBC middleware is complex and prone to errors, leading to difficulties in understanding the middleware stack, which subsequently results in bugs and security vulnerabilities. This complexity stems from the need to intertwine two distinct stacks: the IBC module stack and the ICS4Wrapper stack. The latter is often managed by a module's keeper rather than the module itself, complicating the integration process. Furthermore, the necessity to wire these stacks in reverse order complicates their integration during module initialization.

## Proposal

This proposal suggests a syntactic enhancement aimed at simplifying the wiring process of IBC middlewares without altering their underlying architecture. It introduces a new construct, the `IBCStackBuilder`, designed to streamline the middleware integration process. The proposed API enables a straightforward layering of middlewares, culminating in the base application, as demonstrated below:

```go
transferStack := porttypes.NewIBCStackBuilder().Next(feeMiddleware).Next(callbacksMiddleware).Base(transferApp).Build()
```
To facilitate this approach, several API modifications are proposed:

1. Removing the obligation for `IBCModule` and `IBCMiddleware` interfaces to accept an `ICS4Wrapper` during initialization, which eliminates circular dependencies. Instead, a `SetICS4Wrapper` method will be added to the `IBCModule` interface, allowing for the post-initialization setting of the middleware. (Alternatively, this could be an optional interface that the module can implement to be middleware compatible.)

```diff
type IBCModule interface {
	// ... other methods
+
+	// SetICS4Wrapper sets the ICS4Wrapper. This function may be used after
+	// the module's initialization to set the middleware which is above this
+	// module in the IBC application stack.
+	SetICS4Wrapper(wrapper ICS4Wrapper)
```

This will allow the `IBCStackBuilder` to wire the ICS4Wrapper after the IBC module has been initialized.
 
2. Introduce a new `SetUnderlyingModule` method to the `IBCMiddleware` interface. This allows for the setting of the underlying application post-initialization, thus supporting middleware initializayion without the underlying application.

```diff
type Middleware interface {
	IBCModule
	ICS4Wrapper
+
+	// SetUnderlyingModule sets the underlying IBC module. This function may be used after
[...]

PR #8528, opened 2025-06-13
Middleware Wiring Improvements

AdityaSripal:
## Description

closes: #5814 

---

write a little note why.

### DIFF
--- a/modules/apps/packet-forward-middleware/keeper/keeper_test.go
+++ b/modules/apps/packet-forward-middleware/keeper/keeper_test.go
@@ -198,7 +198,7 @@ func (s *KeeperTestSuite) TestForwardTransferPacket() {
 	initialSender := s.chainA.SenderAccount.GetAddress()
 	finalReceiver := s.chainB.SenderAccount.GetAddress()
 
-	err := pfmKeeper.ForwardTransferPacket(ctx, nil, srcPacket, initialSender.String(), finalReceiver.String(), metadata, sdk.NewInt64Coin("denom", 1000), 2, time.Duration(timeout), nil, nonRefundable)
+	err := pfmKeeper.ForwardTransferPacket(ctx, nil, srcPacket, initialSender.String(), finalReceiver.String(), metadata, sdk.NewInt64Coin("denom", 1000), 2, timeout, nil, nonRefundable)
 	s.Require().NoError(err)
 
 	// Get the inflight packer
@@ -208,7 +208,7 @@ func (s *KeeperTestSuite) TestForwardTransferPacket() {
 	s.Require().Equal(inflightPacket.RetriesRemaining, int32(retries))
 
 	// Call the same function again with inflight packet. Num retries should decrease.
-	err = pfmKeeper.ForwardTransferPacket(ctx, inflightPacket, srcPacket, initialSender.String(), finalReceiver.String(), metadata, sdk.NewInt64Coin("denom", 1000), 2, time.Duration(timeout), nil, nonRefundable)
+	err = pfmKeeper.ForwardTransferPacket(ctx, inflightPacket, srcPacket, initialSender.String(), finalReceiver.String(), metadata, sdk.NewInt64Coin("denom", 1000), 2, timeout, nil, nonRefundable)
 	s.Require().NoError(err)
 
 	// Get the inflight packer


modules/apps/27-interchain-accounts/controller/ibc_middleware.go
```diff
@@ -26,27 +26,37 @@ var (
 // ICA controller keeper and the underlying application.
 type IBCMiddleware struct {
 	app    porttypes.IBCModule
-	keeper keeper.Keeper
+	keeper *keeper.Keeper
```
AdityaSripal, line 29: I made the keepers inside middlewares and modules to now be pointers so that I can call `WithICS4Wrapper` on them and have it persist in other references

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
```diff
@@ -41,17 +41,17 @@ type Keeper struct {
...
 	if strings.TrimSpace(authority) == "" {
 		panic(errors.New("authority must be non-empty"))
 	}
 
-	return Keeper{
+	return &Keeper{
 		storeService:  storeService,
 		cdc:           cdc,
-		ics4Wrapper:   ics4Wrapper,
+		ics4Wrapper:   channelKeeper,
```
AdityaSripal, line 56: `NewKeeper` defaults to using `channelKeeper` as ICS4Wrapper. If you need this changed, you should use the stack builder primitive to set a new ICS4Wrapper

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
gjermundgaraba: Is this a requirement for middlewares wiring up with this?

modules/apps/callbacks/testing/simapp/app.go
```diff
@@ -389,41 +386,27 @@ func NewSimApp(
...
-	// Since the callbacks middleware itself is an ics4wrapper, it needs to be passed to the transfer keeper
-	app.TransferKeeper.WithICS4Wrapper(transferICS4Wrapper)
+	transferStack := porttypes.NewIBCStackBuilder()
+	transferStack.Base(transfer.NewIBCModule(app.TransferKeeper)).Next(
+		ibccallbacks.NewIBCMiddleware(app.MockContractKeeper, maxCallbackGas),
+	)
 
 	// Add transfer stack to IBC Router
-	ibcRouter.AddRoute(ibctransfertypes.ModuleName, transferStack)
+	ibcRouter.AddRoute(ibctransfertypes.ModuleName, transferStack.Build(app.IBCKeeper.ChannelKeeper))
```
gjermundgaraba: Would it make more sense for the channel keeper to just be set on construction instead of being passed into the build method?

modules/core/05-port/types/module.go
```diff
@@ -104,6 +104,15 @@ type IBCModule interface {
...
 		relayer sdk.AccAddress,
 	) error
+
+	// SetICS4Wrapper sets the ICS4Wrapper. This function may be used after
+	// the module's initialization to set the middleware which is above this
+	// module in the IBC application stack.
+	// The ICS4Wrapper **must** be used for sending packets and writing acknowledgements
+	// to ensure that the middleware can intercept and process these calls.
+	// Do not use the channel keeper directly to send packets or write acknowledgements
+	// as this will bypass the middleware.
```
gjermundgaraba, line 114: Just so I understand: this has always been a requirement of middlewares (to not use the channel keeper directly)?

gjermundgaraba (commented):
Did a first pass, left a couple of questions, but this is clearly a big improvement, and it should be much harder to do this incorrectly. It still will require documentation and migration guides. At least the migration guide additions would be good to get sorted in this PR (for both chains and middleware devs).

modules/core/05-port/types/module.go
AdityaSripal: That is correct, though many people don't do it anyway. Which ends up being functionally ok because most of the middlewares are modifying recvpacket and not send and writeack

modules/apps/27-interchain-accounts/controller/ibc_middleware.go
AdityaSripal: If you are using a keeper function to do SendPacket and WriteAck then yes. We have middlewares that don't use keepers (e.g. callbacks) so then its not needed

gjermundgaraba: OK, sounds like it will be quite important to get some migration docs for all of this then!

modules/apps/27-interchain-accounts/controller/keeper/keeper.go
gjermundgaraba: Make sure to document this in the migration guide!

modules/core/05-port/types/module.go
gjermundgaraba: Not sure where we are on docs with this, but we should:
1. Make sure the docs are clear about this
2. Link to that in the migration guide

modules/apps/rate-limiting/ibc_middleware.go
```diff
@@ -1,70 +1,69 @@
 package ratelimiting
```
gjermundgaraba, line 1: @AdityaSripal I merged in rate-limiting, so make sure it looks correct, please :)

e2e/go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/cosmos/ibc-go/e2e
 
 go 1.24.3
 
+// TODO: Remove when v11 release of interchaintest is available (that is where this one is coming from)
+replace github.com/cosmos/interchain-security/v7 => github.com/cosmos/interchain-security/v7 v7.0.0-20250622154438-73c73cf686e5
```
gjermundgaraba, line 6: Had to add this, because interchaintest pulls in this, and the changes here break it :/
The PR that has the commit referenced in the replace: https://github.com/cosmos/interchain-security/pull/2622

modules/core/05-port/types/module.go
```diff
@@ -104,6 +104,15 @@ type IBCModule interface {
...
+	SetICS4Wrapper(wrapper ICS4Wrapper)
```
gjermundgaraba, line 115: A couple of things that might cause confusion:
1. We have a mix of "SetICS4Wrapper" and "WithICS4Wrapper"
2. We have these both in keepers and middleware (which I understand), but it makes for confusing interfaces

Can I suggest thinking about renaming so that perhaps the middleware SetICS4Wrapper is named distinctly from the keepers version of it? Ideally with something in the name that makes it clearer what it's for (and that it is middleware specific). There are also keepers that accidentally fulfill this interface, I don't think they should.

modules/core/05-port/types/stack.go
```diff
@@ -0,0 +1,51 @@
+package types
+
+type IBCStackBuilder struct {
```
gjermundgaraba, line 3: There is an opportunity to give this extra value by adding a query to ibc core (or somewhere where it makes sense) for users to see the stack. Skip go currently keeps a manual repository of which chains support PFM, for instance. If they could just query, it would help them automate this in the future.
Additionally, chain devs could debug their stack more easily.

It could either just take list here and add another function to the Middleware interface to return a name (types.ModuleName type of response), or maybe even have it come out of ICS4Wrapper so it actually flows through the application stack.
wdyt?

modules/apps/callbacks/testing/simapp/app.go
gjermundgaraba: Not super important though...

gjermundgaraba (commented):
This looks great! I left some more comments, but the main thing we need to get in here is docs on the wiring and migration docs for both chains and middleware devs.

FYI: I merged in main and linted, so you should probably take a look through all the code another time yourself too.
You should in particular check out the rate-limiting middleware as I integrated it myself, and made similar changes that you've made to the other middlewares.

modules/core/05-port/types/stack.go
AdityaSripal: This should be a seperate issue and PR if desired

modules/core/05-port/types/module.go
AdityaSripal: That has been the general pattern.

middleware has `SetICS4Wrapper`

and keepers have:

`WithICS4Wrapper`. Is there a place where this pattern is broken or do you take issue with the pattern itself?

modules/core/05-port/types/stack.go
gjermundgaraba: Fair! I'll create an issue for that 👍

docs/docs/05-migrations/15-support-stackbuilder.md
```diff
@@ -0,0 +1,150 @@
...
+
+Refer to the [integration guide](../01-ibc/04-middleware/03-integration.md) to understand how to use this new middleware to improve middleware wiring in the chain application setup.
+
+# Migrations for Application Developers
+
+In order to be wired with the new StackBuilder primitive, applications and middlewares must implement new methods as part of their respective interfaces.
+
+IBC Applications must implement a new `SetICS4Wrapper` which will set the `ICS4Wrapper` through which the application will call `SendPacket` and `WriteAcknowledgement`. It is recommended that IBC applications are initialized first with the IBC ChannelKeeper directly, and then modified with a middleware ICS4Wrapper during the stack wiring. 
+```go
+```
```
gjermundgaraba: ?
```suggestion
```

docs/docs/05-migrations/15-support-stackbuilder.md
```diff
@@ -0,0 +1,150 @@
...
+// The ICS4Wrapper **must** be used for sending packets and writing acknowledgements
+// to ensure that the middleware can intercept and process these calls.
+// Do not use the channel keeper directly to send packets or write acknowledgements
+// as this will bypass the middleware.
+SetICS4Wrapper(wrapper ICS4Wrapper)
+```
+
+Many applications have a stateful keeper that executes the logic for sending packets and writing acknowledgements. In this case, the keeper in the application must be a **pointer** reference so that it can be modified in place after initialization.
+
+The initialization should be modified to no longer take in an addition `ics4Wrapper` as this gets modified later by `SetICS4Wrapper`. The constructor function must also return a **pointer** reference so that it may be modified in-place by the stack builder.
```
gjermundgaraba, line 33: Strictly speaking, it doesn't need to actually return a pointer, but it is important that it uses a pointer to the keeper.

docs/docs/05-migrations/15-support-stackbuilder.md
```diff
@@ -0,0 +1,150 @@
...
+
+// Keeper defines the IBC fungible transfer keeper
+type Keeper struct {
+	...
+	ics4Wrapper   porttypes.ICS4Wrapper
+
+    // Keeper is initialized with ICS4Wrapper
+    // being equal to the top-level channelKeeper
+    // this can be changed by calling WithICS4Wrapper
+    // with a different middleware ICS5Wrapper
```
gjermundgaraba: ICS5Wrapper?

gjermundgaraba (commented):
Left a couple of small comments and fixes on doc stuff, but I think we're very close!! 🙌

docs/docs/01-ibc/04-middleware/03-integration.md
```diff
@@ -55,13 +58,26 @@ customIBCModule1 := custom.NewIBCModule(customKeeper1, "portCustom1")
 customIBCModule2 := custom.NewIBCModule(customKeeper2, "portCustom2")
 
 // create IBC stacks by combining middleware with base application
+// IBC Stack builders are initialized with the IBC ChannelKeeper which is the top-level ICS4Wrapper
 // NOTE: since middleware2 is stateless it does not require a Keeper
 // stack 1 contains mw1 -> mw3 -> transfer
```
womensrights, line 63: from looking at this, although you specify the order I still don't think it is easy to understand conceptually and why one middleware should be before another in the ordering. Just having more examples doesn't really explain how you reason about the order of the stack. I think it is fine to have the mock examples with mw1 etc, but I think it would be useful to give more detail on how to reason about the underlying order. Concretely we have 3 middlewares in the repo and the assumption is users should expect to have them all for transfer so I think we should concretely relate to this.

gjermundgaraba: That is a good point. The code itself helps reason about what order you actually end up getting, but not necessarily which order they _ought_ to be in.

It's hard to get _too_ concrete on this, since it depends on which middlewares you use, but we can certainly use the ones in the repo as an example. We should def do that.

I think we could add a section on middleware ordering where we go through how the ordering works (i.e. how the ordering relates to the actual flow of a packet on send and receive) and use our own middlewares as an example to show how we decided on the order of those.

(@womensrights: Separately from this, the docs for each middleware should probably also make it clear if they have any place in the stack they _should_ be - this is particularly relevant for rate limiting!)

womensrights: yes agreed, like there are likely some heuristics that could help you reason through ordering i.e. should this middleware act like a gate for any further processing (e.g. rate limiting) or should this middleware always be last to execute... just from the top of my head

and yes I think we should just be very concrete with what we have under our maintenance so reasoning should be really clear for pfm, rate limiting and callbacks in a stack

womensrights: also thinking about this more, it seems we are going to have multiple ways to wire middleware if you use the stack builder or not, I think this is going to be messy and confusing and there should only be one recommended way

womensrights: so perhaps this is just a docs thing as people could still choose to use the old way, but I think we only show the new way from v11 docs onwards

AdityaSripal: I don't have good context anymore on exactly what order we should be prescribing. Especially in the context of the new middlewares in the repo.

We only are showing the stack builder approach in the docs from now on and only recommending this approach. Though i imagine some people will still use the old way.

I tried addressing this, but feel im missing the information i need at this point in terms of exactly what you want here. Since its not really in scope for the stack builder approach, Id prefer to merge this and have this comment be a general improvement of the middleware docs that we should tackle separately

docs/docs/05-migrations/15-support-stackbuilder.md
AdityaSripal: It does I believe, since we are modulating the underlying application/ics4wrapper of the middleware itself during stack.Build()

docs/docs/01-ibc/04-middleware/03-integration.md
gjermundgaraba: I'll merge this now, but sounds like there are some doc improvements needed @womensrights

gjermundgaraba (approved):

PR #8545, opened 2025-06-18
Multi payload packet

AdityaSripal:
## Description

closes: IBCGO-2371

---

write a little note why.

### DIFF
--- a/modules/core/04-channel/v2/keeper/packet.go
+++ b/modules/core/04-channel/v2/keeper/packet.go
@@ -175,7 +175,7 @@ func (k *Keeper) recvPacket(
 
 // writeAcknowledgement writes the acknowledgement to the store and emits the packet and acknowledgement
 // for relayers to relay the acknowledgement to the counterparty chain.
-func (k Keeper) writeAcknowledgement(
+func (k *Keeper) writeAcknowledgement(
 	ctx sdk.Context,
 	packet types.Packet,
 	ack types.Acknowledgement,
--- a/modules/core/04-channel/v2/keeper/packet_test.go
+++ b/modules/core/04-channel/v2/keeper/packet_test.go
@@ -22,7 +22,7 @@ import (
 
 var unusedChannel = "channel-5"
 
-func (suite *KeeperTestSuite) TestSendPacket() {
+func (s *KeeperTestSuite) TestSendPacket() {
 	var (
 		path        *ibctesting.Path
 		packet      types.Packet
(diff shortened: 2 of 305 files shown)


gjermundgaraba:
bugbot run

modules/core/04-channel/v2/keeper/packet.go
```diff
@@ -180,6 +180,19 @@ func (k *Keeper) writeAcknowledgement(
...
 	ack types.Acknowledgement,
 ) error {
+	// Validate the acknowledgement
+	if err := ack.Validate(); err != nil {
+		ctx.Logger().Error("write acknowledgement failed", "error", errorsmod.Wrap(err, "invalid acknowledgement"))
+		return errorsmod.Wrap(err, "invalid acknowledgement")
+	}
+
+	// Validate the acknowledgement against the payload length
+	if ack.Success() {
```
gjermundgaraba, line 190: nit: could be a single if statement

modules/core/04-channel/v2/keeper/packet_test.go
```diff
@@ -26,6 +26,7 @@ func (s *KeeperTestSuite) TestSendPacket() {
 	var (
 		path        *ibctesting.Path
 		packet      types.Packet
+		payload     types.Payload
```
gjermundgaraba: I don't think this does anything. There is already a payload in the test function itself that is set with `payload := ...`.

gjermundgaraba: If it's never going to change, but you need access to it, you could just set it once at the top and use it as you do already. But right now I think it will always be empty.

modules/core/04-channel/v2/keeper/packet_test.go
```diff
@@ -244,8 +259,9 @@ func (s *KeeperTestSuite) TestRecvPacket() {
 
 func (s *KeeperTestSuite) TestWriteAcknowledgement() {
 	var (
-		packet types.Packet
-		ack    types.Acknowledgement
+		packet  types.Packet
```
gjermundgaraba, line 261: In this one, it does get set in the test function itself, but as far as I can tell, it never changes, so you could just set it once here and use it directly and not reset it every test

modules/core/04-channel/v2/keeper/packet_test.go
```diff
@@ -457,6 +537,7 @@ func (s *KeeperTestSuite) TestTimeoutPacket() {
 	var (
 		path         *ibctesting.Path
 		packet       types.Packet
+		payload      types.Payload
```
gjermundgaraba: same deal here, I believe

modules/core/04-channel/v2/types/msgs_test.go
```diff
@@ -37,6 +37,7 @@ func TestTypesTestSuite(t *testing.T) {
 
 func (s *TypesTestSuite) TestMsgSendPacketValidateBasic() {
 	var msg *types.MsgSendPacket
+	var payload types.Payload
```
gjermundgaraba, line 40: This one doesn't look like it's used in the test function itself, so probably don't need to define it here?

gjermundgaraba (approved):
This looks very good! I left a couple of nits and minor cleanups with the tests. Feel free to merge if you agree with these and get the fixed 👍

modules/core/04-channel/v2/keeper/packet_test.go
AdityaSripal: The payloads are getting modulated in the test cases

modules/core/04-channel/v2/types/msgs_test.go
AdityaSripal: It gets used in the test cases

modules/core/04-channel/v2/keeper/packet.go
AdityaSripal: I left as is, because i find it more readable as to why the check exists

PR #8573, opened 2025-07-04
imp(transfer,rate-limiting,pfm)!: use AddressCodec instead of Bech32

srdtrk:
## Description

closes: IBCGO-2345

---

write a little note why.


modules/apps/packet-forward-middleware/keeper/keeper.go
```diff
@@ -51,9 +53,18 @@ type Keeper struct {
...
-func NewKeeper(cdc codec.BinaryCodec, storeService corestore.KVStoreService, transferKeeper types.TransferKeeper, channelKeeper types.ChannelKeeper, bankKeeper types.BankKeeper, authority string) *Keeper {
+func NewKeeper(
+	cdc codec.BinaryCodec,
+	addressCodec address.Codec,
+	storeService corestore.KVStoreService,
+	transferKeeper types.TransferKeeper,
+	channelKeeper types.ChannelKeeper,
+	bankKeeper types.BankKeeper,
+	authority string,
+) *Keeper {
```
DeshErBojhaa: We prefer function definitions and function calls on one line. Even if that line gets very long, please refer to the style guide on this repo.

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -78,8 +78,8 @@ func (k *Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types
 	return &types.MsgTransferResponse{Sequence: sequence}, nil
 }
 
-func (k *Keeper) transferV1Packet(ctx sdk.Context, sourceChannel string, token types.Token, timeoutHeight clienttypes.Height, timeoutTimestamp uint64, packetData types.FungibleTokenPacketData) (uint64, error) {
-	if err := k.SendTransfer(ctx, types.PortID, sourceChannel, token, sdk.MustAccAddressFromBech32(packetData.Sender)); err != nil {
+func (k *Keeper) transferV1Packet(ctx sdk.Context, sourceChannel string, token types.Token, timeoutHeight clienttypes.Height, timeoutTimestamp uint64, sender sdk.AccAddress, packetData types.FungibleTokenPacketData) (uint64, error) {
```
DeshErBojhaa, line 81: `sender` parameter looks redundant. Packet data already has the sender. Can packet data have the parsed sender instead of `msg.sender`?

modules/apps/transfer/keeper/msg_server.go
```diff
@@ -47,7 +47,7 @@ func (k *Keeper) Transfer(goCtx context.Context, msg *types.MsgTransfer) (*types
 		return nil, err
 	}
 
-	packetData := types.NewFungibleTokenPacketData(token.Denom.Path(), token.Amount, sender.String(), msg.Receiver, msg.Memo)
+	packetData := types.NewFungibleTokenPacketData(token.Denom.Path(), token.Amount, msg.Sender, msg.Receiver, msg.Memo)
```
DeshErBojhaa, line 50: Please explain why not the parsed sender?

```suggestion
	packetData := types.NewFungibleTokenPacketData(token.Denom.Path(), token.Amount, sender, msg.Receiver, msg.Memo)
```

srdtrk: Because the parsed sender is just bytes now. So doesn't have the String method. Unless you use BytesToString. But is there a need for this?

modules/apps/transfer/keeper/msg_server.go
srdtrk: Then we would have to parse it again. I could make this change if you want

gjermundgaraba: It's probably fine here.

modules/apps/packet-forward-middleware/keeper/keeper_test.go
```diff
@@ -334,3 +366,23 @@ func (*transferMock) DenomPathFromHash(ctx sdk.Context, ibcDenom string) (string
 func (*transferMock) GetPort(ctx sdk.Context) string {
 	return ""
 }
+
+type testAddressCodec struct{}
```
gjermundgaraba: Rather than keep recreating this, you can probably move this to `testing/`, no?

modules/apps/transfer/keeper/relay_test.go
```diff
@@ -30,6 +31,26 @@ var (
 	defaultAmount = ibctesting.DefaultCoinAmount
 )
 
+type TestAddressCodec struct{}
```
gjermundgaraba: Or better yet, move this one into `testing/` and use it in the other tests?

gjermundgaraba (approved):
This looks good to me! If you agree, I would also move the testing address codec into `testing/`.

PR #8587, opened 2025-07-22
chore: remove deprecated ErrPacketTimeout error

avorylli:
Remove deprecated ErrPacketTimeout error constant

Constant was unused in codebase


gjermundgaraba:
Hi, and thanks for the PR, but this change is not needed right now, and is kept around to avoid breaking changes - so it will stay until we are ready to remove it.

PR #8685, opened 2025-10-27
chore(simapp): remove dead intra-tx counter and fix misleading comment in zero-height export

Forostovec:
Removed an unused counter variable and its increment from the validator iteration in prepForZeroHeightGenesis() and updated the comment to accurately reflect behavior (only resetting bond heights). This dead code was a legacy artifact; no code reads the counter, and modern staking does not use a “bond intra-tx counter.” The change clarifies intent and avoids confusion


aljo242:
can you create a detailed issue explaining this?

Forostovec:
> can you create a detailed issue explaining this?

created, #8703

aljo242 (approved):

PR #8717, opened 2025-11-28
feat: attestation light client

gjermundgaraba:
## Description

closes: #XXXX

---

write a little note why.


modules/light-clients/10-attestations/README.md
```diff
@@ -0,0 +1,124 @@
...
+}
+// Encoding: abi.encode(height, timestamp) = 64 bytes (two padded uint64s)
+```
+
+When a valid update is received:
+1. Signatures are verified against the attestor set
+2. A new consensus state is created at the specified height
+3. `latestHeight` is updated if the new height exceeds it
+
+Updates can set consensus states for heights lower than or equal to `latestHeight`, enabling flexible state attestation.
```
gjermundgaraba:
```suggestion
Updates can also set consensus states for heights lower than or equal to `latestHeight`, enabling flexible state attestation.
```

modules/light-clients/10-attestations/abi.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// PacketCompact represents a packet commitment.
+// This type uses ABI encoding (not Protobuf) for cross-platform compatibility.
+type PacketCompact struct {
+	Path       []byte
+	Commitment []byte
+}
+
+func ABIEncodeStateAttestation(sa *StateAttestation) ([]byte, error) {
```
srdtrk: Should this be a method on State attestation instead?

modules/light-clients/10-attestations/abi.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, errorsmod.Wrap(ErrInvalidAttestationData, "invalid timestamp type")
+	}
+
+	return &StateAttestation{
+		Height:    height,
+		Timestamp: timestampSeconds * nanosPerSecond,
+	}, nil
+}
+
+func ABIEncodePacketAttestation(pa *PacketAttestation) ([]byte, error) {
```
srdtrk: ditto: should this be a method?

modules/light-clients/10-attestations/abi.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	return &PacketAttestation{
+		Height:  height,
+		Packets: packets,
+	}, nil
+}
+
+func ABIEncodePacketCompact(pc *PacketCompact) ([]byte, error) {
```
srdtrk: ditto: should this be a method?

modules/light-clients/10-attestations/abi.go
```diff
@@ -0,0 +1,158 @@
...
+	return tupleArrayType.Pack(bytesToBytes32(pc.Path), bytesToBytes32(pc.Commitment))
+}
+
+func bytesToBytes32(b []byte) [32]byte {
+	var result [32]byte
+	copy(result[:], b)
+	return result
+}
+
+func Uint64ToPaddedBytes(v uint64) []byte {
```
srdtrk: where is this used?

modules/light-clients/10-attestations/README.md
```diff
@@ -0,0 +1,134 @@
+# 10-attestations Light Client
```
srdtrk: Did we allocate ICS-10 to attestations?

modules/light-clients/10-attestations/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+An attestor-based IBC light client that verifies state using quorum-signed ECDSA attestations from a fixed set of trusted signers.
```
srdtrk: I was thinking we could call it multisig LC?

modules/light-clients/10-attestations/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Overview
+
+The attestations light client provides a trust model based on a quorum of trusted attestors rather than cryptographic verification of block headers. This design suits scenarios where a set of known, trusted parties can attest to the state of a counterparty chain.
```
srdtrk: It might be good to mention that the attestors attest to IBC packets rather than the state of the counterparty chain which could be interpreted as a merkle root.

modules/light-clients/10-attestations/consensus_state.go
```diff
@@ -0,0 +1,28 @@
...
+
+// ClientType returns Attestations type.
+func (ConsensusState) ClientType() string {
+	return exported.Attestations
+}
+
+// GetTimestamp returns the timestamp in nanoseconds.
+func (cs ConsensusState) GetTimestamp() uint64 {
+	return cs.Timestamp
+}
```
srdtrk: This is deprecated, so we can have this panic. And remove this from the interface in another PR

proto/ibc/lightclients/attestations/v1/attestations.proto
```diff
@@ -0,0 +1,40 @@
...
+  // when true, all verification and updates MUST fail
+  bool is_frozen = 4;
+}
+
+// ConsensusState defines an attestor consensus state. The timestamp of a
+// consensus state is stored per height.
+message ConsensusState {
+  option (gogoproto.goproto_getters) = false;
+  // trusted UNIX timestamp (nanoseconds) for the height
+  uint64 timestamp = 1;
```
srdtrk, line 28: In tendermint LC, we use `google.protobuf.Timestamp`. Maybe we should do the same for consistency?

modules/light-clients/10-attestations/errors.go
```diff
@@ -0,0 +1,23 @@
...
+	ErrInvalidAttestationData  = errorsmod.Register(ModuleName, 7, "invalid attestation data")
+	ErrInvalidHeight           = errorsmod.Register(ModuleName, 8, "invalid height")
+	ErrInvalidTimestamp        = errorsmod.Register(ModuleName, 9, "invalid timestamp")
+	ErrClientFrozen            = errorsmod.Register(ModuleName, 10, "client is frozen")
+	ErrNotMember               = errorsmod.Register(ModuleName, 11, "value not found in attested packet commitments")
+	ErrInvalidPath             = errorsmod.Register(ModuleName, 12, "invalid path")
+	ErrProcessedTimeNotFound   = errorsmod.Register(ModuleName, 13, "processed time not found")
+	ErrProcessedHeightNotFound = errorsmod.Register(ModuleName, 14, "processed height not found")
+	ErrDelayPeriodNotPassed    = errorsmod.Register(ModuleName, 15, "delay period has not been reached")
+	ErrNonMembershipFailed     = errorsmod.Register(ModuleName, 16, "non-membership verification failed: commitment is not zero")
```
srdtrk: The error message is more specific than the error title:
```suggestion
	ErrNonMembershipFailed     = errorsmod.Register(ModuleName, 16, "non-membership verification failed")
```

modules/light-clients/attestations/light_client_module.go
```diff
@@ -0,0 +1,186 @@
...
+
+// VerifyClientMessage obtains the client state associated with the client identifier and calls into the clientState.VerifyClientMessage method.
+func (l LightClientModule) VerifyClientMessage(ctx sdk.Context, clientID string, clientMsg exported.ClientMessage) error {
+	clientStore := l.storeProvider.ClientStore(ctx, clientID)
+	clientState, found := getClientState(clientStore, l.cdc)
+	if !found {
+		return errorsmod.Wrap(clienttypes.ErrClientNotFound, clientID)
+	}
+
+	return clientState.VerifyClientMessage(ctx, l.cdc, clientStore, clientMsg)
```
srdtrk, line 71: We could implement this stuff directly here. Implementing these methods in ClientState was an anti-pattern in tendermint imo

modules/light-clients/attestations/light_client_module.go
```diff
@@ -0,0 +1,186 @@
...
+
+// CheckForMisbehaviour returns false since the attestations client does not support misbehaviour detection.
+func (LightClientModule) CheckForMisbehaviour(ctx sdk.Context, clientID string, clientMsg exported.ClientMessage) bool {
+	return false
+}
+
+// UpdateStateOnMisbehaviour is not supported in this version.
+func (LightClientModule) UpdateStateOnMisbehaviour(ctx sdk.Context, clientID string, clientMsg exported.ClientMessage) {
+	panic(errorsmod.Wrap(ibcerrors.ErrInvalidRequest, "updateStateOnMisbehaviour is not supported"))
+}
```
srdtrk, line 82: We do have some basic checks in the solidity version but I guess this is fine for now.

modules/light-clients/attestations/light_client_module.go
```diff
@@ -0,0 +1,186 @@
...
+
+// UpdateState obtains the client state associated with the client identifier and calls into the clientState.UpdateState method.
+func (l LightClientModule) UpdateState(ctx sdk.Context, clientID string, clientMsg exported.ClientMessage) []exported.Height {
+	clientStore := l.storeProvider.ClientStore(ctx, clientID)
+	clientState, found := getClientState(clientStore, l.cdc)
+	if !found {
+		panic(errorsmod.Wrap(clienttypes.ErrClientNotFound, clientID))
+	}
+
+	return clientState.UpdateState(ctx, l.cdc, clientStore, clientMsg)
```
srdtrk, line 92: ditto about not using client state methods

modules/light-clients/attestations/light_client_module.go
```diff
@@ -0,0 +1,186 @@
...
+	path exported.Path,
+	value []byte,
+) error {
+	clientStore := l.storeProvider.ClientStore(ctx, clientID)
+	clientState, found := getClientState(clientStore, l.cdc)
+	if !found {
+		return errorsmod.Wrap(clienttypes.ErrClientNotFound, clientID)
+	}
+
+	return clientState.verifyMembership(clientStore, l.cdc, height, proof, path, value)
```
srdtrk, line 112: ditto about not using client state methods

modules/light-clients/attestations/light_client_module.go
```diff
@@ -0,0 +1,186 @@
...
+	proof []byte,
+	path exported.Path,
+) error {
+	clientStore := l.storeProvider.ClientStore(ctx, clientID)
+	clientState, found := getClientState(clientStore, l.cdc)
+	if !found {
+		return errorsmod.Wrap(clienttypes.ErrClientNotFound, clientID)
+	}
+
+	return clientState.verifyNonMembership(clientStore, l.cdc, height, proof, path)
```
srdtrk, line 131: ditto about not using client state methods

modules/light-clients/attestations/light_client_module.go
```diff
@@ -0,0 +1,186 @@
...
+		return 0, errorsmod.Wrapf(clienttypes.ErrConsensusStateNotFound, "consensus state not found for height %s", height)
+	}
+
+	return consensusState.Timestamp, nil
+}
+
+// RecoverClient is not supported in this version.
+func (LightClientModule) RecoverClient(ctx sdk.Context, clientID, substituteClientID string) error {
+	return errorsmod.Wrap(ibcerrors.ErrInvalidRequest, "recoverClient is not supported")
+}
```
srdtrk, line 181: We could (and probably should consider) support this

modules/light-clients/attestations/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Wire Compatibility
+
+The attestation data format is ABI-encoded for wire compatibility with:
+- Solidity attestor light client (`solidity-ibc-eureka`)
+- CosmWasm attestor light client (`cw-ics08-wasm-attestor`)
+
+This allows the same attestor infrastructure to generate proofs that work across all three platforms (Cosmos SDK, EVM, CosmWasm).
+
+## Trust Model
```
srdtrk, line 20: We should probably mark this entire module as experimental in the readme (and code), since I doubt this is exactly the design we would have settled on if we were able to spec this out properly. I also assume that there will be lots of changes in the attestor LC in the coming releases

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+
+	clienttypes "github.com/cosmos/ibc-go/v10/modules/core/02-client/types"
+	commitmenttypesv2 "github.com/cosmos/ibc-go/v10/modules/core/23-commitment/types/v2"
+	host "github.com/cosmos/ibc-go/v10/modules/core/24-host"
+	"github.com/cosmos/ibc-go/v10/modules/core/exported"
+)
+
+var _ exported.ClientState = (*ClientState)(nil)
+
+var zeroCommitment = make([]byte, 32)
```
srdtrk: Add godoc to mention it is used in verify non-membership

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+		return errorsmod.Wrap(clienttypes.ErrInvalidClient, "attestor addresses cannot be empty")
+	}
+	if cs.MinRequiredSigs == 0 {
+		return errorsmod.Wrap(clienttypes.ErrInvalidClient, "min required sigs cannot be 0")
+	}
+	if cs.MinRequiredSigs > uint32(len(cs.AttestorAddresses)) {
+		return errorsmod.Wrap(clienttypes.ErrInvalidClient, "min required sigs cannot exceed number of attestors")
+	}
+
+	seen := make(map[string]bool)
```
srdtrk: nit: It might make more sense to use `map[common.Address]bool` for duplicate detection.

modules/light-clients/attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+		if seen[normalizedAddr] {
+			return errorsmod.Wrap(clienttypes.ErrInvalidClient, "duplicate attestor address")
+		}
+		seen[normalizedAddr] = true
+	}
+
+	if cs.LatestHeight == 0 {
+		return errorsmod.Wrap(clienttypes.ErrInvalidClient, "latest height must be greater than 0")
+	}
+
```
srdtrk, line 70: It would be good to check that `isFrozen` is false

modules/light-clients/attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+	path exported.Path,
+	value []byte,
+) error {
+	if cs.IsFrozen {
+		return ErrClientFrozen
+	}
+
+	if path == nil || path.Empty() {
+		return errorsmod.Wrap(ErrInvalidPath, "path cannot be empty")
+	}
```
srdtrk, line 89: Should probably check that the path has length 1 since we want to use the merkle prefix `[""]`

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+	case interface{ Bytes() []byte }:
+		return normalizePathBytes(p.Bytes()), nil
+	default:
+		if stringer, ok := path.(fmt.Stringer); ok {
+			return normalizePathBytes([]byte(stringer.String())), nil
+		}
+	}
+
+	return nil, errorsmod.Wrapf(ErrInvalidPath, "unsupported path type %T", path)
+}
```
srdtrk: I don't like this function at all.
1. I think it should always be a `commitmenttypesv2.MerklePath`.
2. In that case, you should assert this length is = 1 and simply use path[0].

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	return crypto.Keccak256(raw)
+}
+
+func normalizeValueBytes(raw []byte) []byte {
+	if len(raw) == 32 {
+		return raw
+	}
+	return crypto.Keccak256(raw)
+}
```
srdtrk: Is this really true? I thought the attestor doesn't attest to anything except the packet commitment. This function shouldn't really exist. We can simply check that `len(raw) == 32`, else error

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return nil, errorsmod.Wrapf(ErrInvalidPath, "unsupported path type %T", path)
+}
+
+func normalizePathBytes(raw []byte) []byte {
+	if len(raw) == 32 {
+		return raw
+	}
+	return crypto.Keccak256(raw)
+}
```
srdtrk: we should always hash this regardless of path length. I'd rather remove this function too

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+
+	commitmentPath, err := canonicalizePath(path)
+	if err != nil {
+		return err
+	}
+
+	normalizedValue := normalizeValueBytes(value)
+
+	for _, packet := range packetAttestation.Packets {
+		if len(packet.Commitment) == 32 && len(packet.Path) == 32 && bytes.Equal(packet.Commitment, normalizedValue) && bytes.Equal(packet.Path, commitmentPath) {
```
srdtrk: Do we need the length check here?
```suggestion
		if bytes.Equal(packet.Commitment, normalizedValue) && bytes.Equal(packet.Path, commitmentPath) {
```

modules/light-clients/attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+		return err
+	}
+
+	normalizedValue := normalizeValueBytes(value)
+
+	for _, packet := range packetAttestation.Packets {
+		if len(packet.Commitment) == 32 && len(packet.Path) == 32 && bytes.Equal(packet.Commitment, normalizedValue) && bytes.Equal(packet.Path, commitmentPath) {
+			return nil
+		}
+	}
```
srdtrk, line 145: there should be a slices helper for this instead of a for loop? Not that we need to use it

modules/light-clients/attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+
+	if !foundMatchingPath {
+		return ErrNotMember
+	}
+
+	if !allZeroCommitments {
+		return ErrNonMembershipFailed
+	}
+
+	return nil
```
srdtrk, line 226: There must be a better way to do this:
```suggestion
	for _, packet := range packetAttestation.Packets {
		if bytes.Equal(packet.Path, commitmentPath) {
			if !bytes.Equal(packet.Commitment, zeroCommitment) {
				return ErrNonMembershipFailed
			}
			return nil
		}
	}

	return ErrNotMember
```

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+
+	if packetAttestation.Height != height.GetRevisionHeight() {
+		return errorsmod.Wrapf(ErrInvalidHeight, "height mismatch: expected %d, got %d", height.GetRevisionHeight(), packetAttestation.Height)
+	}
+
+	if len(packetAttestation.Packets) == 0 {
+		return errorsmod.Wrap(ErrInvalidAttestationData, "packets cannot be empty")
+	}
+
+	commitmentPath, err := canonicalizePath(path)
```
srdtrk: We probably just want to hash here, instead of canonicalizing. See comment in `canonicalizePath`

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
```
srdtrk: We probably just want to hash here, instead of canonicalizing. See comment in `canonicalizePath`

modules/light-clients/10-attestations/client_state.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return err
+	}
+
+	normalizedValue := normalizeValueBytes(value)
```
srdtrk: Instead of normalizing, we can simply check if the len == 32 since the attestors only support packet commitments anyway

modules/light-clients/10-attestations/signature.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	attestorSet := make(map[string]bool)
+	for _, addr := range cs.AttestorAddresses {
+		attestorSet[strings.ToLower(addr)] = true
+	}
+
+	hash := sha256.Sum256(proof.AttestationData)
+	seenSigners := make(map[string]bool)
+	validSigs := 0
```
srdtrk: You don't need to keep track of the number of valid sigs since you already always return error in invalid or duplicate signatures. You can simply use the `len(proof.Signatures)`

modules/light-clients/10-attestations/signature.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// verifySignatures verifies that the attestation proof has valid signatures from unique attestors
+// meeting the quorum threshold. Signatures cover sha256(attestationData).
+func (cs *ClientState) verifySignatures(proof *AttestationProof) error {
+	if len(proof.Signatures) == 0 {
+		return errorsmod.Wrap(ErrInvalidSignature, "signatures cannot be empty")
+	}
+
+	attestorSet := make(map[string]bool)
```
srdtrk: I think here we should work with common.Address instead of string
```suggestion
	attestorSet := make(map[common.Address]bool)
```

modules/light-clients/10-attestations/signature.go
```diff
@@ -0,0 +1,67 @@
...
+	for _, addr := range cs.AttestorAddresses {
+		attestorSet[strings.ToLower(addr)] = true
+	}
+
+	hash := sha256.Sum256(proof.AttestationData)
+	seenSigners := make(map[string]bool)
```
srdtrk: ditto: common.Address instead of string. You have to do this conversion during the for loop anyway

modules/light-clients/attestations/attestation_proof.go
```diff
@@ -0,0 +1,34 @@
...
+func (AttestationProof) ClientType() string {
+	return exported.Attestations
+}
+
+// ValidateBasic ensures that the attestation data and signatures are initialized.
+func (ap AttestationProof) ValidateBasic() error {
+	if len(ap.AttestationData) == 0 {
+		return errorsmod.Wrap(clienttypes.ErrInvalidHeader, "attestation data cannot be empty")
+	}
+
```
srdtrk, line 37: We could also deserialize the attestation data to ABI here and check that the `len(packets) != 0`

Makefile
```diff
@@ -360,7 +360,7 @@ format:
 
 protoVer=0.17.1
 protoImageName=ghcr.io/cosmos/proto-builder:$(protoVer)
-protoImage=$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace $(protoImageName)
+protoImage=$(DOCKER) run --rm -e GOTOOLCHAIN=auto -v $(CURDIR):/workspace --workdir /workspace $(protoImageName)
```
srdtrk, line 363: what's the story here?

e2e/tests/attestations/attestations_test.go
```diff
@@ -0,0 +1,891 @@
...
+func (s *AttestationsTestSuite) getAttestorAddresses() []string {
+	var addresses []string
+	for _, key := range s.attestorKeys {
+		address := crypto.PubkeyToAddress(key.PublicKey)
+		addresses = append(addresses, address.Hex())
+	}
+	return addresses
+}
+
+func (s *AttestationsTestSuite) signAttestationData(data []byte) [][]byte {
```
srdtrk: nit: better name could be `signData`

e2e/tests/attestations/attestations_test.go
```diff
@@ -0,0 +1,891 @@
...
+}
+
+func (*AttestationsTestSuite) hashPath(path []byte) []byte {
+	return crypto.Keccak256(path)
+}
+
+func (s *AttestationsTestSuite) prefixedPath(key []byte) []byte {
+	prefixedPath := bytes.Join([][]byte{[]byte(ibcexported.StoreKey), key}, []byte("/"))
+	return s.hashPath(prefixedPath)
+}
```
srdtrk: I thought we always use `[][]byte{[]byte("")}` as the counterparty prefix for attestor LC?

e2e/tests/attestations/attestations_test.go
```diff
@@ -0,0 +1,891 @@
...
+
+func (s *AttestationsTestSuite) TestMsgTransfer_Attestations() {
```
srdtrk, line 46: why are we testing with IBC v1??

e2e/tests/attestations/attestations_test.go
```diff
@@ -0,0 +1,891 @@
...
+			clientIDA,
+			connectiontypes.NewCounterparty(clientIDB, "", commitmenttypes.NewMerklePrefix([]byte(ibcexported.StoreKey))),
+			[]*connectiontypes.Version{ibctesting.ConnectionVersion},
+			0,
+		)
+		connEndBz, err := cfg.Codec.Marshal(&connEnd)
+		s.Require().NoError(err)
+		connHash := crypto.Keccak256(connEndBz)
+
+		proofInit := s.createPacketAttestationProof(proofHeight, connPath, connHash)
```
srdtrk: we probably want to use empty merkle prefix. But still, we really should set it up with ibc v2 since the attestor binary wouldn't be produce these proofs anyway

e2e/tests/attestations/attestations_test.go
srdtrk, line 1: We should use ibc v2 tests so that:
1. The attestor never needs to produce connection proofs, so we don't need to worry about normalizing anything.
2. Attestor design assumes it's only attesting to packets
3. We use the easier merkle prefix setup with the empty prefix

modules/light-clients/10-attestations/README.md
gjermundgaraba: I don't know, I just took one. I'll leave that up to you to figure out :)

modules/light-clients/10-attestations/README.md
gjermundgaraba: We call it attestations everywhere, so I don't see a reason to change the name now, unless we agree to rename everything to that. I personally think multisig LC sounds worse, but for now I'd say this is not related to this PR exactly

proto/ibc/lightclients/attestations/v1/attestations.proto
gjermundgaraba: Would we also have to change that in the solidity repo?

e2e/tests/attestations/attestations_test.go
gjermundgaraba: lol, fair 😅

e2e/tests/attestations/attestations_test.go
gjermundgaraba: Hmm, yes, that is probably true, will fix that

e2e/tests/attestations/attestations_test.go
gjermundgaraba: Not sure I agree, feel like it reads better in the code like this, but your call :)

modules/light-clients/attestations/light_client_module.go
gjermundgaraba: I can, but I opted to keep the current pattern so the code is more consistent and discoverable as a result. I would personally have a separate issue to do that across the board. But up to you

modules/light-clients/attestations/light_client_module.go
gjermundgaraba: I would reccommend creating a separate ticket for that to get this working for the current timelines first.

modules/light-clients/10-attestations/client_state.go
gjermundgaraba: I'll just rename to avoid that problem ;)

modules/light-clients/attestations/client_state.go
gjermundgaraba: I don't think that would be correct here. It's not an invalid client state to be frozen

modules/light-clients/10-attestations/client_state.go
gjermundgaraba: Yeah, you are right, removing it

modules/light-clients/10-attestations/client_state.go
gjermundgaraba: Oh year, ur right, will remove

modules/light-clients/10-attestations/client_state.go
gjermundgaraba: mm, yes, will do

modules/light-clients/attestations/README.md
gjermundgaraba: I will not do that in this PR. You can add it, either directly to this branch or separately as you want.

modules/light-clients/attestations/client_state.go
gjermundgaraba: I'm not sure I see how it would help here, so haven't done it

Makefile
gjermundgaraba: I was struggeling with some build issues related to the go toolchain being messed up when using go 1.25 - I think

e2e/tests/attestations/attestations_test.go
gjermundgaraba: Fixed

modules/light-clients/10-attestations/README.md
srdtrk: I decided to remove the number on the client-id. Everything seems to be passing :)

e2e/tests/attestations/attestations_test.go
srdtrk: we can keep it then

srdtrk (approved):

PR #8771, opened 2026-02-02
chore: fix the linter errors

aljo242:


modules/apps/27-gmp/keeper/genesis_test.go
```diff
@@ -1,13 +1,13 @@
...
 import (
-	sdk "github.com/cosmos/cosmos-sdk/types"
-
 	"github.com/cosmos/ibc-go/v10/modules/apps/27-gmp/types"
 	ibcerrors "github.com/cosmos/ibc-go/v10/modules/core/errors"
 	ibctesting "github.com/cosmos/ibc-go/v10/testing"
 )
 
+var invalidAddress = "invalid"
+
```
srdtrk, line 10: We can use `ibctesting.InvalidID` instead

modules/apps/27-gmp/keeper/msg_server_test.go
```diff
@@ -52,7 +52,7 @@ func (s *KeeperTestSuite) TestSendCall() {
 		{
 			"failure: invalid sender address",
 			func() {
-				msg.Sender = "invalid"
+				msg.Sender = invalidAddress
```
srdtrk, line 55: ditto: We can use `ibctesting.InvalidID` instead

modules/apps/27-gmp/ibc_module_test.go
```diff
@@ -27,6 +27,7 @@ type IBCModuleTestSuite struct {
 const (
 	validClientID   = ibctesting.FirstClientID
 	invalidClientID = "invalid"
+	invalidPort     = "invalid-port"
```
srdtrk, line 30: ditto: We can use `ibctesting.InvalidID` instead

modules/core/04-channel/v2/types/commitment.go
srdtrk, line 1: These are critical functions that were audited. Would be good to not modify if not needed.

I only see some semantic changes that doesn't provide any safetly/utility.

modules/light-clients/08-wasm/types/gas_register.go
```diff
@@ -198,7 +198,7 @@ func (g WasmGasRegister) ReplyCosts(discount bool, reply wasmvmtypes.Reply) stor
 	msgLen := len(reply.Result.Err)
 	if reply.Result.Ok != nil {
 		msgLen += len(reply.Result.Ok.Data)
-		var attrs []wasmvmtypes.EventAttribute
+		attrs := make([]wasmvmtypes.EventAttribute, 0, msgLen)
```
srdtrk, line 201: I don't even think this is correct. Neither the length nor capacity should be msgLen. You can see in the for loop below, that to determine the length, you probably need to a for loop first. Here, having variable length feels easier to read and reason about

srdtrk (changes requested):

PR #8937, opened 2026-05-12
fix(rate-limiting)!: Change channel store key length from 16 to 64 bytes

mattac21:
## Description

Increase rate limiting app's `PendingSendPacketChannelLength` from 16 -> 64 bytes. Adds a migration handler to upgrade existing rate limiting store keys from the old 16 byte layout to the 64 byte layout.

closes: #XXXX

---

write a little note why.

- [x] Include changelog entry when appropriate (e.g. chores should be omitted from changelog).
- [x] Wrote unit and integration [tests](https://github.com/cosmos/ibc-go/blob/main/testing/README.md#ibc-testing-package) if relevant.


mattac21:
@greptile review

mattac21:
@greptile review again

mattac21:
@greptile review again

mattac21:
@greptile final review

modules/apps/rate-limiting/keeper/migrator.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/cosmos/ibc-go/v11/modules/apps/rate-limiting/migrations/v2"
+)
+
+// Migrator is a struct for handling in-place store migrations.
+type Migrator struct {
+	keeper *Keeper
+}
+
+func NewMigrator(k *Keeper) Migrator {
```
srdtrk, line 15: add godoc

modules/apps/rate-limiting/migrations/v2/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+	defer func() {
+		err = errors.Join(err, iterator.Close())
+	}()
+
+	for ; iterator.Valid(); iterator.Next() {
+		key := iterator.Key()
+		switch len(key) {
+		case newKeyLen:
+			// already correct length, nothing to do
+			continue
```
srdtrk: Should we be removing this case, since if this is a migration, we never expect to see the newKeyLen?

```suggestion
```

modules/apps/rate-limiting/migrations/v2/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+	return nil
+}
+
+type entry struct {
+	key, value []byte
+}
+
+// collectLegacyEntries returns a list of entries in the prefix store that must
+// be migrated from the oldKeyLen to the newKeyLen.
+func collectLegacyEntries(store prefix.Store, oldKeyLen, newKeyLen int) (legacy []entry, err error) {
```
srdtrk: nit: we generally don't use named return variables if not necessary. Makes the code harder to read imo

modules/apps/rate-limiting/migrations/v2/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+		store.Delete(entry.key)
+		store.Set(newKey, entry.value)
+	}
+
+	return nil
+}
+
+type entry struct {
+	key, value []byte
+}
```
srdtrk, line 77: nit: It seems this is only used for legacy entries. So I think you could use a fix length byte array here for key

modules/apps/rate-limiting/migrations/v2/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// get store entries that need to be migrated
+	legacyEntries, err := collectLegacyEntries(store, oldKeyLen, newKeyLen)
+	if err != nil {
+		return fmt.Errorf("collecting legacy pending send packet entries from prefix store: %w", err)
+	}
+
+	// migrate store entries
+	// old key layout: [16-byte channelID][8-byte sequence]
+	// new key layout: [64-byte channelID][8-byte sequence]
```
srdtrk, line 51: Would it be possible to modify the keeper functions so that if the channel id exceeds 64-bytes, we error. This way, we can prevent unexpected behavior in the future?

modules/apps/rate-limiting/migrations/v2/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, entry := range legacyEntries {
+		newKey := make([]byte, newKeyLen)
+
+		// place 16 byte channel id from old key into first 64 bytes of new key
+		copy(newKey, entry.key[:oldPendingSendPacketChannelLength])
```
srdtrk, line 56: Do we want to first isolate the channel-id from `entry.key` here so that we can add basic validation that it is not corrupted already. If so, we can return or error or simply delete the key?

I can see a case for not doing this as well

modules/apps/rate-limiting/migrations/v2/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+
+		// put remaining 8 bytes sequence from old key into the final 8 bytes
+		// sequence of the new key
+		copy(newKey[newPendingSendPacketChannelLength:], entry.key[oldPendingSendPacketChannelLength:])
```
srdtrk, line 60: I think it is worth validating newKey after this step:
- Construct channel-id from the first 64 bytes
- Construct the sequence from the last 8 bytes

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	tests := []struct {
+		name      string
+		malleate  func(store prefix.Store)
+		expectAll []string
+		postCheck func(store prefix.Store)
+		expectErr string
+	}{
+		{
+			name:     "empty store",
```
srdtrk: nit
```suggestion
			name:     "success: empty store",
```

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+			malleate: func(_ prefix.Store) {},
+		},
+		{
+			name: "legacy entries rewritten",
```
srdtrk: nit
```suggestion
			name: "success: legacy entries rewritten",
```

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+				writeLegacy(store, "channel-1", 0)
+				writeLegacy(store, "channel-1", 1)
+			},
+			expectAll: []string{"channel-1/0", "channel-1/1"},
+			postCheck: func(store prefix.Store) {
+				before := readAllKeys(store)
+				s.Require().NoError(migrator.Migrate1to2(ctx))
+				s.Require().ElementsMatch(before, readAllKeys(store))
+			},
+		},
```
srdtrk, line 93: Why is there a test for this. This is not even desired
```suggestion
```

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+				writeLegacy(store, "channel-11", 0)
+				writeLegacy(store, "channel-11", 7)
+			},
+			expectAll: []string{
+				"channel-1/0", "channel-1/1",
+				"channel-11/0", "channel-11/7",
+			},
+		},
+		{
+			name: "mixed legacy and new-layout entries",
```
srdtrk: Once you address my comments, this should become a failure case.

```suggestion
			name: "failure: mixed legacy and new-layout entries",
```

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+			},
+			expectAll: []string{"channel-1/3"},
+			postCheck: func(store prefix.Store) {
+				longChannelID := fmt.Sprintf("channel-%050d", 1)
+				s.Require().Greater(len(longChannelID), oldPendingSendPacketChannelLength)
+				rlKeeper.SetPendingSendPacket(ctx, longChannelID, 9)
+				s.Require().True(rlKeeper.CheckPacketSentDuringCurrentQuota(ctx, longChannelID, 9))
+				s.Require().True(rlKeeper.CheckPacketSentDuringCurrentQuota(ctx, "channel-1", 3))
+			},
+		},
```
srdtrk, line 107: This test case shouldn't be here. The goal of these unit tests is to test the migration logic, not the application logic. Instead you should add a test case to already existing application tests where you have a long channel id.

BTW, I think this `postCheck` might have passed even before the fix because wouldn't  the long channel-id be truncated? And given that there is only a single channel, there is no collision either?

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+				rlKeeper.SetPendingSendPacket(ctx, idA, 1)
+				rlKeeper.SetPendingSendPacket(ctx, idB, 2)
+
+				s.Require().True(rlKeeper.CheckPacketSentDuringCurrentQuota(ctx, idA, 1))
+				s.Require().True(rlKeeper.CheckPacketSentDuringCurrentQuota(ctx, idB, 2))
+				s.Require().Len(readAllKeys(store), 2, "two distinct keys despite shared first 16 bytes")
+			},
+		},
+	}
```
srdtrk, line 115: I think this should also go to application tests

```suggestion
```

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+		ctx      sdk.Context
+		rlKeeper *keeper.Keeper
+		migrator keeper.Migrator
+	)
+
+	tests := []struct {
+		name      string
+		malleate  func(store prefix.Store)
+		expectAll []string
+		postCheck func(store prefix.Store)
```
srdtrk: I think this is not needed, see my other comments for why

```suggestion
```

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+		return keys
+	}
+
+	var (
+		ctx      sdk.Context
+		rlKeeper *keeper.Keeper
+		migrator keeper.Migrator
+	)
+
+	tests := []struct {
```
srdtrk, line 51: I think you're missing some failure cases, such as incorrect length keys existing in store

modules/apps/rate-limiting/keeper/migrator_test.go
```diff
@@ -0,0 +1,167 @@
...
+			if tc.expectAll == nil {
+				s.Require().Empty(rlKeeper.GetAllPendingSendPackets(ctx))
+			} else {
+				s.Require().ElementsMatch(tc.expectAll, rlKeeper.GetAllPendingSendPackets(ctx))
+			}
+
+			// do custom post checks
+			if tc.postCheck != nil {
+				tc.postCheck(prefixStore)
+			}
```
srdtrk: ditto: no need for postCheck

```suggestion
```

srdtrk (changes requested):

modules/apps/rate-limiting/migrations/v2/migrate.go
mattac21: we dont expect it, but it seemed better to be defensive about it and simply skip these. it was also for the idempotency that felt good to have incase someone runs the migration twice, i.e. on the second run we will hit this case and skip migrating the already migrated entries.

modules/apps/rate-limiting/migrations/v2/migrate.go
mattac21: yeah I dont love it either, just had it since I was using joining the returned error in the defer with the iterator close, the function felt small enough where it wasn't hard to read but I can change it

modules/apps/rate-limiting/migrations/v2/migrate.go
mattac21: > validation that it is not corrupted already

I dont think this is actually possible, the 'corruption' here is that when the key was created it was silently truncated down to length 16. So for example that would be 07-tendermint-100 being truncated to 07-tendermint-10, we can't be sure that the key is actually referring to a 07-tendermint-10 client, or if it was truncated.

mattac21: I guess we could do a lookup of existing clients on chain and ensure that they all exist? that would cover the case where a key that is truncated into a client that doesnt exist at all, but that doesnt cover the (worse) case where a key is truncated into a key that conflicts with an already existing one

modules/apps/rate-limiting/keeper/migrator_test.go
mattac21: seemed like a good property to have, double migrate turns into a non issue

modules/apps/rate-limiting/migrations/v2/migrate.go
mattac21: i can remove it tho in order to address your other comments

modules/apps/rate-limiting/keeper/migrator_test.go
mattac21: makes sense I will move it

modules/apps/rate-limiting/migrations/v2/migrate.go
mattac21: I added some validation to the channelID in new key after we have done the migration using some existing validation functions. it doesn't check for any corruption that would have happened to the channel ID based on the 16 byte limit, but if you want I can add that variation to the existing key as well and just skip over ones that fail (although on im not sure it is actually possible for us to reach a point where we have an existing entry with a channel id that will fail this).

modules/apps/rate-limiting/keeper/migrator_test.go
mattac21: can you be more specific about where you want this test? from my understanding the test here is already spinning up a simp, so that feels 'application test'-y already. Do you want a full e2e test for this using interchain test etc? imo having this in the unit test calling keeper functions feels sufficient.

modules/apps/rate-limiting/keeper/pending_send.go
srdtrk: Given that the function signature now returns an error, we can (and should) return an error if the `iterator.Close()` returns an error

modules/apps/rate-limiting/keeper/migrator_test.go
srdtrk: I just meant location of this, not e2e. I think `modules/apps/rate-limiting/keeper/pending_send_test.go` is a suitable place. You simply test that "long channel keys sharing first 16 bytes don't collide". Since this doesn't have much to do with the migrator

modules/apps/rate-limiting/keeper/migrator_test.go
srdtrk: ditto: I think `modules/apps/rate-limiting/keeper/pending_send_test.go` is a suitable place.

modules/apps/rate-limiting/migrations/v2/migrate.go
```diff
@@ -0,0 +1,150 @@
...
+	rawChannelID := string(newKey[:newPendingSendPacketChannelLength])
+	channelID := strings.TrimRight(rawChannelID, "\x00")
+
+	// validate channelID in the newKey.
+
+	// we are using the client
+	// validator here since in v1 these will be channelID's, and in v2 they
+	// are clientID's, the client validator is slightly less strict and
+	// will accept both.
+	if err := host.ClientIdentifierValidator(channelID); err != nil {
```
srdtrk, line 123: I don't know that we need to accept both v1 and v2 ids. But I think this is a nit, and I'd be happy to leave as is.

```suggestion
	if err := host.ChannelIdentifierValidator(channelID); err != nil {
```

srdtrk (commented):
Lgtm, left a few small comments and I can approve after these are addressed. Will you backport this to `ibc-apps`, since rate-limiting was released for v10 there?

modules/apps/rate-limiting/keeper/pending_send.go
mattac21: Good callout will do

mattac21:
> Lgtm, left a few small comments and I can approve after these are addressed. Will you backport this to `ibc-apps`, since rate-limiting was released for v10 there?

Yes I will backport it there

modules/apps/rate-limiting/migrations/v2/migrate.go
mattac21: I'm going to leave as is, I think the flexibility here is better

modules/apps/rate-limiting/keeper/pending_send.go
```diff
@@ -61,15 +73,22 @@ func (k *Keeper) GetAllPendingSendPackets(ctx sdk.Context) []string {
 		pendingPackets = append(pendingPackets, packetID)
 	}
 
-	return pendingPackets
+	if err := iterator.Close(); err != nil {
```
srdtrk: We want to put the `iterator.Close()` statements inside a defer right after constructing the iterator if possible since defer statements run even if a function panics in the middle. Given that sdk recovers from these panics, this might be relevant somehow. Is it possible to return the error inside the defer?

srdtrk (commented):
Lgtm. Left one question about the `it.Close()`

modules/apps/rate-limiting/keeper/pending_send.go
mattac21: we would need a named return, which you didnt like before, so I opted for this instead. I personally prefer it to be in the defer with a named return (basically the only time I am ok with a named return), but again this was just for your prev preference of no named return values

srdtrk: I agree, I think named return is ok in this case. Let's do that

modules/apps/rate-limiting/keeper/pending_send.go
srdtrk: This isn't modified?

srdtrk (approved):
LGTM. Thanks for the great work!

PR #8959, opened 2026-05-21
chore: use bash array for package iteration

sashaphmn:
Use a proper bash array for package iteration in coverage script. This avoids relying on implicit word splitting and makes ${PKGS[@]} usage semantically correct.


srdtrk:
This change is not needed

PR #8982, opened 2026-07-02
ci: reduce depot usage

dianab-cl:
## Description

Reduces [Depot](https://depot.dev/) usage by migrating GitHub Actions workflows from Depot-hosted runners to standard GitHub-hosted runners, and hardens the workflows to keep CI fast and predictable on the new infrastructure.

- Replaces runs-on: depot-ubuntu-24.04-4 with the standard ubuntu-24.04 runner across all affected workflows (same OS, equivalent 4-vCPU spec).
- docker.yml: adds Docker Buildx + GitHub Actions layer cache (cache-from/cache-to: type=gha) so image builds don't start cold on every push. load: true is set on the build step so the existing "test simd is runnable" step still works under the buildx container driver.
- Adds a concurrency group to every workflow so superseded runs don't pile up:
CI checks (lint, tests, proto, docs-check, codeql) use cancel-in-progress: true — a newer commit cancels the older run.
Publish/deploy workflows (docker, docs-deploy, proto-registry) and manual image builds use cancel-in-progress: false so an in-flight push is never interrupted.
- Adds timeout-minutes to every job so a hung job fails fast instead of squatting a runner slot for the 6-hour default — important now that the runner pool is a fixed shared resource.
- Replaces the hardcoded go-version: '1.25.9' with go-version-file pointing at the relevant go.mod, so bumping a module's Go directive automatically updates CI. In golangci.yml, actions/checkout now runs before actions/setup-go (required for go-version-file to read the file on disk).

---

write a little note why.


dianab-cl:
@greptile review

mattac21 (approved):

Issue #8653, opened 2025-09-25
IBC packet non-timeout with invalid RevisionNumber


lapidix:
## Summary of Bug

When sending an IBC transaction, setting the `RevisionNumber` in the `--packet-timeout-height` flag to a value higher than the actual chain’s `RevisionNumber` can result in a transaction that never times out.

Allowing a high `RevisionNumber` in [test code](https://github.com/cosmos/ibc-go/blob/4cb2ec7a329084f3664a3cbc6ecb3f4a04783498/modules/core/02-client/types/height_test.go#L16-L51) appears be intentional; however, possible to generate transactions without timeout feature, as demonstrated in the following link, appears abnormal and could potentially be a vulnerability: https://www.mintscan.io/milkyway/tx/D2D58B250EA83999D477E9B5E732355D0074FC6FCCE87B88945B7011877CCD93?sector=logs

I’m aware that the `packet-timeout-height` was removed in IBC v2. However, to my knowledge, most chains are currently running with `ibc-go v8.x.x`, which is based on IBC v1 and still includes the issue described above.

Additionally, I know that recent versions no longer support relative timeouts using block height at the CLI level, relying instead on timestamp-based timeouts. However, I believe it would be more robust to strengthen the validation logic at the protocol or state machine level as well.

## Expected Behavior

Timeout prioritizes height over timestamp as shown in the code below:

```go
func (t Timeout) Elapsed(height clienttypes.Height, timestamp uint64) bool {
	return t.heightElapsed(height) || t.timestampElapsed(timestamp)
}

```

Additionally, in the `Compare()` function that validates Height, it unconditionally passes as long as the revision number is greater than the counterparty's revision number.

```go
func (h Height) Compare(other exported.Height) int64 {
	height, ok := other.(Height)
	if !ok {
		panic(fmt.Errorf("cannot compare against invalid height type: %T. expected height type: %T", other, h))
	}
	var a, b big.Int
	if h.RevisionNumber != height.RevisionNumber {
		a.SetUint64(h.RevisionNumber)
		b.SetUint64(height.RevisionNumber)
	} else {
		a.SetUint64(h.RevisionHeight)
		b.SetUint64(height.RevisionHeight)
	}
	return int64(a.Cmp(&b))
}

```

Through this, packet timeout block heights like `9000-123456` are also passing validation checks, and when the timeout timestamp is 0, it's possible to create transactions that never timeout.

## Version

All versions including main

## Steps to Reproduce

[...]

PR #8983, opened 2026-07-02
fix(04-channel): reject packets with an unreachable timeout height revision

renezander030:
## Description

Addresses #8653.

When a packet is sent with a timeout height whose **revision number is greater than the counterparty client's current revision number**, the height can never be reached — revision numbers only increase through a coordinated chain upgrade. `Height.Compare` orders by revision number first, so `heightElapsed` never returns true for such a height. Combined with a zero timeout timestamp (`--packet-timeout-timestamp 0`), this yields a packet that **never times out** (the reporter demonstrated this on a live chain via the CLI).

This PR rejects such timeouts in `SendPacket`, before the packet commitment is written:

```go
if !timeoutHeight.IsZero() && latestHeight.RevisionNumber > 0 &&
    timeoutHeight.RevisionNumber > latestHeight.RevisionNumber {
    return 0, errorsmod.Wrapf(clienttypes.ErrInvalidHeight,
        "packet timeout height revision number (%d) cannot exceed the counterparty client's current revision number (%d)",
        timeoutHeight.RevisionNumber, latestHeight.RevisionNumber)
}
```

### Design notes (worth a reviewer's eye)

- **Solomachine is intentionally exempt.** Solomachine clients report `RevisionNumber == 0` and legitimately send with timeout heights on a different revision. An unconditional revision-equality check breaks the existing `TestSendPacket` "success with solomachine" cases, so the check is scoped to `latestHeight.RevisionNumber > 0` (revision-numbered / Tendermint clients).
- **Only `>` is rejected, not `!=`.** A *lower* timeout-height revision is already handled by the existing `timeout.Elapsed` check (it returns `ErrTimeoutElapsed`); this PR only adds the unreachable *higher*-revision case, keeping the change minimal.
- **Known limitation:** a Tendermint chain whose counterparty revision number is `0` (a chain-id without a revision suffix) is not covered by the `> 0` guard. Fixing that generally, or hardening `heightElapsed` directly (issue's "Approach 2"), is a larger change with broader blast radius on consensus-critical timeout evaluation; happy to take it that direction if maintainers prefer.
- This is **state-machine-breaking** (a `SendPacket` that previously succeeded now errors), so it should target a minor/major release, not a patch.

## Testing

- Added `TestSendPacket` case "timeout height has invalid revision number".
- `go test ./modules/core/04-channel/keeper/` passes (full package), `go vet` clean.

## Author Checklist

- [x] Reviewed the [contributing guidelines](https://github.com/cosmos/ibc-go/blob/main/docs/dev/project-structure.md)
- [x] Added tests
- [x] Added a changelog entry
- [x] Targeted the correct branch (`main`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### State Machine Breaking
 
 * (apps/rate-limiting) [\#8937](https://github.com/cosmos/ibc-go/pull/8937) Increase pending send packet channel length key size from 16 to 64
+* (core/04-channel) [\#8983](https://github.com/cosmos/ibc-go/pull/8983) Reject in `SendPacket` a timeout height whose revision number exceeds the counterparty client's current revision number (an unreachable timeout). See [\#8653](https://github.com/cosmos/ibc-go/issues/8653).
 
 ### Improvements
 
--- a/modules/core/04-channel/keeper/packet.go
+++ b/modules/core/04-channel/keeper/packet.go
@@ -65,6 +65,23 @@ func (k *Keeper) SendPacket(
 		return 0, errorsmod.Wrapf(clienttypes.ErrInvalidHeight, "cannot send packet using client (%s) with zero height", connectionEnd.ClientId)
 	}
 
+	// A timeout height whose revision number is greater than the counterparty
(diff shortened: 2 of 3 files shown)


mattac21 (changes requested):
conflict now in changelog, could you address?

mattac21:
unresponsive closing

PR #8996, opened 2026-07-15
feat: apply pagination for ratelimit queries

### LINKED ISSUE #282: Missing pagination in ratelimit query endpoints
Several endpoints return unbounded results. AllRateLimits, RateLimitsByChainId, RateLimitsByChannelOrClientId, AllBlacklistedDenoms and AllWhitelistedAddresses return full KV store contents without pagination.

mmsqe:
## Description

closes: https://github.com/cosmos/ibc-apps/issues/282

---

write a little note why.


swift1337 (approved):

modules/apps/rate-limiting/keeper/grpc_query.go
```diff
@@ -53,63 +66,86 @@ func (k Querier) RateLimit(c context.Context, req *types.QueryRateLimitRequest)
 	return &types.QueryRateLimitResponse{RateLimit: &rateLimit}, nil
 }
 
-// Query all rate limits for a given chain
+// RateLimitsByChainID returns paginated rate limits whose channel/client resolves to the given chain.
 func (k Querier) RateLimitsByChainID(c context.Context, req *types.QueryRateLimitsByChainIDRequest) (*types.QueryRateLimitsByChainIDResponse, error) {
 	if req == nil {
 		return nil, status.Error(codes.InvalidArgument, "invalid request")
 	}
+	if req.ChainId == "" {
```
gjermundgaraba, line 74: These two new empty-string guards (here and in `RateLimitsByChannelOrClientID`) are currently untested (deleting them still passes the keeper suite). The sibling `RateLimit` endpoint accepts the same empty `channel_or_client_id` at the gRPC level, returning success with a nil rate limit.

modules/apps/rate-limiting/keeper/grpc_query.go
```diff
@@ -53,63 +66,86 @@ func (k Querier) RateLimit(c context.Context, req *types.QueryRateLimitRequest)
...
+	pageRes, err := query.FilteredPaginate(store, req.Pagination, func(_, value []byte, accumulate bool) (bool, error) {
+		var rateLimit types.RateLimit
+		if err := k.k.cdc.Unmarshal(value, &rateLimit); err != nil {
+			return false, err
+		}
 
-	rateLimits := make([]types.RateLimit, 0)
-	for _, rateLimit := range k.k.GetAllRateLimits(ctx) {
 		// Determine the client state from the channel Id
 		_, clientState, err := k.k.channelKeeper.GetChannelClientState(ctx, transfertypes.PortID, rateLimit.Path.ChannelOrClientId)
```
gjermundgaraba: `GetChannelClientState` (channel → connection → client reads, incl. an `Any` decode) runs once per scanned entry, and the scan usually covers the full prefix — a selective filter never reaches the page limit early, and `Pagination: nil` forces `CountTotal=true`.

Pre-PR that cost was paid once per query. With offset paging (`--page N`) it's now paid **per page**: `FilteredPaginate` re-runs the callback from the store head each time, so walking P pages costs ~P× the old single scan. And since `--page-key` doesn't round-trip from the CLI today (the CLI passes the printed base64 `next_key` through without decoding), offset paging is the mode CLI users will actually use.

A per-request memo would keep the re-scan but make it cheap — one graph walk per *distinct* channel/client instead of per entry, which puts this filter in the same cost class as the plain field-compare filters core's `FilteredPaginate` endpoints use:

```go
chainIDs := map[string]string{} // channelOrClientID → chainID, "" = not tendermint/skip
```

consulted at the top of the callback.

modules/apps/rate-limiting/client/cli/query.go
```diff
@@ -76,6 +81,7 @@ Example:
 
 	cmd.Flags().String(FlagDenom, "", "The denom identifying a specific rate limit")
 	flags.AddQueryFlagsToCmd(cmd)
+	flags.AddPaginationFlagsToCmd(cmd, "rate limits associated with a channel or client")
```
gjermundgaraba, line 86: These flags are only read on the `denom == ""` path. With `--denom` set they're accepted and silently dropped, which is easy to misread as "pagination applied". Cobra can enforce this natively, so we should probably do that here for a bit better UX.

CHANGELOG.md
```diff
@@ -44,6 +44,8 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### Improvements
 
+* (apps/rate-limiting) [\#8996](https://github.com/cosmos/ibc-go/pull/8996) Paginate the rate-limiting queries to avoid return the full store contents in a single response.
```
gjermundgaraba: Typo: "to avoid return the full store contents" → "to avoid returning the full store contents".

gjermundgaraba (changes requested):

gjermundgaraba:
@greptileai

gjermundgaraba (approved):
great stuff, thanks!

PR #9006, opened 2026-07-21
fix(apps/pfm): remove legacy params before v3 to v4 migration

### LINKED ISSUE #9004: [PFM] v3-to-v4 migration panics when legacy params remain at key 0x00
## Summary of Bug

  The packet-forward-middleware v3-to-v4 migration iterates over every entry in the PFM module store and attempts to decode each value as a legacy `InFlightPacket`.

  PFM previously stored its module parameters at key `0x00`. The fee functionality and PFM params API were later removed in [cosmos/ibc-apps#202](https://github.com/cosmos/ibc-apps/pull/202), but no store migration was added to delete the already-persisted params value.

  As a result, chains that used the previous PFM params implementation can still have an orphaned value at `0x00` when running the v3-to-v4 migration.

  The params value can be decoded permissively as an `InFlightPacket` with an empty `PacketTimeoutHeight`. `ChannelPacket()` then calls `MustParseHeight("")`, causing the migration to panic:

  ```text
  panic: expected height string format: {revision}-{height}. Got: : invalid height

  github.com/cosmos/ibc-go/v11/modules/core/02-client/types.MustParseHeight
  github.com/cosmos/ibc-go/v11/modules/apps/packet-forward-middleware/types.(*InFlightPacket).ChannelPacket
  github.com/cosmos/ibc-go/v11/modules/apps/packet-forward-middleware/migrations/v4.Migrate
```
  Because this happens during RunMigrations, it deterministically prevents the chain from completing its upgrade.

  Relevant code:

  - PFM v3-to-v4 migration (https://github.com/cosmos/ibc-go/blob/v11.2.0/modules/apps/packet-forward-middleware/migrations/v4/migrate.go#L21-L58)
  - InFlightPacket.ChannelPacket() (https://github.com/cosmos/ibc-go/blob/v11.2.0/modules/apps/packet-forward-middleware/types/types.go#L8-L17)
  - Removal of PFM fee and params functionality (https://github.com/cosmos/ibc-apps/pull/202)

  ## Expected Behaviour

  The v3-to-v4 migration should handle obsolete PFM store entries without panicking.

  A legacy params value at key 0x00 should be deleted or skipped before in-flight packet conversion.

  ## Version

  Observed with:

  github.com/cosmos/ibc-go/v11 v11.2.0

  The affected migration upgrades packet-forward-middleware consensus version 3 to version 4.

  ## Steps to Reproduce

  1. Create a PFM module store containing the legacy params entry at key 0x00.
  2. Set the packet-forward-middleware module consensus version to 3.
  3. Run the ibc-go v11 module migrations.
  4. The v3-to-v4 migration iterates over 0x00 and decodes its value as an
     InFlightPacket.

  5. ChannelPacket() calls MustParseHeight with an empty timeout height and
     panics.

  ## Root Cause

  The failure is caused by two migration paths interacting:

  1. PFM params functionality was removed without deleting the existing params value from the module store.

  2. The v3-to-v4 migration treats every PFM store entry as a legacy InFlightPacket.

  Therefore, this can affect chains following the historical PFM upgrade path and is not limited to manually corrupted state.

  ## Suggested Fix

  Update the PFM v3-to-v4 migration to remove or skip the known legacy params key
  before decoding store values as in-flight packets.

  For example:
```go
  legacyParamsKey := []byte{0x00}

  hasLegacyParams, err := store.Has(legacyParamsKey)
  if err != nil {
      return fmt.Errorf("failed to check legacy PFM params key: %w", err)
  }

  if hasLegacyParams {
      if err := store.Delete(legacyParamsKey); err != nil {
          return fmt.Errorf("failed to delete legacy PFM params key: %w", err)
      }
  }

  itr, err := store.Iterator(nil, nil)
  if err != nil {
      return err
  }
  defer itr.Close()

  for ; itr.Valid(); itr.Next() {
      // Existing InFlightPacket migration logic
  }
  }
```
  A migration test should also cover a v3 store containing both:

  - the legacy params entry at 0x00
  - a valid legacy in-flight packet

  The migration should delete or ignore the obsolete params entry and correctly
  migrate the valid packet without panicking.

nowooj:
## Description

  Fix the packet-forward-middleware v3-to-v4 migration when the module store still contains the deprecated PFM params entry under key `0x00`.

  Previously, the migration iterated over every store entry and decoded each value as a legacy `InFlightPacket`. Historical params bytes could therefore reach the packet conversion path and cause a panic while parsing an empty timeout height.

  This PR:

  - defines the historical params key as the migration-owned `LegacyParamsKey`;
  - checks for and deletes that key before creating the store iterator;
  - preserves the existing conversion and validation behavior for genuine in-flight packets;
  - continues to reject malformed or invalid packet entries;
  - adds regression coverage for a store containing both legacy params and a valid in-flight packet.

  The regression test verifies that the legacy params entry is removed while the packet remains readable through the current keeper and retains all fields supported by the v4 schema.

  closes: #9004

  ## Testing

  - `go test ./modules/apps/packet-forward-middleware/keeper -run TestKeeperTestSuite/TestMigrate3to4
  -count=1`
  - `go test ./modules/apps/packet-forward-middleware/... -count=1`
  - `make tidy-all`
  - `make build`
  - `make lint`
  - `make test-unit`
  - `git diff --check`

  ---

  - [x] Linked to GitHub issue with discussion and accepted design.
  - [x] Included a changelog entry.
  - [x] Added relevant unit regression coverage.
  - [x] Documentation update is not applicable because this fixes internal upgrade behavior without
  changing public APIs or integration instructions.
  - [x] Added a godoc comment for the exported migration key.
  - [x] Self-reviewed the final changes.
  - [x] Proposed conventional commit message:


modules/apps/packet-forward-middleware/migrations/v4/migrate.go
```diff
@@ -12,12 +12,25 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("failed to check legacy PFM params key: %w", err)
+	}
+	if hasLegacyParams {
+		if err := store.Delete(LegacyParamsKey); err != nil {
+			return fmt.Errorf("failed to delete legacy PFM params key: %w", err)
+		}
+	}
+
 	itr, err := store.Iterator(nil, nil)
```
gjermundgaraba, line 30: Non-blocking: `ChannelPacket()` panics in `MustParseHeight` on any malformed `PacketTimeoutHeight` (empty string included), before `ValidateBasic()` runs — so the `invalid in-flight packet ... for key %q` error below is unreachable, and a corrupt entry from a fork or imported genesis still halts the upgrade with a panic naming no store key. Using `clienttypes.ParseHeight` in the loop and returning the existing wrapped error keeps bad entries attributable (and makes that error path testable).

modules/apps/packet-forward-middleware/migrations/v4/migrate.go
```diff
@@ -12,12 +12,25 @@ import (
 	"github.com/cosmos/ibc-go/v11/modules/apps/packet-forward-middleware/types"
 )
 
+// LegacyParamsKey is the store key used by the deprecated PFM params state.
+var LegacyParamsKey = []byte{0x00}
```
gjermundgaraba: `0x00` can't be verified from inside this repo (PFM was vendored in after upstream removed params), and the test asserts through this same symbol — a wrong byte would keep CI green. One line citing the origin makes it checkable:

```suggestion
// LegacyParamsKey is the store key the upstream packetforward middleware used for
// its params (ParamsKey in packetforward/types/keys.go, removed by cosmos/ibc-apps#202).
var LegacyParamsKey = []byte{0x00}
```

modules/apps/packet-forward-middleware/migrations/v4/migrate.go
```diff
@@ -12,12 +12,25 @@ import (
...
 
+	hasLegacyParams, err := store.Has(LegacyParamsKey)
+	if err != nil {
+		return fmt.Errorf("failed to check legacy PFM params key: %w", err)
+	}
+	if hasLegacyParams {
+		if err := store.Delete(LegacyParamsKey); err != nil {
+			return fmt.Errorf("failed to delete legacy PFM params key: %w", err)
+		}
+	}
```
gjermundgaraba: `Delete` on an absent key is a no-op (and `Has`/`Delete` here can't actually return errors), so the guard is dead code — same shape as `v10.deleteParams` in core:

```suggestion
	if err := store.Delete(LegacyParamsKey); err != nil {
		return fmt.Errorf("failed to delete legacy PFM params key: %w", err)
	}
```

gjermundgaraba (changes requested):
Thank you so much for opening this! Apologies for the delay in reviewing. I left a couple of minor comments. If you prefer me to just take those over the finish line, just lemme know.

modules/apps/packet-forward-middleware/migrations/v4/migrate.go
nowooj: https://github.com/cosmos/ibc-go/pull/9006/commits/f7629812eae6bf7b8fcd5d7a7d920e86c4ceeba8

gjermundgaraba:
@greptileai

gjermundgaraba (approved):

PR #9007, opened 2026-07-21
imp(transfer): add per-channel escrow accounting

srdtrk:
## Description

closes: #XXXX

---

write a little note why.


modules/apps/transfer/types/genesis.go
```diff
@@ -35,5 +39,41 @@ func (gs GenesisState) Validate() error {
...
 		return err
 	}
-	return gs.TotalEscrowed.Validate() // will fail if there are duplicates for any denom
+	if err := gs.TotalEscrowed.Validate(); err != nil {
+		return err
+	}
+	// ChannelEscrows is nil when importing a legacy genesis created before
+	// per-channel escrow accounting was added.
+	if gs.ChannelEscrows != nil {
+		if err := gs.validateChannelEscrows(); err != nil {
```
gjermundgaraba, line 48: importing an old genesis with `total_escrowed` set but no `channel_escrows` passes validation, but `InitGenesis` never fills in the per-channel entries. and since a fresh chain start stamps the module version map at 7, the v6 to v7 migration won't run either. every channel starts at zero escrow and receives/refunds on old channels fail with `ErrInsufficientEscrow`. we should either rebuild the entries from bank balances on import (like the migration does), or reject this genesis and make people migrate the file first.

modules/apps/transfer/keeper/relay.go
```diff
@@ -284,33 +278,39 @@ func (k *Keeper) refundPacketTokens(
...
 
 	return nil
 }
 
-// EscrowCoin will send the given coin from the provided sender to the escrow address. It will also
-// update the total escrowed amount by adding the escrowed coin's amount to the current total escrow.
-func (k *Keeper) EscrowCoin(ctx sdk.Context, sender, escrowAddress sdk.AccAddress, coin sdk.Coin) error {
+// EscrowCoin sends the coin to the channel or client escrow account and updates its accounting.
+func (k *Keeper) EscrowCoin(ctx sdk.Context, portID, channelOrClientID string, sender sdk.AccAddress, coin sdk.Coin) error {
+	k.AddToChannelEscrow(ctx, channelOrClientID, coin)
```
gjermundgaraba, line 291: this updates the accounting before `SendCoins` checks the sender has the funds (on main it was the other way around). the msg server only special-cases the exact `UnboundedSpendLimit`, so a transfer for one below the max slips through, and if the channel escrow already holds 2 or more of that denom the `Add` panics on overflow instead of returning insufficient funds. not a halt since the sdk recovers it, but it should fail cleanly. can we send first and update the accounting after, like before?

modules/apps/transfer/keeper/migrations.go
```diff
@@ -62,6 +64,59 @@ func (m Migrator) MigrateTotalEscrowForDenom(ctx sdk.Context) error {
...
+	portID := m.keeper.GetPort(ctx)
+	// Use a map to deduplicate identifiers that may be discovered through more than one source
+	// before sorting them for deterministic iteration.
+	identifiers := map[string]struct{}{exported.LocalhostClientID: {}}
+	// IBC v2 aliases are keyed only by v1 channel ID and may originate from any port.
+	// An empty prefix returns all v1 channels so every possible aliased escrow identifier is included.
+	for _, channel := range m.keeper.channelKeeper.GetAllChannelsWithPortPrefix(ctx, "") {
+		identifiers[channel.ChannelId] = struct{}{}
+	}
+	for _, client := range m.keeper.clientKeeper.GetAllGenesisClients(ctx) {
```
gjermundgaraba, line 78: `GetAllGenesisClients` decodes every client state via `MustUnmarshalClientState`, so if the chain still has state for a light client type it no longer registers (a dropped solomachine or wasm client, say), this panics mid-upgrade. the migration only needs the client ids, so iterating the store keys without decoding the values avoids this.

modules/apps/transfer/keeper/keeper.go
```diff
@@ -42,7 +48,7 @@ type Keeper struct {
 }
 
 // NewKeeper creates a new IBC transfer Keeper instance
-func NewKeeper(cdc codec.BinaryCodec, addressCodec address.Codec, storeService corestore.KVStoreService, channelKeeper types.ChannelKeeper, msgRouter types.MessageRouter, authKeeper types.AccountKeeper, bankKeeper types.BankKeeper, authority string) *Keeper {
+func NewKeeper(cdc codec.BinaryCodec, addressCodec address.Codec, storeService corestore.KVStoreService, channelKeeper types.ChannelKeeper, clientKeeper types.ClientKeeper, msgRouter types.MessageRouter, authKeeper types.AccountKeeper, bankKeeper types.BankKeeper, authority string) *Keeper {
```
gjermundgaraba, line 51: `clientKeeper` is only used for the one-time v6 to v7 migration, but adding it here changes `NewKeeper` for every downstream chain permanently. could it be wired into the migrator instead? if it has to stay, it should at least be called out as an api break in the changelog.

modules/apps/packet-forward-middleware/keeper/keeper.go
```diff
@@ -128,6 +128,11 @@ func (k *Keeper) WriteAcknowledgementForForwardedPacket(ctx sdk.Context, packet
 
 	// Sender chain is source
 	if !denom.HasPrefix(packet.SourcePort, packet.SourceChannel) {
+		channelEscrow := k.transferKeeper.GetChannelEscrowForDenom(ctx, packet.SourceChannel, coin.Denom)
+		if channelEscrow.Amount.LT(coin.Amount) {
+			return fmt.Errorf("insufficient escrow on channel %s/%s: has %s, requested %s", packet.SourcePort, packet.SourceChannel, channelEscrow, coin)
```
gjermundgaraba, line 133: this is the same condition the transfer keeper returns as `transfertypes.ErrInsufficientEscrow`, but the plain `fmt.Errorf` loses `errors.Is` matching and the registered error code. pfm already imports transfertypes, so this could just wrap that error.

proto/ibc/applications/transfer/v1/query.proto
```diff
@@ -42,6 +42,16 @@ service Query {
...
     option (google.api.http).get = "/ibc/apps/transfer/v1/total_escrow/{denom=**}";
   }
+
+  // ChannelEscrowForDenom returns the amount of tokens in escrow for a channel or client and denomination.
+  rpc ChannelEscrowForDenom(QueryChannelEscrowForDenomRequest) returns (QueryChannelEscrowForDenomResponse) {
+    option (google.api.http).get = "/ibc/apps/transfer/v1/channel_escrow/{channel_or_client_id}/{denom=**}";
+  }
+
+  // AllChannelEscrows returns all channel and client escrow amounts.
+  rpc AllChannelEscrows(QueryAllChannelEscrowsRequest) returns (QueryAllChannelEscrowsResponse) {
```
gjermundgaraba, line 52: all six existing query rpcs have a command in `GetQueryCmd`, but these two new ones don't. can we add cli commands for them too?

modules/apps/transfer/module.go
```diff
@@ -139,7 +143,7 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
 }
 
 // ConsensusVersion implements AppModule/ConsensusVersion defining the current version of transfer.
-func (AppModule) ConsensusVersion() uint64 { return 6 }
+func (AppModule) ConsensusVersion() uint64 { return 7 }
```
gjermundgaraba, line 146: this bumps the consensus version, adds two public queries, and changes the `NewKeeper` signature, but there's no changelog entry, docs update, or breaking marker in the pr title. can we add those before merge?

gjermundgaraba (changes requested):

PR #9033, opened 2026-08-11
fix(04-channel): reject packet timeouts on a future client revision

### LINKED ISSUE #8653: IBC packet non-timeout with invalid RevisionNumber

lapidix:
## Description

Fixes #8653.

When a packet is sent with a timeout height whose **revision number is greater
than the counterparty client's current revision number**, the height can never
be reached — revision numbers only increase through a coordinated chain upgrade,
and `Height.Compare` orders by revision number first, so `heightElapsed` never
returns true. Combined with a zero timeout timestamp, this yields a packet that
**never times out** (reproduced on a live chain; PoC in the issue).

This PR rejects such timeouts in `SendPacket`, before the packet commitment is
written:

```go
if !timeoutHeight.IsZero() && timeoutHeight.RevisionNumber > latestHeight.RevisionNumber {
	clientType, _, err := clienttypes.ParseClientIdentifier(connectionEnd.ClientId)
	if err != nil {
		return 0, errorsmod.Wrapf(err, "unable to parse client identifier %s", connectionEnd.ClientId)
	}

	var unreachable bool
	switch clientType {
	case exported.Tendermint, exported.Localhost:
		unreachable = true
	case exported.Solomachine:
		unreachable = false
	default:
		unreachable = latestHeight.RevisionNumber > 0
	}

	if unreachable {
		return 0, errorsmod.Wrapf(
			clienttypes.ErrInvalidHeight,
			"packet timeout height revision number (%d) cannot exceed the counterparty client's current revision number (%d)",
			timeoutHeight.RevisionNumber, latestHeight.RevisionNumber,
		)
	}
}
```

The check is gated on the client type because solomachine heights carry no
revision semantics and legitimately mismatch the timeout revision. Tendermint
and localhost always use revisioned heights, so they are checked even at
revision 0 (a chain-id without a revision suffix); unknown client types are
checked only when they report a revision-numbered latest height. A *lower*
timeout revision needs no new check — it already fails with
`ErrTimeoutElapsed`.

The client type is parsed from the client identifier rather than read from the
client state — client identifiers are generated from the client type at
creation, and the sentinel localhost client stores no client state at all. The
branch performs no state reads, and it is only entered when the timeout
revision exceeds the client's current revision, so all other sends are
unaffected.

This change is state-machine breaking: a `SendPacket` that previously
succeeded can now return an error.

## Testing

- Added `TestSendPacket` cases "timeout height has unreachable revision number"
  against a regular tendermint client, a revision-0 counterparty, and the
  localhost client.
- Existing solomachine success cases exercise the exemption and pass.
- `go test ./modules/core/04-channel/...` passes (full package), `go vet` clean.

---

- [x] Linked to GitHub issue with discussion and accepted design (#8653).
- [x] Include changelog entry when appropriate.
- [x] Wrote unit and integration tests.
- [ ] Updated documentation (`docs/`) — n/a, no API change.
- [x] Added `godoc` comments — inline comments on the new check.
- [x] Self-reviewed `Files changed` in the GitHub PR explorer.
- [x] Provide a conventional commit message.


modules/core/04-channel/keeper/packet.go
```diff
@@ -65,6 +65,39 @@ func (k *Keeper) SendPacket(
...
+	// with a zero timeout timestamp the packet has no other timeout condition.
+	// see https://github.com/cosmos/ibc-go/issues/8653
+	if !timeoutHeight.IsZero() && timeoutHeight.RevisionNumber > latestHeight.RevisionNumber {
+		// the localhost client stores no client state, so the type is taken from the identifier
+		clientType, _, err := clienttypes.ParseClientIdentifier(connectionEnd.ClientId)
+		if err != nil {
+			return 0, errorsmod.Wrapf(err, "unable to parse client identifier %s", connectionEnd.ClientId)
+		}
+
+		var unreachable bool
```
srdtrk:
```suggestion
```
I think this variable is not needed since we can just return the error within the switch statement

modules/core/04-channel/keeper/packet.go
```diff
@@ -65,6 +65,39 @@ func (k *Keeper) SendPacket(
...
+		switch clientType {
+		case exported.Tendermint, exported.Localhost:
+			// these clients use revision-numbered heights, including revision 0
+			unreachable = true
```
srdtrk:
```suggestion
			return 0, errorsmod.Wrapf(
				clienttypes.ErrInvalidHeight,
				"packet timeout height revision number (%d) cannot exceed the counterparty client's current revision number (%d)",
				timeoutHeight.RevisionNumber, latestHeight.RevisionNumber,
			)
```
ditto: removing unreachable var

modules/core/04-channel/keeper/packet.go
```diff
@@ -65,6 +65,39 @@ func (k *Keeper) SendPacket(
...
+		case exported.Solomachine:
+			// solomachine heights have no revision semantics
+			unreachable = false
+		default:
+			// a non-zero revision number indicates the client uses revision-numbered heights
+			unreachable = latestHeight.RevisionNumber > 0
```
srdtrk: I think for other light clients, we can just assume that they have no revision semantics. If we have other light clients with revision semantics, we can always add them later:

```suggestion
		default:
			// no-op: other clients are assumed to have no revision semantics
			// update this sanity check if new light clients with revision sematics are added
```
Maybe it is time to use `if` rather than `switch` after this change

modules/core/04-channel/keeper/packet.go
```diff
@@ -65,6 +65,39 @@ func (k *Keeper) SendPacket(
...
+		}
+
+		if unreachable {
+			return 0, errorsmod.Wrapf(
+				clienttypes.ErrInvalidHeight,
+				"packet timeout height revision number (%d) cannot exceed the counterparty client's current revision number (%d)",
+				timeoutHeight.RevisionNumber, latestHeight.RevisionNumber,
+			)
+		}
```
srdtrk:
```suggestion
```
ditto: removing unreachable var

srdtrk (changes requested):
Thanks for adding this sanity check. I've requested some minor changes.

lapidix:
@srdtrk Thanks for the review. I dropped the `unreachable` variable and return the error directly, and simplified the check to a plain `if` covering tendermint and localhost only.

srdtrk (dismissed):

dianab-cl:
@greptile review

dianab-cl (dismissed):

modules/core/04-channel/keeper/packet.go
```diff
@@ -67,6 +67,30 @@ func (k *Keeper) SendPacket(
 		return 0, errorsmod.Wrapf(clienttypes.ErrInvalidHeight, "cannot send packet using client (%s) with zero height", connectionEnd.ClientId)
 	}
 
+	// a packet timeout height must not be set on a revision number the counterparty
+	// client has not reached: the client cannot consider such a height elapsed, and
+	// with a zero timeout timestamp the packet has no other timeout condition.
+	// see https://github.com/cosmos/ibc-go/issues/8653
+	if !timeoutHeight.IsZero() && timeoutHeight.RevisionNumber > latestHeight.RevisionNumber {
```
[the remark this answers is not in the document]
srdtrk: no-need. We wanna be as restrictive as possible

dianab-cl (approved):

srdtrk:
@lapidix We are ready to merge this PR but the repo requires signed commits. I must ask you to sign your commits (and force push). Please don't drop my changes since I adjusted it after reviews from other contributors. Thank you

lapidix:
@srdtrk 
Thanks for letting me know. I've signed all commits and your changes are unchanged. The tree matches 5960bf2.

srdtrk (approved):

